PR #1, opened 2019-06-07
CI: move to travis

benma:


.ci/ci.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -e
+
+# Check style
+
+./.ci/check-style
+make -C py
+./.ci/check-pep8
+
+# Bootloader variants
+make -j8 bootloader
```
NickeZ: Do we need `make clean` in between every call to make?

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+services:
+  - docker
+
+os:
+  - linux
+
+addons:
+  apt:
+    packages:
+      - make
```
NickeZ: I don't think you need `make` here, is this from some earlier attempts?

.ci/ci.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -e
```
NickeZ: It is recommended to have `set -e` on a separate line so that the scripts exits on fail even if it is called like so: `bash ci.sh`.

IMO names of shell scripts should also not be suffixed with `.sh`. The fact that the application is written as a shell script is an implementation detail.

I typically use [google style guide](https://google.github.io/styleguide/shell.xml) for shell scripts.

.ci/ci.sh
benma: Probably, but I want to avoid CI time to go up that much :/ Do you think we can get away like it is now?

It should work fine once those are proper cmkae targets, right?

.travis.yml
benma: I copied it from our other travis files where we need `make`. I will try pushing a commit without it and seeing if it will work.

.ci/ci.sh
benma: :+1: 
Done

benma: > bash ci.sh.

...

> an implementation detail.

:laughing:

PR #4, opened 2019-06-11
Fix sensor intial calibration

NickeZ:


src/drivers/qtouch/qtouch.c
```diff
@@ -273,6 +273,29 @@ static touch_ret_t touch_sensors_config(void)
...
 
+
+/*
+ * Force calibrate
+ *
+ * Call this function when the user "probably" isn't touching the device to reset all the
+ * calibration values.
+ * TODO: Disable if key is active
+ */
+//#include <screen.h>
```
benma: remove

src/drivers/qtouch/qtouch.c
```diff
@@ -273,6 +273,29 @@ static touch_ret_t touch_sensors_config(void)
 	return (touch_ret);
 }
 
+
+/*
```
benma: move documentation to the header file

src/ui/components/orientation_arrows.c
```diff
@@ -61,18 +62,21 @@ static void _render(component_t* component)
     int16_t position = data->screen_count / SCALE;
     if (position < COUNT_CHANGE_DIRECTION) {
         // Horizontal motion
+        qtouch_force_calibrate();
```
benma: instead of calling it in each if clause, why not just at the top of the render function?

src/drivers/qtouch/qtouch.c
```diff
@@ -273,6 +273,29 @@ static touch_ret_t touch_sensors_config(void)
...
+//#include <screen.h>
+void qtouch_force_calibrate(void) {
+    qtm_touch_key_data_t* key;
+    for (uint16_t i = 0u; i < DEF_NUM_CHANNELS; i++) {
+        key = &qtlib_key_data_set1[i];
+        uint16_t value = key->node_data_struct_ptr->node_acq_signals;
+        uint16_t reference = key->channel_reference;
+        int32_t diff = (int32_t)reference - (int32_t)value;
+        //screen_sprintf_debug(100, "%u %u %d", value, reference, diff);
+        if(diff > 10 || diff < -10){
```
douglasbakkum: nit would be to make the `10` a macro 

also remove the debug printing before merge

src/ui/components/orientation_arrows.c
douglasbakkum: I agree with Marko that would be cleaner.

src/drivers/qtouch/qtouch.c
douglasbakkum: Haven't tested this PR, so this is speculation:

The scenario of touching a sensor on plugin should only give `reference` > `value`. So it is safer to only re-calibrate on that scenario. Then the other scenario would occur if a touch is happening, and so re-calibrating then could lead to a 'dead' sensor later.

src/drivers/qtouch/qtouch.c
NickeZ: This is not consistent with the rest of `qtouch.c/.h`

src/ui/components/orientation_arrows.c
NickeZ: I misunderstood how many times it would be called there before. My goal at that time was to only call calibrate at the end of the animation. However this is not needed anymore.

src/drivers/qtouch/qtouch.c
NickeZ: I agree, changed to only check `reference > value` and also check key state to avoid spurious triggers.

src/drivers/qtouch/qtouch.c
NickeZ: would be inconsistent with `qtouch.c/.h`

NickeZ:
I had to make some quite awkward preprocessor ifs. Is there a better way?

douglasbakkum:
lgtm

src/ui/components/orientation_arrows.c
```diff
@@ -17,6 +17,13 @@
 #include <hardfault.h>
 #include <util.h>
 
+#if !defined(TESTING)
+#include <qtouch.h>
+#else
+void qtouch_force_calibrate(void);
```
benma: this should move here:
https://github.com/digitalbitbox/bitbox02-firmware/blob/pr%2F4/test/unit-test/framework/mock_qtouch.c#L22

benma: And then include mock_qtouch, like here:

https://github.com/digitalbitbox/bitbox02-firmware/blob/f8c348a9ab61175b06a348d6b7d17c3c41c8cc84/src/touch/gestures.c#L24

PR #5, opened 2019-06-11
Attestation

benma:


src/hww.c
```diff
@@ -28,14 +75,32 @@
...
+            out_packet->len = 0;
+            return;
+        default:
+            break;
+        }
+    }
+
+    // No other message than the attestation and unlock calls shall pass until the device is
+    // unlocked or ready to be initialized.
+    if (memory_is_initialized() && !keystore_is_unlocked()) {
```
NickeZ: IMHO: it is easier to read code if the boolean function is positive. `keystore_is_locked()` is easer to understand than `!keystore_is_unlocked()`.

src/workflow/unlock.c
```diff
@@ -132,6 +145,12 @@ void workflow_unlock_enter_done(const char* password)
 
 void workflow_unlock(void)
 {
-    // "Enter password"
-    ui_screen_stack_switch(entry_screen_create("Enter password", _enter));
+    if (!memory_is_initialized() || keystore_is_unlocked()) {
```
NickeZ: Is this a sanity check or are there valid code paths that lead to this return? If sanity-check, should it debug print or abort?

py/bitbox02/devices.py
```diff
@@ -34,3 +37,11 @@ def get_bitbox02_devices(product_string=BITBOX02):
         and (info["usage_page"] == 0xFFFF or info["interface_number"] == 0)
         and info["product_string"] == product_string
     ]
+
+
+def parse_device_version(serial_number):
```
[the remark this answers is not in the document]
benma: I like that the regexp works no matter what comes before (e.g. it will work if we strip the prefix), and that it works robustly also with the bootloader version. It happens to have the same length prefix (`bl` instead of `fw`), but that sounds pretty fragile.

I will extend the regexp to go to the end of the line instead if ok.

src/hww.c
benma: Done.

src/workflow/unlock.c
benma: It is a valid code path. OP_UNLOCK calls this, and it does nothing if we are already unlocked (can happen when if you restart the desktop app or switch wallets while the device is is unlocked).

py/bitbox02/devices.py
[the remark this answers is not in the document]
benma: Done.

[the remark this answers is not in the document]
NickeZ: regexes can also be hard. It has to be `.*` as there might not be any more characters after the last number, and it should end with `$` so that it takes the whole string until the end. `r"v([0-9]+\.[0-9]+\.[0-9]+.*$)"`

src/workflow/unlock.c
NickeZ: OK

py/bitbox02/devices.py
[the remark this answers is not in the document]
benma: :flushed: fixed

PR #7, opened 2019-06-16
CMakeLists: update libwally (and secp256k1)

benma:
Pulled in libwally and rebased our patches, and put them into the
bitbox02-firmare branch of
https://github.com/shiftdevices/libwally-core (reviously firmware_v2
branch).


NickeZ:
long term, should we move the dependencies to this github organization instead?

benma:
> long term, should we move the dependencies to this github organization instead?

yeah!

PR #14, opened 2019-06-21
workflow/verify: combine amount and unit

benma:
It makes it easier to tack on a separate unit for the total and the fee.

### DIFF
--- a/src/apps/btc/btc_common.c
+++ b/src/apps/btc/btc_common.c
@@ -151,9 +151,9 @@ static void _sprint_uint64(uint64_t value, char* out, char** out_start)
     *out_start = p;
 }
 
-bool btc_common_format_amount(uint64_t satoshi, char* out, size_t out_len)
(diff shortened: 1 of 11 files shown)


src/apps/btc/btc_common.c
```diff
@@ -165,28 +165,33 @@ bool btc_common_format_amount(uint64_t satoshi, char* out, size_t out_len)
...
     _sprint_uint64(remainder, remainder_str, &remainder_str_start);
+    char amount_str[30] = {0};
     int sprintf_result = snprintf(
-        out,
-        out_len,
+        amount_str,
+        sizeof(amount_str),
         "%s.%.*s%s",
         quotient_str_start,
         // pad up to 8 zeroes
```
NickeZ: Could the comment say `// left pad 'remainder_str' up to 8 zeros`?

benma: Done

PR #15, opened 2019-06-24
workflow/verify: combine amount and unit

benma:
It makes it easier to tack on a separate unit for the total and the fee.


benma:
Approved in https://github.com/digitalbitbox/bitbox02-firmware/pull/14

PR #18, opened 2019-06-25
Tidy

benma:


scripts/tidy
```diff
@@ -22,4 +22,8 @@ fi
 sed -i 's/-mcpu=cortex-m4//g; s/-mfloat-abi=softfp//g; s/-mlong-calls//g; s/-mthumb//g;
         s/-Wno-cast-function-type//g; s/-mfpu=fpv4-sp-d16//g' build/compile_commands.json
 
-${RUNCLANGTIDY} -p build
+# Exclude drivers and external for now. We get the base file set from compile_commands.json, as
+# clang-tidy unfortunately tries to lint a provided source anyway even if it is not part of the
+# build.
+SOURCES=`cat build/compile_commands.json | jq 'map_values(.file) | .[]' | grep -v drivers/ | grep -v external/ | grep -v tools/nanopb/ | xargs echo`
```
NickeZ: You should check if `jq` is available like it is done for `clang-tidy-8`. I think you can assume that `cat`, `grep` and `xargs` are available.

NickeZ: You can do all the greps with one line `grep -v -e 'pattern1' -e 'pattern2'`.

NickeZ: Is it really necessary to `| xargs echo`?

benma: 1) and 2): done

3) yes, if i missing there are quotes around each line, which I thought would not hurt, but there are lots of errors without it. I didn't spend more time trying to figure out why and if there is a cleaner way to solve it.

NickeZ: Interesting way to remove the quotes.

I realise that I forgot to write that I also prefer `$()` over ` `` ` because of readability.

NickeZ:
Could we merge https://github.com/digitalbitbox/bitbox02-firmware/pull/19 before this? So that I don't have to rebase right now to run CI on my cmake-cleanup branch? You should then create a docker image with tag `2` and use that in this branch

PR #19, opened 2019-06-25
Launch specific version of docker container in CI

NickeZ:


NickeZ:
Are there any conventions to docker tags? Should it be `1` or `v1` ? I assume we won't use semver and simply use a counter.

hkjn:
> Are there any conventions to docker tags? Should it be `1` or `v1` ? I assume we won't use semver and simply use a counter.

Fwiw, I like using semver for tagging Docker images, e.g starting at `0.1.0` and increasing the minor version by one each time there's a change. See e.g [here](https://github.com/digitalbitbox/bitbox-base/pull/90/files#diff-b6791165).

NickeZ:
> > Are there any conventions to docker tags? Should it be `1` or `v1` ? I assume we won't use semver and simply use a counter.
> 
> Fwiw, I like using semver for tagging Docker images, e.g starting at `0.1.0` and increasing the minor version by one each time there's a change. See e.g [here](https://github.com/digitalbitbox/bitbox-base/pull/90/files#diff-b6791165).

We will probably break with every release of the docker image since we don't have a single build machine and we use `apt`. So every time debian/ubuntu releases a new version of any packages that are installed into our image we will break backwards comp. Which means that every release would have redundant `.0.0`. In general I also prefer not to use semver for things that doesn't have an API which I plan to keep strict control over.

benma:
@hkjn when would you update the minor vs major version? A Dockerfile is not an API, where semver is usually used (major bump for an incompatible API change, minor for a compatible change).

The desktop app is also "semver", but I realized relatively quickly that it does not carry any real meaning, as the app is also not an API. Should probably switch that to a single number like FF, Chrome, etc. at some point.

hkjn:
> > Fwiw, I like using semver for tagging Docker images, e.g starting at `0.1.0` and increasing the minor version by one each time there's a change. See e.g [here](https://github.com/digitalbitbox/bitbox-base/pull/90/files#diff-b6791165).
> 
> We will probably break with every release of the docker image since we don't have a single build machine and we use `apt`. So every time debian/ubuntu releases a new version of any packages that are installed into our image we will break backwards comp. Which means that every release would have redundant `.0.0`. In general I also prefer not to use semver for things that doesn't have an API which I plan to keep strict control over.

Yeah, this is true, by fetching external resources i.e via `apt` even rebuilding from the same `Dockerfile` can produce different / breaking results. I guess I just like things being "shaped like" semver, since it simplifies parsing and reasoning about version numbers, but these are good counterarguments against using it.

hkjn:
> Should probably switch that to a single number like FF, Chrome, etc. at some point.

Right, my Firefox is `60.7.1esr` currently, so they both use Semver but in practice seem to just bump the "major" version.. presumably minor version denotes release candidates or similar. CoreOS does something similar, and are on version `2079.6.0` for their stable channel at time of writing: :)

* https://coreos.com/releases/

NickeZ:
> Right, my Firefox is `60.7.1esr` currently, so they both use Semver but in practice seem to just bump the "major" version.. presumably minor version denotes release candidates or similar. CoreOS does something similar, and are on version `2079.6.0` for their stable channel at time of writing: :)
> 
>     * https://coreos.com/releases/

It could be that Firefox is using semver (the line you printed is invalid though, you are not allowed to put `esr` onto the patch number). But simply using MAJOR.MINOR.PATCH is not "using semver" to me. The meaning of the numbers, the "semantic" part of semantic versioning is what makes something semver to me.

Using MAJOR.MINOR.PATCH if you are not using semver can trick others into thinking that you are using semver. So I actively don't use that format for stuff which doesn't have an API.

hkjn:
> It could be that Firefox is using semver (the line you printed is invalid though, you are not allowed to put `esr` onto the patch number). But simply using MAJOR.MINOR.PATCH is not "using semver" to me. The meaning of the numbers, the "semantic" part of semantic versioning is what makes something semver to me.

Apologies for speaking unclearly. I meant to say that Firefox appears to be using version numbers that appear similar to semver. (Good point about the `esr` in the `PATCH`.)

I didn't mean to imply that Firefox has a stable API and that the version numbers of their releases can be used to infer anything about backwards compatibility of that API, or anything along those lines.
> 
> Using MAJOR.MINOR.PATCH if you are not using semver can trick others into thinking that you are using semver. So I actively don't use that format for stuff which doesn't have an API.

PR #21, opened 2019-06-25
src/util: Improve min/max

NickeZ:
This stops double evaluation of arguments to min/max. With previous macro, `MIN(i++,16)` would have increased `i` twice.

### DIFF
--- a/test/unit-test/test_util.c
+++ b/test/unit-test/test_util.c
@@ -1,0 +1,61 @@
+#include "util.h"
+
+// Copyright 2019 Shift Cryptosecurity AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <setjmp.h>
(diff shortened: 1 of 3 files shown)


benma:
@NickeZ A thought: wouldn't it be easier to just turn those into functions instead of macros? Maybe an `inline static` full function in the header so it is inlined like the macro is.

NickeZ:
> @NickeZ A thought: wouldn't it be easier to just turn those into functions instead of macros? Maybe an `inline static` full function in the header so it is inlined like the macro is.

I thought about that as well, but since C doesn't have generics you cannot have min/max functions that work with different types :/

test/unit-test/test_util.c
```diff
@@ -0,0 +1,61 @@
+#include "util.h"
+
+// Copyright 2019 Shift Cryptosecurity AG
```
benma: license is usually the very first thing in the file

test/unit-test/test_util.c
```diff
@@ -0,0 +1,61 @@
...
+#include <stddef.h>
+#include <cmocka.h>
+
+#include "util.h"
+
+static void test_minmax(void** state)
+{
+    (void)state;
+
+    assert_true(1);
```
benma: ??

test/unit-test/test_util.c
```diff
@@ -0,0 +1,61 @@
...
+    assert_int_equal(res, 10);
+
+    res = MAX(10, 5);
+    assert_int_equal(res, 10);
+
+    res = MIN(5, 10);
+    assert_int_not_equal(res, 10);
+
+    res = MIN(10, 5);
+    assert_int_not_equal(res, 10);
```
benma: redudant with line 37? this is basically testing that 10 !=5 ;)

same with the following two

PR #26, opened 2019-06-29
Smallf fixes

NickeZ:


src/backup.c
```diff
@@ -27,6 +27,34 @@
 #include <util.h>
 #include <version.h>
 
+/**
```
benma: please move docstring to the header file

src/backup.c
```diff
@@ -27,6 +27,34 @@
...
+ */
+
+const char* backup_error_str(backup_error_t err)
+{
+    switch (err) {
+    case BACKUP_OK:
+        return "OK";
+    case BACKUP_STALE:
+        return "STALE";
+    case BACKUP_SEED_UNACCESSIBLE:
```
benma: INACCESSIBLE

src/restore.c
```diff
@@ -273,10 +273,8 @@ restore_error_t restore_list_backups(ListBackupsResponse* backups)
             continue;
         }
         backups->info[backups->info_count].timestamp = backup.backup_v1.content.metadata.timestamp;
-        memcpy(
-            backups->info[backups->info_count].id,
-            dir,
-            sizeof(backups->info[backups->info_count].id));
+        int max_size = MIN(sizeof(backups->info[backups->info_count].id), strlen(dir) + 1);
+        memcpy(backups->info[backups->info_count].id, dir, max_size);
```
benma: hmm if the 2nd arg is bigger than the first, it won't be null terminated?

Why not snprintf like on the next line? you can also check the result of snprintf to see if the string was truncated and Abort()/return an error, like for example here: https://github.com/digitalbitbox/bitbox02-firmware/blob/22807a82670d9e62f2874095cf9eb768f36d634b/src/sd.c#L48

NickeZ: Yeah, I'll switch to snprintf.

PR #29, opened 2019-07-01
[wip] Polish cmake

NickeZ:
Todo

- [ ] Port device tests
- [x] Document that the outer `Makefile` is a hack. It is used as a command runner. Probably remove all dependencies in it.
- [x] ? Run tests with ctests, also enable valgrind support: https://stackoverflow.com/questions/9303711/how-do-i-make-ctest-run-a-program-with-valgrind-without-dart/9843594
- [x] Add `bootstrap` in command runner makefile target that does `git submodule init...`.
- [x] Remove `external/*-config` directories.
- [ ] Finally check shasums of all generated targets so that output didn't change.
- [ ] Which targets should be in default? (currently bootloader, bootloader-development, firmware, factory-setup)
- [ ] Do we prefer indent level of 2 (seems to be more standard in the community) or 4 (more consistent with our C code)


.gitignore
```diff
@@ -1,30 +1,11 @@
...
-.DS_Store
-.ycm*
-*.o
-*.pyc
-*.sw*
-*.bak
-*.bin
-!*.c
+/py/**/*.pyc
+.mypy_cache
```
benma: should be `py/.mypy_cache`, as CI runs mypy inside the `py` folder.

PR #30, opened 2019-07-01
keystore: unit test and harden unlock()

benma:
Be able to use it as a PW check only.

### DIFF
--- a/src/keystore.c
+++ b/src/keystore.c
@@ -19,6 +19,7 @@
 #include "keystore.h"
 #include "memory.h"
 #include "random.h"
+#include "reset.h"
 #include "salt.h"
 #include "screen.h"
 #include "securechip/securechip.h"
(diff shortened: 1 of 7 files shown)


src/keystore.c
```diff
@@ -292,8 +297,15 @@ keystore_error_t keystore_unlock(const char* password, uint8_t* remaining_attemp
         return KEYSTORE_ERR_GENERIC;
     }
     if (password_correct) {
-        memcpy(_retained_seed, seed, sizeof(_retained_seed));
-        _is_unlocked_device = true;
+        if (_is_unlocked_device) {
+            // Already unlocked. Fail if the seed changed under our feet (should never happen).
```
NickeZ: should we have an abort if it should never happen?

src/workflow/restore.c
```diff
@@ -40,13 +40,11 @@ static bool _result = false;
 static bool _restore(const char* password)
 {
     bool res = restore_seed(&_backup_data, password);
-    component_t* status;
     if (res) {
```
NickeZ: Maybe `res == true`?

src/keystore.c
benma: Maybe. In this case, it was a bit easier to unit test, and the caller Aborts() anyway on a generic error.

src/workflow/restore.c
benma: hmm why?

PR #31, opened 2019-07-01
Enable coverage reports from codecov

NickeZ:


.codecov.yml
```diff
@@ -0,0 +1,4 @@
+ignore:
+  - 'test'
+  - 'external'
+  - 'src/drivers/**/*'
```
benma: also `tools`?

.ci/travis-ci
```diff
@@ -1,9 +1,24 @@
...
 git fetch
 
-docker run -v ${TRAVIS_BUILD_DIR}:/bitbox02-firmware/ \
-       -i shiftcrypto/firmware_v2:2 \
-       bash -c "make -C /bitbox02-firmware ci"
+docker run \
+	--cap-add SYS_PTRACE \
+	--volume ${TRAVIS_BUILD_DIR}:/bitbox02-firmware/ \
+	--workdir /bitbox02-firmware \
+	shiftcrypto/firmware_v2:2 \
```
benma: can this be in a variable so you can't forget to update the tag in the 2nd place?

.codecov.yml
NickeZ: AFAIK we don't compile any code in tools, so not needed for now.

.ci/travis-ci
NickeZ: ack

.codecov.yml
benma: turns out we do, nanopb encoding/decoding ;)
fix here: https://github.com/digitalbitbox/bitbox02-firmware/pull/47

PR #32, opened 2019-07-01
Password

benma:


src/workflow/password_enter.h
```diff
@@ -0,0 +1,29 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef _PASSWORD_ENTER_H_
+#define _PASSWORD_ENTER_H_
+
+#include <ui/components/set_password.h> // for SET_PASSWORD_MAX_PASSWORD_LENGTH
```
NickeZ: This include should be in `password_enter.c`.

src/workflow/password_enter.h
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Starts the enter password workflow.
+ * This call blocks.
+ * @param[in] title screen
+ * @param[out] password_out must be SET_PASSWORD_MAX_PASSWORD_LENGTH bytes (including null
+ * terminator). otherwise.
```
NickeZ: `otherwise.`?

src/workflow/unlock.c
```diff
@@ -52,94 +44,63 @@ static void _finish_bip39(const char* passphrase)
...
     ui_screen_stack_pop_all();
-    ui_screen_stack_push(entry_screen_create("Enter password", _enter));
+    char password[SET_PASSWORD_MAX_PASSWORD_LENGTH] = {0};
+    password_enter("Enter password", password);
+    workflow_unlock_enter_done(password);
+    util_zero(password, sizeof(password));
 }
 
-static void _start(void)
+static void _get_mnemonic_passphrase(char* passphrase_out)
```
NickeZ: IIUC I think `mnemonic` is a bad name here.

> 
> noun
> noun: mnemonic; plural noun: mnemonics
> 
>     1.
>     a system such as a pattern of letters, ideas, or associations which assists in remembering something.

src/workflow/password_enter.h
benma: I put it here because the docstring of password_enter says the caller needs to use `SET_PASSWORD_MAX_PASSWORD_LENGTH`, so this makes it available to the caller (without having to look for the source and including a second header).

good/bad? :)

src/workflow/password_enter.h
benma: removed

src/workflow/unlock.c
benma: true! let's do this in another PR as there are a lot of places to change.

src/workflow/password_enter.h
NickeZ: I don't need it for my workflow in vim. But if it floats your boat ;).

The include should be in the c-file though. Files should include what they use.

src/workflow/unlock.c
NickeZ: After reading bip39 I'm a little bit more OK with this name. The passphrase protects the mnemonic sentence.

src/workflow/password_enter.h
NickeZ: Since this is the current coding style, that it is acceptable to include files in the "matching" header file we will continue like this. The only drawback is that it is not possible to include the c-file (which could be necessary for unit testing).

PR #37, opened 2019-07-02
commander: set device name user confirmation

benma:


test/unit-test/commander/test_commander_set_mnemonic_passphrase_enabled.c
```diff
@@ -0,0 +1,79 @@
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-prototypes"
+#pragma GCC diagnostic ignored "-Wimplicit-function-declaration"
+
+bool __wrap_workflow_confirm(const char* title, const char* body, bool accept_only)
+{
+    check_expected(title);
+    check_expected(body);
+    check_expected(accept_only);
+    return mock();
```
NickeZ: You could use `mock_type(bool)` here I believe.

benma: TIL

PR #42, opened 2019-07-03
Py

benma:


NickeZ:
I think we should start using mypy in CI. WDYT?

```
mypy --ignore-missing-imports --strict py
```

benma:
> I think we should start using mypy in CI. WDYT?

Yes! I've been using it locally actually, but I did not know about the `--ignore-missing-imports` flag. Nice.

PR #43, opened 2019-07-16
py: fix typing issues and enforce mypy

benma:
mypy typechecks the code.

### DIFF
--- a/py/bitbox02/bitbox02.py
+++ b/py/bitbox02/bitbox02.py
@@ -21,7 +21,7 @@ import base64
 import binascii
 from datetime import datetime
 import hashlib
-import typing
+from typing import Optional, Callable, List, Dict, Tuple, Any, Generator
 
 import ecdsa
 from noise.connection import NoiseConnection, Keypair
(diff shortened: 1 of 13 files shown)


py/load_firmware.py
```diff
@@ -17,17 +17,18 @@
...
+import typing
 from time import sleep
 
 from bitbox02 import devices
 from bitbox02 import Bootloader, BitBox02
 
 
-def eprint(*args, **kwargs):
-    print(*args, **kwargs, file=sys.stderr)
+def eprint(msg: str) -> None:
```
NickeZ: The purpose of this function is to transparently wrap `print(*objects, sep=' ', end='\n', file=sys.stdout, flush=False)`. So I suggest that we either do `*args: Any, **kwargs: Any` or `print(*objects: object, sep: str = ' ', end: str = '\n', flush: boolean = False) -> None`

py/bitbox02/bitbox02.py
```diff
@@ -96,9 +98,9 @@ class BitBox02:
 
     def __init__(
         self,
-        device_info,
+        device_info: DeviceInfo,
         show_pairing_callback: typing.Callable[[str], None],
-        attestation_check_callback: typing.Callable[[bool], None] = None,
+        attestation_check_callback: typing.Optional[typing.Callable[[bool], None]] = None,
```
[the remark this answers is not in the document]
benma: Done

py/load_firmware.py
benma: Done

PR #47, opened 2019-07-22
.codecov: remove unneeded ignores

benma:
Seems like codecov only looks in src/ anyway.


benma:
See https://github.com/digitalbitbox/bitbox02-firmware/pull/31#discussion_r305953059

benma:
:man_shrugging:

PR #49, opened 2019-07-23
Cipher

benma:


src/cipher/cipher.h
```diff
@@ -23,15 +23,17 @@
  * @param[out] must have inlen + 64 bytes.
  * @param[inout] out_len must be inlen + 64 incoming, and will be the actual
  * length of the output.
+ * @param[in] secret 32 bytes secret
```
NickeZ: document the whole interface

src/cipher/cipher.h
```diff
@@ -23,15 +23,17 @@
...
     const uint8_t* in,
-    int in_len,
+    size_t in_len,
     uint8_t* out,
-    int* out_len,
+    size_t* out_len,
     const uint8_t* secret);
 
 /**
+ * @param[in] must have at least 64 bytes.
```
NickeZ: document the whole interface

src/cipher/cipher.h
benma: Done

src/cipher/cipher.h
benma: Done

PR #60, opened 2019-07-26
Blocking

benma:


src/u2f.c
```diff
@@ -279,20 +279,8 @@ static void _register(const USB_APDU* apdu, Packet* out_packet)
...
-
-    if (keystore_is_locked()) {
-        workflow_unlock();
-    }
-
-    // TODO: Remove this check when workflow_unlock resturns boolean
-    if (!memory_is_initialized()) {
+    if (!workflow_unlock()) {
         _error(U2F_SW_CONDITIONS_NOT_SATISFIED, out_packet);
-        return;
```
NickeZ: After every call to `_error()` there must be a `return` otherwise later stages can modify the `out_packet`.

PR #63, opened 2019-07-29
build system: reorder some code

NickeZ:


NickeZ:
just some preparation for big build system change

PR #69, opened 2019-08-05
Cleanup

benma:


benma:
@NickeZ this is a part of https://github.com/digitalbitbox/bitbox02-firmware/pull/70, split out for easier review

This is turning a callback into an event so both forms of confirmation (hold gesture, tap) can be handled the same way.

PR #70, opened 2019-08-05
Mnemonic

benma:


benma:
Can be tested via send_message.py

NickeZ:
The last commit (workflow/password: use trinary input ) looks a bit WIP with commented out lines and so on.

benma:
> The last commit (workflow/password: use trinary input ) looks a bit WIP with commented out lines and so on.

Done

PR #72, opened 2019-08-06
components/insert_sd_card: use button directly

benma:
Instead of an empty screen containing the button.


benma:
Fixes an issue introduced by https://github.com/digitalbitbox/bitbox02-firmware/commit/75db35ff35570c1afd1219f534acfc770a2cbd98. That change correctly changes which component to use, but I missed to update this instance.

PR #75, opened 2019-08-07
Makefile,Docker: small fixes

benma:


benma:
Not 100% sure but it seems nothing has changed: https://codecov.io/gh/digitalbitbox/bitbox02-firmware/pull/75

PR #79, opened 2019-08-08
u2f: Enable tests without hardware

NickeZ:
This PR hooks up the `hid_read`/`hid_write` to internal functions to simulate a connected device.

### DIFF
--- a/src/queue.h
+++ b/src/queue.h
@@ -30,11 +30,13 @@ uint8_t queue_push(const uint8_t* data);
 /**
  * Return the first data that was added to the queue.
  */
-uint8_t* queue_pull(void);
+const uint8_t* queue_pull(void);
 
 /**
  * Clear the queue.
  */
 void queue_clear(void);
+
+const uint8_t* queue_peek(void);
 
 #endif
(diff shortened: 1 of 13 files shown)


src/queue.h
```diff
@@ -37,4 +37,6 @@ uint8_t* queue_pull(void);
  */
 void queue_clear(void);
 
+uint8_t* queue_peek(void);
```
benma: can the return be made const?

test/unit-test/framework/mock_securechip.c
```diff
@@ -36,3 +38,19 @@ bool securechip_kdf(securechip_slot_t slot, const uint8_t* msg, size_t len, uint
...
+
+bool securechip_u2f_counter_inc(uint32_t* counter)
+{
+    *counter = _u2f_counter++;
+    return true;
+}
+
+bool securechip_ecc_unsafe_sign(const uint8_t* priv_key, const uint8_t* msg, uint8_t* sig)
+{
+    return false;
```
benma: do the tests not exercise signing?

NickeZ: not right now, I guess I could mock it somehow, but I can wait with that.

src/queue.h
NickeZ: fixed

PR #87, opened 2019-08-13
CI: Run tidy on modified files

NickeZ:


benma:
Hmm I am really not a fan of the approach of making the person who touches a file for an unrelated reason go through the process of fixing all linting issues in the file.

When I want to make a small change somewhere, I don't want to have a surprise of having to spend potentially a long time time on something like this. 

The better approach imho is to exclude files, and dedicate some time to fix linting issues in them, and then take them out of the blacklist so they are checked going forward,

NickeZ:
Yeah that is an unintended side effect of this approach.

The benefit of this approach is that we save some 20 minutes of every CI run because there is no need to test files which hasn't been touched.

NickeZ:
I'll fix all the tidy issues in this PR as well, to avoid the problem you mention.

NickeZ:
This PR now includes both enabling of tidy for modified files and fixes all tidy issues in the whole project. You can check by running `scripts/tidy`.

src/restore.c
```diff
@@ -249,10 +249,10 @@ restore_error_t restore_from_directory(const char* dir, Backup* backup, BackupDa
 restore_error_t restore_list_backups(ListBackupsResponse* backups)
 {
     sd_list_t list_subdirs __attribute__((__cleanup__(sd_free_list)));
+    backups->info_count = 0;
```
benma: I wonder what tidy was complaining about here?

NickeZ:
Uninitialized memory. If the error code is returned you are not allowed to
look at info_count.

src/screen.c
```diff
@@ -47,12 +47,14 @@ void screen_print_debug(const char* message, int duration)
...
+void screen_sprintf_debug(int duration, const char* fmt, ...)
 {
     va_list args;
-    va_start(args, message);
+    va_start(args, fmt);
     char print[100] = {0};
-    vsnprintf(print, sizeof(print) - 1, message, args);
+    // There is a bug in clang-tidy
+    // See https://bugs.llvm.org/show_bug.cgi?id=41311
+    vsnprintf(print, sizeof(print) - 1, fmt, args); // NOLINT
```
benma: >  The  functions  snprintf()  and  vsnprintf()  write  at most size bytes
>       (including the terminating null byte ('\0')) to str.

So actually the `-1` here is not needed?

src/screen.c
```diff
@@ -38,7 +38,7 @@ slider_location_t bottom_slider = 0;
 void screen_print_debug(const char* message, int duration)
 {
     char print[100];
-    strncpy(print, message, sizeof(print));
+    strncpy(print, message, sizeof(print) - 1);
     print[sizeof(print) - 1] = 0;
```
benma: could use `snprintf` to avoid manually handling the terminator?

src/workflow/restore_from_mnemonic.c
```diff
@@ -113,27 +113,28 @@ static bool _get_mnemonic(char* mnemonic_out)
...
-        strcat(mnemonic_out, word);
+        strncat(mnemonic_out, word, WORKFLOW_TRINARY_INPUT_MAX_WORD_LENGTH);
     }
     return true;
 }
 
 bool workflow_restore_from_mnemonic(void)
 {
-    // same as: (MAX_WORD_LENGTH-1) * MAX_WORDS + (MAX_WORDS-1) + 1
+    // same as: MAX_WORD_LENGTH * MAX_WORDS + MAX_WORDS - 1 + 1
```
benma: please keep the parens around `(MAX_WORDS - 1)` for better grouping with the next line. `-1 + 1` just looks silly :P

test/device-test/src/framework/test_common.c
```diff
@@ -318,7 +318,9 @@ static void read_and_print2screen(struct test_usb_metadata* metadata)
 
     int ret = read_single_frame((USB_FRAME*)metadata->usb_hid_out_report, &metadata->packet);
     if (ret == 0) {
-        screen_print_debug(data_to_string(&metadata->packet), 1000);
+        char* tmp = data_to_string(&metadata->packet);
```
benma: `data_to_string` could also return a static string, so using it keeps being a one-liner.

test/unit-test/framework/mock_hidapi.c
```diff
@@ -97,7 +97,10 @@ static void _hid_open(void)
     pthread_mutex_init(&mutex, NULL);
 }
 
-hid_device* hid_open(unsigned short vid, unsigned short pid, const wchar_t* path)
+hid_device* hid_open(
```
benma: looks unused to me -> delete whole function?

src/restore.c
benma: > Uninitialized memory. If the error code is returned you are not allowed to
look at info_count.

But that is okay? One should not inspect the out param at all in case of an error and treat it as undefined.

NickeZ: Right, another way to fix it, which probably is cleaner is to initialize the struct in the caller before actually calling `restore_list_backups`.

src/screen.c
NickeZ: ack

src/screen.c
NickeZ: ack!

src/workflow/restore_from_mnemonic.c
NickeZ: ha, you commented on the comment but I think there is a bug in the next chunk. 

```
(WORKFLOW_TRINARY_INPUT_MAX_WORD_LENGTH + 1) * WORKFLOW_RESTORE_FROM_MNEMONIC_MAX_WORDS + 1
```

should be

```
(WORKFLOW_TRINARY_INPUT_MAX_WORD_LENGTH + 1) * WORKFLOW_RESTORE_FROM_MNEMONIC_MAX_WORDS
```

right?

test/device-test/src/framework/test_common.c
NickeZ: ack

test/unit-test/framework/mock_hidapi.c
NickeZ: ack

src/workflow/restore_from_mnemonic.c
NickeZ: @benma After you have double checked this I'll merge this branch

test/unit-test/test_restore.c
```diff
@@ -241,7 +241,7 @@ static void test_restore_list_backups_single_seed(void** state)
  */
 static void test_restore_list_backups_multiple_seeds(void** state)
 {
-    ListBackupsResponse list_backups_response;
+    ListBackupsResponse list_backups_response = {0};
```
benma: i see. the alternative (maybe more correct way) would be to not perform the follow up tests which inspect this struct in case it is not RESTORE_OK.

For my info, is tidy intelligent enough to not complain if you only access the struct in case of RESTORE_OK?

If so, it's a pity that cmocka's assert does not help tidy here.

src/workflow/restore_from_mnemonic.c
benma: `(WORKFLOW_TRINARY_INPUT_MAX_WORD_LENGTH + 1) * WORKFLOW_RESTORE_FROM_MNEMONIC_MAX_WORDS` looks correct to me. can't  recall seeing anything else before, but the previous commit is gone now.

benma:
ACK

test/unit-test/test_restore.c
NickeZ: No I tried that and tidy still thinks that there can be an access of uninit memory.

PR #90, opened 2019-08-14
Py: Add semver to requirements file

TheCharlatan:


benma:
Thanks, but it's already there, one folder down:

https://github.com/digitalbitbox/bitbox02-firmware/blob/master/py/bitbox02/requirements.txt#L5

PR #91, opened 2019-08-15
Create tracing/logging framework.

NickeZ:
This PR introduces two things:

* [x] Debugging with semihosting
* [x] ~Retarget stdout/stderr to the screen.~ (future feature)
* [x] Introduces `trace(args...)`
 
# Debugging with semihosting

The detailed instructions are in `/scripts/jlink-gdb-debug/README.md`. Essentially this allows you to add regular `printf`s and they are printed on a screen on the computer running the GDB Server.

Semihosting also allows us to open files and dump binary data with normal `fopen`/`fwrite` libc functions.

# Retarget stdout/stderr

`printf` uses a function called `_write` to actually write the resulting bytes. This PR implements it by printing the bytes to the screen. The default implementation of `_write` is a no-op and does nothing. 

# `trace(args...)`
`trace` is a macro that expands to printf in a semihosted build and expands to no-op in a non-semihosted build. This allows us to leave some debugging functionality in the firmware and reduce the churn while implementing new features.

# Test

You can test it with a device test firmware:

```
make -C build fw_test_trace.elf
JLinkExe -if SWD -device ATSAMD51J20 -speed 4000 -autoconnect 1 -CommanderScript ./build/test/device-test/fw_test_trace.jlink
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,11 @@ endif()
 
 project(bitbox02 C)
 
+# nosys is set in arm.cmake so that `project(c)` above works. Remove it since it interferes with compile options
+if(CMAKE_CROSSCOMPILING)
+  string(REPLACE "--specs=nosys.specs" "" CMAKE_EXE_LINKER_FLAGS ${CMAKE_EXE_LINKER_FLAGS})
+endif()
+
 # Where to find custom cmake modules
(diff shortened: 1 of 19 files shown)


benma:
utACK; I checked that the production bootloader/firmware is not changed by this commit.

PR #95, opened 2019-08-20
.gitmodules: moved cryptoauthlib repo

benma:


NickeZ:
nACK for now, add length to function signature.

benma:
discussed offline, proceeding like this for now, as upstream does not accept PRs directly; so we will just wait and see how upstream fixes it.

PR #97, opened 2019-08-20
Workflow ifs

benma:


benma:
It should be okay, I tested it.

PR #100, opened 2019-08-21
securechip: better split between factory setup and normal init

benma:


src/securechip/securechip.c
```diff
@@ -309,20 +318,70 @@ bool securechip_setup(securechip_interface_functions_t* ifs)
...
-        if (atcab_cmp_config_zone(_configuration, &same_config) != ATCA_SUCCESS) {
-            return false;
-        }
-        if (!same_config) {
-            return false;
-        }
+static bool _verify_config(void)
+{
+    bool is_config_locked = false;
+    ATCA_STATUS result = atcab_is_locked(LOCK_ZONE_CONFIG, &is_config_locked);
```
NickeZ: This could be a `_zone_is_locked` function to be consistent with `_slot_is_locked`. But I don't feel super strongly about it.

PR #102, opened 2019-08-22
Create btc-only firmware target

NickeZ:
This is the initial PR to create the btc only firmware.

This creates the new targets in CMake and ifdefs out all functionality except BTC.

It also creates a new iProduct name for the btc only firmware called `BitBox02BTC`.

Next step will be to create an interface for the app to get the supported apps.


src/commander/commander.c
```diff
@@ -264,10 +264,12 @@ static commander_error_t _api_process(const Request* request, Response* response
     case Request_btc_pub_tag:
         response->which_response = Response_pub_tag;
         return commander_btc_pub(&(request->request.btc_pub), &(response->response.pub));
+#if defined(APP_BTC)
```
benma: move one case up?

should be `defined(APP_BTC) || defined(APP_LTC)`? Otherwise `APP_LTC` without `APP_BTC` does not work. Then more granular filtering is needed in the relevant functions.

benma: What about ifdefing the relevant includes at the top of the file too?

NickeZ: fixed both

Makefile
```diff
@@ -35,6 +35,9 @@ firmware: | build
 firmware-semihosting: | build
 	$(MAKE) -C py
 	$(MAKE) -C build firmware-semihosting.elf
+firmware-btc: | build
```
benma: please also add `jlink-flash-firmware-btc`

NickeZ: ack

src/commander/commander_btc.c
```diff
@@ -30,6 +30,9 @@ static const char* _coin_tltc = "LTC Testnet";
 
 commander_error_t commander_btc_pub(const BTCPubRequest* request, PubResponse* response)
 {
+    if (!app_btc_enabled(request->coin)) {
+        return COMMANDER_ERR_DISABLED;
```
benma: this line is now unreachable, since you already check in the main commander function.

Imho we should have those checks either all in the main commander func, or in the individual functions, but not both.

src/commander/commander_btc.c
```diff
@@ -101,6 +104,9 @@ commander_error_t commander_btc_sign(const Request* request, Response* response)
     app_btc_sign_error_t result;
     switch (request->which_request) {
     case Request_btc_sign_init_tag:
+        if (!app_btc_enabled(request->request.btc_sign_init.coin)) {
+            return COMMANDER_ERR_DISABLED;
```
benma: same here I think.

NickeZ:
I verified with `nm` that all `u2f` and `eth` functions are removed.

PR #103, opened 2019-08-26
py/bitbox02: add btc_sign()

benma:


py/bitbox02/bitbox02.py
```diff
@@ -360,6 +395,110 @@ def btc_pub(
...
         return self._msg_query(request).pub.pub
 
+    # pylint: disable=too-many-arguments
+    def btc_sign(
+        self,
+        coin: hww.BTCCoin,
+        script_type: hww.BTCScriptType,
+        bip44_account: int,
+        num_inputs: int,
+        get_input: Callable[[int], BTCInputType],
```
achow101: Why not just pass in a list of BTCInputType (and BTCOutputType later) rather than using a callback?

benma: I figured then there is no preprocessing needed to transform from source format to target format.

Happy to change it be a list instead next week.

benma: @achow101 updated the function now to take lists for inputs/outputs instead

PR #106, opened 2019-08-26
Restore u2f counter

NickeZ:
Restore u2f counter to current unix timestamp. Implemented both for SD backup restore and mnemonic phrase.

@benma, could you specifically look at if this is the correct places to ask the question about the current date? Another question is, should we abort the whole thing if the date is wrong or should we simply skip resetting the U2F counter?


messages/hww.proto
```diff
@@ -129,7 +131,7 @@ message PubResponse {
 message BackupInfo {
   string id = 1;
   uint32 timestamp = 2;
-  uint32 timezone_offset = 3;
+  int32 timezone_offset = 3;
```
benma: get rid of it somehow :P

NickeZ: I got rid of it. I kept the tag numbers to increase forward/backwards compatibility.

py/bitbox02/bitbox02.py
```diff
@@ -546,6 +548,8 @@ def restore_from_mnemonic(self) -> bool:
         request = hww.Request()
         # pylint: disable=no-member
         request.restore_from_mnemonic.CopyFrom(hww.RestoreFromMnemonicRequest())
```
benma: i think this line is not needed anymore since you set the vars below.

Alternatively, you can set the fields as args to `hww.RestoreFromMnemonicRequest(timestamp=...)`.

src/workflow/restore.c
```diff
@@ -47,6 +49,10 @@ bool workflow_restore_backup(const RestoreBackupRequest* restore_request)
     if (!memory_set_initialized()) {
         return false;
     }
+    if (!workflow_confirm_time(restore_request->timestamp, restore_request->timezone_offset)) {
+        return false;
+    }
+    securechip_u2f_counter_set(restore_request->timestamp);
```
benma: error not checked (can also be explicit ignore).

maybe add USE_RESULT to securechip.h

src/workflow/restore.c
```diff
@@ -47,6 +49,10 @@ bool workflow_restore_backup(const RestoreBackupRequest* restore_request)
     if (!memory_set_initialized()) {
         return false;
     }
+    if (!workflow_confirm_time(restore_request->timestamp, restore_request->timezone_offset)) {
+        return false;
```
benma: this and the following (counter_set) should be before `memory_set_initialized`, otherwise you are in some semi/undefined state for U2F.

src/workflow/restore_from_mnemonic.c
```diff
@@ -170,6 +172,10 @@ bool workflow_restore_from_mnemonic(void)
     if (!memory_set_initialized()) {
         return false;
     }
+    if (!workflow_confirm_time(request->timestamp, request->timezone_offset)) {
```
benma: same comments as for restore from sd.

src/workflow/confirm_time.c
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "confirm_time.h"
+#include <time.h>
+#include <workflow/confirm.h>
+
+bool workflow_confirm_time(uint32_t timestamp, int32_t timezone_offset)
```
benma: this should be named `workflow_confirm_date`, as the time is not confirmed.

For the backup, this is okay, but for U2F, you could run into issues if you restore on the same day as when you register on a u2f site, but the host machine time is too early (but same day) :nerd_face: 

Not sure if this edge case is worth handling :man_shrugging: I guess not (and the edge case is fixable by re-doing it a day later).

benma:
We are now repeating the timestamp/timezone protobuf fields and checking code twice.

It might be less and easier code to make a new protobuf api call: `U2FResetCounter(timestamp, timezone)` which the host app can call after either restore operation. Would also be bit more modular in case we ever want to change how we handle the counter.

What do you think?

NickeZ:
> We are now repeating the timestamp/timezone protobuf fields and checking code twice.
> 
> It might be less and easier code to make a new protobuf api call: `U2FResetCounter(timestamp, timezone)` which the host app can call after either restore operation. Would also be bit more modular in case we ever want to change how we handle the counter.
> 
> What do you think?

If we want to allow the u2f to be restored independent of the seed then sure. But if we always want the App to call this endpoint after the restore endpoint I think that would be bad design because the interface would contain state.

src/workflow/restore.c
NickeZ: ack

src/workflow/restore.c
NickeZ: ack

src/workflow/restore_from_mnemonic.c
NickeZ: ack

benma:
> But if we always want the App to call this endpoint after the restore endpoint I think that would be bad design because the interface would contain state.

:+1: then let's keep it like this

src/workflow/confirm_time.c
benma: Maybe we should make this case / assumption explicit in comments in the code, so it's clear this was deliberate and not an accident.

py/bitbox02/bitbox02.py
NickeZ: ack

src/workflow/confirm_time.c
NickeZ: I added support for checking the actual time. Should we do that for u2f?

benma: Not sure tbh. I'd be a bit worried that for some people the time might be off by a few hours, which would cause some anxiety when trying to restore. I did have this problem in a previous project.

Maybe let's go with time anyway and see what happens, as it would be more correct.

src/workflow/restore.c
```diff
@@ -44,6 +46,13 @@ bool workflow_restore_backup(const RestoreBackupRequest* restore_request)
         workflow_status_create("Could not\nrestore backup", false);
         return false;
     }
+    if (!workflow_confirm_time(
+            restore_request->timestamp, restore_request->timezone_offset, true)) {
+        return false;
+    }
+    if (!securechip_u2f_counter_set(restore_request->timestamp)) {
```
benma: in securechip.c, disable the two u2f functions for btc-only

src/workflow/restore_from_mnemonic.c
```diff
@@ -167,6 +169,12 @@ bool workflow_restore_from_mnemonic(void)
         workflow_status_create("Could not\nrestore backup", false);
         return false;
     }
+    if (!workflow_confirm_time(request->timestamp, request->timezone_offset, true)) {
```
benma: same

src/workflow/restore.c
```diff
@@ -44,6 +46,13 @@ bool workflow_restore_backup(const RestoreBackupRequest* restore_request)
         workflow_status_create("Could not\nrestore backup", false);
         return false;
     }
+    if (!workflow_confirm_time(
```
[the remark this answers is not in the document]
NickeZ: good point! ack

src/workflow/restore.c
NickeZ: ack

src/workflow/restore_from_mnemonic.c
NickeZ: ack

src/workflow/confirm_time.c
NickeZ: ack

src/securechip/securechip.c
```diff
@@ -688,15 +688,16 @@ static ATCA_STATUS _write_data_slot_block(uint8_t* bytes, uint16_t slot, uint8_t
     return atcab_write_enc(slot, block, bytes, encryption_key, SECURECHIP_SLOT_ENCRYPTION_KEY);
 }
 
-bool securechip_u2f_counter_set(uint32_t value)
+// Set is needed for factory setup
+bool securechip_u2f_counter_set(uint32_t counter)
```
benma: so `#if defined(APP_U2F) || defined(FACTORYSETUP)`?

NickeZ: yeah I suppose..

PR #111, opened 2019-09-04
u2f: reduce time to response

NickeZ:
with our current firmware architecture and the fact that windows
requires a response within 2-3s we need to reduce the time for the user
to respond to less than 1s.

This timeout was empirically found using windows 1903.

### DIFF
--- a/src/u2f/u2f_app.c
+++ b/src/u2f/u2f_app.c
@@ -1,5 +1,6 @@
 #include "u2f_app.h"
 
+#include <hardfault.h>
 #include <util.h>
 #include <workflow/confirm.h>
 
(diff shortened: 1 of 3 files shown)


src/u2f/u2f_app.c
```diff
@@ -42,12 +42,25 @@ static void _app_string(const uint8_t* app_id, char* out, size_t out_len)
...
+        _app_string(app_id, app_string, sizeof(app_string));
+        break;
+    case U2F_APP_BOGUS:
+        title = "";
+        snprintf(app_string, sizeof(app_string), "%s", "Use U2F?");
+        break;
+    default:
+        return false;
+    }
+    return workflow_confirm_with_timeout(title, app_string, false, 75);
```
benma: maybe a comment here that it should be ~1s, as it's important, and the "framerate" might change over time.

src/u2f/u2f_app.c
```diff
@@ -42,12 +42,25 @@ static void _app_string(const uint8_t* app_id, char* out, size_t out_len)
...
+    case U2F_APP_AUTHENTICATE:
+        title = "U2F Authenticate";
+        _app_string(app_id, app_string, sizeof(app_string));
+        break;
+    case U2F_APP_BOGUS:
+        title = "";
+        snprintf(app_string, sizeof(app_string), "%s", "Use U2F?");
+        break;
+    default:
+        return false;
```
benma: maybe `Abort()` for unreachable lines.

src/u2f/u2f_app.c
NickeZ: ack, added a comment

src/u2f/u2f_app.c
NickeZ: ack

PR #114, opened 2019-09-09
build-system: Add BTC only tag

NickeZ:
Tag BTC only edition with `firmware-btc/vX.Y.Z`


CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ endif()
 # Versions MUST contain three parts and start with lowercase 'v'.
 # Example 'v1.0.0'. They MUST not contain a pre-release label such as '-beta'.
 set(FIRMWARE_VERSION "v4.0.0")
+set(FIRMWARE_BTC_VERSION "v6.0.0")
```
benma: Should be v4.0.0, we can diverge later, but for the initial version should be the same so all previous version checks work.

CMakeLists.txt
```diff
@@ -137,6 +148,7 @@ if(GIT_FOUND AND PYTHONINTERP_FOUND)
   endif()
 else()
   set(GIT_FIRMWARE_VERSION_STRING "v0.0.0")
+  set(GIT_FIRMWARE_BTC_VERSION_STRING "v0.0.0")
```
benma: For consistency, I'd prefer `BTCONLY` everywhere instead of `BTC`.

CMakeLists.txt
NickeZ: oh yeah, I was just testing, needed it to be different than `4.0.0`. I forgot to set it to 4 again.

CMakeLists.txt
NickeZ: for consistency I named it `BTC_ONLY`.

CMakeLists.txt
```diff
@@ -104,6 +105,16 @@ if(GIT_FOUND AND PYTHONINTERP_FOUND)
   if(NOT exit_code EQUAL "0")
     set(GIT_FIRMWARE_VERSION_STRING "v0.0.0")
   endif()
+  execute_process(
+    COMMAND ${PYTHON_EXECUTABLE} ./scripts/get_version firmware-btc --check-semver --check-gpg
```
benma: also here? `btc` -> `btc_only` or `btc-only`.

NickeZ: ok, no strong opinion from me.

PR #116, opened 2019-09-10
cipher: Add some upper bound checks on stack allocated variables

NickeZ:


NickeZ:
tested manually by creating and restoring from backup, unlocking the device and so on.

PR #117, opened 2019-09-10
Update FatFs

NickeZ:
Added the 5 released patches to the FatFs dependency.


NickeZ:
Tested it manually, it works!

PR #119, opened 2019-09-11
factorysetup: add attestation keys

benma:


benma:
bitbox app side: https://github.com/digitalbitbox/bitbox-wallet-app/pull/513

PR #121, opened 2019-09-13
py: recognize btconly firmware

benma:


py/bitbox02/devices.py
```diff
@@ -24,6 +24,7 @@
 BB02_BOOTLOADER = "bb02-bootloader"
 BB02BTC_BOOTLOADER = "bb02btc-bootloader"
 BITBOX02 = "BitBox02"
+BITBOX02BTC = "BitBox02BTC"
```
NickeZ: I think we should avoid exposing these variables as our external API and instead have one function per device type.

py/load_firmware.py
```diff
@@ -50,7 +50,9 @@ def main() -> int:
 
     bootloaders = devices.get_bitbox02_devices(devices.BB02_BOOTLOADER)
     bootloaders.extend(devices.get_bitbox02_devices(devices.BB02BTC_BOOTLOADER))
-    bitboxes = devices.get_bitbox02_devices()
+    bitboxes = devices.get_bitbox02_devices(devices.BITBOX02) + devices.get_bitbox02_devices(
+        devices.BITBOX02BTC
+    )
```
NickeZ: this isn't needed if we merge my PR first.

py/send_message.py
```diff
@@ -308,7 +309,9 @@ def main() -> int:
     parser.add_argument("--debug", action="store_true", help="Print messages sent and received")
     args = parser.parse_args()
 
-    bitboxes = bitbox02.get_bitbox02_devices()
+    bitboxes = bitbox02.get_bitbox02_devices(devices.BITBOX02) + bitbox02.get_bitbox02_devices(
+        devices.BITBOX02BTC
+    )
```
NickeZ: Same here

PR #126, opened 2019-09-23
Api doc

NickeZ:
I'm already thinking if I should group a few together again:

* commands/system.proto
  * DeviceInfo
  * SetDeviceName
  * SetDeviceLanguage
  * SetPassword
  * Reset
  * Reboot

See example here (don't know how long it is up): https://pste.eu/p/x13U.html


messages/commands/reboot.proto
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+message RebootRequest {
```
benma: Not sure I like this level of file splitting. Could for example put all the backup related messages in one file, the other "platform" messages in one file, etc.

messages/commands/eth/eth.proto
```diff
@@ -0,0 +1,68 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+// TODO: untangle this mess...
```
benma: please be specific

.ci/travis-ci
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-CONTAINER=shiftcrypto/firmware_v2:5
+CONTAINER=shiftcrypto/firmware_v2:6
```
benma: I think this has to be rebased, next is :8

messages/commands/reboot.proto
NickeZ: ack

messages/commands/eth/eth.proto
NickeZ: ack

NickeZ:
OK, I reorganized it in fewer files. You can check the generated docs with `make doc` in `build` directory. The result is in `build/pbdocs`.

messages/commands/eth.proto
```diff
@@ -0,0 +1,68 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+// TODO(nc): This should go into the main proto file in a future refactor
+message PubResponse {
```
benma: hmm what about putting inti common.proto and including it here and in the file with the btc messages?

NickeZ: Interesting suggestion. I'll try it.

benma: i'll have one more of those with the nonce covert channel fix, which will be used for in btc and eth, so it would be nice if it worked :)

NickeZ: It worked perfectly

NickeZ:
moved pubrequest, error and success over to a `common.proto` file.

PR #127, opened 2019-09-24
Truncatevalues

benma:


src/apps/eth/eth_common.c
```diff
@@ -89,10 +89,21 @@ void eth_common_format_amount(
...
+    const size_t truncate_len = 13;
+
+    if (out == NULL || out_len < min_out_len || strlen(unit) > 10) {
         Abort("eth_common_format_amount");
     }
     char unit_with_space[strlen(unit) + 2];
     snprintf(unit_with_space, sizeof(unit_with_space), " %s", unit);
-    bn_format(scalar, "", unit_with_space, decimals, 0, false, out, out_len);
+    bn_format(scalar, "", "", decimals, 0, false, out, out_len);
+    if (strlen(out) > truncate_len) {
```
NickeZ: Could use a local variable for `strlen(out)` to avoid recalculating the length three times.

PR #128, opened 2019-09-24
python fixes

NickeZ:
This is a pretty big reorganization of the python stuff. Tested everything with bb02 and bb02btc firmware.

* Reorganize
* Merge similar code into one code base
* Package bitbox02 module
* Add support for btc-only
* Add support for bootloader (boot command)


py/load_firmware.py
```diff
@@ -48,43 +66,58 @@ def main() -> int:
...
-        bitbox = BitBox02(device_info=bitboxes[0], show_pairing_callback=show_pairing)
+        bitbox = BitBox02(device_info=bitbox, show_pairing_callback=_show_pairing)
         bitbox.reboot()
-        bootloaders = _wait_for_bootloaders()
 
-    if len(bootloaders) > 1:
-        eprint("Multiple bootloaders detected. Only one supported")
-        return 1
+        # wait for it to reboot
+        while True:
```
benma: if you have multiple it might take the wrong one

py/load_firmware.py
```diff
@@ -48,43 +66,58 @@ def main() -> int:
...
+
+    if bootloader is None:
+        try:
+            bootloader = devices.get_bitbox02btc_bootloader()
+        except TooManyFoundException:
+            eprint("Found multiple bb02 bootloader btc editions. Only one supported.")
+
+    if bootloader is None:
+        bitbox = None
+        try:
```
benma: a bit strange semantics, it will work if one device is plugged in, but also if 2 + 1 are plugged in.

What about having a function which returns all devices of all editions? The distinction is usually not important, and if it is we can have a `@property edition` to query for it per device/bootloader.

benma:
It would be much easier to review if this many commits with a description per commit.

With big PRs I tend to not look very carefully as it's a bit overwhelming. Manual testing is much harder because I don't know where to focus.

Here even multiple PRs would have helped, as some seems unrelated (like the packaging).

NickeZ:
> It would be much easier to review if this many commits with a description per commit.
> 
> With big PRs I tend to not look very carefully as it's a bit overwhelming. Manual testing is much harder because I don't know where to focus.
> 
> Here even multiple PRs would have helped, as some seems unrelated (like the packaging).

The goal was to do the packaging but when I started to shave that jak I ended up modifying much more than expected. At least I split out the packaging specific stuff in a separate commit.

py/load_firmware.py
NickeZ: ack

py/load_firmware.py
NickeZ: ack

NickeZ:
There is still some work to be done here. mypy linter currently fails.

NickeZ:
Fixed the linter issues on the flight yesterday and I tested send_message and load_firmware. I think this is ready to be merged.

scripts/lint-python
```diff
@@ -28,7 +28,7 @@ fi
 # behavior).
 # We must typecheck the whole `py` directory even if only a few files are modified so that mypy
 # sees all types
-${MYPY} --implicit-reexport --namespace-packages --ignore-missing-imports --strict py
+${MYPY} --implicit-reexport --namespace-packages --ignore-missing-imports --strict py py/bitbox02
```
benma: why is this needed? doesn't it do `py` and all imported dependencies?

py/bitbox02/bitbox02/bitbox02.py
```diff
@@ -269,6 +272,9 @@ def _msg_query(
             print(response)
         return response
 
+    def u2f_query(self, msg: bytes) -> bytes:
```
benma: What's the purpose of this? Seems like it ca be deleted as it's u2f, and empty body.

py/bitbox02/bitbox02/devices.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    devices = get_any_bitboxes()
+    if len(devices) > 1:
+        raise TooManyFoundException(len(devices))
+    if not devices:
+        raise NoneFoundException()
+    return devices[0]
+
+
+def get_any_bitbox_bootloaders() -> List[DeviceInfo]:
```
benma: `bitbox02`

py/bitbox02/bitbox02/devices.py
```diff
@@ -0,0 +1,165 @@
...
+    Searches for both btc-only and non-btc-only devices
+    Returns:
+        List of devices
+    """
+    devices = get_bitbox02_bootloaders()
+    devices.extend(get_bitbox02btc_bootloaders())
+    return devices
+
+
+def get_any_bitbox_bootloader() -> DeviceInfo:
```
benma: same, `bitbox02`

py/bitbox02/setup.py
```diff
@@ -0,0 +1,39 @@
...
+    long_description_content_type="text/markdown",
+    url="https://github.com/digitalbitbox/bitbox02-firmware",
+    python_requires=">=3.6",
+    classifiers=[
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python :: 3.6",
+    ],
+    keywords="digitalbitbox bitbox bitbox02 bitcoin litecoin ethereum erc20 u2f",
+    install_requires=["hidapi", "noiseprotocol", "protobuf", "ecdsa", "semver"],
```
benma: this is missing version bounds like in requirements.py. I think the newest protobuf does not work yet with our code, for example.

Speaking of, is `requirements.py` still needed?

scripts/lint-python
NickeZ: It's needed because py/bitbox02 is a package now. So the imports in for example `send_message.py` requires you to actually install bitbox02 before you can use it.

NickeZ:
> I think the linter is not working properly. I can't run `./py/send_message.py`:
> 
> ```
> Traceback (most recent call last):
>   File "./py/send_message.py", line 25, in <module>
>     from bitbox02 import HARDENED
> ImportError: cannot import name 'HARDENED'
> ```
> 
> Running `./scripts/lint-python py/send_message.py` gives a lot of similar issues.

You need to install bitbox02. Run `python3 -m pip install -e .` in the `py/bitbox02` directory.

py/bitbox02/setup.py
NickeZ: Oh, I didn't know that you could put versions in the `setup.py` nowadays.

After reading `https://packaging.python.org/discussions/install-requires-vs-requirements/` it seems like we can remove the requirements files yes.

py/bitbox02/bitbox02/bitbox02.py
NickeZ: ack

py/bitbox02/bitbox02/devices.py
NickeZ: ack

NickeZ:
Force pushed a new version

PR #131, opened 2019-09-25
refactoring

NickeZ:
moving some code around, should reduce build time a lot.


external/CMakeLists.txt
```diff
@@ -113,5 +113,11 @@ add_library(secp256k1 STATIC IMPORTED GLOBAL)
 set_property(TARGET secp256k1
   PROPERTY IMPORTED_LOCATION ${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}/libsecp256k1.a)
 
-add_library(CMSIS INTERFACE)
-target_include_directories(CMSIS INTERFACE CMSIS/Include)
+if(CMAKE_CROSSCOMPILING)
+  add_library(CMSIS INTERFACE)
```
benma: for my information, what does this do? why is there a library target with no c files?

src/securechip/securechip.c
```diff
@@ -692,14 +692,14 @@ static ATCA_STATUS _write_data_slot_block(uint8_t* bytes, uint16_t slot, uint8_t
 bool securechip_u2f_counter_set(uint32_t counter)
 {
     data_9_0_t data = {0};
-    ATCA_STATUS result = _read_data_slot_block(&data.bytes, SECURECHIP_SLOT_DATA0, 0);
+    ATCA_STATUS result = _read_data_slot_block(&data.bytes[0], SECURECHIP_SLOT_DATA0, 0);
```
benma: seems unrelated, can it be in a different PR? then there is less stuff to test per PR ;)

which linter complained? ;)

external/CMakeLists.txt
```diff
@@ -54,6 +54,16 @@ ExternalProject_Add(noise-c
 add_library(noiseprotocol STATIC IMPORTED GLOBAL)
 set_property(TARGET noiseprotocol
   PROPERTY IMPORTED_LOCATION ${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}/libnoiseprotocol.a)
+set_target_properties(noiseprotocol
+  PROPERTIES
+    INTERFACE_INCLUDE_DIRECTORIES
```
benma: TIL!

Though still not sure how it works, are those automatically added as include headers by add_dependency? 

https://cmake.org/cmake/help/v3.15/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html is very cryptic in how this works.

external/CMakeLists.txt
```diff
@@ -54,6 +54,16 @@ ExternalProject_Add(noise-c
 add_library(noiseprotocol STATIC IMPORTED GLOBAL)
 set_property(TARGET noiseprotocol
   PROPERTY IMPORTED_LOCATION ${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}/libnoiseprotocol.a)
+set_target_properties(noiseprotocol
```
benma: isn't this the same as target_include_direcetories with the INTERFACE keyword?

https://stackoverflow.com/questions/26243169/cmake-target-include-directories-meaning-of-scope

benma:
This looks like a kick-ass improvement!

external/CMakeLists.txt
NickeZ: Those are the headers provided by ARM for cortex-m4 development. Full description is here: https://www.arm.com/why-arm/technologies/cmsis

The point of creating an interface-only library is just build-system cleanliness. Separate dependencies should show up as different entities in the graph. Makes it easier to reason about.

src/securechip/securechip.c
NickeZ: The normal compiler complained. For some reason before the refactoring `-Wno-incompatible-pointer-types` got applied to our files even though I think the purpose was to apply it to cryptoauthlib.

external/CMakeLists.txt
NickeZ: heh, it is even more weird they are added as `include_directories` because of `target_link_libraries`. `add_dependency` only forces those targets to be built before this target is built. `add_dependency` is a hack becuase the dependencies aren't CMake projects.

You can read more about it here: https://cmake.org/cmake/help/v3.10/command/target_include_directories.html

external/CMakeLists.txt
NickeZ: yeah but that isn't allowed for imported libraries. https://cmake.org/cmake/help/v3.10/command/target_include_directories.html

> The named <target> must have been created by a command such as add_executable() or add_library() and must not be an IMPORTED target.

NickeZ:
> This looks like a kick-ass improvement!

yeah! It significantly speeds up compilation time.

src/securechip/securechip.c
NickeZ: > seems unrelated, can it be in a different PR? then there is less stuff to test per PR ;)

Then this PR will never pass CI. So I don't think so.

benma: technically it would if the other pr would be merged first :P

but i see now, so it's ok, i will take care to test it :)

src/usb/class/hid/hid.c
```diff
@@ -43,11 +43,11 @@ static int32_t _enable(struct usbd_descriptors* desc, struct usbdf_driver* drv)
     if (HID_CLASS == ifc_desc.bInterfaceClass) {
         if (func_data->func_iface == ifc_desc.bInterfaceNumber) { // Initialized
             return ERR_ALREADY_INITIALIZED;
-        } else if (func_data->func_iface != 0xFF) { // Occupied
```
benma: just wondering, this code is not by us, right? so maybe we shouldn't spend time on stylistic changes to this.

mainly asking because it's slightly hard to review.

speaking of: should we expect the compiled binary to stay exactly the same after this PR? that would make me more comfortable that everything is still working.

NickeZ: This is written by us, otherwise I would've put it in the `external` folder.

Very unlikely. getting the order of the binaries the same might not even be possible. But the binary size should stay the same. So I'm pretty confident it is the same result if that happens. I don't really see what can go wrong anyway, if the thing compiles and runs I can't see how it would be broken.

NickeZ: > This is written by us, otherwise I would've put it in the external folder.

This is also partly why I badly want this reorganization. It is hard as a newcomer to understand what is our code and not when it is all mixed up.

external/asf4-drivers/driver_init.c
```diff
@@ -281,21 +266,44 @@ void system_init(void)
     gpio_set_pin_direction(PIN_SD_PWON, GPIO_DIRECTION_OUT);
     gpio_set_pin_level(PIN_SD_PWON, PIN_HIGH);
     gpio_set_pin_function(PIN_SD_PWON, GPIO_PIN_FUNCTION_OFF);
+}
 
+void system_init(void)
```
benma: maybe name this `firmware_init()` (as opposed to `bootloader_init()` below)?

src/usb/class/hid/hid.c
```diff
@@ -186,36 +186,34 @@ int32_t hid_req(
...
+        switch (req->bRequest) {
+        case 0x03: /* Get Protocol */
+            return usbdc_xfer(ep, &func_data->protocol, 1, 0);
+        case 0x0B: /* Set Protocol */
+            func_data->protocol = req->V.wValue;
+            return usbdc_xfer(ep, NULL, 0, 0);
+        case USB_REQ_HID_SET_REPORT:
+            if (USB_SETUP_STAGE == stage) {
+                return usbdc_xfer(ep, ctrl_buf, len, false);
             } else {
```
benma: how did this `else` survive the linter? :laughing:

NickeZ:
File sizes actually increased a little bit. I would like to look into that (probably with diffoscope) and see if it is to be expected or if something important has changed.

```
master:
-rwxr-xr-x 1 dockeruser dockergroup   50548 Oct  2 15:29 build/bin/bootloader.bin
-rwxr-xr-x 1 dockeruser dockergroup  820716 Oct  2 15:29 build/bin/bootloader.elf
-rw-r--r-- 1 dockeruser dockergroup  300176 Oct  2 15:29 build/bin/bootloader.map
-rwxr-xr-x 1 dockeruser dockergroup  225460 Oct  2 15:31 build/bin/factory-setup.bin
-rwxr-xr-x 1 dockeruser dockergroup 2527212 Oct  2 15:31 build/bin/factory-setup.elf
-rw-r--r-- 1 dockeruser dockergroup  934832 Oct  2 15:31 build/bin/factory-setup.map
-rwxr-xr-x 1 dockeruser dockergroup  405968 Oct  2 15:32 build/bin/firmware-btc.bin
-rwxr-xr-x 1 dockeruser dockergroup 4344652 Oct  2 15:32 build/bin/firmware-btc.elf
-rw-r--r-- 1 dockeruser dockergroup  983119 Oct  2 15:32 build/bin/firmware-btc.map
-rwxr-xr-x 1 dockeruser dockergroup  422336 Oct  2 15:34 build/bin/firmware.bin
-rwxr-xr-x 1 dockeruser dockergroup 4616812 Oct  2 15:34 build/bin/firmware.elf
-rw-r--r-- 1 dockeruser dockergroup  977411 Oct  2 15:34 build/bin/firmware.map



move-drivers:
-rwxr-xr-x 1 dockeruser dockergroup   50876 Oct  2 15:36 build/bin/bootloader.bin
-rwxr-xr-x 1 dockeruser dockergroup  839168 Oct  2 15:36 build/bin/bootloader.elf
-rw-r--r-- 1 dockeruser dockergroup  252461 Oct  2 15:36 build/bin/bootloader.map
-rwxr-xr-x 1 dockeruser dockergroup  225100 Oct  2 15:37 build/bin/factory-setup.bin
-rwxr-xr-x 1 dockeruser dockergroup 2466524 Oct  2 15:37 build/bin/factory-setup.elf
-rw-r--r-- 1 dockeruser dockergroup  880894 Oct  2 15:37 build/bin/factory-setup.map
-rwxr-xr-x 1 dockeruser dockergroup  406212 Oct  2 15:37 build/bin/firmware-btc.bin
-rwxr-xr-x 1 dockeruser dockergroup 3948304 Oct  2 15:37 build/bin/firmware-btc.elf
-rw-r--r-- 1 dockeruser dockergroup  929467 Oct  2 15:37 build/bin/firmware-btc.map
-rwxr-xr-x 1 dockeruser dockergroup  422548 Oct  2 15:38 build/bin/firmware.bin
-rwxr-xr-x 1 dockeruser dockergroup 4174144 Oct  2 15:38 build/bin/firmware.elf
-rw-r--r-- 1 dockeruser dockergroup  931452 Oct  2 15:38 build/bin/firmware.map


```

NickeZ:
I think the slight file differences come from the fact that the compiler optimizes a little differently depending on the linking order of the object files and archives. I see that a few symbols differ when I look at the symbol list with `nm`.

PR #132, opened 2019-09-25
accessible reproducible builds

benma:
[rendered](https://github.com/benma/bitbox02-firmware/blob/reproducible/releases/README.md)


TheCharlatan:
Bitcoin and Monero use gitian to generate their reproducible builds. Gitian has the benefit that it generates an "assertion" file that contains the hashes of the produced binaries, used toolchain, system images and libraries. I think we can skip this for the firmware repro builds.
It should be enough to take the sha256sum of the binary and write it to a file. Then this file can be signed for example with: 

    gpg --detach-sign firmware_$VERSION.assert

I propose we write a short script that takes the version and signer name as an argument, then computes the firmware checksum, signs it and places both the checksum and signature file into a directory structure similar to https://github.com/bitcoin-core/gitian.sigs

releases/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## What is the purpose of this?
+
+The purpose is twofold:
+
+- It is much more technically involved to actually build the binaries from source, than it is to
+  verify signatures using `gpg` on the command line or in a GUI. This gives users who are willing
+  and eager to verify the option to verify our releases even if they are unable to build the
+  binaries themselves.
+- The firmware binaries are only truly reproducible for a limited amount of time after the
```
NickeZ: Only a problem with the instructions. They are truly reproducible forever if you specify the docker container with a numbered tag. Reproducibly rebuilding the docker container on the other hand is not possible forever, but I think that is out of scope for reproducible builds.

releases/README.md
```diff
@@ -0,0 +1,86 @@
+# Reproducible builds
+
+The bitbox02 firmware [releases](https://github.com/digitalbitbox/bitbox02-firmware/releases/) are
+tagged: `firmware/vX.Y.Z` for the Multi edition, and `firmware-btc-only/vX.Y.Z` for the Bitcoin-only
+edition. The binaries are built from those tags in a reproducible manner, based on Docker.
+
+*Note*: it is possible to reproduce the binaries without Docker, the only really relevant dependency
```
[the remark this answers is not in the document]
benma: correct, fill fix

releases/README.md
benma: As the paragraph says, in the future Docker might not even be compatible with older images, etc. etc.

I added `practically` as a qualifier, otherwise the rabbit hole goes deep :smile:

benma:
Added a simple-to-review script to automate the build and an assertion file for firmware/v4.1.0. I ran  the whole thing and verified the hash and contributed my sig to kick it off.

releases/firmware-v4.1.0/assertion.txt
```diff
@@ -0,0 +1,8 @@
+By signing this file, the signer confirms that the firmware binary built from:
+
+git tag firmware/v4.1.0
+git commit hash 26b9e05b3127deb795bfb220fab3693ca19e12c3
+
+resulted, at the time of signing, in a firmware binary file with the following sha256sum:
+
+47fc9ef2ff754455e350606de2ba5684b634455bd52ac12ba6a9fb792f5c229c
```
TheCharlatan: Maybe add a comment somewhere that this file needs to be generated by the maintainer every time we tag a release?

benma: good idea. where though?

TheCharlatan:
I missed something: The gpg public keys of the signers should be inside the repository, but there are no instructions in there on how to add them.

benma:
good point!!

will add

benma:
> I missed something: The gpg public keys of the signers should be inside the repository, but there are no instructions in there on how to add them.

added in new commit

releases/README.md
```diff
@@ -0,0 +1,84 @@
...
+e.g. [firmware-v4.1.0/assertion.txt](firmware-v4.1.0/assertion.txt). If you agree to its contents
+and verified that the sha256 hash therein matches the one you got, please sign the file using:
+
+```sh
+cd firmware/v4.1.0 # go to the relevant subfolder
+gpg -o assertion-YOURNAME.sig --detach-sign assertion.txt
+```
+
+Open a PR adding your signature file to this folder. Also add your pgp pubkey to the
+[./pubkeys](./pubkeys) folder.
```
TheCharlatan: Can you add this line for instructions?

gpg --export --armor YOUR_PGP_KEY_ID > benma.asc

benma: Done!

PR #133, opened 2019-09-25
Dockerfile: A few more deps

NickeZ:
tested by rebuilding and then building firmware.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,39 +67,51 @@ RUN apt-get update && apt-get install -y \
     libudev-dev \
     libhidapi-dev
 
+RUN apt-get update && apt-get install -y \
+    doxygen
+
 # Set gcc-8 as the default gcc
 RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 100
 RUN update-alternatives --install /usr/bin/gcov gcov /usr/bin/gcov-8 100
 
 # Tools for CI
 RUN apt-get update && apt-get install -y \
+    python \
     python3 \
     python3-pip \
     clang-format-8 \
     clang-tidy-8
 
+RUN python3 -m pip install --upgrade pip
+
 # Python modules
-RUN pip3 install \
-    hidapi \
-    noiseprotocol \
-    protobuf \
-    ecdsa \
-    tzlocal \
-    semver
+COPY py/bitbox02/requirements.txt /tmp
+RUN python3 -m pip install --upgrade --requirement /tmp/requirements.txt
+COPY py/requirements.txt /tmp
+RUN python3 -m pip install --upgrade --requirement /tmp/requirements.txt
+RUN rm /tmp/requirements.txt
 
 # Python modules for CI
-RUN pip3 install \
-    pylint \
-    pylint-protobuf \
-    black \
+# TODO: This emits a warning about astroid that is safe to ignore?
+#       https://github.com/PyCQA/astroid/issues/699
+RUN python3 -m pip install --upgrade \
+    pylint==2.4.1 \
+    pylint-protobuf==0.11 \
+    black==19.3b0 \
     mypy==0.720 \
     mypy-protobuf==1.13
+
+# Python modules for packaging
+RUN python3 -m pip install --upgrade \
+    setuptools==41.2.0 \
+    wheel==0.33.6 \
+    twine==1.15.0
 
 # Developer tools
 RUN apt-get update && apt-get install -y \
     bash-completion
 # Install gcovr from PIP to get a newer version than in apt repositories
-RUN pip3 install gcovr
+RUN python3 -m pip install gcovr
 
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -83,23 +87,29 @@ RUN pip3 install \
     hidapi \
     noiseprotocol \
     protobuf \
+    mypy-protobuf \
```
benma: why move this here? imho the pip version below is better because it pins the version, and it's easier to maintain because all the python packages are in one place.

Dockerfile
```diff
@@ -83,23 +87,29 @@ RUN pip3 install \
...
 
 # Python modules for CI
-RUN pip3 install \
+RUN python3 -m pip install \
     pylint \
     pylint-protobuf \
     black \
-    mypy==0.720 \
-    mypy-protobuf==1.13
+    mypy
```
benma: why drop the version? ideally every software that Docker gets is pinned, otherwise a simple rebuild could break CI (happened frequently in the BitBox App until I pinned the relevant packages).

Dockerfile
NickeZ: my next pr will use `python3 pip install -r requirements.txt` so that we don't have that list in two places.

Dockerfile
NickeZ: shouldn't have versions here, they should be in `requirements.txt`

NickeZ: oh, or if packages only needed for CI it should take the latest and break so that we don't get stuck on old versions. This won't break the build system because it's only related to linting.

NickeZ:
I tried to enforce versions of everything but then I have to downgrade to mypy 0.660 which doesn't even support all the features we use. There is a conflict between pylint and mypy.

I also realized that it is very anti-social to have `==X.Y.Z` requirements on our python package. This means that a project that includes our module with a very high probability will have conflicts with other modules. My suggestion is that we assume that our required modules use semver and have `>=X.Y.Z, < X+1` as our requirements.

Dockerfile
benma: They are in requirements.txt and ideally we'd be installing from that file in the Dockerfile.

Still disagree. Rebuilding a Docker image should not cause any breakage if we can help it, and not pinning versions will cause breaks.

Not getting stuck on old versions: you want to upgrade them at your own pace and controlled, and not deal with it when it hits you at an inconvenient time.

Trust me, it was very inconvenient in the BitBox App until I fixed the versions.

benma:
> I also realized that it is very anti-social to have ==X.Y.Z requirements on our python package. This means that a project that includes our module with a very high probability will have conflicts with other modules. My suggestion is that we assume that our required modules use semver and have >=X.Y.Z, < X+1 as our requirements.

Yeah this is an issue. Yeah let's add more relaxed version bounds like you suggest.

Dockerfile
benma: To elaborate: similar to the lazy linting approach, I very much dislike having to deal with unrelated issues when I am working on something.

scripts/lint-python
```diff
@@ -0,0 +1,15 @@
...
+
+# Run pylint on all python code
+
+# Exit on error
+set -e
+
+# Exit on pipe fail
+set -o pipefail
+
+PYLINT=${PYLINT:-pylint}
```
benma: What's the overlap with `.ci/check-pep8`? that script should probably use this script, right?

maybe even use the other script as a base, so mypy etc is also checked, but then have the .ci/ script just call it with the diff files (or on all of them to keep it simple, as linting Python does not take any time).

NickeZ: I guess it wouldn't hurt to have this script take a list of files as optional input.

Dockerfile
NickeZ: did it in this PR.

Dockerfile
NickeZ: Yeah I see your point and added hard coded versions. Like mentioned below we are unfortunately using incompatible versions right now. The error doesn't seem fatal though, so my suggestion is that we live with it.

NickeZ: I also get that you don't prefer lazy linting and want to move the whole project at once due to the reasons you stated (we will have to agree to disagree on that one). I wouldn't call it lazy that CI only looks at modified files though. It is however lazy to not check and fix all the files at the time new checks are introduced. But to me those are two unrelated things.

NickeZ:
I did some more refactoring and moved most of `check-pep8` out into two separate scripts that now can be used as convenience scripts.

Dockerfile
benma: I don't mind not moving the whole project at once, file by file or similar is fine, as long as it doesn't force me to do it when I don't expect it :smile: 

Thanks!

benma:
> I did some more refactoring and moved most of `check-pep8` out into two separate scripts that now can be used as convenience scripts.

:clap: i like

PR #135, opened 2019-09-26
.ci: check sigs in the ./releases folder

benma:


.ci/check-release-sigs
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
```
NickeZ: I think you need `set -e` otherwise it won't actually fail the build.

benma: Good catch!

PR #141, opened 2019-10-01
usb_desc: fix descriptors for btc-only edition

NickeZ:


x1ddos:
Such a small change. Big impact. This could've been more descriptive compared to just "fix descriptors". Might be obvious to you guys. Not so much to some occasional readers like me.

This is the most significant change [between v4.1.0 and v.4.1.1](https://github.com/digitalbitbox/bitbox02-firmware/compare/firmware/v4.1.0...firmware-btc-only/v4.1.1).

benma:
It's the only functional change in fact since 4.1.0.

Agree, we could describe our commits better. The release notes however describe the reason for the patch release: https://github.com/digitalbitbox/bitbox02-firmware/releases/tag/firmware-btc-only%2Fv4.1.1

You are right in that it was obvious to us, as we discussed it internally ;)

Post-merge explanation: before thix hotfix, Windows would not recognize the Bitcoin-only edition (Linux and macOS worked past the invalid descriptor).

PR #147, opened 2019-10-04
Minor improvements

conte91:
Preliminary cleanup of `screen_process.c` to allow for easier refactoring in the next PR


src/ui/screen_process.c
```diff
@@ -34,7 +34,7 @@ void ui_screen_render_component(component_t* component)
     UG_SendBuffer();
 }
 
-static component_t *get_waiting_screen(void)
+static component_t* _get_waiting_screen(void)
```
benma: avoid merging fixup commits, rather squash them into the right commit, to keep a nice commit history

`git blame` should show relevant commits mostly, not "fixed this and that" :)

conte91: Got it :)

PR #148, opened 2019-10-04
Separate screen-, workflow- and usb-related functions

conte91:
Moved the usb processing and is_done() logic outside of `_screen_process()`.

Only the main loop should be responsible for calling the USB handlers - does this have anything to do with the GUI after all? Same for the workflow-related logic (blocking, timeouts, etc).

Note that this is just code wrangling so the new implementation should be 100% equivalent to the current one - if it isn't, that's a bug.

Depends on #147.

### DIFF
--- a/src/firmware_main_loop.c
+++ b/src/firmware_main_loop.c
@@ -1,0 +1,37 @@
+#include "firmware_main_loop.h"
+
+#include "hardfault.h"
+#include "ui/screen_process.h"
+#include "usb/usb_processing.h"
+
+#include <stdbool.h>
+
+/*
+ * "is_done" callback that only spins the UI
+ * once.
+ */
+static bool _is_done_run_once(void* param)
+{
+    if (!param) {
+        Abort("_is_done_run_once called\nwith NULL param.");
+    }
+    bool* already_run = (bool*)param;
+    if (!(*already_run)) {
+        *already_run = true;
+        return false;
+    }
+    return true;
+}
+
+void firmware_main_loop(void)
+{
+    bool already_run;
+    while (1) {
+        already_run = false;
+        ui_screen_process(_is_done_run_once, &already_run);
+        usb_processing_process(usb_processing_hww());
+#if defined(APP_U2F)
+        usb_processing_process(usb_processing_u2f());
+#endif
+    }
+}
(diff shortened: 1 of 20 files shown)


src/firmware_main_loop.c
```diff
@@ -0,0 +1,15 @@
+#include "firmware_main_loop.h"
```
benma: license header missing

src/ui/screen_process.c
```diff
@@ -34,6 +31,23 @@ void ui_screen_render_component(component_t* component)
...
 }
 
+/**
+ * Detects if the screen component being displayed has changed
+ * since the last time this function was called.
+ * This stores the last observed component into a global.
+ *
+ * @param[in] current_comp Current on-screen component.
+ */
+static bool _screen_has_changed(const component_t* current_comp)
```
benma: please no short names, as explicit long names are easier to understand, is less ambiguous etc.. `current_component`

src/ui/screen_process.c
```diff
@@ -46,52 +60,44 @@ static component_t* _get_waiting_screen(void)
...
+        return _get_waiting_screen();
+    }
+    return result;
 }
 
-void ui_screen_process_with_timeout(
-    bool (*is_done)(void),
-    void (*on_timeout)(void),
-    uint32_t timeout)
+void ui_spin(void)
```
benma: this function should be called `screen_process_spin` (usually same as filename).

can rename the whole file and the functions in here in a separate PR, but would keep `screen_process_*` now to keep consistent.

furthermore, `spin` to me means it is looping, but it is not looping. Please find a better name (`screen_process()` ?)

src/ui/screen_process.c
```diff
@@ -46,26 +62,39 @@ static component_t* _get_waiting_screen(void)
     return waiting_screen;
 }
 
-static void _screen_process(bool (*is_done)(void), void (*on_timeout)(void), const uint32_t timeout)
+/**
+ * \param is_done WTF
```
benma: :thinking: 

not sure this comment helps  :laughing:

benma:
This looks like a very good cleanup. Code looks good to me!

I will test it some more tomorrow.

src/ui/screen_process.c
conte91: Fixed. `screen_process` wins.

src/ui/screen_process.c
conte91: Fixed

src/ui/screen_process.c
conte91: Comment? Which comment? :D

PR #150, opened 2019-10-04
py/u2f: Fix order of cid/cmd and add debug prints

NickeZ:
cid and cmd where mixed up in the u2f python testing script so that didn't work.

### DIFF
--- a/py/bitbox02/u2fhid/u2fhid.py
+++ b/py/bitbox02/u2fhid/u2fhid.py
@@ -84,9 +84,17 @@ def write(hid_device: SupportsReadWrite, data: bytes, cmd: int, cid: int) -> Non
         data: Data to send
         cmd: U2F HID command
         cid: U2F HID channel ID
+    Throws:
+        ValueError: In case any value is out of range
     """
+    if cmd < 0 or cmd > 0xFF:
+        raise ValueError("Channel command is out of range '0 < cmd <= 0xFF'")
+    if cid < 0 or cid > 0xFFFFFFFF:
+        raise ValueError("Channel id is out of range '0 < cid <= 0xFFFFFFFF'")
     data = bytearray(data)
     data_len = len(data)
+    if data_len > 0xFFFF:
+        raise ValueError("Data is too large 'size <= 0xFFFF'")
     seq = 0
     idx = 0
     buf = b""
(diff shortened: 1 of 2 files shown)


py/bitbox02/u2fhid/u2fhid.py
```diff
@@ -98,7 +98,7 @@ def write(hid_device: SupportsReadWrite, data: bytes, cmd: int, cid: int) -> Non
             buf = data[idx : idx + min(data_len, USB_REPORT_SIZE - 7)]
             hid_device.write(
                 b"\0"
-                + struct.pack(">IBH", cid, cmd, data_len & 0xFFFF)
+                + struct.pack(">IBH", cid, cmd & 0xFF, data_len & 0xFFFF)
```
benma: instead of clamping, maybe check the range of `cmd` in the beginning of the function and raise a ValueError exception?

NickeZ: good idea, I was just lazy and did what was already present. Should probably warn for both cid, cmd and data_len if they are to large.

NickeZ: fixed and updated comments

NickeZ:
force pushed with updates according to review.

py/bitbox02/u2fhid/u2fhid.py
```diff
@@ -84,9 +84,17 @@ def write(hid_device: SupportsReadWrite, data: bytes, cmd: int, cid: int) -> Non
         data: Data to send
         cmd: U2F HID command
         cid: U2F HID channel ID
+    Throws:
+        ValueError: In case any value is out of range
     """
+    if cmd < 0 or cmd > 0xFF:
+        raise ValueError("Channel command is out of range '0 > cmd > 0xFF'")
```
benma: you mean `0 <= cmd <= 0xFF`? Fyi, you can use this as a valid Python expression as well:

`if not (0 <= cmd <= 0xFF): ...`

py/bitbox02/u2fhid/u2fhid.py
```diff
@@ -84,9 +84,17 @@ def write(hid_device: SupportsReadWrite, data: bytes, cmd: int, cid: int) -> Non
...
+    if cid < 0 or cid > 0xFFFFFFFF:
+        raise ValueError("Channel id is out of range '0 > cid > 0xFFFFFFFF'")
```
benma: same, `>` should be `<=`

py/bitbox02/u2fhid/u2fhid.py
NickeZ: Yeah the exception had the wrong comment. Oh, I forgot about that. Anyway I prefer the current syntax. The python syntax is a bit weird.

py/bitbox02/u2fhid/u2fhid.py
NickeZ: ack

PR #153, opened 2019-10-07
ui/keyboard_switch: use one source of truth for the current keyboard

benma:
The current keyboard mode was tracked twice, once in keyboard_switch.c
to display the title, and once in trinary_input_string.c to select the
actual keyboard.


src/ui/components/keyboard_switch.h
```diff
@@ -18,7 +18,7 @@
 #include <screen.h>
 #include <ui/component.h>
 
-typedef enum { DIGITS, LOWER_CASE, UPPER_CASE, NUM_INPUT_TYPES } keyboard_mode_t;
+typedef enum { DIGITS, LOWER_CASE, UPPER_CASE } keyboard_mode_t;
```
conte91: `NUM_INPUT_TYPES` can be used as a tracker for the size of the enum, without having to hardcode it in the code or change it if additional values are added to the enum itself (see below)

src/ui/components/set_password.c
```diff
@@ -193,7 +193,7 @@ static void _on_event(const event_t* event, component_t* component)
     case EVENT_TOGGLE_ALPHANUMERIC:
         data->auto_forward_active = true;
         data->auto_forward_counter = 0;
-        data->input_mode = (data->input_mode + 1) % NUM_INPUT_TYPES;
+        data->input_mode = (data->input_mode + 1) % 3;
```
conte91: (See above) `NUM_INPUT_TYPES` is quite useful as a sentinel value for computing the size of the enum automatically. If we want to remove it from the enum let's define it on a side so not to leave the `3` hardcoded...

benma: As the commit message says, set_password.c is dead code and will be removed soon (it is outdated anyway in various ways).

Other than here, this sentinel value is not used anywhere currently.

Happy to remove set_password.c today.

src/ui/components/keyboard_switch.h
benma: Yeah I was aware; see the commit message / my comment below.

src/ui/components/set_password.c
conte91: Ok, let's ditch all of it together then?

benma: ACK, done

PR #154, opened 2019-10-07
Specialchars

benma:


benma:
First few commits are from https://github.com/digitalbitbox/bitbox02-firmware/pull/153

src/ui/components/keyboard_switch.c
```diff
@@ -82,14 +86,17 @@ static void _on_event(const event_t* event, component_t* component)
...
             break;
         case LOWER_CASE:
             ks_data->mode = UPPER_CASE;
             break;
         case UPPER_CASE:
             ks_data->mode = DIGITS;
             break;
+        case SPECIAL_CHARS:
+            ks_data->mode = LOWER_CASE;
+            break;
```
conte91: Can we move this up as the 2nd entry - so that one can follow what the toggle order will be? It's confusing having to jump around to make sure that all cases are covered. Maybe let's reorder the values in the enum so that the order is preserved there as well?

benma: Good idea. Done, see new commit.

PR #156, opened 2019-10-07
Build instruction fixes

KasparEtter:


BUILD.md
```diff
@@ -119,7 +121,7 @@ make docs
 
 To view the results, open `build/docs/html/index.html` in a web browser.
 
-### Python Lib
+### BitBox02 Python Library
```
benma: uuh didn't the other commit just remove title case? :D

KasparEtter: Yes, but only with the last commit, this one was earlier… 😇

benma: Indeed!

PR #157, opened 2019-10-08
workflow.c: Remove the useless "workflow_change_state" function.

conte91:
`workflow_change_state` was supposed to be "switching workflows based on
user inputs" (as per the docs), but in practice all what it was doing was switching to
the "Tap this side" screen.

### DIFF
--- a/src/workflow/workflow.c
+++ b/src/workflow/workflow.c
@@ -43,6 +43,17 @@ void workflow_confirm_dismiss(const char* title, const char* body)
     ui_screen_stack_switch(confirm_create(title, body, false, _confirm_dismiss, NULL));
 }
 
+void workflow_start(void)
+{
+    usb_start(hww_setup);
+    ui_screen_stack_pop_all();
+    ui_screen_stack_push(info_centered_create("See the BitBox App", NULL));
+}
+
+/**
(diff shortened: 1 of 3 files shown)


src/workflow/workflow.c
```diff
@@ -52,23 +63,8 @@ static void _select_orientation_done(bool upside_down)
...
 
-void workflow_start(void)
-{
-    usb_start(hww_setup);
-    ui_screen_stack_pop_all();
-    ui_screen_stack_push(info_centered_create("See the BitBox App", NULL));
-}
-
-void workflow_change_state(workflow_state_t state)
+void start_orientation_screen(void)
```
benma: prefix with `workflow_` based on the filename. We generally do this as a poor man's namespacing ;) and it helps to know roughly in what area a function lives when reading code.

conte91: Amended e35498b24bf0ba762babf224fdc8e08edd2d4e

conte91:
> You have a commit `Merge remote-tracking branch 'upstream/master'` in this PR now. Should just be your commit.

I've manually merged `master` in to fix possible conflicts and test on the updated codebase - I'll rebase instead.

PR #161, opened 2019-10-10
usb/hid: Separate the code sending and receiving packets.

conte91:
Remove the assumption spread throughout hid_hww.c, hid_u2f.c, and
usb_processing.c, that sends should start only after a receive and
receives should start only when a frame has been sent.

Split the single code flow performing the send->rcv->send->rcv... operations
into two separate call stacks. The read part will reissue a (async) read
as soon as data arrives and finished processing. The write part will
continuously issue (async) sends until the write buffer is flushed
completely. Whenever somebody queues new data to write, a new series of
sends will start.

### DIFF
--- a/src/u2f/u2f_packet.c
+++ b/src/u2f/u2f_packet.c
@@ -127,7 +127,7 @@ void u2f_packet_timeout(uint32_t cid)
     usb_frame_prepare_err(FRAME_ERR_MSG_TIMEOUT, cid, _queue_push);
 }
 
-bool u2f_packet_process(const USB_FRAME* frame, void (*send_packet)(void))
+bool u2f_packet_process(const USB_FRAME* frame)
 {
     struct usb_processing* ctx = usb_processing_u2f();
     switch (usb_frame_process(frame, &_in_state)) {
@@ -155,6 +155,7 @@ bool u2f_packet_process(const USB_FRAME* frame, void (*send_packet)(void))
             // Do not send a message yet
             return true;
         }
+        /* We have received a complete frame. Buffer it for processing. */
         if (usb_processing_enqueue(ctx, &_in_state)) {
             // Queue filled and will be sent during usb processing
             _reset_state();
@@ -171,6 +172,5 @@ bool u2f_packet_process(const USB_FRAME* frame, void (*send_packet)(void))
         _queue_err(FRAME_ERR_OTHER, frame->cid);
         break;
     }
-    send_packet();
     return false;
 }
--- a/src/u2f/u2f_packet.h
+++ b/src/u2f/u2f_packet.h
@@ -24,10 +24,9 @@
 /**
  * Processes an incoming USB packet.
  * @param[in] frame The frame that is to be processed.
- * @param[in] send_packet The function to be called to send the response packet.
  * @return true if we are waiting for more frames to complete a packet, false otherwise.
  */
-bool u2f_packet_process(const USB_FRAME* frame, void (*send_packet)(void));
+bool u2f_packet_process(const USB_FRAME* frame);
 
 /**
  * Checks if there has been a timeout
--- a/src/usb/class/hid/hww/hid_hww.c
+++ b/src/usb/class/hid/hww/hid_hww.c
@@ -62,17 +62,25 @@ static int32_t _read(void)
     return hid_read(&_func_data, _out_report, USB_HID_REPORT_OUT_SIZE);
 }
 
+/** Set when the send channel is busy sending data. */
+static bool _send_busy = false;
+
 /**
  * Sends the next frame, if the USB interface is ready.
  */
 static void _send_next(void)
 {
+    if (_send_busy) {
+        /*
+         * We can't send yet. Whenever the current sender finished, it will
+         * flush anything that's still queued.
+         */
+        return;
+    }
     const uint8_t* data = queue_pull(queue_hww_queue());
     if (data != NULL) {
+        _send_busy = true;
         hid_write(&_func_data, data, USB_HID_REPORT_OUT_SIZE);
-    } else {
-        // Read again after we sent everything.
-        _read();
     }
 }
 
@@ -86,11 +94,9 @@ static uint8_t _out(const uint8_t ep, const enum usb_xfer_code rc, const uint32_
     (void)ep;
     (void)rc;
     (void)count;
-    bool need_more = usb_packet_process((const USB_FRAME*)_out_report, _send_next);
-    if (need_more) {
-        _read();
-    }
-
+    usb_packet_process((const USB_FRAME*)_out_report);
+    /* Incoming data has been processed completely. Start a new read. */
+    _read();
     return ERR_NONE;
 }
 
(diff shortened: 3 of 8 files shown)


conte91:
As a side effect, this PR makes the U2F and HWW receiving USB stacks entirely independent. This solves an existing issue for which incoming U2F packets are silently dropped if there is some HWW operation in progress.

Scenario: The user opens the "confirm receive address" window. While it's open, it requests U2F auth.
Current situation: the U2F request is silently dropped - this breaks Firefox (you have to reload the page and start U2F from scratch).
New situation: the U2F request is queued - so when the HWW window exits, the U2F process starts.

However, the solution is probably not ideal: it has the downside of having two separate (7kB!) buffers allocated for each stack. Also, from the UI side, having the U2F request being processed with such a large delay could confuse the user (maybe he just closed the U2F window entirely 10 minutes ago, and suddenly a U2F registration request pops out when he finishes his HWW operation?)

Maybe a good solution could be to modify this situation slightly, so that the device reports as busy when something else is being processed on the other stack?

I don't think I know enough details of U2F (or HWW for what matters) not to make a mistake here :) @NickeZ could you give your opinion?

Also @benma we've discussed this offline.

benma:
> Maybe a good solution could be to modify this situation slightly, so that the device reports as busy when something else is being processed on the other stack?

:+1:

NickeZ:
Yes! Fantastic!

I hope the u2f and the hww usb frame stacks can be merged eventually.

If the device is busy (with another u2f request, or a hww request) the correct response is "ERR_CHANNEL_BUSY" on the u2fhid layer.

https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-u2f-hid-protocol-v1.2-ps-20170411.html#u2fhid_error

NickeZ:
FYI: you can run the test_u2f_hid_with_hardware and test_u2f_standard_with_hardware with the bitbox to double check that everything works.

Eventually we should have some regression tests so that we can make changes to the usb stack with more confidence.

conte91:
> If the device is busy (with another u2f request, or a hww request) the correct response is "ERR_CHANNEL_BUSY" on the u2fhid layer.
> 
There is some comment by @douglasbakkum in the u2f layer that says we send something else instead:
`usb_processing.c:175` says:

>     if (!cmd_valid) {
>         // TODO: if U2F is disabled, we used to return a 'channel busy' command.
>         // now we return an invalid cmd, because there is not going to be a matching
>         // cmd in '_registered_cmds' if the U2F bit it not set (== U2F disabled).
>         // TODO: figure out the consequences and either implement a solution or
>         // inform U2F hijack vendors.
>         _global_queue = ctx->out_queue();
>         usb_frame_prepare_err(FRAME_ERR_INVALID_CMD, ctx->in_packet.cid, _queue_push);
>     }
> 

Is this what you were talking about?

`

conte91:
> FYI: you can run the test_u2f_hid_with_hardware and test_u2f_standard_with_hardware with the bitbox to double check that everything works.
> 
> Eventually we should have some regression tests so that we can make changes to the usb stack with more confidence.

+1 - the more modularization PRs I open, the more I wish we had some automatic testing. Maybe we should get some _before_ we push the big ones in (u2f modularization is WIP, workflow modularization will come next)?

NickeZ:
> Ja that passes - although it makes assumptions about the exact number of usb_processing calls that need to happen, which is going to break the tests if an event loop goes in.

The hardware-in-the-loop tests should not make any such assumptions.

NickeZ:
> > If the device is busy (with another u2f request, or a hww request) the correct response is "ERR_CHANNEL_BUSY" on the u2fhid layer.
> 
> There is some comment by @douglasbakkum in the u2f layer that says we send something else instead:
> `usb_processing.c:175` says:
> 
> > ```
> > if (!cmd_valid) {
> >     // TODO: if U2F is disabled, we used to return a 'channel busy' command.
> >     // now we return an invalid cmd, because there is not going to be a matching
> >     // cmd in '_registered_cmds' if the U2F bit it not set (== U2F disabled).
> >     // TODO: figure out the consequences and either implement a solution or
> >     // inform U2F hijack vendors.
> >     _global_queue = ctx->out_queue();
> >     usb_frame_prepare_err(FRAME_ERR_INVALID_CMD, ctx->in_packet.cid, _queue_push);
> > }
> > ```
> 
> Is this what you were talking about?
> 
> `

Thats probably just some copy-paste leftover from bitbox01.

If the bitbox02 is busy we should respond with the busy message. AFAIU

conte91:
> > Ja that passes - although it makes assumptions about the exact number of usb_processing calls that need to happen, which is going to break the tests if an event loop goes in.
> 
> The hardware-in-the-loop tests should not make any such assumptions.

`mock_hidapi.c` seems to do that - `hid_read_timeout()` function "knows" when to poll for data based on the internal behaviour of `usb_processing_process`. This should be easy to fix but can go in another PR.

conte91:
> If the device is busy (with another u2f request, or a hww request) the correct response is "ERR_CHANNEL_BUSY" on the u2fhid layer.

Ok, implementing this on the U2F layer is pretty tough. Currently the check is performed at the raw USB frame level (`{usb,u2f}_packet.c`). So we send out a `FRAME_ERR_CHANNEL_BUSY` response, which AFAIK means the _usb stack_ itself is busy. What we want to say is that the UI/device is busy, over a normal (non-error) USB frame. But with the current code, a blocking UI does indeed imply that the USB port is stalling - so there's really no way around it atm.

I've left the "old" solution (responding with `FRAME_ERR_CHANNEL_BUSY`) and moved the packet buffers to a shared space in the meanwhile (as each packet buffer occupies 7kB RAM, and we're only using one at the time!).

I'll open a separate issue for plugging in the correct responses once blocking operations become independent from the USB processing.

NickeZ:
> > If the device is busy (with another u2f request, or a hww request) the correct response is "ERR_CHANNEL_BUSY" on the u2fhid layer.
> 
> Ok, implementing this on the U2F layer is pretty tough. Currently the check is performed at the raw USB frame level (`{usb,u2f}_packet.c`). So we send out a `FRAME_ERR_CHANNEL_BUSY` response, which AFAIK means the _usb stack_ itself is busy. What we want to say is that the UI/device is busy, over a normal (non-error) USB frame. But with the current code, a blocking UI does indeed imply that the USB port is stalling - so there's really no way around it atm.
> 
> I've left the "old" solution (responding with `FRAME_ERR_CHANNEL_BUSY`) and moved the packet buffers to a shared space in the meanwhile (as each packet buffer occupies 7kB RAM, and we're only using one at the time!).
> 
> I'll open a separate issue for plugging in the correct responses once blocking operations become independent from the USB processing.

I think I understand what you mean. If there is anything on the screen we should respond with `SW_CONDITIONS_NOT_SATISFIED` to any U2F requests.

For our own communications we should probably add a "BUSY" error code here: https://github.com/digitalbitbox/bitbox02-firmware/blob/730c5b956ec1e9a421fa76aec8353f95ff6ff9e1/src/commander/commander.h#L26

conte91:
> Did you also test flashing the bootloader and using the bootloader to flash the firmware? (since this is touching the usb stack which also affects the bootloader).

Just did, still working it seems!

NickeZ:
Cool, i bricked a few devices when i touched the usb stack last time ;)

PR #162, opened 2019-10-11
[wip] Base buildtarget

NickeZ:
- [ ] Bump dockerfile accordingly


external/CMakeLists.txt
```diff
@@ -273,7 +276,12 @@ if(CMAKE_CROSSCOMPILING)
     cryptoauthlib/lib/crypto/hashes/sha2_routines.c
     cryptoauthlib/lib/crypto/atca_crypto_sw_sha2.c
   )
-  target_link_libraries(cryptoauthlib asf4-drivers-min)
+  target_link_libraries(cryptoauthlib
+    PRIVATE
+      CMSIS
```
benma: why does cryptoauthlib even need any of those?

external/asf4-drivers/driver_init.c
```diff
@@ -290,6 +290,31 @@ void system_init(void)
     _is_initialized = true;
 }
 
+void base_init(void)
```
benma: does this refer to bitbox base? i would spell it would everywhere tbh, `base` is so generic that it is easy to mistake it for something else.

messages/hww.proto
```diff
@@ -264,6 +264,25 @@ message RestoreFromMnemonicRequest {
   int32 timezone_offset = 2;
 }
 
+message DisplayBase32Request {
```
benma: Make one `oneof` request/response type for all Base things, same as ETHRequest and ETHResponse. Then it is better contained and easier to read/change/maintain.

messages/hww.proto
```diff
@@ -264,6 +264,25 @@ message RestoreFromMnemonicRequest {
...
+  bytes msg = 1;
+}
+
+
+message BaseSetConfigRequest {
```
benma: also here, albeit very verbose, i'd be explicit: `BitBoxBaseSetConfigRequest`.

src/commander/commander.c
```diff
@@ -203,25 +206,46 @@ static commander_error_t _api_set_mnemonic_passphrase_enabled(
...
+#if defined(FIRMWARE_BASE)
+static commander_error_t _api_display_base32(const DisplayBase32Request* request) {
+    if (!workflow_display_base32(request->msg, sizeof(request->msg))) {
         return COMMANDER_ERR_GENERIC;
     }
     return COMMANDER_OK;
 }
 
-static commander_error_t _api_restore_from_mnemonic(const RestoreFromMnemonicRequest* request)
+static commander_error_t _api_base_set_config(const BaseSetConfigRequest* request) {
```
benma: move all the base api's to `commander/commander_bitboxbase.c`. this works nicely if all the api messages are in on Request/Response message. check `commander/commander_eth.c` as a reference, whose main commander function is much like the one in this file, just nested one level deeper:

https://github.com/digitalbitbox/bitbox02-firmware/blob/730c5b956ec1e9a421fa76aec8353f95ff6ff9e1/src/commander/commander_eth.c#L89-L102

src/commander/commander.c
```diff
@@ -318,6 +340,21 @@ static commander_error_t _api_process(const Request* request, Response* response
     case Request_restore_from_mnemonic_tag:
         response->which_response = Response_success_tag;
         return _api_restore_from_mnemonic(&(request->request.restore_from_mnemonic));
+#endif
+#if defined(FIRMWARE_BASE)
+    case Request_display_base32_tag:
```
benma: same as my other  comments: here you can collapse all into one case: `Request_bitboxbase_tag`

src/rust/base/src/config.rs
```diff
@@ -0,0 +1,40 @@
+// Copyright 2019 Shift Cryptosecurity AG
```
benma: A historic moment :tada:

src/rust/base/src/config.rs
```diff
@@ -0,0 +1,40 @@
...
+        Config {
+            hostname: None,
+        }
+    }
+
+    pub fn set_hostname(&mut self, hostname: &str) -> Result<(), Error>  {
+        if hostname.chars().any(|c| !c.is_alphanumeric()) {
+            return Err(Error::InvalidHostname);
+        }
+        // We need to use unsafe to access globals mutably
```
benma: what is unsafe? where is the mutable global? sorry, i'm new to this :P

src/rust/base/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[no_mangle]
+pub extern fn workflow_display_base32(bytes: *const u8, bytes_len: usize) -> bool {
+    assert!(!bytes.is_null());
+    assert!(bytes_len > 0 && bytes_len <= 32);
+    let bytes = unsafe { core::slice::from_raw_parts(bytes, bytes_len) };
+    workflow::display_base32(bytes)
+}
+
+static mut CONFIG: config::Config = config::Config::new();
```
benma: is there no type inference for something like this?

also what is `CONFIG:` vs `config::Config`?

src/rust/base/src/screen.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2019 Shift Cryptosecurity AG
```
benma: `rust/base/src/screen.rs`

I want to use Rust not only for the base over time, so can we structure it so that this will be easier?

`rust/src/screen.rs`
`rust/src/apps/bitboxbbase/config.rs`

etc.

src/CMakeLists.txt
```diff
@@ -296,7 +341,18 @@ if(CMAKE_CROSSCOMPILING)
   target_compile_definitions(bootloader-btc-production.elf PRIVATE BOOTLOADER_PRODUCTION BOOTLOADER_BTC_ONLY)
   set_property(TARGET bootloader-btc-production.elf PROPERTY EXCLUDE_FROM_ALL ON)
 
-  set(FIRMWARES firmware firmware-semihosting firmware-btc factory-setup)
+  target_compile_definitions(bootloader-base.elf PRIVATE BOOTLOADER_BASE)
```
benma: `BOOTLOADER_BITBOXBASE`

src/CMakeLists.txt
```diff
@@ -345,20 +402,26 @@ if(CMAKE_CROSSCOMPILING)
...
   target_link_libraries(firmware-semihosting.elf PRIVATE --specs=rdimon.specs)
-  target_compile_definitions(firmware-semihosting.elf PRIVATE APP_BTC APP_LTC APP_ETH APP_U2F)
+  target_compile_definitions(firmware-semihosting.elf PRIVATE FIRMWARE_MULTI APP_BTC APP_LTC APP_ETH APP_U2F)
   target_compile_definitions(firmware-semihosting.elf PRIVATE SEMIHOSTING)
 
   target_sources(firmware-btc.elf PRIVATE firmware.c)
   target_compile_definitions(firmware-btc.elf PRIVATE FIRMWARE_BTC_ONLY APP_BTC)
 
+  target_sources(firmware-base.elf PRIVATE base.c)
+  target_compile_definitions(firmware-base.elf PRIVATE FIRMWARE_BASE APP_BTC)
```
benma: `FIRMWARE_BITBOXBASE`.

reading `#ifdef FIRMWARE_BASE` is not descriptive enough

external/CMakeLists.txt
NickeZ: I think it's because it's a wrapper around those libs.

external/asf4-drivers/driver_init.c
NickeZ: ack

messages/hww.proto
NickeZ: ack

messages/hww.proto
NickeZ: ack

src/commander/commander.c
NickeZ: ack

src/commander/commander.c
NickeZ: ack

src/rust/base/src/config.rs
NickeZ: stray comment. Moved the code, forgot about the comment ;)

src/rust/base/src/lib.rs
NickeZ: `CONFIG` is the variable name `config::Config` is the type. Globals don't have type inference yet IIRC.

src/rust/base/src/screen.rs
NickeZ: ack

src/CMakeLists.txt
NickeZ: ack

src/CMakeLists.txt
NickeZ: ack

NickeZ:
Thanks for the comments!

src/rust/base/src/config.rs
NickeZ: Indeed!

NickeZ:
force pushed, needs another round of review.

src/rust/base/src/screen.rs
benma: thanks for moving the files, but I think it makes more sense to structure it like I proposed, e.g. mirroring the existing directory structure.

- clearer where to add new Rust files
- bitboxbase can be treated as an app alongside btc/eth

src/CMakeLists.txt
```diff
@@ -248,13 +248,73 @@ set(BOOTLOADER-SOURCES
...
+
+# TODO(nc): Move to root cmakefile when api-docs is merged
+add_custom_target(doc-rust
+  COMMAND
+    ${CMAKE_COMMAND} -E env SYSROOT=${CMAKE_SYSROOT} INCLUDES="${RUST_INCLUDES}" ${CARGO} doc --manifest-path ${LIBBITBOX02_NEW_SOURCE_DIR}/Cargo.toml --target-dir ${CMAKE_BINARY_DIR}/doc-rust ${RUST_CARGO_FLAG} --target thumbv7em-none-eabi
+  COMMAND
+    echo "See docs at file://${CMAKE_BINARY_DIR}/doc-rust/thumbv7em-none-eabi/doc/bitbox02_new/index.html"
+)
+
+add_library(bitbox02_new STATIC IMPORTED)
```
benma: What does `_new` mean? Would be good to add a comment to explain what this is, and maybe pick a descriptive name.

src/CMakeLists.txt
```diff
@@ -345,20 +417,26 @@ if(CMAKE_CROSSCOMPILING)
...
   target_link_libraries(firmware-semihosting.elf PRIVATE --specs=rdimon.specs)
-  target_compile_definitions(firmware-semihosting.elf PRIVATE APP_BTC APP_LTC APP_ETH APP_U2F)
+  target_compile_definitions(firmware-semihosting.elf PRIVATE FIRMWARE_MULTI APP_BTC APP_LTC APP_ETH APP_U2F)
   target_compile_definitions(firmware-semihosting.elf PRIVATE SEMIHOSTING)
 
   target_sources(firmware-btc.elf PRIVATE firmware.c)
   target_compile_definitions(firmware-btc.elf PRIVATE FIRMWARE_BTC_ONLY APP_BTC)
 
+  target_sources(firmware-base.elf PRIVATE bitboxbase/bitboxbase.c commander/commander_bitboxbase.c)
+  target_compile_definitions(firmware-base.elf PRIVATE FIRMWARE_BITBOXBASE APP_BTC)
```
benma: I'd also add `APP_BITBOXBASE`, and then use that to include/exclude the bitboxbase api calls (similar to ETH).

That would be more consistent with how we do it with the other editions.

src/rust/base/src/screen.rs
NickeZ: I wouldn't really say that base is an app. hopefully a lot of code can be shared between the base and the other targets in the future.

src/CMakeLists.txt
NickeZ: Basically it means "functionality implemented in rust". Should I call it `bitbox02_rust` instead?

src/CMakeLists.txt
NickeZ: I would differentiate "platform" from "app". Maybe `PLATFORM_BITBOXBASE`?

NickeZ: Maybe we should remove all `FIRMWARE_` prefixed defines and use `TARGET_BITBOXMULTI`, `TARGET_BITBOXBTCONLY` and `TARGET_BITBOXBASE`.

benma: the base is an edition on the same level as multi, btconly, right?

i was just saying that the base api calls could be in the `bitboxbase` app, to be consistent with rest.

e.g. [here](https://github.com/digitalbitbox/bitbox02-firmware/pull/162/files#diff-82b64e48d7dd3ab4dbc37ae5d9ff3dceR310) it is also `#if defined(APP_ETH)`, not `#if defined(FIRMWARE_MULTI)`. The `#if defined(FIRMWARE_BITBOXBASE)` a few lines below is kind of breaking the pattern.

Not a big deal though.

src/rust/base/src/screen.rs
benma: true, but the api particular api calls of the base could be considered an app. i see this could be confusing though, was just going for consistency in packaging specific api calls in apps, but we can also not do that :+1:

src/CMakeLists.txt
NickeZ: Yeah, but I'm just saying that we have two types of abstractions right now, firmware targets such as multi/bitcoin-only/bitboxbase and that we have apps such as BTC, ETH, U2F, perhaps "autosigning with whitelisting". And I agree that calling it `FIRMWARE_BITBOXBASE` might be misleading, but calling it `APP_BITBOXBASE` would be completely wrong IMO.

src/CMakeLists.txt
benma: yes please :)

NickeZ:
Closing this as it became to big. Doing separate PRs instead.

PR #164, opened 2019-10-14
Make the U2F-handling code asynchronous. 

conte91:
Make most of the U2F-handling code asynchronous. 

All non-trivial U2F operations (i.e. auth, registration) have been split into multiple steps. The state machine controlling the execution of these steps is run as part of the main loop in `firmware_main_loop.c`. This allow the code processing incoming USB frames to release control immediately, and run the actual operations asynchronously.


src/u2f.c
```diff
@@ -249,56 +330,104 @@ static int _sig_to_der(const uint8_t* sig, uint8_t* der)
...
+ *
+ * @param[in] app_id
+ * @param[out] mac Buffer to fill with the generated message authentication code.
+ * @param[out] nonce Buffer to fill with the generated nonce.
+ * @param[out] pubkey Buffer to fill with the generated public key.
+ * @return Whether the generation was successful.
+ */
+static bool _generate_registration_keys(const uint8_t* app_id, uint8_t* nonce, uint8_t* mac, uint8_t* pubkey) {
+    uint8_t privkey[U2F_EC_KEY_SIZE];
+    for (int i = 0; i < 10; ++i) {
```
benma: can use `UTIL_CLEANUP_32(privkey)` on the next line and skip the manual zeroing.

src/u2f.c
```diff
@@ -249,56 +330,104 @@ static int _sig_to_der(const uint8_t* sig, uint8_t* der)
...
-            _error(U2F_SW_CONDITIONS_NOT_SATISFIED, out_packet);
-            return;
-        }
+/**
+ * Tries to generate a set of keys for signing a registration request.
+ *
+ * Fails after 10 attempts.
+ *
+ * @param[in] app_id
+ * @param[out] mac Buffer to fill with the generated message authentication code.
```
benma: order the doc params like the actual params; also add what sizes the buffers must be (app_id, nonce, mac 32 bytes, pubkey 64 bytes).

src/u2f.c
```diff
@@ -249,56 +330,104 @@ static int _sig_to_der(const uint8_t* sig, uint8_t* der)
...
+ * @param[out] nonce Buffer to fill with the generated nonce.
+ * @param[out] pubkey Buffer to fill with the generated public key.
+ * @return Whether the generation was successful.
+ */
+static bool _generate_registration_keys(const uint8_t* app_id, uint8_t* nonce, uint8_t* mac, uint8_t* pubkey) {
```
benma: even though it's kind of redundant, I like the convention we mostly follow to suffix `_out` to the out params. e.g. `nonce_out`, `mac_out`, etc.

src/u2f.c
```diff
@@ -39,6 +41,9 @@
 #define APPID_BOGUS_CHROMIUM "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
 #define APPID_BOGUS_FIREFOX "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
 
+#define U2F_ASSERT(cond, err) do { if (!(cond)) { Abort("U2F error: " err); } } while(0);
+#define U2F_USB_SEND(a, b) do { if (usb_send_packet(a, b) != QUEUE_ERR_NONE) {Abort("DAMMIT!");} } while(0);
```
benma: better abort message please, can't have sth like this pop up on production devices :smile: 

please use functions over macros

src/u2f.c
```diff
@@ -101,6 +106,82 @@ typedef struct __attribute__((__packed__)) {
...
+    U2F_STATE_REGISTER,
+} u2f_state_t;
+
+/**
+ * Holds state data for the U2F_STATE_AUTH workflow.
+ */
+typedef struct {
+    /** Part of the auth we're currently processing. */
+    enum {
+        U2F_STATE_AUTH_REQUESTED,
```
benma: please document each of those as clearly as possible

src/u2f.c
```diff
@@ -101,6 +106,82 @@ typedef struct __attribute__((__packed__)) {
...
+    U2F_REGISTER_REQ reg_request;
+    /** Cmd that originated this request. */
+    uint8_t cmd;
+    /** Cid that originated this request. */
+    uint32_t cid;
+    /** apdu->p1 from the original request. */
+    uint8_t apdu_p1;
+    /**
+     * U2F usage counter, incremented by the secure chip.
+     * Valid only if state == U2F_STATE_AUTH_COMPLETED.
```
benma: AUTH could be a typo here.

a lot of this struct is the same as in the above struct, consider refactoring the common keys into a struct which you embed in `u2f_auth_state_t` and `u2f_register_state_t`

src/u2f.c
```diff
@@ -101,6 +106,82 @@ typedef struct __attribute__((__packed__)) {
...
+/**
+ * Holds state data for the U2F_STATE_AUTH workflow.
+ */
+typedef struct {
+    /** Part of the auth we're currently processing. */
+    enum {
+        U2F_STATE_AUTH_REQUESTED,
+        U2F_STATE_AUTH_COMPLETED,
+        U2F_STATE_AUTH_ERROR,
+        U2F_STATE_AUTH_HIJACK
```
benma: especially this, not sure what hijack means here. also consider renaming it, because "u2f hijack" is kind of a reserved word for smuggling data over u2f when a webwallet (like myetherwallet) bridges communication to the hardware wallet.

src/usb/noise.c
```diff
@@ -214,7 +214,7 @@ void noise_rand_bytes(void* bytes, size_t size)
 // processes client messages. The first two messages are handshake messages
 // (see XX in https://noiseprotocol.org/noise.html#interactive-handshake-patterns-fundamental).
 // After, all incoming messages are decrypted and outgoing messages encrypted.
-bool bb_noise_process_msg(
+bb_noise_result_t bb_noise_process_msg(
```
benma: why was this change needed? seems like a 1:1 mapping, with `BB_NOISE_DEFER_REPLY` being unused.

src/hww.c
```diff
@@ -101,9 +101,17 @@ static void _msg(const Packet* in_packet, Packet* out_packet, const size_t max_o
     }
 
     // Process protofbuf/noise api calls.
-    if (!bb_noise_process_msg(in_packet, out_packet, max_out_len, commander)) {
+    bb_noise_result_t noise_process_result = bb_noise_process_msg(in_packet, &out_packet, USB_DATA_MAX_LEN, commander);
+    if (noise_process_result == BB_NOISE_ERROR) {
         workflow_status_create("Could not\npair with app", false);
     }
+    if (noise_process_result != BB_NOISE_DEFER_REPLY) {
```
benma: impossible result as far as i can tell

NickeZ:
Hmm I just had an idea. What if we simply print "unlock with BitBoxApp" instead of starting the unlock workflow? Then u2f would be completely async and not horrible to use anymore? We can do this in a separate PR if you wish.

It's a bit worse UX but at least you don't have to refresh the browser to trigger u2f again.

Hmm on the other hand, if the user has both the app and a u2f login running at the same time the browser will anyway lose track of the bitbox because we won't respond when the user is in the unlock workflow...

conte91:
> Hmm I just had an idea. What if we simply print "unlock with BitBoxApp" instead of starting the unlock workflow? Then u2f would be completely async and not horrible to use anymore? We can do this in a separate PR if you wish.
> 
Still that wouldn't solve the other two "flashing" screens, i.e. the confirmation screens for registration/authentication.

src/usb/noise.c
conte91: As per the commit description:
> Having the possibility of returning ok + don't send responses will allow to
defer the responses for later, thus moving the blocking calls outside of
the USB rx stack.

The idea is that this allows to start decoupling the need for responding immediately to all requests. If we want to make any workflow asynchronous in the future (so we can e.g. still detect disconnection from the app when we're in a blocking user operation and cancel it), we need to have a way to tell the USB stack that the operation is not completed yet, and so no response must be sent over USB (which is the third option for `bb_noise_result_t`). On a separate side, I think it should be possible to move this to a separate PR as it only affects the HWW stack.

conte91: This has been moved to a separate branch and will come as a future PR.

src/u2f.c
conte91: Problem is: I have no idea what this is about as well! This has been refactored from the same "hijack" code in `master`, which seems to be a very convoluted no-op. It's been left in the code not to break the existing code flow. See https://github.com/digitalbitbox/bitbox02-firmware/blob/544a6690de4b7f8fe220852849ada0cd87a500aa/src/u2f.c#L441 and https://github.com/digitalbitbox/bitbox02-firmware/blob/544a6690de4b7f8fe220852849ada0cd87a500aa/src/u2f.c#L58

NickeZ:
I tested it and on windows the "refresh website" screen dissappears by itself and you still have to refresh the webpage to make it work. Not sure if this PR introduced the regression though.

conte91:
> I tested it and on windows the "refresh website" screen dissappears by itself and you still have to refresh the webpage to make it work. Not sure if this PR introduced the regression though.

This is weird, I've fixed a similar bug just a few days ago. Maybe out of date code? I'll double check tomorrow.

NickeZ:
I can try with master tomorrow.

conte91:
> I can try with master tomorrow.

I've just retested on Windows 10 and everything seems to be working smoothly. 

> > I tested it and on windows the "refresh website" screen dissappears by itself and you still have to refresh the webpage to make it work. 

Hah, I've also read your comment better. Yes this is intended behaviour. We need to keep the BB02 locked for each request*, but it also implies we need "terminate" the "refresh" screen on our own, to prevent the BB02 being stuck in the request (as we have no way to detect whether the host has given up on the request or is just taking a long time to refresh the page).

\* If we don't lock the BB02 for the whole authentication process, we potentially allow a third party to fool the user into accepting an authentication request it didn't make - so this seems to me like the safest path to go by.

conte91:
> I tested it and on windows the "refresh website" screen dissappears by itself and you still have to refresh the webpage to make it work. Not sure if this PR introduced the regression though.

Fixed by increasing the timeout value so it's easier to notice for the user.

src/u2f.c
conte91: Old code, was refactored.

src/u2f.c
conte91: Old code, was refactored.

src/u2f.c
conte91: Old code, was refactored.

src/u2f.c
conte91: Old code, was refactored.

src/u2f.c
conte91: Old code, was refactored.

src/u2f.c
conte91: Old code, was refactored.

PR #167, opened 2019-10-17
securechip: contribute external randomness to the tempkey

benma:


benma:
Cryptoauhlib made a new releases containing many of the fixes. I made a new rebased branch for this [here](https://github.com/digitalbitbox/cryptoauthlib/tree/bitbox02-20191202) and updated this PR to use it.

src/securechip/securechip.c
```diff
@@ -484,13 +484,25 @@ static ATCA_STATUS _update_kdf_key(void)
     UTIL_CLEANUP_32(encryption_key);
     _interface_functions->get_encryption_key(encryption_key);
 
+    uint8_t nonce_contribution[32] = {0};
+    UTIL_CLEANUP_32(nonce_contribution);
```
conte91: Why do we clean this up if we're filling it just below?

src/securechip/securechip.c
```diff
@@ -484,13 +484,25 @@ static ATCA_STATUS _update_kdf_key(void)
...
+    _interface_functions->random_32_bytes(nonce_contribution);
+#if NONCE_NUMIN_SIZE > 32
```
conte91: `!= 32`? Or improve the error message if `< 32` is also allowed.

src/securechip/securechip.c
```diff
@@ -484,13 +484,25 @@ static ATCA_STATUS _update_kdf_key(void)
...
 
     return atcab_write_enc(
-        SECURECHIP_SLOT_KDF, 0, new_key, encryption_key, SECURECHIP_SLOT_ENCRYPTION_KEY);
+        SECURECHIP_SLOT_KDF,
+        0,
+        new_key,
+        encryption_key,
+        SECURECHIP_SLOT_ENCRYPTION_KEY,
+        nonce_contribution);
 }
```
conte91: Should we cleanup the nonce on exit? (Does this matter at all or it's just a waste of cycles?)

src/securechip/securechip.c
benma: THE TIL_CLEANUP_ macros mark a stack var to be zeroed when the var goes out of scope, e.g. when the function returns, using the ` __attribute__((__cleanup__())` extension.

src/securechip/securechip.c
benma: We gather 32 random bytes and pass it to the cryptoauthlib function, which uses NONCE_NUMIN_SIZE bytes of that. Technically it can be less (probably never would), so I could make the check `!=32`, but it would definitely be a problem if it was more, then it would be an overflow bug.

src/securechip/securechip.c
benma: Yes, that's what UTIL_CLEANUP_32 does.

src/securechip/securechip.c
conte91: TIL `TIL_CLEANUP`

src/securechip/securechip.c
conte91: Got it - I guess from our side we aren't guaranteed that `NONCE_NUMIN_SIZE` is 32 exactly, so a more relaxed check is better.

PR #170, opened 2019-10-21
Enable R/W access to pages 2000-2047 in the MPU.

conte91:
The default MPU configuration left by the bootloader disables
pages 2000-2047. This is pointless as it means losing 48 pages
at the end of the NVM main address space to no real benefit.

Extend the APP_DATA_1 MPU region until the end of the flash address
space; the protected bootloader section will remain R/W blocked for
the firmware as it has higher priority in the MPU table.

### DIFF
--- a/src/mpu.h
+++ b/src/mpu.h
@@ -85,5 +85,13 @@ void mpu_set_region(uint32_t rbar, uint32_t rasr);
 /** Disable the region with the given number */
 void mpu_disable_region(uint32_t region_number);
 
+/**
+ * Sets the correct MPU configuration for BitBox firmwares.
+ *
+ * Extends the APPDATA_1 MPU region up to the end of the
+ * flash memory, and sets it to R/W.
+ */
+void mpu_bitbox02_init(void);
+
 #endif
 #endif
--- a/src/mpu.h
+++ b/src/mpu.h
@@ -1,0 +1,89 @@
+// Copyright 2019 Shift Cryptosecurity AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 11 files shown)


src/mpu.h
```diff
@@ -0,0 +1,89 @@
...
+ * of its region size. Regions can overlap.
+ */
+
+/**
+ * Region numbers - as set in the MPU_RNR register (0-7).
+ * MPU rules will be evaluated using the order of the region number.
+ * The last match for a region will be the one that is enforced; so
+ * regions with the higher number have higher priority.
+ */
+#define MPU_REGION_NUMBER_FLASH (0u)
```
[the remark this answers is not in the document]
conte91: Fixed

src/firmware.c
```diff
@@ -25,11 +27,34 @@
...
+    /* Reset the last selected region to the new setting. */
+    mpu_set_region(rbar, rasr);
+}
+
 int main(void)
 {
     init_mcu();
     system_init();
     peripherals_init();
+    _memory_init();
```
benma: This should be in `common_main()` I think, as it should be handled the same way in the factorysetup firmware.

src/firmware.c
```diff
@@ -25,11 +27,34 @@
 
 uint32_t __stack_chk_guard = 0;
 
+/**
+ * Sets the correct memory configuration for this firmware.
+ *
+ * Extends the APPDATA_1 MPU region up to the end of the
+ * flash memory, and sets it to R/W.
+ *
+ * FUTURE: maybe add SmartEEPROM support here?
```
benma: remove this?

src/firmware.c
conte91: Fixed.

src/firmware.c
conte91: That will come with https://github.com/digitalbitbox/bitbox02-firmware/pull/205

conte91: Removed, as this has been moved to `mpu.c`

benma:
Please fix CI :)

PR #171, opened 2019-10-21
Fix the behaviour of deletions in trinary_input_char.

conte91:
After inserting the first letter into a trinary_input_char,
the following letters would always require backspace to be pressed
twice in order to be effective, even in the case in which one was
in the main screen (i.e. not selecting a sub-alphabet).
This was due to the code setting the in_progress variable to true
even when a letter had just been selected.

### DIFF
--- a/src/ui/components/trinary_input_char.c
+++ b/src/ui/components/trinary_input_char.c
@@ -60,21 +60,34 @@ typedef struct {
     UG_S16 horiz_space;
 } data_t;
 
-static void _set_alphabet(component_t* component, const char* alphabet)
+/**
+ * Called when a selection on one of the alphabet options has been made.
+ *
+ * If the selected alphabet is a single letter, the input is completed.
+ * Otherwise, the selected alphabet replaces the current one and the
+ * input continues.
+ */
+static void _alphabet_selected(component_t* component, const char* alphabet)
 {
     data_t* data = (data_t*)component->data;
     size_t len = strlens(alphabet);
     if (len == 0) {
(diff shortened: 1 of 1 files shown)


src/ui/components/trinary_input_char.c
```diff
@@ -61,20 +61,39 @@ typedef struct {
 } data_t;
 
 static void _set_alphabet(component_t* component, const char* alphabet)
+{
+    data_t* data = (data_t*)component->data;
```
benma: remove this function? it is too small to be worth it ;)

src/ui/components/trinary_input_char.c
```diff
@@ -61,20 +61,39 @@ typedef struct {
...
     }
     if (len == 1) {
-        // Reset elements so that the previous letters don't slide back on a new alphabet.
+        /*
+         * Reset elements so that the previous letters don't slide back
+         * on a new alphabet.
+         */
         memset(data->elements, 0, sizeof(data->elements));
         data->character_chosen_cb(component, alphabet[0]);
         data->in_progress = false;
```
benma: > This was due to the code setting the in_progress variable to true
> even when a letter had just been selected.

the letter is selected here, so in_progress=false. where was it set to true?

benma:
Oh yeah and please prefix the commit title with the context :nerd_face:

src/ui/components/trinary_input_char.c
conte91: Yeah will do.

PR #172, opened 2019-10-24
Fix touch

conte91:
Various improvements for the touch sensor behaviour. Calibration changes are still to consider.

### DIFF
--- a/external/asf4-drivers/qtouch/qtouch.c
+++ b/external/asf4-drivers/qtouch/qtouch.c
@@ -625,7 +625,7 @@ void qtouch_process_scroller_positions(void)
         // Put a threshold on this to detect whether we're touching or not.
         if (max_sensor_reading >= DEF_SCROLLER_TOUCH_THRESHOLD) {
             scroller_active[scroller] = true;
-        } else {
+        } else if (max_sensor_reading <= DEF_SCROLLER_UNTOUCH_THRESHOLD) {
             scroller_active[scroller] = false;
         }
     }
--- a/external/asf4-drivers/qtouch/qtouch.c
+++ b/external/asf4-drivers/qtouch/qtouch.c
@@ -517,24 +517,16 @@ static uint16_t sensor_previous_filtered_reading[DEF_NUM_SENSORS][DEF_SENSOR_NUM
 /* Custom sensor signal filter. */
 uint16_t qtouch_get_sensor_node_signal_filtered(uint16_t sensor_node)
 {
-    // Filter the sensor signal using the equation:
-    //     Y = X / (1 + 0.0001 * X^2)
+    // Filter the sensor signal.
     //
-    // The maximum Y value is 50 when X = 100, with a decreasing tail for X > 100.
-    // The equation can be plotted using: https://www.desmos.com/calculator
-    // and is an approximation of `Y = arctan(X)` but scaled in Y.
+    // Smooth it out and saturate it so that values never go beyond 50.
+    // This helps to mitigate 'jumpy' channels that exist at higher sensor readings when
+    // in noisy environments.
     //
-    // Smaller sensor readings will have higher relative weights, improving sensitivity of
-    // neighboring channels for better linearity and less deadband, while helping to mitigate
-    // 'jumpy' channels that exist at higher sensor readings when in noisy environments.
-    //
-    // The above empirically worked better than a sigmoid function Y = X / (1 + 0.015 * X).
-    //
-    uint8_t j;
-    double X, Y;
-    uint16_t Y_ave;
-    X = qtouch_get_sensor_node_signal(sensor_node) - qtouch_get_sensor_node_reference(sensor_node);
-    X = (X < 0) ? 0 : X;
+    uint16_t X;
+    uint16_t sensor_raw = qtouch_get_sensor_node_signal(sensor_node);
+    uint16_t sensor_reference = qtouch_get_sensor_node_reference(sensor_node);
+    X = sensor_raw < sensor_reference ? 0 : sensor_raw - sensor_reference;
     // Add more weight to edge buttons because they are physically smaller (smaller readings).
     if ((sensor_node == DEF_SCROLLER_OFFSET_0) ||
         (sensor_node == DEF_SCROLLER_OFFSET_1) ||
@@ -542,25 +534,26 @@ uint16_t qtouch_get_sensor_node_signal_filtered(uint16_t sensor_node)
         (sensor_node == DEF_SCROLLER_OFFSET_1 + DEF_SCROLLER_NUM_CHANNELS - 1)) {
         X = X * (1 + DEF_SENSOR_EDGE_WEIGHT);
     }
-    Y = X / (1 + 0.0001 * X * X);
+    // Saturate out-of-range readings.
+    X = (X > 50) ? 50 : X;
 
     // Calculate sensor readout using a moving average
     // The moving average wieghts previous N readings twice current reading
     uint16_t moving_average_cummulative_weight = 1;// Add one for current reading calculated above
-    Y_ave = (uint16_t) Y;
-    for (j = 0; j < DEF_SENSOR_NUM_PREV_POS; j++) {
+    uint16_t X_ave = X;
+    for (size_t j = 0; j < DEF_SENSOR_NUM_PREV_POS; j++) {
         moving_average_cummulative_weight += 2;
-        Y_ave += sensor_previous_filtered_reading[sensor_node][j] * 2;
+        X_ave += sensor_previous_filtered_reading[sensor_node][j] * 2;
     }
-    Y_ave = Y_ave / moving_average_cummulative_weight;
+    X_ave = X_ave / moving_average_cummulative_weight;
 
     // Update recorded previous positions
-    for (j = 0; j < DEF_SENSOR_NUM_PREV_POS - 1; j++) {
+    for (size_t j = 0; j < DEF_SENSOR_NUM_PREV_POS - 1; j++) {
         sensor_previous_filtered_reading[sensor_node][j] = sensor_previous_filtered_reading[sensor_node][j + 1];
     }
-    sensor_previous_filtered_reading[sensor_node][DEF_SENSOR_NUM_PREV_POS - 1] = Y;
+    sensor_previous_filtered_reading[sensor_node][DEF_SENSOR_NUM_PREV_POS - 1] = X;
 
-    return ((uint16_t) Y_ave);
+    return X_ave;
 }
 
 bool qtouch_is_scroller_active(uint16_t scroller)
--- a/external/asf4-drivers/qtouch/qtouch.c
+++ b/external/asf4-drivers/qtouch/qtouch.c
@@ -573,6 +573,7 @@ void qtouch_process_scroller_positions(void)
     {
         uint8_t i, j;
         uint16_t sum = 0;
+        uint16_t max_sensor_reading = 0;
         uint16_t weighted_sum = 0;
         uint16_t sensor_location[DEF_SCROLLER_NUM_CHANNELS] = {
             1,// Offset by `1` because a `0` location cannot be weight-averaged
@@ -586,6 +587,7 @@ void qtouch_process_scroller_positions(void)
             value = qtouch_get_sensor_node_signal_filtered(i + (scroller ? DEF_SCROLLER_OFFSET_1 : DEF_SCROLLER_OFFSET_0));
             sum += value;
             weighted_sum += value * sensor_location[i];
+            max_sensor_reading = (value > max_sensor_reading) ? value : max_sensor_reading;
         }
 
         // Compensate for deadband (i.e. when only a single edge button gives a reading and neighbors do not)
@@ -616,9 +618,14 @@ void qtouch_process_scroller_positions(void)
         }
         if (sum >= DEF_SCROLLER_DET_THRESHOLD) {
             scroller_previous_position[scroller][DEF_SCROLLER_NUM_PREV_POS - 1] = scaled_value;
-            scroller_active[scroller] = true;
         } else {
             scroller_previous_position[scroller][DEF_SCROLLER_NUM_PREV_POS - 1] = DEF_SCROLLER_OFF;
+        }
+        // Use the maximum value of all sensor readings as an estimate of pressure.
+        // Put a threshold on this to detect whether we're touching or not.
+        if (max_sensor_reading >= DEF_SCROLLER_TOUCH_THRESHOLD) {
+            scroller_active[scroller] = true;
+        } else {
             scroller_active[scroller] = false;
         }
     }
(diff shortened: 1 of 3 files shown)


conte91:
The commit history is pretty broken atm, will be rebased soon :)

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -604,17 +608,25 @@ void qtouch_process_scroller_positions(void)
 
         // Calculate scroller position using a moving average
         // The moving average wieghts previous N readings twice current reading
+        uint16_t estimated_scroller_position;
```
douglasbakkum: Here to L629: What is the need for these edits? Looks like no functional change except for setting the position to zero if below threshold. However, I dont think it makes sense to set it to zero - makes more sense that `qtouch_get_scroller_position` reports the last scroller position that was hit. I dont remember off hand, but it is possible that other code relies on this value being the last position hit.

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -604,17 +608,25 @@ void qtouch_process_scroller_positions(void)
...
-                    scroller_position[scroller] += scroller_previous_position[scroller][j] * 2;
+                    estimated_scroller_position += scroller_previous_position[scroller][j] * 2;
                 }
             }
-            scroller_position[scroller] += scaled_value;// Most recent signal is half weight of others to help avoid bounce when finger is released
-            scroller_position[scroller] = scroller_position[scroller] / (moving_average_cummulative_weight + 1);
+            estimated_scroller_position += scaled_value;// Most recent signal is half weight of others to help avoid bounce when finger is released
+            estimated_scroller_position = estimated_scroller_position / (moving_average_cummulative_weight + 1);
+        } else {
+            estimated_scroller_position = 0;
```
douglasbakkum: see above comment about deleting this line

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -604,17 +608,25 @@ void qtouch_process_scroller_positions(void)
 
         // Calculate scroller position using a moving average
         // The moving average wieghts previous N readings twice current reading
+        uint16_t estimated_scroller_position;
         if (sum >= DEF_SCROLLER_DET_THRESHOLD) {
```
douglasbakkum: nit for code consistency: use MAX instead of SUM?
```
if (adjusted_touch_pressure >= DEF_SCROLLER_TOUCH_THRESHOLD)
```

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -604,17 +608,25 @@ void qtouch_process_scroller_positions(void)
...
-            scroller_position[scroller] += scaled_value;// Most recent signal is half weight of others to help avoid bounce when finger is released
-            scroller_position[scroller] = scroller_position[scroller] / (moving_average_cummulative_weight + 1);
+            estimated_scroller_position += scaled_value;// Most recent signal is half weight of others to help avoid bounce when finger is released
+            estimated_scroller_position = estimated_scroller_position / (moving_average_cummulative_weight + 1);
+        } else {
+            estimated_scroller_position = 0;
+        }
+
+        // Position-independent pressure estimation.
+        if (sum >= DEF_SCROLLER_DET_THRESHOLD) {
```
douglasbakkum: can be part of above conditional (which is the same conditional) to reduce lines of code

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -623,9 +635,12 @@ void qtouch_process_scroller_positions(void)
         }
         if (sum >= DEF_SCROLLER_DET_THRESHOLD) {
```
douglasbakkum: see above nit on code consistency, which would apply here too

external/asf4-drivers/qtouch/qtouch.h
```diff
@@ -242,13 +242,15 @@ extern "C" {
 #define DEF_NUM_SCROLLERS              2// Number of scrollers (sliders or wheels)
 #define DEF_SCROLLER_NUM_CHANNELS   (DEF_NUM_CHANNELS / DEF_NUM_SCROLLERS)// Assumes same number for each scroller
 #define DEF_SCROLLER_RESOLUTION      256// Scroller resolution in bits
-#define DEF_SCROLLER_DET_THRESHOLD    32// Scroller detect threshold
+#define DEF_SCROLLER_DET_THRESHOLD    25// Scroller detect threshold
+#define DEF_SCROLLER_TOUCH_THRESHOLD    25 // Scroller active threshold
+#define DEF_SCROLLER_UNTOUCH_THRESHOLD    20 // Scroller active threshold
```
douglasbakkum: i guess these may need some trial and error with different fingers...

src/touch/gestures.c
```diff
@@ -94,6 +102,8 @@ static void _slider_state_update(gestures_detection_state_t* state, uint16_t pos
     state->velocity_sum = state->velocity_sum - velocity_removed + velocity_current;
     state->velocity_values[state->velocity_index] = velocity_current;
     state->velocity_index = (state->velocity_index + 1) % MAX_HISTORY;
+    uint16_t distance_from_start = abs((int)position - (int)state->position_start);
```
douglasbakkum: Is the `int` typecast needed? (or if not needed, then considered safer) 
If so, check for other places using `abs()` which should then also add the typecast (e.g. a couple lines below).

src/touch/gestures.c
```diff
@@ -59,11 +61,16 @@ typedef struct {
...
+     * The maximum distance the finger has travelled
+     * since starting the gesture.
+     */
+    uint16_t max_slide_travel;
     int32_t velocity_sum;
     uint16_t velocity_index;
     // Stores the velocities.
     int16_t velocity_values[MAX_HISTORY];
-    // The gesture type (tap, slide).
+    // The gesture type (tap, slide). FUTURE: remove this?
```
douglasbakkum: Looks like the changes below allow TAP and SLIDE to be independent of each other (i.e. can occur at the same time). Then best to remove gesture_type as its existence implies only one state at a time (i.e. will be confusing).

src/touch/gestures.c
```diff
@@ -35,6 +35,8 @@
 // The minimum amount of sliding difference required, so that the gesture is detected as slide, not
 // touch
 static const uint8_t SLIDE_DETECTION_DIFF = MAX_SLIDER_POS * 0.04; // Percent of slider range
+static const uint8_t TAP_NO_SLIDE_DIFF = MAX_SLIDER_POS * 0.1; // Percent of slider range
```
douglasbakkum: should update the comment above
- remove ", not touch"
- warn developers about potential future issues if both TAP and SLIDE gesture events are needed at the same time on a given slider (i.e. alternative pw entry UX concepts)

douglasbakkum: i guess this was done such that slide gesture is still responsive enough. makes sense to me (i was worried about how changes could effect slide UX but this keeps it the same as before)

douglasbakkum: nit: `TAP_NO_SLIDE_DIFF` --> `TAP_SLIDE_TOLERANCE`?

src/touch/gestures.c
conte91: My typecast expertise is a bit rusty - wouldn't the case in which state->position > position underflow the uint to <big number>, which then gets casted to int? As in the evaluation order, the subtraction will come first.

src/touch/gestures.c
conte91: It is confusing but it is what the slide processing is currently relying on. Better not to push further functional changes atm. Maybe I can open a separate issue?

src/touch/gestures.c
benma: C implicit casting is very non-intuitive. It might just be that the operands are implicitly cast as the resulting expression is cast to `int`, but I'm also not sure anymore.

I prefer explicit casting no matter the rules, for clarity.

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -517,50 +517,43 @@ static uint16_t sensor_previous_filtered_reading[DEF_NUM_SENSORS][DEF_SENSOR_NUM
...
     // Add more weight to edge buttons because they are physically smaller (smaller readings).
     if ((sensor_node == DEF_SCROLLER_OFFSET_0) ||
         (sensor_node == DEF_SCROLLER_OFFSET_1) ||
         (sensor_node == DEF_SCROLLER_OFFSET_0 + DEF_SCROLLER_NUM_CHANNELS - 1) ||
         (sensor_node == DEF_SCROLLER_OFFSET_1 + DEF_SCROLLER_NUM_CHANNELS - 1)) {
         X = X * (1 + DEF_SENSOR_EDGE_WEIGHT);
     }
-    Y = X / (1 + 0.0001 * X * X);
+    // Saturate out-of-range readings.
+    X = X > 50 ? 50 : X;
```
NickeZ: Use the `MIN()` macro instead. It conveys the meaning better.

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -517,50 +517,43 @@ static uint16_t sensor_previous_filtered_reading[DEF_NUM_SENSORS][DEF_SENSOR_NUM
...
 
     // Calculate sensor readout using a moving average
     // The moving average wieghts previous N readings twice current reading
     uint16_t moving_average_cummulative_weight = 1;// Add one for current reading calculated above
-    Y_ave = (uint16_t) Y;
-    for (j = 0; j < DEF_SENSOR_NUM_PREV_POS; j++) {
+    uint16_t X_ave = X;
+    for (int j = 0; j < DEF_SENSOR_NUM_PREV_POS; j++) {
```
NickeZ: `j` should be an `usize` since it is used in indexing.

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -517,50 +517,43 @@ static uint16_t sensor_previous_filtered_reading[DEF_NUM_SENSORS][DEF_SENSOR_NUM
...
         moving_average_cummulative_weight += 2;
-        Y_ave += sensor_previous_filtered_reading[sensor_node][j] * 2;
+        X_ave += sensor_previous_filtered_reading[sensor_node][j] * 2;
     }
-    Y_ave = Y_ave / moving_average_cummulative_weight;
+    X_ave = X_ave / moving_average_cummulative_weight;
 
     // Update recorded previous positions
-    for (j = 0; j < DEF_SENSOR_NUM_PREV_POS - 1; j++) {
+    for (int j = 0; j < DEF_SENSOR_NUM_PREV_POS - 1; j++) {
```
NickeZ: `usize` instead of `int`

external/asf4-drivers/qtouch/qtouch.h
NickeZ: For me (with my device) it worked better with the thresholds 18 and 15. Maybe we should let the user choose, like have a low/medium/high sensitivity option?

src/touch/gestures.c
```diff
@@ -94,6 +104,10 @@ static void _slider_state_update(gestures_detection_state_t* state, uint16_t pos
     state->velocity_sum = state->velocity_sum - velocity_removed + velocity_current;
     state->velocity_values[state->velocity_index] = velocity_current;
     state->velocity_index = (state->velocity_index + 1) % MAX_HISTORY;
+    uint16_t distance_from_start = abs((int)position - (int)state->position_start);
+    state->max_slide_travel = (distance_from_start > state->max_slide_travel)
```
NickeZ: use `MAX()` from `util.h`.

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -533,7 +533,9 @@ uint16_t qtouch_get_sensor_node_signal_filtered(uint16_t sensor_node)
     uint8_t j;
     double X, Y;
```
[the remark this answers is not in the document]
conte91: Fixed.

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -542,7 +544,7 @@ uint16_t qtouch_get_sensor_node_signal_filtered(uint16_t sensor_node)
         (sensor_node == DEF_SCROLLER_OFFSET_1 + DEF_SCROLLER_NUM_CHANNELS - 1)) {
         X = X * (1 + DEF_SENSOR_EDGE_WEIGHT);
     }
-    Y = X / (1 + 0.0001 * X * X);
+    Y = X > 50 ? 50 : X;//Y = X / (1 + 0.0001 * X * X);
```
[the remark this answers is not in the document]
conte91: Fixed.

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -580,6 +582,7 @@ void qtouch_process_scroller_positions(void)
     {
         uint8_t i, j;
         uint16_t sum = 0;
+        uint16_t adjusted_touch_pressure = 0;
```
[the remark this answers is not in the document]
conte91: Fixed.

external/asf4-drivers/qtouch/qtouch.c
conte91: Dropped to a separate branch.

external/asf4-drivers/qtouch/qtouch.h
conte91: +1 for the sensitivity option. But that will require additional time and cooperation with the app. How about we leave the current values and we track this in a separate issue?

src/touch/gestures.c
conte91: Fixed.

external/asf4-drivers/qtouch/qtouch.c
conte91: As in `size_t`?

I'd rather have `int`s as much as possible in this simple situations, they make the code simpler to read and less of a mental overhead (imho). Happy to swap to `size_t` or similar if we like it better.

external/asf4-drivers/qtouch/qtouch.c
conte91: The `util.h` file is in our main `src` folder, so it's not seen by this file. (The issue is that this code shouldn't be located here to start with...) Maybe we can just reimplement the macro at the top of this file?

src/touch/gestures.c
conte91: Yep I'll leave explicit casting to the places in which it's unsure what the signedness of a subtraction might be.

external/asf4-drivers/qtouch/qtouch.c
```diff
@@ -593,6 +596,7 @@ void qtouch_process_scroller_positions(void)
             value = qtouch_get_sensor_node_signal_filtered(i + (scroller ? DEF_SCROLLER_OFFSET_1 : DEF_SCROLLER_OFFSET_0));
             sum += value;
             weighted_sum += value * sensor_location[i];
+            adjusted_touch_pressure = value > adjusted_touch_pressure ? value : adjusted_touch_pressure;
```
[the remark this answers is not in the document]
conte91: Fixed.

src/touch/gestures.c
conte91: Fixed.

external/asf4-drivers/qtouch/qtouch.c
conte91: Fixed.

external/asf4-drivers/qtouch/qtouch.c
conte91: Fixed.

external/asf4-drivers/qtouch/qtouch.h
NickeZ: Yes for sure, let's do that in a later stage!

PR #173, opened 2019-10-25
release: offbyn pubkey and signatures v4.1.x

offbyn:


benma:
Thanks, this is much appreciated!

PR #176, opened 2019-10-28
Systeminfo

benma:


src/hww.c
```diff
@@ -67,6 +70,52 @@ static void _api_attestation(const Packet* in_packet, Packet* out_packet)
...
+ * - For the BitBoxBase platform:
+ " - 0x00 - Standard
+ * 1 byte: 0x00 if the device is locked, 0x01 if the device is unlocked.
+ * @param[out] buf serialize info to this buffer.
+ * @return number of bytes written
+ */
+static size_t _api_info(uint8_t* buf)
+{
+    uint8_t* current = buf;
+    { // 12 bytes version string, 0x00-padded
```
NickeZ: I don't think the introduced scope here helps with anything. There are no local variables. Same applies below.

src/hww.c
```diff
@@ -67,6 +70,52 @@ static void _api_attestation(const Packet* in_packet, Packet* out_packet)
...
+        memset(current, 0x00, 12);
+        strncpy((char*)current, DIGITAL_BITBOX_VERSION_SHORT, 12);
```
NickeZ: Is the idea here to not write a null-terminated string in case the version takes 12 bytes? Because strncpy will not null terminate.

src/hww.c
```diff
@@ -67,6 +70,52 @@ static void _api_attestation(const Packet* in_packet, Packet* out_packet)
     memcpy(data, result.challenge_signature, sizeof(result.challenge_signature));
 }
 
+/**
+ * Serializes sytem information to the buffer. The first 12 bytes are considered stable to extract
+ * the version and should never change.
+ * The following bytes are written:
+ * 12 bytes: short firmware version string, ascii encoded, right-padded with 0x00. E.g. "v4.12.2".
```
NickeZ: I would prefer if it was three bytes or maybe 6 `uint16_t` to encode the version. A string is quite strange in this case.

If you want to send a string I think it should be length prefixed so that we can have arbitrary length strings. Which would enable us to send the "full version". Maybe @conte91 has some opinions here too.

conte91: We can assume data integrity at this point of the protocol, so sending the string _should_ be fine; I second that prefixing it with the length field allows for easier future changes though. IMHO we should have:

- Version ID (4 byte release date, like decimal 20191028)
- Version string (12 bytes, right padded with 0s)

Also on the same topic: I usually like to prepend a version byte to all packets that are being transmitted, this allows for super-easy backwards compatibility! So my actual ideal would be 1 byte packet version, 4 bytes app version ID, 12 bytes app version string.

src/hww.c
benma: no, as the comment says it's zero padded only.

src/hww.c
benma: @conte91 the firmware version itself can define all the rest of the api. per-packet versioning sounds like overkill to me. keep in mind that we have protobuf for the encrypted communication, where it's easy to extend on a per-packet basis.

not sure what the 4 date bytes would accomplish - it would be something to diligently bookkeep before every release, but I would avoid that unless there is a machine which reads and uses this data.

@NickeZ i could length-prefix it, but is it necessary? the short version of the semver will never be longer than 12 chars. with a prefix. it makes parsing and serializing more complicated too.

I used the string because this is what the usb descriptor uses, and version.h defines. Does not seem like it's worth it to add a different encoding now. But if we did, why do you suggest 6 uint16? Don't you mean 3 (major, minor, patch)?

NickeZ: Yeah i did mean 3 uint16. This semver version: "1000.1000.1000" is longer than 12 chars.

benma: then again we will never have a version like this ;) i'm ok though adding a prefix len byte :+1:

NickeZ: Yeah I fixed length strings are easier to parse, but zero terminated strings/arrays are the root of all evil ;)
If you think 12 characters are enough then fine. The only drawback I see is that we can't print a full semver version such as X.Y.Z-prerelease+metadata, which might come in handy during development.

NickeZ: > keep in mind that we have protobuf for the encrypted communication, where it's easy to extend on a per-packet basis.

Yeah we use protobuf, but do we always ensure backwards compatibility of the messages?

> not sure what the 4 date bytes would accomplish - it would be something to diligently bookkeep before every release, but I would avoid that unless there is a machine which reads and uses this data (I very much dislike unused code :P).

You could put in the date of the last commit. This would both be reproducible and could come in handy.

NickeZ: > You could put in the date of the last commit. This would both be reproducible and could come in handy.

But not for this PR, that would be something we could add later in that case.

src/hww.c
benma: done. even with no local vars i did it so it's visually easy to make out the individual fields in the response.

src/hww.c
benma: @NickeZ i updated with a len prefix pls check

> but zero terminated strings/arrays are the root of all evil ;)

it was not zero terminated though ;) just zero padded :laughing:

NickeZ: yeah but how do you find the end of the string? ;)

PR #186, opened 2019-10-31
Add USART stack for the BBBase firmware

conte91:
Still to do some cleanup/integration - the `#ifdef BBBASE_HMS_BOARD`s are also to be updated to the "real" ones.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -61,6 +61,8 @@ set(DBB-FIRMWARE-SOURCES
   ${CMAKE_SOURCE_DIR}/src/apps/btc/btc_params.c
   ${CMAKE_SOURCE_DIR}/src/apps/btc/btc_sign.c
   ${CMAKE_SOURCE_DIR}/src/apps/btc/btc_bip143.c
+  ${CMAKE_SOURCE_DIR}/src/queue.c
+  ${CMAKE_SOURCE_DIR}/src/usb/usb_processing.c
 )
 set(DBB-FIRMWARE-SOURCES ${DBB-FIRMWARE-SOURCES} PARENT_SCOPE)
 
@@ -69,8 +71,6 @@ set(DBB-FIRMWARE-USB-SOURCES
   ${CMAKE_SOURCE_DIR}/src/usb/usb_frame.c
   ${CMAKE_SOURCE_DIR}/src/usb/usb_packet.c
   ${CMAKE_SOURCE_DIR}/src/u2f/u2f_packet.c
-  ${CMAKE_SOURCE_DIR}/src/usb/usb_processing.c
-  ${CMAKE_SOURCE_DIR}/src/queue.c
 )
 set(DBB-FIRMWARE-USB-SOURCES ${DBB-FIRMWARE-USB-SOURCES} PARENT_SCOPE)
 
@@ -122,6 +122,8 @@ set(DBB-BOOTLOADER-SOURCES
   ${CMAKE_SOURCE_DIR}/src/bootloader/bootloader.c
   ${CMAKE_SOURCE_DIR}/src/bootloader/startup.c
   ${CMAKE_SOURCE_DIR}/src/bootloader/mpu.c
+  ${CMAKE_SOURCE_DIR}/src/queue.c
+  ${CMAKE_SOURCE_DIR}/src/usb/usb_processing.c
   ${CMAKE_SOURCE_DIR}/src/ui/ugui/ugui.c
   ${CMAKE_SOURCE_DIR}/src/ui/fonts/font_a_9X9.c
   ${CMAKE_SOURCE_DIR}/src/screen.c
@@ -148,6 +150,9 @@ set(PLATFORM-BITBOX02-SOURCES ${PLATFORM-BITBOX02-SOURCES} PARENT_SCOPE)
 
 set(PLATFORM-BITBOXBASE-SOURCES
   ${CMAKE_SOURCE_DIR}/src/platform/bitboxbase/leds.c
+  ${CMAKE_SOURCE_DIR}/src/usart/usart.c
+  ${CMAKE_SOURCE_DIR}/src/usart/usart_frame.c
+  ${CMAKE_SOURCE_DIR}/src/usart/usart_hww.c
 )
 set(PLATFORM-BITBOXBASE-SOURCES ${PLATFORM-BITBOXBASE-SOURCES} PARENT_SCOPE)
 
--- a/src/bitboxbase/bitboxbase.c
+++ b/src/bitboxbase/bitboxbase.c
@@ -16,11 +16,14 @@
 #include "driver_init.h"
 #include "firmware_main_loop.h"
 #include "hardfault.h"
+#include "hww.h"
 #include "platform_init.h"
 #include "qtouch.h"
 #include "screen.h"
 #include "ui/oled/oled.h"
 #include "ui/screen_process.h"
+#include "usart/usart.h"
+#include "usb/usb_processing.h"
 #include "util.h"
 #include "workflow/workflow.h"
 
@@ -38,10 +41,14 @@ int main(void)
     screen_init();
     screen_splash();
     // qtouch_init();
+    usart_start();
+    hww_setup();
     common_main();
     traceln("%s", "Device initialized");
     for (;;) {
         screen_process();
+        usart_receive();
+        usb_processing_process(usb_processing_hww());
     }
     return 0;
 }
--- a/src/bootloader/bootloader.c
+++ b/src/bootloader/bootloader.c
@@ -31,6 +31,7 @@
 #include <ui/ugui/ugui.h>
 #if PLATFORM_BITBOXBASE == 1
 #include <usart/usart.h>
+#include <usart/usart_frame.h>
 #elif PLATFORM_BITBOX02 == 1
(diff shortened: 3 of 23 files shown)


conte91:
Depends on https://github.com/digitalbitbox/bitbox02-firmware/pull/184, https://github.com/digitalbitbox/bitbox02-firmware/pull/191.

benma:
@conte91 I can't get this to build

```
':
/home/marko/coding/dbb/bitbox02-firmware/src/usart/usart_frame.c:196:5: error: switch missing default case [-Werror=switch-default]
     switch (_usart_frame_parser_state.state) {
     ^~~~~~
/home/marko/coding/dbb/bitbox02-firmware/src/usart/usart_frame.c:226:5: error: 'else' without a previous 'if'
     else
     ^~~~
/home/marko/coding/dbb/bitbox02-firmware/src/usart/usart_frame.c:232:5: error: break statement not within loop or switch
     break;
     ^~~~~
/home/marko/coding/dbb/bitbox02-firmware/src/usart/usart_frame.c:233:1: error: 'default' label not within a switch statement
 default:
 ^~~~~~~
/home/marko/coding/dbb/bitbox02-firmware/src/usart/usart_frame.c: At top level:
/home/marko/coding/dbb/bitbox02-firmware/src/usart/usart_frame.c:236:1: error: expected identifier or '(' before '}' token
 }

```

Some of those look like proper syntax errors unless it's some esoteric C I don't know ;)

conte91:
> @conte91 I can't get this to build
> 

Crapped out the code while merging it with the newest master... Repushing in a minute.

src/bitboxbase/bitboxbase.c
```diff
@@ -37,11 +40,15 @@ int main(void)
...
-    // qtouch_init();
+    qtouch_init();
+    usart_start();
+    hww_setup();
     common_main();
     traceln("%s", "Device initialized");
     for (;;) {
         screen_process();
+        usart_receive();
+        usb_processing_process(usb_processing_hww());
```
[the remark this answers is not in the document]
NickeZ: OK, future refactoring then!

src/CMakeLists.txt
```diff
@@ -148,6 +151,13 @@ set(PLATFORM-BITBOX02-SOURCES ${PLATFORM-BITBOX02-SOURCES} PARENT_SCOPE)
 
 set(PLATFORM-BITBOXBASE-SOURCES
   ${CMAKE_SOURCE_DIR}/src/platform/bitboxbase/leds.c
+  ${CMAKE_SOURCE_DIR}/src/usart/usart.c
+  ${CMAKE_SOURCE_DIR}/src/usart/usart_frame.c
+  ${CMAKE_SOURCE_DIR}/src/usart/usart_hww.c
+  ${CMAKE_SOURCE_DIR}/external/asf4-drivers/hal/src/hal_usart_async.c
```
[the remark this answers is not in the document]
conte91: Fixed.

src/bitboxbase/bitboxbase.c
```diff
@@ -37,11 +40,15 @@ int main(void)
     __stack_chk_guard = common_stack_chk_guard();
     screen_init();
     screen_splash();
-    // qtouch_init();
+    qtouch_init();
```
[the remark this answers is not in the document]
conte91: Fixed.

src/usart/usart_hww.c
```diff
@@ -0,0 +1,141 @@
...
+static uint8_t _usart_tx_buf[USART_TX_BUFFER_SIZE];
+
+/** Set when the send channel is busy sending data. */
+static bool _send_busy = false;
+
+/** Pointer to our UART descriptor. */
+static struct usart_async_descriptor* _usart_descriptor;
+
+/**
+ * Sends the next frame, if the USB interface is ready.
```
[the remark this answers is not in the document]
conte91: Fixed.

src/CMakeLists.txt
```diff
@@ -69,7 +71,6 @@ set(DBB-FIRMWARE-USB-SOURCES
   ${CMAKE_SOURCE_DIR}/src/usb/usb_frame.c
   ${CMAKE_SOURCE_DIR}/src/usb/usb_packet.c
   ${CMAKE_SOURCE_DIR}/src/u2f/u2f_packet.c
-  ${CMAKE_SOURCE_DIR}/src/usb/usb_processing.c
   ${CMAKE_SOURCE_DIR}/src/queue.c
```
NickeZ: `queue.c` should probably be removed here since it was added above.

src/bootloader/bootloader.c
```diff
@@ -929,7 +959,7 @@ void bootloader_jump(void)
 #ifdef BOOTLOADER_DEVDEVICE
 #if PLATFORM_BITBOXBASE == 1
         // We don't have touch on base dev bootloader yet
-        (void)_devdevice_enter;
+        _base_wait_fwupdate();
```
NickeZ: Could you remove all special handling in `bootloader_jump()`. Since this PR will add the missing functionality (communication to the bootloader). I think it makes sense to revert to the bitbox02 behaviour.

We should stay in the bootloader until we are instructed to boot via the `OP_REBOOT` command. If it is a dev bootloader we should print the same text as on BB02 and wait for user to either boot or to enter bootloader.

You can probably just check out this file from the `upstream/master` branch.

src/firmware_main_loop.c
```diff
@@ -15,6 +15,7 @@
 #include "firmware_main_loop.h"
 
 #include "ui/screen_process.h"
+#include "usart/usart.h"
```
NickeZ: stray header?

src/usart/usart_frame.c
```diff
@@ -0,0 +1,347 @@
...
+        return;
+    }
+    _usart_frame_parser_state.buf[_usart_frame_parser_state.packet_size] = b;
+    _usart_frame_parser_state.packet_size++;
+}
+
+static void _usart_frame_process_byte(uint8_t b)
+{
+    switch (_usart_frame_parser_state.state) {
+    case USART_FRAME_PARSE_IDLE:
```
NickeZ: I don't think this will work if the packets come "back-to-back", i.e. with only 1 flag byte between the payloads. Is this by intention or? because I guess it would be quite simple to just start eating bytes if you are in IDLE and receive anything other than `FLAG_BYTE`.

NickeZ: You could basically have a fall-thru to `USART_FRAME_PARSE_RX`. if you break in the conditional.

src/usb/usb_processing.c
```diff
@@ -58,35 +62,27 @@ static Packet _in_packet;
 // TODO: remove this global in future refactoring
 static struct queue* _global_queue;
```
NickeZ: Could you remove this `_global_queue`? I guess it isn't used anymore.

src/workflow/workflow.c
```diff
@@ -45,7 +49,13 @@ void workflow_confirm_dismiss(const char* title, const char* body)
 
 void workflow_start(void)
 {
+#if PLATFORM_BITBOXBASE == 1
+    usart_start();
+    hww_setup();
+    usart_endpoint_enable(USART_ENDPOINT_HWW);
+#else
```
NickeZ: I prefer to be explicit here. Use `#elif PLATFORM_BITBOX02` so that it doesn't work by default if we add a new platform in the future.

src/workflow/workflow.c
```diff
@@ -65,6 +75,10 @@ static void _select_orientation_done(bool upside_down)
 
 void workflow_start_orientation_screen(void)
 {
+#if PLATFORM_BITBOXBASE == 1
+    _select_orientation_done(false);
+#else
```
NickeZ: See above comment about `#else`.

src/bootloader/bootloader.c
```diff
@@ -28,6 +29,8 @@
 #include <ui/components/ui_images.h>
 #include <ui/oled/oled.h>
 #include <ui/ugui/ugui.h>
+#include <usart/usart.h>
```
benma: ifdef guards

src/usart/usart_frame.c
conte91: That won't work for b2b packets. The correct way of moving forward here is to jump from end of packet straight into rx. `USART_FRAME_PARSE_IDLE` should only be entered when starting or after an error.

src/bootloader/bootloader.c
conte91: Fixed

src/usart/usart_frame.c
conte91: Fixed.

src/workflow/workflow.c
conte91: Fixed

src/workflow/workflow.c
conte91: Fixed

src/usb/usb_processing.c
conte91: Fixed

src/bootloader/bootloader.c
conte91: Semi-fixed - Some changes are still necessary for the USART to be initialized instead of the USB stack.

NickeZ:
Overall looks good to me now. Could you add some unit-tests of queue and the parser?

PR #190, opened 2019-11-01
release: offbyn signatures for v4.2.x

offbyn:


benma:
Thanks again, this is helpful!

PR #191, opened 2019-11-01
Fix BitBoxBase API for LEDs

conte91:
Improve the firmware API for turning LEDs on/off in the BitBoxBase.

Depends on https://github.com/digitalbitbox/bitbox02-firmware/pull/184.

### DIFF
--- a/src/platform/bitboxbase/leds.c
+++ b/src/platform/bitboxbase/leds.c
@@ -15,36 +15,74 @@
 #include "leds.h"
 #include "bitboxbase_pins.h"
 
-static void _turn_small_led(int led, bool level)
+#include <stdint.h>
+
+/** List of small LEDs (leftmost to rightmost) */
+static uint8_t small_leds[] = {PIN_LED_SMALL(0),
+                               PIN_LED_SMALL(1),
+                               PIN_LED_SMALL(2),
+                               PIN_LED_SMALL(3),
+                               PIN_LED_SMALL(4)};
+
+void leds_turn_small_led(int led, bool enabled)
 {
-    gpio_set_pin_level(led, level);
+    gpio_set_pin_level(small_leds[led], enabled);
 }
 
-static void _turn_big_led(int led, bool level)
+/**
+ * Turn on/off the given RBG LED pin. Note that these pins
+ * are active low!
+ */
+static void _leds_turn_big_led_component(uint8_t pin, bool level)
 {
-    gpio_set_pin_level(led, !level);
+    gpio_set_pin_level(pin, !level);
 }
 
-typedef enum { LED_WHITE, LED_RED, LED_GREEN, LED_BLUE } led_color_t;
+/** Position of each component in each RGB LED. */
+static const uint8_t big_leds[2][3] = {{PIN_BLED0_R, PIN_BLED0_G, PIN_BLED0_B},
+                                       {PIN_BLED1_R, PIN_BLED1_G, PIN_BLED1_B}};
 
-static const uint8_t big_leds[] =
-    {PIN_BLED0_R, PIN_BLED0_G, PIN_BLED0_B, PIN_BLED1_R, PIN_BLED1_G, PIN_BLED1_B};
-
-static void _set_pin_directions(void)
+void leds_turn_big_led(int led, led_color_t color)
 {
-    for (int i = 0; i < 5; ++i) {
-        gpio_set_pin_direction(PIN_LED_SMALL(i), GPIO_DIRECTION_OUT);
-        _turn_small_led(PIN_LED_SMALL(i), false);
+    switch (color) {
+    case LED_COLOR_WHITE:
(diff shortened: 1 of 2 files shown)


src/platform/bitboxbase/leds.c
```diff
@@ -15,33 +15,69 @@
...
+    gpio_set_pin_level(pin, !level);
 }
 
-typedef enum { LED_WHITE, LED_RED, LED_GREEN, LED_BLUE } led_color_t;
+static const uint8_t big_leds[2][3] = {{PIN_BLED0_R, PIN_BLED0_G, PIN_BLED0_B},
+                                       {PIN_BLED1_R, PIN_BLED1_G, PIN_BLED1_B}};
 
-static const uint8_t big_leds[] =
-    {PIN_BLED0_R, PIN_BLED0_G, PIN_BLED0_B, PIN_BLED1_R, PIN_BLED1_G, PIN_BLED1_B};
+void turn_big_led(int led, led_color_t color)
```
NickeZ: I would like to have something like `enum led_big_t {LED_BIG_LEFT, LED_BIG_RIGHT}` and then take `enum led_big_t` instead of `int` in `turn_big_led`.

NickeZ: `turn_big_led` should be prefixed with `leds` to be consistent. We always prefix non-static functions with the filename.

src/platform/bitboxbase/leds.h
```diff
@@ -12,6 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#ifndef __BITBOXBASE_LEDS_H
+#define __BITBOXBASE_LEDS_H
+
+#include <stdbool.h>
 typedef enum { SMALL_LED, BIG_LED } led_t;
```
NickeZ: I guess this `led_t` could be removed.

src/platform/bitboxbase/leds.h
```diff
@@ -12,6 +12,18 @@
...
 
+typedef enum { LED_WHITE, LED_RED, LED_GREEN, LED_BLUE, LED_NONE } led_color_t;
```
NickeZ: I think it would be nice to prefix the enumerations with `LEDS` (filename). Maybe `LEDS_COLOR_WHITE` or just `LEDS_WHITE`?

src/platform/bitboxbase/leds.h
```diff
@@ -12,6 +12,18 @@
...
+
 void leds_init(void);
+
+void turn_small_led(int led, bool level);
```
NickeZ: `bool level` sounds a bit awkard to me. I would change the variable name to `bool enabled`.

src/platform/bitboxbase/leds.c
```diff
@@ -15,33 +15,69 @@
...
 
-static void _turn_small_led(int led, bool level)
+#include <stdint.h>
+static uint8_t small_leds[] = {PIN_LED_SMALL(0),
+                               PIN_LED_SMALL(1),
+                               PIN_LED_SMALL(2),
+                               PIN_LED_SMALL(3),
+                               PIN_LED_SMALL(4)};
+
+void turn_small_led(int led, bool level)
```
NickeZ: Change `int led` to something like `enum leds_small_led_t`. You should then switch over the enumeration instead of directly passing the pin to `gpio_set_pin_level`.

NickeZ:
I will avoid touching the leds in my PR, so that you don't get any complicated rebases.

src/platform/bitboxbase/leds.h
conte91: `LED_COLOR_WHITE`? As the data type is `led_color_t` (Not prefixing the filename, but this is not a function...)

src/platform/bitboxbase/leds.c
conte91: +1 for left/right LED instead of 0/1. At the same time though, the small LEDs are better indexed by number 0-4, so consistency might be a factor here.

src/platform/bitboxbase/leds.h
conte91: yeah that's a leftover. Removed.

src/platform/bitboxbase/leds.c
conte91: Passing the index of the LED (0 -> leftmost, 4 -> rightmost) is the best solution imho.

src/platform/bitboxbase/leds.h
conte91: Removed.

src/platform/bitboxbase/leds.c
conte91: I've put in a consistent numbering for both small and big leds, I'd stick to that if everybody agrees.

conte91: Fixed.

PR #193, opened 2019-11-04
ui/oled/oled: simplify drawing code

NickeZ:
By configuring the device correctly we don't need to move around bits.

### DIFF
--- a/src/ui/oled/oled.c
+++ b/src/ui/oled/oled.c
@@ -137,7 +137,7 @@ void oled_init(void)
     _write_command(OLED_CMD_SET_DISPLAY_OFF);
     // Set brightness
     _write_command(OLED_CMD_SET_CONTRAST_CONTROL_FOR_BANK0);
-    _write_command(0x80); /* 0x00..0xff */
+    _write_command(0xff); /* 0x00..0xff */
     _write_command(0x21); // Set vertical addressing mode
     // Set scan directions for our non-mirrored orientation
     _write_command(OLED_CMD_SET_SEGMENT_RE_MAP_COL127_SEG0);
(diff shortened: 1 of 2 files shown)


NickeZ:
I did also maximize the brightness. I could turn that down again.

src/ui/oled/oled.c
```diff
@@ -134,27 +135,30 @@ void oled_init(void)
 
     // Initialize
     _write_command(OLED_CMD_SET_DISPLAY_OFF);
-    _write_command(OLED_CMD_SET_LOW_COL(0));
-    _write_command(OLED_CMD_SET_HIGH_COL(0));
-    _write_command(OLED_CMD_SET_PAGE_START_ADDRESS(0));
-    _write_command(0xdc); /*set display start line*/
-    _write_command(0x00); // +0x7f
     _write_command(OLED_CMD_SET_CONTRAST_CONTROL_FOR_BANK0);
-    _write_command(0x80); /* 0x00..0xff */
```
benma: could you please split this particular change into a new commit and combine it with using `OLED_CMD_SET_CONTRAST_CONTROL_FOR_BANK0` in `oled_brightness()`, and using `oled_brightness()` here? for clarity.

NickeZ: I removed that function instead. It was dead code and I'll have to make larger changes to this code anyway due to the bitboxbase.

NickeZ:
> tested ACK. Don't notice any difference in animation speeds.

Yeah, that was a bit of a surprise to me. I guess we have `delay_ms` somewhere that has a greater impact. Or maybe it is all the work we are doing in interrupts.

The similar changes on the base branch made the animation blazingly fast.

conte91:
utACK.

PR #194, opened 2019-11-05
bootloader: OP_VERSIONS returns wrong length

benma:
Judging by the out of date comment, it returns 12 bytes instead of 8
bytes by mistake from a previous iteration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,7 @@ endif()
 # Example 'v1.0.0'. They MUST not contain a pre-release label such as '-beta'.
 set(FIRMWARE_VERSION "v4.3.0")
 set(FIRMWARE_BTC_ONLY_VERSION "v4.3.0")
-set(BOOTLOADER_VERSION "v1.0.1")
+set(BOOTLOADER_VERSION "v2.0.0")
 
 find_package(PythonInterp 3.6 REQUIRED)
 
--- a/src/bootloader/bootloader.c
+++ b/src/bootloader/bootloader.c
@@ -702,7 +702,7 @@ static size_t _api_set_show_firmware_hash(const uint8_t* input, uint8_t* output)
 }
 
 /*
- * output filled with bootloader version | firmware version | signing pubkeys version
+ * output filled with firmware version | signing pubkeys version
  */
 static size_t _api_versions(uint8_t* output)
 {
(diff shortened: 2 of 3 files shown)


benma:
The two clients we have handle both versions (before and after fix) the same.

I also added RELEASE-NOTES, where we can document api and feature changes as we go. This should help in figuring out if and when a new release is necessary and what went into it when we release.

CMakeLists.txt
```diff
@@ -87,7 +87,7 @@ endif()
 # Example 'v1.0.0'. They MUST not contain a pre-release label such as '-beta'.
 set(FIRMWARE_VERSION "v4.3.0")
 set(FIRMWARE_BTC_ONLY_VERSION "v4.3.0")
-set(BOOTLOADER_VERSION "v1.0.1")
+set(BOOTLOADER_VERSION "v2.0.0")
```
benma: wasn't sure about this one; technically the api is not compatible so I bumped the major version, even though all clients continue to work without any change. :man_shrugging:

src/bootloader/bootloader.c
```diff
@@ -713,7 +713,7 @@ static size_t _api_versions(uint8_t* output)
         (const uint8_t*)&data->fields.signing_pubkeys_version,
         sizeof(version_t));
     _report_status(OP_STATUS_OK, output);
-    return BOOT_OP_LEN + sizeof(version_t) * 3;
+    return BOOT_OP_LEN + sizeof(version_t) * 2;
```
NickeZ: naked number 2 should be a define.

CMakeLists.txt
NickeZ: This seems like a bugfix that isn't breaking backwards compatibility. So I would bump the patch number. Any client "following" the old length-number would be reading out of bounds right?

src/bootloader/bootloader.c
benma: a define? not sure how it helps, the two versions are written out just before.

maybe the `len += ...` pattern after each write is more robust.

CMakeLists.txt
benma: It's an edge case; purely technically it is a backwards incompatible change. Just that our existing clients are not parsing the responses strictly enough for it to matter in practice (they could check for the expected length and raise an error on mismatch, in which case the new clients would actually break without a version check).

In the end, major or minor does not have any real impact or consequence, does it?

the third option is just add a comment to the buggy line and state why it is the wrong length, without changing it, as it doesn't have any real impact.

NickeZ: It still seems to me that undocumented features are not part of the API. Bumping the bootloader version will enforce all clients to have two code paths which seems like a maintenance burden.

Actually reading the documentation it seems like the bug is that bootloader version is missing and that the other fields have an invalid offset. If you want to bump the version, why don't you add the bootloader version to the command?

NickeZ:
We agreed to not break backwards compatibility over 4 bytes.

PR #196, opened 2019-11-05
Base qtouch

NickeZ:


NickeZ:
To test this you can build and run the `fw_bitboxbase_touch_demo.elf`. The jlink script to flash is in `build/test/device-test`.

src/hardfault.c
```diff
@@ -14,6 +14,7 @@
 
 #include "hardfault.h"
 #include "util.h"
+#include <platform_config.h>
```
conte91: This should go in a separate commit I guess?

src/touch/gestures.c
```diff
@@ -44,7 +46,7 @@ extern volatile uint8_t measurement_done_touch;
 
 /********************************** STATE **********************************/
 
-static const uint16_t LONG_TOUCH = 500;
+static const uint16_t LONG_TOUCH = 500 * 2;
```
conte91: ...1000? Also this will affect the BB02 behaviour!

src/touch/gestures.h
```diff
@@ -39,6 +39,11 @@ typedef struct {
     int32_t velocity;
 } gestures_slider_data_t;
 
+enum button_id_t {
```
conte91: `bitboxbase_button_id_t`? Or guard it with `#ifdef`s so that we're sure it doesn't affect BB02 stuff.

src/touch/gestures.h
```diff
@@ -39,6 +39,11 @@ typedef struct {
...
+    BUTTON_LEFT,
```
conte91: `BITBOXBASE_BUTTON_{LEFT,RIGHT}`

src/hardfault.c
NickeZ: It is ;)

src/touch/gestures.c
NickeZ: ack

src/touch/gestures.h
NickeZ: ack

src/touch/gestures.h
NickeZ: ack

PR #197, opened 2019-11-05
workflow/reboot: Reboot immediately on BitBoxBase

NickeZ:


src/workflow/reboot.c
```diff
@@ -39,9 +39,12 @@ static void _reboot(void)
 
 bool workflow_reboot(void)
 {
+#if PLATFORM_BITBOX02 == 1
```
benma: maybe slightly more future-robust if you checked for the base platform instead, as this could misbehave if we add another platform in the future :sweat_smile:

PR #200, opened 2019-11-06
py: The protobufs have become out of sync

NickeZ:


benma:
Is this due to an update in the generator, or something else?

NickeZ:
Don't know, Is it different when you compile it in your docker container?

benma:
I pulled shiftcrypto/firmware_v2:7, did `make clean; make firmware`, and did not get any modified files. When I re-ran this after deleting all the Python protobuf files, I did get similar modifications.

So it seems like a small issue with the cmake files, probably the files should be generated every time.

NickeZ:
Did you do `touch messages/hww.proto` or something like this? git might checkout the python files with younger dates than the proto files. Nothing Make can do about it..

benma:
> Did you do `touch messages/hww.proto` or something like this? git might checkout the python files with younger dates than the proto files. Nothing Make can do about it..

No I didn't. I see that could be the issue. 

I don't even see where the Python protobuf generator is installed, or if/how it was updated so that the files change. Anyway, it's not a big deal :P

PR #202, opened 2019-11-07
Unit test for the USART frame parser

conte91:


benma:
Here are some Go unit tests in case it helps:

https://github.com/digitalbitbox/bitbox02-api-go/blob/master/communication/usart/usart_test.go

The main one basically creates random data and checks that decode(encode(x))==x, and some additional stuff.

conte91:
> Could you write some more simple tests, where I can see some encoded strings and that they get properly decoded? And also some tests where an invalid input is detected as invalid.

That should be the exact purpose of this test :) As the mocked `usb_processing_enqueue` will check that the (decoded) buffer matches what was originally encoded by the test (`expect_memory` stuff). Also the `test_short_msg` and `test_long_msg` tests will create invalid messages that should get dropped (no `expect_anything` calls), so anything going through will create an error.

test/unit-test/test_usart.c
```diff
@@ -0,0 +1,377 @@
...
+// limitations under the License.
+
+#include <setjmp.h>
+#include <stdarg.h>
+#include <stddef.h>
+#include <stdlib.h>
+#include <cmocka.h>
+
+// including C file to test static functions
+#include <commander/commander.c>
```
NickeZ: Is this needed? We haven't typically tested static functions before. The motivation has been that all code should be accessible through public functions.

conte91: Nope, these are stray includes from `test_hww.h`. I'm pushing a cleanup for these.

conte91: Fixed.

PR #210, opened 2019-11-20
py: add pyserial to requirements.txt

benma:
Needed for the UART connection to the BitBoxBase HSM.

Dockerfile including this was build and pushed with :8.


NickeZ:
8 is already taken by my rust container ;)

PR #212, opened 2019-11-20
bitboxbase: Add new message definitions

NickeZ:
Adding new message definitions in line with our discussions.

cc @benma @conte91 @stadicus


NickeZ:
You can see the new definitions in `messages/bitboxbase.proto`

messages/bitboxbase.proto
```diff
@@ -0,0 +1,79 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+// Should be sent every X seconds (TBD) unless the firmware already is busy with a command.
+message BitBoxBaseHeartBeatRequest {
+  enum StateCode {
+    IDLE = 0;
+    WORKING = 1;
+    ALERT_WARNING = 2;
```
benma: alert_ prefix seems superfluous

messages/bitboxbase.proto
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// This will display the first 20 characters of the base32 encoded version of
+// the provided msg
+message BitBoxBaseConfirmPairingRequest {
+  bytes msg = 1;
+}
+
+// Optional fields can be represented by a "oneof" with only one field in it.
+// All fields are technically optional. But in reality the default value for the type will be set.
```
benma: interesting! the empty value could work though, right? empty ip or hostname is invalid.

messages/bitboxbase.proto
```diff
@@ -0,0 +1,79 @@
...
+  }
+  StatusLedMode status_led_mode = 1;
+  enum StatusScreenMode {
+    SCREEN_ALWAYS = 0; // display on screen when status is IDLE, WORKING, WARNING and ERROR
+    SCREEN_ON_WORKING = 1; // display on screen when status is WORKING, WARNING and ERROR
+    SCREEN_ON_WARNING = 2; // display on screen when status is WARNING and ERROR
+    SCREEN_ON_ERROR = 3; // display on screen when status is ERROR
+  }
+  StatusScreenMode status_screen_mode = 2;
+  uint32 default_display_duration = 3;
```
benma: let's add it once needed

messages/bitboxbase.proto
```diff
@@ -0,0 +1,79 @@
...
+  }
+}
+
+message BitBoxBaseDisplayStatusRequest {
+  uint32 duration = 1;
+}
+
+message BitBoxBaseRequest {
+  oneof request {
+    BitBoxBaseHeartBeatRequest heartbeat = 1;
```
benma: nit: inconsistency with using underscore or not. with no underscore, the msg would be `BitBoxBaseHeartbeatRequest`

messages/bitboxbase.proto
NickeZ: yeah good point

messages/bitboxbase.proto
NickeZ: empty ip would be 0.0.0.0 and that is a valid ip. not sure if empty string is a valid hostname. But I guess not

messages/bitboxbase.proto
NickeZ: ack

messages/bitboxbase.proto
NickeZ: aha good catch

PR #217, opened 2019-11-25
py: Set show firmware hash

NickeZ:


py/send_message.py
```diff
@@ -364,11 +364,19 @@ def _get_versions(self) -> None:
...
+
+    def _dont_show_fw_hash(self) -> None:
+        self._device.set_show_firmware_hash(False)
+
     def _menu(self) -> None:
         choices = (
             ("Boot", self._boot),
             ("Print versions", self._get_versions),
             ("Erase firmware", self._erase),
+            ("Show firmware hash", self._show_fw_hash),
```
benma: could do a toggle like with the passphrase (can check with `show_firmware_hash_enabled()`).

NickeZ: yeah we can do that in a future optimization

PR #219, opened 2019-11-25
bitboxbase: Don't abort on wrong screen orientation

NickeZ:
Also don't rotate screen in bootloader and always set it to not flipped
when rebooting

### DIFF
--- a/src/ui/oled/oled.c
+++ b/src/ui/oled/oled.c
@@ -67,6 +67,7 @@
 
 #include <driver_init.h>
 #include <hardfault.h>
+#include <screen.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <string.h>
@@ -253,7 +254,8 @@ void oled_mirror(bool mirror)
     }
 #elif PLATFORM_BITBOXBASE == 1
     (void)mirror;
-    Abort("BitBoxBase cannot mirror screen");
+    // DON'T ABORT HERE, this is called in the bootloader
+    screen_print_debug("BitBoxBase cannot mirror screen", 5000);
 #endif
 }
 
--- a/src/workflow/reboot.c
+++ b/src/workflow/reboot.c
@@ -26,7 +26,13 @@ static void _reboot(void)
         .value = sectrue_u8,
     };
     upside_down_t upside_down = {
+#if PLATFORM_BITBOX02 == 1 || defined(TESTING)
         .value = screen_is_upside_down(),
+#elif PLATFORM_BITBOXBASE == 1
+        .value = false,
+#else
+#error "No platform"
+#endif
     };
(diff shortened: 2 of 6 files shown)


src/workflow/reboot.c
```diff
@@ -26,7 +26,11 @@ static void _reboot(void)
         .value = sectrue_u8,
     };
     upside_down_t upside_down = {
+#if PLATFORM_BITBOX02 == 1
         .value = screen_is_upside_down(),
+#elif PLATFORM_BITBOXBASE == 1
+        .value = false,
```
benma: `#else #error` 

or `#if PLATFORM_BITBOXBASE == 1 ... #else ... `

NickeZ: ack

benma: still missing `#else #error`

conte91: > or `#if PLATFORM_BITBOXBASE == 1 ... #else ... `

This wins

src/workflow/workflow.c
```diff
@@ -73,11 +74,13 @@ static void _select_orientation_done(bool upside_down)
...
     ui_screen_stack_switch(show_logo);
 }
+#endif
 
 void workflow_start_orientation_screen(void)
 {
 #if PLATFORM_BITBOXBASE == 1
-    _select_orientation_done(false);
+    component_t* show_logo = show_logo_create(workflow_start, 200);
+    ui_screen_stack_switch(show_logo);
```
conte91: Direct call to `_select_orientation_done(false)` instead? So we don't have to track future changes in the latter.

src/workflow/reboot.c
NickeZ: I think I added that, must have force pushed over it

src/workflow/workflow.c
NickeZ: Yeah, it's either that and doing ifdefs around "upside_down" or copying these two lines. I did the second. I think the whole "orientation screen" doesn't make sense from the BBB point of view.

src/ui/oled/oled.c
```diff
@@ -253,7 +254,8 @@ void oled_mirror(bool mirror)
     }
 #elif PLATFORM_BITBOXBASE == 1
     (void)mirror;
-    Abort("BitBoxBase cannot mirror screen");
+    // DON'T ABORT HERE, this is called in the bootloader
+    screen_print_debug("BitBoxBase cannot mirror screen", 5000);
```
benma: `#else #error` also missing here :laughing: but let's merge this so we can get on with it :D

NickeZ: But here it is definitely a compile error because "mirror" would be unused!

PR #221, opened 2019-11-25
Bbb versions

NickeZ:
this has a `FACTORYSETUP` in it that has to be changed if we merge the other PR first. https://github.com/digitalbitbox/bitbox02-firmware/pull/220


CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ endif()
 # Example 'v1.0.0'. They MUST not contain a pre-release label such as '-beta'.
 set(FIRMWARE_VERSION "v4.3.0")
 set(FIRMWARE_BTC_ONLY_VERSION "v4.3.0")
+set(FIRMWARE_BITBOXBASE_VERSION "v1.0.0")
```
benma: 4.3.0 please - there is a lot of api stuff in both clients that rely on this.

CMakeLists.txt
```diff
@@ -114,6 +115,16 @@ if(GIT_FOUND AND PYTHONINTERP_FOUND)
   if(NOT exit_code EQUAL "0")
     set(GIT_FIRMWARE_BTC_ONLY_VERSION_STRING "v0.0.0")
   endif()
+  execute_process(
+    COMMAND ${PYTHON_EXECUTABLE} ./scripts/get_version firmware-bitboxbase --check-semver --check-gpg
```
benma: `firmware-bitboxbase-standard` (now that we have platform+edition; otherwise we could mistake bitboxbase to be an edition)

Here and in the rest of the diff.

CMakeLists.txt
NickeZ: ack

CMakeLists.txt
NickeZ: the tags follow the file names so far. do we really want them to follow the defines?

PR #222, opened 2019-11-25
python scripts: add SIGDATA_MAGIC_BITBOXBASE_STANDARD magic.

conte91:


py/bitbox02/bitbox02/bootloader.py
```diff
@@ -64,7 +65,7 @@ def __init__(self, transport: TransportLayer, device_info: DeviceInfo):
         self.expected_magic = {
             "bb02-bootloader": SIGDATA_MAGIC_STANDARD,
             "bb02btc-bootloader": SIGDATA_MAGIC_BTCONLY,
-            "bb02-base": SIGDATA_MAGIC_STANDARD,
+            "bb02-base": SIGDATA_MAGIC_BITBOXBASE_STANDARD,
```
benma: bb02-base is a weird product string, mixing two platforms. sounds like `bitboxbase-standard` is more fitting.

conte91: `bitboxbase-bootloader` to match the other ones?

conte91: Fixed.

PR #225, opened 2019-11-26
Py

benma:


py/bitbox02/communication/usart/usart.py
```diff
@@ -29,8 +29,8 @@ class SerialPort(PhysicalLayer):
     (i.e. to make it implement communication.PhysicalLayer).
     """
 
-    def __init__(self, *args: Any, **kwargs: Any):
-        self.port = serial.Serial(*args, **kwargs)
+    def __init__(self, port: str, timeout: Optional[int] = None):
```
conte91: The timeout is set for each `read` operation explicitely (see the `read` method). If we want to change it to be in the constructor (or remove it completely if you ask me...) we should change the definition of `PhysicalLayer` as well (note that this has been defined to be compatible with the `hid` class, so there's no reason not to adapt both to our needs).

py/load_firmware.py
```diff
@@ -203,7 +203,7 @@ def main() -> int:
         return 1
 
     if args.usart is not None:
-        serial_port = usart.SerialPort(args.usart, 115200, timeout=3)
+        serial_port = usart.SerialPort(args.usart, timeout=3)
```
conte91: For the same reason as above, this stray `timeout` argument should be removed (as it's going to be ignored)

py/bitbox02/communication/usart/usart.py
benma: fixed

py/load_firmware.py
benma: fixed

PR #232, opened 2019-11-29
py/bitbox02: peg attestation pubkey to allowed bootloader hash

benma:
Equivalent of https://github.com/digitalbitbox/bitbox02-api-go/pull/18


NickeZ:
utACK

PR #235, opened 2019-11-30
Py: Move noise verification callback after send, before response

TheCharlatan:
The noise verification callback might not return if the callback function does a blocking call to a client (for example to render the pairing code on a window). If that is the case, the client would display the pairing code before it is displayed on the device. To evade this, place the callback after making the pairing verification request to the device, and read the response once the callback returns.

Why is the 2ufhid interface used here in the first place? Is this just the unencrypted interface, or is it a separate hid handler?

### DIFF
--- a/py/bitbox02/communication/bitbox_api_protocol.py
+++ b/py/bitbox02/communication/bitbox_api_protocol.py
@@ -134,7 +134,7 @@ class BitBoxCommonAPI:
         self,
         transport: TransportLayer,
         device_info: DeviceInfo,
-        show_pairing_callback: Callable[[str], None],
+        show_pairing_callback: Callable[[str], bool],
         attestation_check_callback: Optional[Callable[[bool], None]] = None,
     ):
         self.debug = False
@@ -180,11 +180,6 @@ class BitBoxCommonAPI:
         send_msg = noise.write_message()
         assert noise.handshake_finished
         pairing_code = base64.b32encode(noise.get_handshake_hash()).decode("ascii")
-        show_pairing_callback(
-            "{} {}\n{} {}".format(
-                pairing_code[:5], pairing_code[5:10], pairing_code[10:15], pairing_code[15:20]
-            )
-        )
         response = self._query(send_msg)
 
         # Can be set to False if the remote static pubkey was previously confirmed.
@@ -192,7 +187,18 @@ class BitBoxCommonAPI:
 
         pairing_verification_required_by_device = response == b"\x01"
         if pairing_verification_required_by_host or pairing_verification_required_by_device:
-            pairing_response = self._query(OP_I_CAN_HAS_PAIRIN_VERIFICASHUN)
+            cid = self._transport.generate_cid()
+            self._transport.write(OP_I_CAN_HAS_PAIRIN_VERIFICASHUN, HWW_CMD, cid)
+            client_response_success = show_pairing_callback(
+                "{} {}\n{} {}".format(
+                    pairing_code[:5], pairing_code[5:10], pairing_code[10:15], pairing_code[15:20]
+                )
+            )
+            if not client_response_success:
+                self.close()
+                raise Exception("pairing rejected by the user on client")
+            pairing_response = self._transport.read(HWW_CMD, cid)
+
             if pairing_response == RESPONSE_SUCCESS:
                 pass
             elif pairing_response == RESPONSE_FAILURE:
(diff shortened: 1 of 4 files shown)


benma:
Have you observed a bad consequence of not closing?

> Why is the 2ufhid interface used here in the first place? Is this just the unencrypted interface, or is it a separate hid handler?

The usb communication protocol is u2f framing over hid (encrypted and non encrypted).

TheCharlatan:
> Have you observed a bad consequence of not closing?
Yes, if you use it in a 3rd party library and do not close the handler if the initialization fails, the handler will remain open. If you try initializing again in the same process, the handler will still be open and you cannot initialize the device again.

TheCharlatan:
> Have you observed a bad consequence of not closing?
Not closing what?

py/bitbox02/bitbox02/bitbox02.py
```diff
@@ -229,19 +229,22 @@ def __init__(
...
 
         # Can be set to False if the remote static pubkey was previously confirmed.
         pairing_verification_required_by_host = True
 
         pairing_verification_required_by_device = response == b"\x01"
         if pairing_verification_required_by_host or pairing_verification_required_by_device:
-            pairing_response = self._query(OP_I_CAN_HAS_PAIRIN_VERIFICASHUN)
+            cid = u2fhid.generate_cid()
+            u2fhid.write(self.device, OP_I_CAN_HAS_PAIRIN_VERIFICASHUN, HWW_CMD, cid)
+            show_pairing_callback(
```
benma: the callback should be extended also to get a user response (accept, reject) and handle that properly.

now or in another PR?

TheCharlatan: I can do it now, but not sure how to do it properly. Where do you do that in the bitbox02-api-go package?

benma: the rejection should probably just raise a UserRejectedException.

NickeZ: the callback should also not take a formatted string. It should take `pairing_code[:20]`. Formatting should be done in the application.

benma: I think I disagree. The formatting with with spaces and two lines is fixed by the bb02, any client has to match it. Re-doing the same code in every client does not add any value.

benma:
> > Have you observed a bad consequence of not closing?
> 
> Not closing what?

commented in the wrong PR, I was wondering why this comment disappeared from your other PR ;)

py/bitbox02/bitbox02/bitbox02.py
TheCharlatan: I raise an exception now. Please merge the other PR first, I already added a `self.close()` here in anticipation.

TheCharlatan:
Force-pushed some python linter fixes.

py/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -180,19 +180,25 @@ def __init__(
...
+            cid = self._transport.generate_cid()
+            self._transport.write(OP_I_CAN_HAS_PAIRIN_VERIFICASHUN, HWW_CMD, cid)
+            client_response_success = show_pairing_callback(
+                "{} {}\n{} {}".format(
+                    pairing_code[:5], pairing_code[5:10], pairing_code[10:15], pairing_code[15:20]
+                )
+            )
+            if not client_response_success:
+                self.close()
+                raise Exception("pairing rejected by the user on client")
```
benma: you should raise a `PairingFailedException` or something like that, as that is an exception that is supposed to be handled by the client, e.g. by showing a (translated) error message. The other exceptions are kind of internal errors which cannot be handled specially.

TheCharlatan: I would have guessed that if the client is capable of determining if the user rejected the pairing or not, there is no need for a specific exception.

TheCharlatan: If I'd name the exceptions specifically, I would rather handle the user rejecting the pairing on the device properly, since that it is no error internal to the library, but the user executing an action. That would not be in scope of this commit though.

benma: What's the plan then, to silently ignore all exceptions raised here? If you plan on showing the error message in the exception, then treating this specially makes sense, as it is something we expect the user to see, versus the others which should not happen if everything is bug-free.

I guess for now it's good enough to just show the exception message (English) but the "proper" way is to have a translation entry.

benma: Handling a failed pairing is client side only, the device is not affected. Assume a MITM, you don't need to engage with it anymore when you reject a pairing. The app just needs to warn the user of what happened, or try again in case it was a transient error (which I hope will not happen again after we fix the one outstanding known bug related to this).

PR #237, opened 2019-12-01
Py: Close hid device on pairing failure

TheCharlatan:
In case the pairing fails in the BitBox02's init function, the hid handler was never closed. This commit closes it in case of an exception.


benma:
Did you see any consequence of not closing? Just for my information.

TheCharlatan:
Yes, if you use the library and try to re-initialize the BitBox02 in the same process, the handler stays open and you can't open it again. I ran into this in Electrum.

PR #238, opened 2019-12-01
Py: Fix sendmessage eth and btc pubkey pb2 import

TheCharlatan:
I also saw that in https://github.com/digitalbitbox/bitbox02-firmware/blob/master/messages/hww.proto the btc requests are included, but the eth requests are not. What is the reason for that? I also did not understand, why I could not do `bitbox02.eth.ETHPubRequest.Address`, even though `eth` is aliased as an import of `generated.eth_pb2` in the bitbox02 package.

### DIFF
--- a/py/bitbox02/bitbox02/__init__.py
+++ b/py/bitbox02/bitbox02/__init__.py
@@ -50,6 +50,7 @@ from .bitbox02 import (
     HARDENED,
     hww,
     btc,
+    eth,
     UserAbortException,
 )
 from .bootloader import Bootloader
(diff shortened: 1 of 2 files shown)


NickeZ:
You probably need to reexport it here: https://github.com/digitalbitbox/bitbox02-firmware/blob/2355fcd6bd04afa0bff5eb83ebf75c61f116da62/py/bitbox02/bitbox02/__init__.py#L41

Maybe in more places as well. The python import rules never made sense to me.

TheCharlatan:
Perfect, fixed.

PR #239, opened 2019-12-02
Add testnet segwit upub and vpub

TheCharlatan:
BIP49: Testnet uses 0x044a5262 "upub" and 0x044a4e28 "uprv."
BIP84: Testnet uses 0x045f1cf6 "vpub" and 0x045f18bc "vprv."
Update the generated python protobuf code.
Add extra fields to the BTCPubRequest for each of them.

### DIFF
--- a/messages/btc.proto
+++ b/messages/btc.proto
@@ -36,6 +36,8 @@ message BTCPubRequest {
     YPUB = 2;
     ZPUB = 3;
     ADDRESS = 4;
+    VPUB = 5;
+    UPUB = 6;
   }
   BTCScriptType script_type = 2; // only applies for ADDRESS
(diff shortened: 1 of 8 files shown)


messages/btc.proto
```diff
@@ -35,6 +35,8 @@ message BTCPubRequest {
     XPUB = 1;
     YPUB = 2;
     ZPUB = 3;
+    VPUB = 5;
```
benma: Please put them in order of number, so the next person does not add `= 5` for the next var ;) It's a bit unfortunate but it is normal to append the new thing at the end.

TheCharlatan: Done.

PR #241, opened 2019-12-02
WIP: cmake cleanup

conte91:
Extract common files in the CMakeLists.txt when possible, to avoid recompiling the same source file over and over with different flags.


NickeZ:
In principle nothing wrong with this PR, but it has become stale. Closing for now.

PR #242, opened 2019-12-02
bitboxbase: Wake screen on more interactions

NickeZ:


src/bitboxbase/bitboxbase_screensaver.c
```diff
@@ -53,16 +54,16 @@ void bitboxbase_screensaver_init(void)
 }
 
 // Checks the timer
```
benma: i would never add a docstring in the .c file, it is bound to diverge with the one in the .h file, like in this case :D

maybe the function could also be renamed, as "_check" is not accurate anymore.

NickeZ: would `bitboxbase_screensaver_poll` be a good name?

NickeZ: I'll name it `process`. Everything else is already named process

PR #245, opened 2019-12-03
docker_exec.sh: Fail the docker subprocess returns nonzero.

conte91:
Running docker_exec.sh <command> will always exit 0, even if the invoked command has failed. This prevents using it as e.g. a git hook to verify commits.


conte91:
> I assume you tested it?

Currently using it in my git hooks :)

benma:
> Running docker_exec.sh will always exit 0, even if the invoked command has failed. This prevents using it as e.g. a git hook to verify commits.

This would be an excellent commit message :smile:

PR #247, opened 2019-12-04
Cmake cleanup: stage1

conte91:
Trying to recover from the multitude of added compilations that the BitBoxBase firmware introduced. This PR adds modularization to the code, and tries to bulid with as little `PRODUCT_*`, `APP_*` etc. definitions as possible. In order to make sure that nothing breaks when `undef`ining those macros, they are enforced to be `#define`d in the relevant headers (so if a code unit doesn't really need the macro, it won't be able to include the headers that depend on it).


NickeZ:
One thing I noticed was that I intended that we should put the definitions of `APP_` macros in a config file (such as `platform_config.h`) because it is easier to review it there than in the cmake files.

However I also realized now that `platform_config.h` probably isn't an appropriate name anymore. What do you think about renaming it to `product_config.h`?

NickeZ:
Also just redefinig the `APP_` macros should be its own PR. that can be done independently of any other refactoring.

NickeZ:
we could also have a separate `app_config.h` with the `APP_` macros in it. WDYT?

conte91:
> However I also realized now that `platform_config.h` probably isn't an appropriate name anymore. What do you think about renaming it to `product_config.h`?

The `PRODUCT_` macros are imho a wrong concept anyway. We use the "bigger" enum (all the products we have or might have in the future) to deduce the "smaller" enums (the board that the firmware shall run on). IMHO we should split this into two configs (`PLATFORM`, `EDITION`) to allow separation of the hardware-dependent code that doesn't depend on, for example, whether the firmware is multi or btc-only.

conte91:
> we could also have a separate `app_config.h` with the `APP_` macros in it. WDYT?

+1

NickeZ:
> > However I also realized now that `platform_config.h` probably isn't an appropriate name anymore. What do you think about renaming it to `product_config.h`?
> 
> The `PRODUCT_` macros are imho a wrong concept anyway. We use the "bigger" enum (all the products we have or might have in the future) to deduce the "smaller" enums (the board that the firmware shall run on). IMHO we should split this into two configs (`PLATFORM`, `EDITION`) to allow separation of the hardware-dependent code that doesn't depend on, for example, whether the firmware is multi or btc-only. See [436611b](https://github.com/digitalbitbox/bitbox02-firmware/commit/436611b2a58829523396b13fbceb208ed3dc1c54) for a draft of how this could look like.

That makes sense to me in case it reduces the built object files.

conte91:
> Also just redefinig the `APP_` macros should be its own PR. that can be done independently of any other refactoring.

Makes sense - I'll open a separate request for this.

src/CMakeLists.txt
```diff
@@ -399,7 +491,9 @@ if(CMAKE_CROSSCOMPILING)
   foreach(bootloader ${BOOTLOADERS})
     set(elf ${bootloader}.elf)
     add_executable(${elf} ${BOOTLOADER-SOURCES})
-    target_link_libraries(${elf} PRIVATE c asf4-drivers-min -Wl,-u,exception_table)
+    add_dependencies(${elf} libwally-core)
+    add_dependencies(${elf} noise-c)
+    target_link_libraries(${elf} PRIVATE c asf4-drivers-min noiseprotocol -Wl,-u,exception_table)
```
NickeZ: I think noise and libwally shouldn't be needed in the bootloader.

src/CMakeLists.txt
```diff
@@ -512,40 +609,54 @@ if(CMAKE_CROSSCOMPILING)
       target_compile_options(${elf} PRIVATE --specs=rdimon.specs)
       target_link_libraries(${elf} PRIVATE --specs=rdimon.specs)
     endif()
+    string(FIND ${elf} bitboxbase BITBOXBASE_FOUND)
+    if(BITBOXBASE_FOUND EQUAL -1)
+      target_sources(${elf} PRIVATE ${BITBOX02-FIRMWARE-SOURCES})
+    else()
+      target_sources(${elf} PRIVATE ${BITBOXBASE-FIRMWARE-SOURCES})
+    endif()
```
NickeZ: I don't think this pattern (find string in elf name) works well here. It works well for the "semhosting" string because it is binary. I think it would be better if you created two variables: "FIRMWARES-BITBOX02" and "FIRMWARE-BITBOXBASE" a bit like there is for the bootloaders. Then you could iterate only over the bb02 firmwares or the bbb firmwares and add sources.

src/CMakeLists.txt
conte91: Fixed. There was a missing dependency between `libwally.a` and `libwally-core` (the target actually building libwally!), so the dep tree was broken.

src/CMakeLists.txt
conte91: Fixed.

src/touch/gestures_bitboxbase.c
```diff
@@ -0,0 +1,385 @@
+// Copyright 2019 Shift Cryptosecurity AG
```
benma: this file is nowhere to be found in cmakelists.txt, how does that work?

src/touch/gestures_bitboxbase.c
```diff
@@ -0,0 +1,385 @@
...
+
+#ifndef TESTING
+#include "qtouch.h"
+#include <driver_init.h>
+#else
+#include "mock_qtouch.h"
+#endif
+
+#include "gestures.h"
+#if PLATFORM_BITBOXBASE == 1
```
benma: is this redundant? hopefully this file would only be compiled for bitboxbase only.

src/touch/gestures.c
```diff
@@ -25,6 +25,10 @@
 #endif
 
 #include "gestures.h"
+#include "gestures_impl.h"
+#if PLATFORM_BITBOXBASE == 1
+#include "gestures_bitboxbase.h"
```
benma: what's the difference to gestures_bitboxbase.c?

PR #249, opened 2019-12-04
ui/components/label: fix center top offset

NickeZ:
The labels size was incorrectly calculated when updating the label

### DIFF
--- a/src/ui/components/label.c
+++ b/src/ui/components/label.c
@@ -39,14 +39,16 @@ typedef struct {
     int16_t text_position_last;
 } data_t;
 
+static void _measure_label_dimensions(component_t* label);
+
 void label_update(component_t* component, const char* text)
 {
     data_t* data = (data_t*)component->data;
     snprintf(data->text, sizeof(data->text), "%s", text);
-    UG_MeasureString(&(component->dimension.width), &(component->dimension.height), text);
     if (component->parent == NULL) {
         return;
     }
+    _measure_label_dimensions(component);
     component_t* parent = component->parent;
     switch (data->position) {
     case CENTER:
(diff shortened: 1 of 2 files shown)


src/ui/components/label.c
```diff
@@ -175,6 +177,28 @@ static void _cleanup(component_t* component)
...
+
+    UG_FontSetVSpace(2);
+    UG_FontSelect(data->font);
+    if (data->scrollable) {
+        UG_MeasureStringNoBreak(&(label->dimension.width), &(label->dimension.height), data->text);
+        if (label->dimension.width < SCREEN_WIDTH) {
+            // Do not scroll if text already fits in the screen
+            data->scrollable = false;
+        }
+    } else if (
```
benma: else if has no condition?

src/ui/components/label.c
```diff
@@ -39,14 +39,16 @@ typedef struct {
     int16_t text_position_last;
 } data_t;
 
+static void _measure_label_dimensions(component_t* label);
```
benma: could also move the function here instead.

src/ui/components/label.c
```diff
@@ -175,6 +177,28 @@ static void _cleanup(component_t* component)
     ui_util_component_cleanup(component);
 }
 
+void _measure_label_dimensions(component_t* label)
```
benma: needs `static` too I think

src/ui/components/label.c
NickeZ: I just moved the code. I'll fix it

src/ui/components/label.c
NickeZ: sure

src/ui/components/label.c
NickeZ: nope not allowed.

src/ui/components/label.c
NickeZ: I think you are mis-seeing a left parenthesis as a left bracket.

PR #251, opened 2019-12-04
cmake: Enforce correct usage of the APP_* macros.

conte91:
Enforce all checks for APP_{BTC,LTC,ETH,U2F} to check
the actual value of the macro, instead of just checking
for definition. This avoids silly errors due to misconfiguration.


conte91:
> I assume you grepped for the changed variables and double-checked that there are no `ifdef` nor `if defined()` left.
> 
That's correct

PR #253, opened 2019-12-08
firmware: BTC sign, supported locktime and sequence bytes

TheCharlatan:
The following pairs of locktime and sequence bytes are now supported:

locktime = 0 , nsequence = 0xffffffff
-> as before

locktime > 0, nsequence = 0xffffffff - 1
-> nsequence < 0xffffffff indicates a locktime on a transaction.

locktime > 0, nsequence = 0xffffffff - 2
-> nsequence == 0xffffffff - 2 indicates that the transaction is RBF. Since
we have the above rule on the locktime, the RBF rule only makes sense if the locktime is not zero again.

All other combinations of nsequence and locktime bytes are still not allowed. Other combinations should eventually be allowed as well to support BIP68 (nseuqnece byte definitions), BIP65
(OP_CHECKLOCKTIMEVERIFY) and BIP112 (OP_CHECKSEQUENCEVERIFY).

The relaxed constraints were chosen to accommodate for electrum's usage of RBF. The relation between locktime, nsequence and RBF is explained in this bitcoinwiki article: https://en.bitcoinwiki.org/wiki/NSequence .

I initially wrote this code for myself, but since it seems to work reasonably well, I thought I might as well post a pull request. My C is horrible, so fill your pockets with grains of salt before reviewing. At some point I am converting a `uint32` to a `const char *` with `snprintf(formatted_locktime, sizeof formatted_locktime, "\n\n\n%lu", (unsigned long)_locktime);`, which feels especially wrong.

EDIT: There is still one case that is supported with extra verification, which is when the sequence bytes are 0xffffffff and the locktime is > 0. I am not sure if this case should be rejected or not.

### DIFF
--- a/src/apps/btc/btc_sign.c
+++ b/src/apps/btc/btc_sign.c
@@ -16,6 +16,7 @@
 #include "btc_bip143.h"
 #include "btc_common.h"
 #include "btc_params.h"
+#include "confirm_locktime_rbf.h"
 
 #include <crypto/sha2/sha256.h>
 #include <keystore.h>
@@ -44,6 +45,8 @@ static uint32_t _num_inputs;
 static uint32_t _num_outputs;
 static uint32_t _locktime;
 static uint32_t _index;
+static enum apps_btc_rbf_flag _rbf;
+static bool _locktime_applies;
 
 // used during STATE_INPUTS_PASS1. Will contain the sum of all spent output
 // values.
@@ -92,6 +95,8 @@ static void _reset(void)
     _num_outputs = 0;
     _locktime = 0;
     _index = 0;
+    _rbf = CONFIRM_LOCKTIME_RBF_OFF;
+    _locktime_applies = false;
     _inputs_sum_pass1 = 0;
     _inputs_sum_pass2 = 0;
     _outputs_sum_out = 0;
@@ -114,16 +119,16 @@ app_btc_sign_error_t app_btc_sign_init(
     if (_state != STATE_INIT) {
         return _error(APP_BTC_SIGN_ERR_STATE);
     }
+    // Currently we do not support time-based nlocktime
+    if (request->locktime >= 500000000) {
+        return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
+    }
     // currently only support version 1 or version 2 tx.
     // version 2: https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki
     if (request->version != 1 && request->version != 2) {
         return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
     }
     if (request->num_inputs < 1 || request->num_outputs < 1) {
-        return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
-    }
-    // Locktime not yet supported.
-    if (request->locktime != 0) {
         return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
     }
     const app_btc_coin_params_t* coin_params = app_btc_params_get(request->coin);
@@ -305,8 +310,17 @@ app_btc_sign_error_t app_btc_sign_input(
     if (_state != STATE_INPUTS_PASS1 && _state != STATE_INPUTS_PASS2) {
         return _error(APP_BTC_SIGN_ERR_STATE);
     }
-    if (request->sequence != 0xffffffff) {
+    // relative locktime and sequence nummbers < 0xffffffff-2 are not supported
+    if (request->sequence < 0xffffffff - 2) {
         return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
+    }
+    if (_coin_params->rbf_support) {
+        if (request->sequence == 0xffffffff - 2) {
+            _rbf = CONFIRM_LOCKTIME_RBF_ON;
+        }
+    }
+    if (request->sequence < 0xffffffff) {
+        _locktime_applies = true;
     }
     if (request->prevOutValue == 0) {
         return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
@@ -433,7 +447,21 @@ app_btc_sign_error_t app_btc_sign_output(
         next_out->type = BTCSignNextResponse_Type_OUTPUT;
         next_out->index = _index;
     } else {
-        // Done with outputs. Verify total and fee.
+        // Done with outputs. Verify locktime, total and fee.
+        //
+        // This is not a security feature, a transaction that is not rbf
+        // and has a locktime of 0 will not be verified.
+        if (_locktime_applies || _rbf == CONFIRM_LOCKTIME_RBF_ON) {
+            // The RBF nsequence bytes are often set in conjunction with a locktime,
+            // so verify both simultaneously.
+            // There is no RBF in Litecoin, so make sure it is disabled.
+            if (!_coin_params->rbf_support) {
+                _rbf = CONFIRM_LOCKTIME_RBF_DISABLED;
+            }
+            if (!apps_btc_confirm_locktime_rbf(_locktime, _rbf)) {
+                return _error(APP_BTC_SIGN_ERR_USER_ABORT);
+            }
+        }
 
         // total_out, including fee.
         if (_inputs_sum_pass1 < _outputs_sum_ours) {
--- a/src/apps/btc/confirm_locktime_rbf.c
+++ b/src/apps/btc/confirm_locktime_rbf.c
@@ -1,0 +1,50 @@
+// Copyright 2019 Shift Cryptosecurity AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "confirm_locktime_rbf.h"
+
+#include <workflow/blocking.h>
+#include <workflow/status.h>
+
+#include <ui/components/confirm_transaction.h>
+#include <ui/screen_stack.h>
+#include <util.h>
+#include <workflow/confirm.h>
+
+bool apps_btc_confirm_locktime_rbf(uint32_t locktime, enum apps_btc_rbf_flag rbf)
+{
+    char formatted_locktime_rbf[100] = {0};
+    const char* locktime_text = "Locktime on block:";
+    const char* rbf_text;
+    if (rbf == CONFIRM_LOCKTIME_RBF_ON) {
+        rbf_text = "Transaction is RBF";
+    } else if (rbf == CONFIRM_LOCKTIME_RBF_OFF) {
+        rbf_text = "Transaction is not RBF";
+    } else {
+        rbf_text = "";
+    }
+    snprintf(
+        formatted_locktime_rbf,
+        sizeof formatted_locktime_rbf,
+        "%s\n%lu\n%s",
+        locktime_text,
+        (unsigned long)locktime,
+        rbf_text);
+
+    bool result = workflow_confirm("", formatted_locktime_rbf, NULL, false, false);
+    if (!result) {
+        workflow_status_create("Transaction\ncanceled", false);
+    }
+    return result;
+}
--- a/test/unit-test/test_btc_sign.c
+++ b/test/unit-test/test_btc_sign.c
@@ -20,6 +20,7 @@
 #include <stdio.h>
 
 #include <apps/btc/btc_sign.h>
+#include <apps/btc/confirm_locktime_rbf.h>
 #include <keystore.h>
 #include <wally_bip32.h>
 
@@ -34,6 +35,13 @@ bool __wrap_workflow_verify_total(const char* total, const char* fee)
 {
     check_expected(total);
     check_expected(fee);
+    return mock();
+}
+
+bool __wrap_apps_btc_confirm_locktime_rbf(uint32_t locktime, enum apps_btc_rbf_flag rbf)
+{
+    check_expected(locktime);
+    check_expected(rbf);
     return mock();
 }
 
@@ -106,6 +114,12 @@ static void _test_btc_sign_init(void** state)
             assert_int_equal(APP_BTC_SIGN_ERR_INVALID_INPUT, app_btc_sign_init(&invalid, &next));
         }
     }
+    { // test invalid locktime
+        tst_app_btc_reset();
+        BTCSignInitRequest invalid = init_req_valid;
+        invalid.locktime = 500000000;
+        assert_int_equal(APP_BTC_SIGN_ERR_INVALID_INPUT, app_btc_sign_init(&invalid, &next));
+    }
     { // test invalid inputs
         tst_app_btc_reset();
         BTCSignInitRequest invalid = init_req_valid;
@@ -116,12 +130,6 @@ static void _test_btc_sign_init(void** state)
         tst_app_btc_reset();
         BTCSignInitRequest invalid = init_req_valid;
         invalid.num_outputs = 0;
-        assert_int_equal(APP_BTC_SIGN_ERR_INVALID_INPUT, app_btc_sign_init(&invalid, &next));
-    }
-    { // test invalid locktime
-        tst_app_btc_reset();
-        BTCSignInitRequest invalid = init_req_valid;
-        invalid.locktime = 1;
         assert_int_equal(APP_BTC_SIGN_ERR_INVALID_INPUT, app_btc_sign_init(&invalid, &next));
     }
     { // test invalid coin
@@ -163,13 +171,17 @@ typedef struct {
     bool state_init_after_init;
     // wrong state transition
     bool state_output_after_init;
-    // sequence number other than 0xffffffff
+    // sequence number below 0xffffffff - 2
     bool wrong_sequence_number;
     // value 0 is invalid
     bool wrong_input_value;
     bool wrong_output_value;
     // when a user aborts on an output verification
     bool user_aborts_output;
+    // check workflow when a locktime applies
+    bool locktime_applies;
+    // when a user aborts on a locktime verification
+    bool user_aborts_locktime_rbf;
     // when a user aborts on total/fee verification.
     bool user_aborts_total;
     // if value addition in inputs would overflow
@@ -188,7 +200,7 @@ static void _sign(const _modification_t* mod)
     // Need keystore to derive change and input scripts
     mock_state(mod->seeded ? _mock_seed : NULL, mod->seeded ? _mock_bip39_seed : NULL);
 
-    const BTCSignInitRequest init_req = {
+    BTCSignInitRequest init_req = {
         .coin = BTCCoin_BTC,
         .script_type = BTCScriptType_SCRIPT_P2WPKH,
         .bip44_account = BIP32_INITIAL_HARDENED_CHILD + 10,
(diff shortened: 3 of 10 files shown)


src/ui/components/confirm_transaction.h
```diff
@@ -43,4 +43,13 @@ component_t* confirm_transaction_fee_create(
     void (*confirm_callback)(void),
     void (*cancel_callback)(void));
 
+/**
+ * Creates a confirm screen.
+ */
+component_t* confirm_transaction_locktime_create(
```
benma: move to a new file, `confirm_locktime.h`.

src/ui/components/confirm_transaction.h
```diff
@@ -43,4 +43,13 @@ component_t* confirm_transaction_fee_create(
     void (*confirm_callback)(void),
     void (*cancel_callback)(void));
 
+/**
```
benma: docstring incomplete, describe all params and the expected behavior of the component.

src/workflow/verify_locktime.h
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef _WORKFLOW_VERIFY_LOCKTIME_H_
+#define _WORKFLOW_VERIFY_LOCKTIME_H_
+
+#include <stdbool.h>
+
+/**
+ * Shows a confirmation dialog to the user to confirm the total amount being
```
benma: wrong docstring

py/bitbox02/bitbox02/bitbox02.py
```diff
@@ -267,9 +267,6 @@ def btc_sign(
         """
         # pylint: disable=too-many-locals,no-member
 
-        # Reserved for future use.
-        assert version == 1 and locktime == 0
```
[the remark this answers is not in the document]
benma: actually, the firmware check was not extended. i suggest to restore this here and we open a new PR for extending the version check.

src/ui/components/confirm_transaction.h
TheCharlatan: Should I really move this? It displays only in combination with the other two functions in there and I'll have to redefine the functions internal to that file.

src/apps/btc/btc_sign.c
```diff
@@ -405,6 +412,16 @@ app_btc_sign_error_t app_btc_sign_output(
         }
     }
 
+    if (_locktime > 0 && !request->ours) {
+        // the RBF nseuqence bytes only make sense in conjunction with a locktime,
+        // so verify both simultaneously
+        char formatted_locktime[100] = {0};
+        snprintf(
+            formatted_locktime, sizeof formatted_locktime, "\n\n\n%lu", (unsigned long)_locktime);
+        if (!workflow_verify_locktime(formatted_locktime, _rbf)) {
```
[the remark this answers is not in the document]
TheCharlatan: I'm copying from the behaviour of passing the amount to the workflow here. I did remove the newlines now.

src/apps/btc/btc_sign.c
```diff
@@ -306,7 +305,15 @@ app_btc_sign_error_t app_btc_sign_input(
         return _error(APP_BTC_SIGN_ERR_STATE);
     }
     if (request->sequence != 0xffffffff) {
-        return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
+        // rbf without a locktime is ill-defined
+        if (request->sequence == 0xffffffff - 2 && _locktime != 0) {
+            _rbf = 1;
+        }
+        // sequence bytes indicate a locktime, so check that it is defined
+        else if (request->sequence == 0xffffffff - 1 && _locktime != 0) {
```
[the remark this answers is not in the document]
TheCharlatan: I left out the coin check, do you think we really need it? Electrum-ltc still sets the sequence bytes and locktime for example.

src/ui/components/confirm_transaction.h
benma: Hang on why don't you just use the already existing `workflow_confirm()`?

src/apps/btc/btc_sign.c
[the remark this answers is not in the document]
benma: yeah we need it. you can't ask the user to confirm that they are using rbf in litecoin. 

i would even add a `_handle_rbf` bool that is true only for Bitcoin. reason:

we should think about even confirming with the user when they are *not* using rbf (at least in the expert mode i proposed in my blog post), and that should not slip into Litecoin either, so separating rbf tracking from whether it should be confirmed in the first place seems more robust. 

instead of the `_handle_rbf` var, maybe it even makes sense to `#if` guard all the rbf stuff using the `APP_BTC` macro.

src/apps/btc/btc_sign.c
```diff
@@ -433,7 +447,17 @@ app_btc_sign_error_t app_btc_sign_output(
         next_out->type = BTCSignNextResponse_Type_OUTPUT;
         next_out->index = _index;
     } else {
-        // Done with outputs. Verify total and fee.
+        // Done with outputs. Verify locktime, total and fee.
+        if (_locktime > 0 && _locktime_applies) {
+            // the RBF nsequence bytes only make sense in conjunction with a locktime,
+            // so verify both simultaneously
+            char formatted_locktime[100] = {0};
+            snprintf(
```
benma: still a bit confused why you format the locktime here, but rbf inside the function. anyway, check if you can use `workflow_confirm` instead, which would cut down on a lot of code.

TheCharlatan: Is there an easy way to move the confirmation in `workflow_confirm` to the top? Why is it on the bottom in the first place actually?

src/apps/btc/btc_sign.c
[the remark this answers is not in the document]
TheCharlatan: This seems a bit premature if we don't know yet how the expert mode will look like. How about just adding an `#if defined(APP_BTC)` guard around the `_rbf=true`?

src/apps/btc/btc_sign.c
TheCharlatan: It feels awkward to accept the locktime at the bottom and to the rest of the transaction workflow at the top.

benma: i would say

```uint8_t slider_position = scrollable || longtouch ? top_slider : bottom_slider;```

can be made an argument instead in `_confirm_create()`, have `confirm_create()` and `confirm_create_scrollable()` provide the right position, and make a new one `confirm_create_top()` or sth that forces it to the top.

Comment (TheCharlatan) [reply src/apps/btc/btc_sign.c]: Mmh, this is getting a bit complicated for what was before a straight forward change.

File: src/apps/btc/btc_sign.c (hunk 5 above)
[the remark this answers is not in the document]
Comment (TheCharlatan) [reply src/apps/btc/btc_sign.c]: Passing `_locktime` now.

File: src/ui/components/confirm_transaction.h (hunk 1 above)
Comment (TheCharlatan) [reply src/ui/components/confirm_transaction.h]: Using it now.

File: src/apps/btc/btc_sign.c (hunk 6 above)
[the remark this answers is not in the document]
Comment (TheCharlatan) [reply src/apps/btc/btc_sign.c]: Added `#if defined(APP_BTC)` guard around the `_rbf=true`

File: src/apps/btc/btc_sign.c (hunk 7 above)
Comment (TheCharlatan) [reply src/apps/btc/btc_sign.c]: I'll leave it at the bottom for now, can always improve it later.

File: src/workflow/confirm_locktime.c [hunk 8]
```diff
@@ -0,0 +1,48 @@
...
+#include <workflow/confirm.h>
+
+bool workflow_confirm_locktime(uint32_t locktime, bool rbf)
+{
+    char formatted_locktime[100] = {0};
+    if (rbf) {
+        snprintf(
+            formatted_locktime,
+            sizeof formatted_locktime,
+            "Locktime set to block:\n%lu\nTransaction is RBF",
```
benma: This text is inaccurate, as the locktime can also specify a timestamp: https://en.bitcoin.it/wiki/NLockTime

I would still keep this, but reserve the unix timestamp formatting for the future, by adding a check here: `if (locktime > LOCKTIME_THRESHOLD) { Abort("unsupported locktime") }` with `LOCKTIME_THRESHOLD = 500000000`, and add the same check in btc_sign.c (`return _error(APP_BTC_SIGN_ERR_INVALID_INPUT)`).

benma: Did you test this on a device?I think `%lu` might not work.

src/workflow/confirm_locktime.c
```diff
@@ -0,0 +1,48 @@
...
+            (unsigned long)locktime);
+    } else {
+        snprintf(
+            formatted_locktime,
+            sizeof formatted_locktime,
+            "Locktime set to block:\n%lu\nTransaction is RBF",
```
benma: transaction is not rbf here?

src/apps/btc/btc_sign.c
```diff
@@ -433,7 +449,14 @@ app_btc_sign_error_t app_btc_sign_output(
         next_out->type = BTCSignNextResponse_Type_OUTPUT;
         next_out->index = _index;
     } else {
-        // Done with outputs. Verify total and fee.
+        // Done with outputs. Verify locktime, total and fee.
+        if (_locktime > 0 && _locktime_applies) {
+            // the RBF nsequence bytes only make sense in conjunction with a locktime,
```
benma: comment is not technically accurate, rbf makes sense independent of the locktime value.

test/unit-test/test_btc_sign.c
```diff
@@ -88,7 +88,6 @@ static void _test_btc_sign_init(void** state)
         .version = 1,
         .num_inputs = 1,
         .num_outputs = 1,
-        .locktime = 0,
```
benma: why remove this?

src/apps/btc/btc_sign.c
```diff
@@ -306,7 +307,22 @@ app_btc_sign_error_t app_btc_sign_input(
         return _error(APP_BTC_SIGN_ERR_STATE);
     }
     if (request->sequence != 0xffffffff) {
-        return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
+        if (request->sequence < 0xffffffff - 2) {
+            return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
+        }
+        // rbf without a locktime is ill-defined
+        if (request->sequence == 0xffffffff - 2 && _locktime != 0) {
```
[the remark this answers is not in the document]
TheCharlatan: Consensus enforces the truth values of the following implications:
if locktime > 0 -> nsequence < MAX_INT
if nsequence < MAX_INT - 1 -> RBF
I changed the code to only check for these values. 
This makes me think whether we should let the user verify the nlocktime at all. The rbf check makes sense to an extent, but the nlocktime seems to have little actual impact on the user.

My Opinion: Maybe we should only verify this content in some future "expert" signing mode, since it is data that cannot be verified meaningfully by the hardware wallet after all. However, in light of the things that you showed me today @benma I see no reason to verify it with a normal user. 

I pushed a change to the logic now, that just asserts the above implications, and passes through otherwise.

src/workflow/confirm_locktime.c
TheCharlatan: Yes, also seems to work fine for a large number, like a timestamp. But as I said initially, I am very  unsure of the conversion.

src/workflow/confirm_locktime.c
TheCharlatan: Fixed.

src/workflow/confirm_locktime.c
TheCharlatan: Added support for timestamps now.

test/unit-test/test_btc_sign.c
TheCharlatan: It's only used in that single test case. That said, it would be much better to actually repair whatever is breaking the test currently.

benma: In that case please restore it. this is just the struct initialization for the tx that is used in all tests. Sure, `0` is the default, but I prefer explicitness.

TheCharlatan: Done.

src/apps/btc/btc_sign.c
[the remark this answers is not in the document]
benma: > Consensus enforces the truth values of the following implications:
> if locktime > 0 -> nsequence < MAX_INT

I understood it differently. If nsequence < MAX_INT, locktime is in effect, and it can be any value, including 0.

If nsequence = MAX_INT for all inputs, locktime can be any value, but it is ignored (not compared to current block or time).

I'll have to think about not verifying the locktime with the user at all.

src/apps/btc/btc_sign.c
```diff
@@ -305,9 +307,17 @@ app_btc_sign_error_t app_btc_sign_input(
     if (_state != STATE_INPUTS_PASS1 && _state != STATE_INPUTS_PASS2) {
         return _error(APP_BTC_SIGN_ERR_STATE);
     }
-    if (request->sequence != 0xffffffff) {
+    if (request->sequence == 0xffffffff && _locktime > 0) {
```
benma: this is wrong, locktime can be >0. if any other input has sequence < MAX INT it is compared to the blockheight or time, otherwise it can be anything but will be ignored.

TheCharlatan: I'll remove the check and move the sequence check to `_locktime_applies`.

src/apps/btc/btc_sign.c
```diff
@@ -305,8 +307,13 @@ app_btc_sign_error_t app_btc_sign_input(
     if (_state != STATE_INPUTS_PASS1 && _state != STATE_INPUTS_PASS2) {
         return _error(APP_BTC_SIGN_ERR_STATE);
     }
-    if (request->sequence != 0xffffffff) {
-        return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
+    if (request->sequence < 0xffffffff - 1) {
+#if defined(APP_BTC)
```
benma: I think this has changed in https://github.com/digitalbitbox/bitbox02-firmware/pull/251

TheCharlatan:
Changed the PR to contain what I have discussed with @benma. The following rules are now enforced:
- nsequence bytes need to be 0xffffffff-2, 0xffffffff-1 or 0xffffffff
- nlocktime can span from 0 to to 500000000-1 , this makes sure that the BitBox is not signing a nlocktime based on median time, but rather based on block height.
- If nlocktime > 0 and the nsequence is <0xffffffff then the nlocktime is displayed to the user during transaction signing verification
- If nsequence == 0xffffffff-2 then the transaction is RBF and the user is informed on this

src/apps/btc/btc_sign.c
```diff
@@ -44,6 +46,8 @@ static uint32_t _num_inputs;
 static uint32_t _num_outputs;
 static uint32_t _locktime;
 static uint32_t _index;
+static bool _rbf;
```
benma: i would also guard this with APP_BTC, so it's clear that it is not handled in LTC.

src/workflow/confirm_locktime.c
```diff
@@ -0,0 +1,54 @@
...
+
+#include "blocking.h"
+#include "status.h"
+
+#include <ui/components/confirm_transaction.h>
+#include <ui/screen_stack.h>
+#include <util.h>
+#include <workflow/confirm.h>
+
+bool workflow_confirm_locktime_rbf(uint32_t locktime, bool rbf)
```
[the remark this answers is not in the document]
benma: This is still outstanding, it seems Litecoin will still show an RBF nessage.

benma:
In bitbox02.py, in `btc_sign()`:

> version, locktime: reserved for future use.

Needs to be updated with a more specific doc.

src/apps/btc/confirm_locktime_rbf.c
```diff
@@ -0,0 +1,58 @@
...
+    const char* locktime_text = "Locktime on block:";
+    const char* rbf_text;
+    if (rbf == rbf_on) {
+        rbf_text = "Transaction is RBF";
+    } else if (rbf == rbf_off) {
+        rbf_text = "Transaction is not RBF";
+    } else {
+        rbf_text = "";
+    }
+    if (locktime == 0 && rbf == rbf_disabled) {
```
benma: it's awkward that part of the condition to show a dialog is at call site (`if (_locktime_applies || _rbf == rbf_on) {`) and part of it is here. please move both to a single place, so it's easier to follow the logic.

src/apps/btc/confirm_locktime_rbf.c
```diff
@@ -0,0 +1,58 @@
...
+    if (locktime > 0) {
+        snprintf(
+            formatted_locktime_rbf,
+            sizeof formatted_locktime_rbf,
+            "%s\n%lu\n%s",
+            locktime_text,
+            (unsigned long)locktime,
+            rbf_text);
+    } else {
+        snprintf(formatted_locktime_rbf, sizeof formatted_locktime_rbf, "\n%s", rbf_text);
```
benma: Drop newline?

src/apps/btc/btc_sign.c
```diff
@@ -305,9 +311,18 @@ app_btc_sign_error_t app_btc_sign_input(
     if (_state != STATE_INPUTS_PASS1 && _state != STATE_INPUTS_PASS2) {
         return _error(APP_BTC_SIGN_ERR_STATE);
     }
-    if (request->sequence != 0xffffffff) {
+    // relative locktime and sequence nummbers < 0xffffffff-2 are not supported
+    if (request->sequence < 0xffffffff - 2) {
         return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
     }
+#if APP_BTC == 1
```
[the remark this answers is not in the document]
TheCharlatan: Uff, `request->unit` is stringy, so I cannot compare that. There is also bip44_coin, but it seems weird to compare against a BIP44 account code. I could add a field `bool rbf_support` to the `app_btc_coin_params_t` struct. I think it makes sense if we put feature differences between Bitcoin forks, clones and networks there.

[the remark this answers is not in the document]
TheCharlatan: I went with putting the rbf setting in `app_btc_coin_params_t` now.

src/apps/btc/btc_sign.c
```diff
@@ -16,13 +16,15 @@
...
 #include "btc_params.h"
+#include "confirm_locktime_rbf.h"
 
 #include <crypto/sha2/sha256.h>
 #include <keystore.h>
 #include <util.h>
 #include <workflow/verify_recipient.h>
 #include <workflow/verify_total.h>
 
+#include <stdio.h>
```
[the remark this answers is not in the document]
TheCharlatan: Nothing :P

src/apps/btc/confirm_locktime_rbf.c
TheCharlatan: Removed the check in confirm_locktime_rbf, added a comment toe the function explaining what is required from the input.

src/apps/btc/confirm_locktime_rbf.c
TheCharlatan: I like it with the newline.

TheCharlatan: Although, looks fine without the newline as well. Removing.

TheCharlatan:
I squashed the commits and added two more test cases for btc_sign, but did not add a test case for checking if rbf is really disabled for litecoin yet.

TheCharlatan:
Pushed a separate commit for testing Litecoin with rbf and locktime > 0 now.

test/unit-test/test_btc_sign.c
```diff
@@ -381,6 +383,15 @@ static void _sign(const _modification_t* mod)
     if (mod->overflow_output_ours) {
         outputs[4].value = ULLONG_MAX;
     }
+    if (mod->litecoin_rbf_disabled) {
+        init_req.coin = BTCCoin_LTC;
+        init_req.locktime = 1;
+        inputs[0].sequence = 0xffffffff - 2;
+        inputs[0].keypath[1] += 2;
```
benma: slight preference for `= 0 + BIP32_INITIAL_HARDENED_CHILD`, as `+=` means this value actually depends on `inputs[0].keypath[1]` (even if it changes), but it doesn't.

test/unit-test/test_btc_sign.c
```diff
@@ -250,6 +254,9 @@ static void _sign(const _modification_t* mod)
     if (mod->wrong_sequence_number) {
         inputs[0].sequence = 0;
     }
+    if (mod->user_aborts_locktime_rbf) {
```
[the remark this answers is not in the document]
TheCharlatan: I think I am testing that already with the litecoin_rbf_disabled case.

PR #254, opened 2019-12-10
firmware/btc: Support for transaction byte version 2

TheCharlatan:
Transaction byte version 2 was introduced with the CHECKSEQUENCEVERIFY soft fork and has been in usage with many clients (including Bitcoin Core as a default) since a long time.

For the sake of transaction uniformity we should change this in the desktop app as well. Electrum for example also uses the version 2 transaction byte since nearly a year now: https://github.com/spesmilo/electrum/pull/5039 and bitcoin since 2014 https://github.com/bitcoin/bitcoin/pull/7562

### DIFF
--- a/py/bitbox02/bitbox02/bitbox02.py
+++ b/py/bitbox02/bitbox02/bitbox02.py
@@ -268,7 +268,7 @@ class BitBox02(BitBoxCommonAPI):
         # pylint: disable=too-many-locals,no-member
 
         # Reserved for future use.
-        assert version == 1 and locktime == 0
+        assert version in (1, 2)
 
         sigs: List[Tuple[int, bytes]] = []
 
--- a/src/apps/btc/btc_sign.c
+++ b/src/apps/btc/btc_sign.c
@@ -114,9 +114,9 @@ app_btc_sign_error_t app_btc_sign_init(
     if (_state != STATE_INIT) {
         return _error(APP_BTC_SIGN_ERR_STATE);
     }
-    // currently only support version 1 tx.
+    // currently only support version 1 or version 2 tx.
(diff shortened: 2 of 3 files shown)


src/apps/btc/btc_sign.c
```diff
@@ -116,7 +116,7 @@ app_btc_sign_error_t app_btc_sign_init(
     }
     // currently only support version 1 tx.
     // version 2: https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki
-    if (request->version != 1) {
+    if (request->version != 1 && request->version != 2) {
```
benma: comment above outdated

TheCharlatan:
Just merge the locktime pr first for the correct version assert. Force-pushed the comment fix.

benma:
> Just merge the locktime pr first for the correct version assert. Force-pushed the comment fix.

no sorry, there's no reason to make those two PRs dependent on each other and scatter one change across two PRs.

TheCharlatan:
Ok, done.

py/bitbox02/bitbox02/bitbox02.py
```diff
@@ -268,7 +268,7 @@ def btc_sign(
         # pylint: disable=too-many-locals,no-member
 
         # Reserved for future use.
-        assert version == 1 and locktime == 0
+        assert version in (1, 2)
```
benma: technically locktime was not touched in this pr and the assertion should remain

TheCharlatan: god damn.

benma:
tested ACK (made a v2 tx that was mined successfully)

PR #260, opened 2019-12-12
py/send_message: remove unusued _search_for_device

benma:


benma:
See also https://github.com/digitalbitbox/bitbox02-firmware/pull/259

benma:
Actually I merged the PRs as they are so small and depend on each other.

PR #261, opened 2019-12-12
Passphrase visual confirmation

benma:


src/workflow/unlock.c
```diff
@@ -31,24 +32,77 @@
...
+        if (!password_enter("Enter\noptional passphrase", true, passphrase_out)) {
             return false;
         }
-        if (!password_enter("Confirm\noptional passphrase", true, mnemonic_passphrase_repeat)) {
+        if (strlen(passphrase_out) == 0) {
+            // No need to confirm the empty passphrase.
+            break;
+        }
+        if (_is_only_spaces(passphrase_out)) {
+            workflow_status_create("Invalid passphrase\nPlease try again", false);
```
conte91: Is this really the case? Can't I be sadistic enough to want my passphrase to be `␣␣␣␣`?

src/workflow/unlock.c
```diff
@@ -31,24 +32,77 @@
...
+            continue;
+        }
+        if (!workflow_confirm(
+                "", "You will be asked to\nvisually confirm your\npassphrase now.", false, true)) {
```
conte91: I would skip this and move on to the confirmation, the user has wasted quite some time already :) (Not a UI expert though, so that's just my opinion)

src/workflow/unlock.c
```diff
@@ -31,24 +32,77 @@
...
+        if (_has_dangerous_spaces(passphrase_out)) {
+            if (!workflow_confirm_scrollable(
+                    "Danger",
+                    "Your passphrase starts or ends with a space, or contains consecutive spaces.",
+                    true)) {
+                return false;
+            }
+        }
+        bool cancel_forced = false;
+        if (workflow_confirm_scrollable_longtouch("Confirm", passphrase_out, &cancel_forced)) {
```
conte91: (See the other comment) we can avoid double confirmation by just changing the confirm message: `Please confirm your passphrase`

src/workflow/unlock.c
```diff
@@ -31,24 +32,77 @@
...
+/**
+ * @return true if the passphrase starts or ends with a space or contains consecutive spaces.
+ */
+static bool _has_dangerous_spaces(const char* passphrase)
+{
+    size_t len = strlen(passphrase);
+    if (len == 0) {
+        return false;
+    }
+    if (passphrase[0] == ' ' || passphrase[len - 1] == ' ') {
```
conte91: We can avoid this being dangerous if we put quotes around the passphrase when we show it (so that leading/trailing space is well visible)

src/workflow/unlock.c
benma: Yeah it's not technically invalid. Can get rid of it later when we have underbrackets.

src/workflow/unlock.c
benma: For the record, I made this as a middleground solution for what @NickeZ asked. Without warning, you might accidentally expose it.

src/workflow/unlock.c
benma: Let's postpone and make the right fix (underbrackets) sometime.

PR #265, opened 2019-12-18
Mono font

NickeZ:
Added Courier New


src/workflow/confirm.c
```diff
@@ -105,11 +105,16 @@ enum workflow_async_ready workflow_confirm_async(
     }
 }
 
-bool workflow_confirm(const char* title, const char* body, bool longtouch, bool accept_only)
+bool workflow_confirm(
+    const char* title,
+    const char* body,
+    bool longtouch,
+    bool monospaced,
```
benma: What about providing directly the desired font, like here? https://github.com/digitalbitbox/bitbox02-firmware/blob/4dfc14acc8d38f61e6c5e66c495b0325eddbe9df/src/ui/components/label.c#L286

Can be passed down all the way to the confirm label component, which default to the old one if NULL.

src/workflow/confirm.c
```diff
@@ -105,11 +105,16 @@ enum workflow_async_ready workflow_confirm_async(
     }
 }
 
-bool workflow_confirm(const char* title, const char* body, bool longtouch, bool accept_only)
+bool workflow_confirm(
```
benma: We get more and more combinations of the confirm functions and params. Later we could think about making only one function that takes a config struct instead with good defaults, then calling it is more readable.

https://github.com/digitalbitbox/bitbox02-firmware/pull/253 will also need a new combination where the confirmation buttons are at the top during transaction confirmation :sweat_smile:

benma: If you do this, then using the password font here would be possible too:
https://github.com/digitalbitbox/bitbox02-firmware/pull/266#issuecomment-567472999

src/workflow/confirm.c
NickeZ: ack

src/workflow/confirm.c
NickeZ: sounds like a good refactoring issue. Can you create it?

NickeZ:
I've changed to propagating the `UG_FONT*`. So let's merge this before the password font.

src/workflow/confirm.c
benma: Done

src/workflow/unlock.c
```diff
@@ -84,19 +84,25 @@ static bool _get_mnemonic_passphrase(char* passphrase_out)
...
                     "Your passphrase starts or ends with a space, or contains consecutive spaces.",
+                    NULL,
                     true)) {
                 return false;
             }
         }
         bool cancel_forced = false;
-        if (workflow_confirm_scrollable_longtouch("Confirm", passphrase_out, &cancel_forced)) {
+        if (workflow_confirm_scrollable_longtouch(
+                "Confirm", passphrase_out, NULL, &cancel_forced)) {
```
benma: This should use the password font to make the spaces visible.

benma: Nevermind, I'm in the wrong PR.

NickeZ: Not in this PR. That comes with the password font PR.

thisconnect:
@NickeZ isn't the Courier New License an issue?

https://docs.microsoft.com/en-us/typography/font-list/courier-new

NickeZ:
Yes obviously we need licenses to modify and redistribute the fonts we use.

NickeZ:
For now we will go with a public domain font instead of purchasing license for Courier New.

PR #266, opened 2019-12-19
Pw space [depends on mono font]

NickeZ:


benma:
You are my hero :heart_eyes: 

In unlock.c in the confirmation of the passphrase, you can also use this font and remove the checks and warnings involving spaces, since they would become visible.

NickeZ:
> You are my hero heart_eyes

Life goal achieved ;)

> 
> In unlock.c in the confirmation of the passphrase, you can also use this font and remove the checks and warnings involving spaces, since they would become visible.

Done

src/ui/fonts/password_9X9.c
```diff
@@ -0,0 +1,108 @@
+// Converted from password.ttf
```
benma: Would be useful to commit this file as well? Eventually those files could be generated at build time from the ttfs.

NickeZ: That is a "search and replace" effect. The original file is the one referenced in the normal "a" fonts and that file is lost afaik. This comment should be updated to mention that it is almost a copy of the "font_a" c/h files.

NickeZ: Updated the comment.

PR #267, opened 2019-12-19
workflow/restore: Confirm name/time

NickeZ:


src/workflow/restore.c
```diff
@@ -36,6 +37,20 @@ bool workflow_restore_backup(const RestoreBackupRequest* restore_request)
...
     }
 
+    char buf[MEMORY_DEVICE_NAME_MAX_LEN + 20] = {0};
+    char* name = backup.backup_v1.content.metadata.name;
+    uint32_t* timestamp = &backup.backup_v1.content.metadata.timestamp;
+    time_t local_timestamp = *timestamp + restore_request->timezone_offset;
+    struct tm* local_time = localtime(&local_timestamp);
+    size_t len = snprintf(buf, sizeof(buf), "%s", name);
+    if (len < sizeof(buf)) {
+        strftime(&buf[len], sizeof(buf) - len, "\n(%Y-%m-%d)", local_time);
```
benma: should include the time, as it is displayed in the app.

src/workflow/restore.c
```diff
@@ -36,6 +37,20 @@ bool workflow_restore_backup(const RestoreBackupRequest* restore_request)
...
+    }
+
+    if (!workflow_confirm("Restore?", buf, true, false)) {
```
benma: this does not work for long backup names. try making a backup with the longest possible name. confirming the name is a scrollable confirm, but here it will be truncated from both sides.

Probably best to split it into two confirmation screens: confirm ID and datetime, then confirm name (scrollable).

Alternatively, truncate the name and add `...` ellipsis, similar to the long ETH amounts [here](https://github.com/digitalbitbox/bitbox02-firmware/blob/945b50bb702b1b0b9b159bc6c182111e56ac8377/src/apps/eth/eth_common.c#L94)

benma: btw, check backup also has the same issue with too long names :/

NickeZ:
Closing this as a unified approach probably should be implemented for both creating and restoring from backup.

PR #268, opened 2019-12-19
workflow/backup: Show long names correctly

NickeZ:
This has one potential blocker. The three confirmation screens have the confirm button on opposite sides. Especially weird when the name is short and the scrollable slider isn't needed.

### DIFF
--- a/src/workflow/backup.c
+++ b/src/workflow/backup.c
@@ -132,8 +132,15 @@ bool workflow_backup_check(char* id_out, bool silent)
         return false;
     case BACKUP_OK:
         if (!silent) {
-            workflow_status_create("Backup valid\nConfirm details", true);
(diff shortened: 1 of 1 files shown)


benma:
> This has one potential blocker. The three confirmation screens have the confirm button on opposite sides. Especially weird when the name is short and the scrollable slider isn't needed.

Same issue in https://github.com/digitalbitbox/bitbox02-firmware/pull/253 -> you (or I) can make add a position argument to the confirm functions.

Maybe time to consider moving all buttons to the top for consistency.

NickeZ:
> Maybe time to consider moving all buttons to the top for consistency.

:+1:

benma:
> > Maybe time to consider moving all buttons to the top for consistency.
> 
> +1

So let's to this then before merging this.

PR #270, opened 2019-12-19
Hash monospaced

NickeZ:
The firmware hash looks really nice with the monospaced font.


benma:
This will live on in https://github.com/digitalbitbox/bitbox02-firmware/pull/632

PR #273, opened 2019-12-23
Monogram font

NickeZ:
This PR adds a public domain font ([monogram](https://datagoblin.itch.io/monogram)) and enables it for the pairing request.


src/rust/bitbox02/src/lib.rs
```diff
@@ -96,7 +96,13 @@ pub fn delay(duration: Duration) {
 }
 
 // Safe wrapper for workflow_confirm
-pub fn workflow_confirm(title: &str, body: &str, longtouch: bool, accept_only: bool) -> bool {
+pub fn workflow_confirm(
+    title: &str,
+    body: &str,
+    longtouch: bool,
+    monospaced: bool,
```
benma: would be good to mirror the C version of this function and have the font param here. maybe this needs some re-exporting from C to Rust, but it will be worth it as we will have workflow_confirm take a struct with this param anyway soon.

NickeZ: hmm. I think taking a global static pointer is a very C-centric function signature. I would prefer some other signature, like having an enumeration of the fonts as an argument. But we can discuss that later.

NickeZ: accessing external statics was unsafe (which is understandable because rust doesn't construct or destruct the variable) so now calling the rust function `workflow_confirm` will involve using unsafe:

```rust
    bitbox02::workflow_confirm(
        "Base pairing code",
        &formatted,
        false,
        unsafe { &bitbox02::font_monogram_5X9 },
        false,
    )
```

A future interface could be

```rust
enum Font {
    Monospace9,
    Monospace11,
    Regular9,
    Regular11,
    Regular....,
}
```

Callsite would then look like:

```rust
    bitbox02::workflow_confirm(
        "Base pairing code",
        &formatted,
        false,
        Font::Monospace9,
        false,
    )
```

fabooowy:
Nice find @NickeZ , I like the look of the font. Curious to see what it looks on like on the device.

PR #274, opened 2019-12-24
Limit the amounts of events created [wip]

NickeZ:


NickeZ:
I can't reproduce the issue right now. We will have to wait with reviewing this issue until we can reliably reproduce it. I did manage to see the issue when I wrote the fix.

PR #275, opened 2020-01-02
Add ability to get the root fingerprint

TheCharlatan:
This adds the ability to get a fingerprint of the key at `m/0'` in the form of what electrum implements `https://github.com/spesmilo/electrum/blob/master/electrum/bip32.py#L240` to support the bitbox02 electrum integration.


py/bitbox02/bitbox02/bitbox02.py
```diff
@@ -362,6 +363,16 @@ def remove_sdcard(self) -> None:
         )
         self._msg_query(request, expected_response="success")
 
+    def root_fingerprint(self) -> bytes:
+        """
+        Get the root fingerprint form the bitbox02
```
benma: form->from

src/commander/commander.c
```diff
@@ -196,6 +197,17 @@ static commander_error_t _api_insert_remove_sdcard(const InsertRemoveSDCardReque
     return COMMANDER_OK;
 }
 
+static commander_error_t _api_get_root_fingerprint(RootFingerprintResponse* response)
+{
+    char fingerprint[4];
+    bool success = get_root_fingerprint(fingerprint, sizeof(fingerprint));
+    if (!success) {
+        return COMMANDER_ERR_INVALID_STATE;
```
benma: `COMMANDER_ERR_GENERIC` is better, as invalid state is used only by commander_states.c.

src/commander/commander.c
```diff
@@ -196,6 +197,17 @@ static commander_error_t _api_insert_remove_sdcard(const InsertRemoveSDCardReque
...
+    }
+    memcpy(&response->fingerprint, fingerprint, sizeof(fingerprint));
```
benma: no need for this, you can use `response->fingerprint` directly above instead of using a local `fingerprint` var.

src/keystore.c
```diff
@@ -427,6 +427,20 @@ static bool _get_xprv(const uint32_t* keypath, const size_t keypath_len, struct
     return true;
 }
 
+bool get_root_fingerprint(char* fingerprint, int fingerprint_len)
```
benma: `keystore_get_rooot_fingerprint` -> it is good to prefix everything as in C, everything is in global namespace :(

src/keystore.h
```diff
@@ -125,6 +125,8 @@ USE_RESULT bool keystore_bip39_mnemonic_to_seed(
     uint8_t* seed_out,
     size_t* seed_len_out);
 
+USE_RESULT bool get_root_fingerprint(char* fingerprint, int fingerprint_len);
```
benma: add a docstring

src/keystore.c
```diff
@@ -427,6 +427,20 @@ static bool _get_xprv(const uint32_t* keypath, const size_t keypath_len, struct
...
+{
+    static const uint32_t keypath[] = {
+        0 + BIP32_INITIAL_HARDENED_CHILD,
```
benma: this is `m/0'`, but that is not the root? `m/` is. any reasoning I am missing?

src/keystore.c
benma: you can drop `fingerprint_len` and state in the docstring that the passed buffer must be 4 bytes long. This is what we usually do.

Alternatively, use `size_t fingerprint_len`, and Abort() if it isn't `4`. This might make it a bit safer for the programmer, but is technically not needed as it's a constant `4` anyway.

benma: fingerprint: use `uint8_t*` for bytes for consistency.

src/keystore.c
```diff
@@ -427,6 +427,20 @@ static bool _get_xprv(const uint32_t* keypath, const size_t keypath_len, struct
     return true;
 }
 
+bool get_root_fingerprint(char* fingerprint, int fingerprint_len)
+{
+    static const uint32_t keypath[] = {
```
benma: why `static`? it is enough to have it be a normal stack var.

src/keystore.c
```diff
@@ -427,6 +427,20 @@ static bool _get_xprv(const uint32_t* keypath, const size_t keypath_len, struct
...
+        0 + BIP32_INITIAL_HARDENED_CHILD,
+    };
+
+    struct ext_key derived_xpub __attribute__((__cleanup__(keystore_zero_xkey))) = {0};
+    if (!keystore_get_xpub(keypath, sizeof(keypath) / sizeof(uint32_t), &derived_xpub)) {
+        return false;
+    }
+    snprintf(fingerprint, fingerprint_len, "%s", derived_xpub.hash160);
```
benma: hash160 is a bytes array, not a string, it is meaningless to use snprintf here and null terminate it. This will even cut your 4 bytes fingerprint to 3 bytes + `\0`.

Use `memcpy` instead to copy the first 4 bytes.

test/unit-test/test_keystore.c
```diff
@@ -209,6 +209,17 @@ static void _test_keystore_get_xpub(void** state)
     wally_free_string(xpub_string);
 }
 
+static void _test_keystore_get_root_fingerprint(void** state)
+{
+    mock_state(_mock_seed, _mock_bip39_seed);
+    char fingerprint[4];
+    assert_true(get_root_fingerprint(fingerprint, sizeof(fingerprint)));
+    assert_int_equal(fingerprint[0], 0xffffffffffffffca);
```
benma: `assert_memory_equal` easier to compare the whole fingerprint at once? it is weird with the 0xfff... padding / int conversion.

test/unit-test/test_keystore.c
```diff
@@ -209,6 +209,17 @@ static void _test_keystore_get_xpub(void** state)
...
+    assert_int_equal(fingerprint[1], 0x6d);
+    assert_int_equal(fingerprint[2], 0xffffffffffffff8b);
+    assert_int_equal(fingerprint[3], 0);
```
benma: did you check this against expected fingerprints from other libraries? this is probably the null temrinator from your snprintf, not the real fingerprint.

benma:
> This adds the ability to get a fingerprint of the key at `m/0'` in the form of what electrum implements `https://github.com/spesmilo/electrum/blob/master/electrum/bip32.py#L240` to support the bitbox02 electrum integration.

why `m/0'` over `m/`? also the whole description would be good to have in the commit message itself.

src/keystore.c
TheCharlatan: No, just me getting the naming wrong. How about `get_accounts_fingerprint`?

TheCharlatan:
Thanks for the patient review at @benma I think I addressed most points now. I'll cross check the produced fingerprint with electrum now and then amend the commit message.

src/keystore.c
TheCharlatan: On the other hand it does get the fingerprint of the parent public key of `m/0'`. That should then be the first four bytes of `m/`, right?

TheCharlatan:
I cross checked  the fingerprint with electrum by first removing the keypath checks in the bitbox02 firmware and checking then that the electrum derived fingerprint of the `m/0'` pubkey is the same as the one returned by getting the fingerprint through the python library as implemented by this pull request.

src/keystore.c
benma: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#key-identifiers

first 4 bytes of hash160 is the key fingerprint. children also encode the parent key fingerprint (see [here](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#serialization-format)).

To get the root fingerprint, you take the xpub at m/ at take the 4 bytes of hash160, which I believe you already did, but for `m/0'` instead.

TheCharlatan: so rename to `get_accounts_fingerprint`?

src/keystore.c
```diff
@@ -427,6 +427,23 @@ static bool _get_xprv(const uint32_t* keypath, const size_t keypath_len, struct
     return true;
 }
 
+bool keystore_get_root_fingerprint(uint8_t* fingerprint)
+{
+    if (sizeof(fingerprint) < sizeof(uint32_t)) {
```
benma: this does not work as expected; `sizeof(singerprint)` is the size of the pointer, which happens to be 4 bytes on the device, but 8 bytes on x86_64. That's why my comment said to either pass the length and check it here, or drop it completely and just describe the required size in the docstring :)

src/keystore.c
benma: It's still static.

src/keystore.c
```diff
@@ -427,6 +427,23 @@ static bool _get_xprv(const uint32_t* keypath, const size_t keypath_len, struct
...
+    }
+    static const uint32_t keypath[] = {
+        0 + BIP32_INITIAL_HARDENED_CHILD,
+    };
+
+    struct ext_key derived_xpub __attribute__((__cleanup__(keystore_zero_xkey))) = {0};
+    if (!keystore_get_xpub(keypath, sizeof(keypath) / sizeof(uint32_t), &derived_xpub)) {
+        return false;
+    }
+    memcpy(fingerprint, derived_xpub.parent160, sizeof(uint32_t));
```
benma: I see you take the parent160 here, but it's more straight forward to get the xpub at `m/` and copy from `derived_xpub.hash160` instead, which saves some work and should be the same result.

benma: not sure `sizeof(uint32_t)` is useful, as there is no `uint32_t` type used anywhere. I'd just put `4` here.

src/keystore.h
```diff
@@ -125,6 +125,13 @@ USE_RESULT bool keystore_bip39_mnemonic_to_seed(
     uint8_t* seed_out,
     size_t* seed_len_out);
 
+/**
+ * Fills a uint8_t buffer with a fingerprint,
```
benma: "a fingerprint" is a bit vague, what about "with the fingerprint of the root xpub as specified in https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#serialization-format"?

src/keystore.c
TheCharlatan: It's what libwally uses to calculate the fingerprint that is encode with each extended key. I also found more `sizeof(uint32_t)` than `4`, so I thought I was being consistent here.

TheCharlatan: Mmh, I disagree here, since it is the hash of the parent public key that is included in the xpub, so getting the fingerprint of a public key always implies taking to the first four bytes of the hash160 of its parent. Further, I don't know how to get from `m/` with the usual keypath syntax :P

TheCharlatan: Anyway, changed it to `4` now, since it is more readable.

src/keystore.h
TheCharlatan: Done.

src/keystore.c
TheCharlatan: I dropped the check entirely now.

src/keystore.c
benma: Hmm there shouldn't be anyting to disagree on, you have ```fingerprint(root xpub) == parentfingerprint(child xpub)```, so deriving a child to get the parent fingerprint when you already have the parent fingerprint in memory is both physical overhead and mental overhead when reading/maintaining the code :P

> Further, I don't know how to get from m/ with the usual keypath syntax :P

`keystore_get_xpub(NULL, 0, ...)`. Could add a small comment that this is allowed in the docstring of `keystore_get_xpub()`.

src/keystore.c
```diff
@@ -427,6 +428,20 @@ static bool _get_xprv(const uint32_t* keypath, const size_t keypath_len, struct
     return true;
 }
 
+bool keystore_get_root_fingerprint(uint8_t* fingerprint)
+{
+    struct ext_key derived_xpub __attribute__((__cleanup__(keystore_zero_xkey))) = {0};
+    if (!keystore_get_xpub(NULL, 0, &derived_xpub)) {
+        return false;
+    }
+    unsigned char fingerprint_buffer[4];
```
benma: i think you can drop this variable and the memcpy below, and just pass `fingerprint` to `bip32_key_get_fingerprint`.

If you want to keep the buffer size safety check, in commander.c, you could call `keystore_get_root_fingerprint(response->fingerprint, sizeof(response->fingerprint))` and pass the size to `bip32_key_get_fingerprint()`, which will make sure it is 4 bytes.

TheCharlatan: Removed the extra buffer, I won't add an explicit size check.

benma:
:clap: :clap:

PR #282, opened 2020-01-13
py/bitbox02: fix check sd card

benma:


benma:
@NickeZ why doesn't our CI not catch this with mypy? :(

NickeZ:
> @NickeZ why doesn't our CI not catch this with mypy? :(

Python is a dynamically typed language and we are even disabling one check that might have caught it: `# pylint: disable=no-member`

PR #285, opened 2020-01-15
eth: Allow unsupported tokens [wip]

NickeZ:
I have to check that the confirmation screen looks good before we merge this.

### DIFF
--- a/src/apps/eth/eth_verify.c
+++ b/src/apps/eth/eth_verify.c
@@ -4,6 +4,7 @@
 
 #include <hardfault.h>
 #include <util.h>
+#include <workflow/confirm.h>
 #include <workflow/verify_recipient.h>
 #include <workflow/verify_total.h>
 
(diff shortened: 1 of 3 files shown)


src/apps/eth/eth_verify.c
```diff
@@ -83,15 +84,23 @@ static app_eth_sign_error_t _verify_total_fee(
...
-        return APP_ETH_SIGN_ERR_INVALID_INPUT;
+        if (!workflow_confirm(
+                "Continue?",
+                "Unknown token\nplease verify amount\nseparately.",
+                NULL,
+                false,
+                true)) {
+            return APP_ETH_SIGN_ERR_INVALID_INPUT;
+        }
+        erc20_params = app_eth_erc20_unknown_params_get();
```
benma: imho this feels like a hack; the two follow up confirmation screens will probably look weird or confusing, and it is also a bit hard to read as one has to think through what empty/default params imply.

for better readability i would be more explicit:
```if (erc20_params == NULL) { return _verify_unknown_erc20(...); } ``` and not use a "fake" eth_params value.

NickeZ:
Leaning towards printing "Unknown value" instead of decimals at 0.

PR #288, opened 2020-01-16
Py: Add min firmware version check to the bitbox02 lib

TheCharlatan:


py/bitbox02/bitbox02/bitbox02.py
```diff
@@ -43,6 +43,7 @@
 except ModuleNotFoundError:
     pass
 
+MIN_BITBOX02_FIRMWARE_VERSION = (5, 0, 0)
```
benma: `BITBOX02` is redundant

py/bitbox02/bitbox02/bitbox02.py
```diff
@@ -108,6 +109,11 @@ def device_info(self) -> Dict[str, Any]:
             "monotonic_increments_remaining": response.device_info.monotonic_increments_remaining,
         }
 
+    def check_bitbox02_firmware_version(self) -> bool:
+        bitbox02_info = self.device_info()
```
benma: You need to use `self.version` and do the check in the constructor.

Reason:
`self.version` is set in the constructor by OP_INFO - `device_info()` might not be around in a future change. Actually we should remove the `version` field from there, it was added in the beginning but never used.

In the Go lib, the min bound check is done after pairing, as Reboot() needs to be available which runs over noise. In the Python lib, you could do both checks in the constructor and abort early with the right message (requires app/library upgrade, requires firmware upgrade).

TheCharlatan: Updated:
The communication lib now checks that the firmware is newer than a minimum firmware version and raises an exeption if this is not the case.
    
Additionally a function is added in the bitbox02 to check whether the firmware is newer than what is required as a minimum, meaning that the library should probably be updated.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -407,3 +419,22 @@ def get_info(self) -> Tuple[str, Platform, Union[BitBox02Edition, BitBoxBaseEdit
         unlocked_byte = response[0]
         unlocked = {0x00: False, 0x01: True}[unlocked_byte]
         return (version_str, platform, edition, unlocked)
+
+    def check_version(self) -> Tuple[bool, semver.VersionInfo]:
```
benma: returning the constant `MIN_BITBOX02_MULTI_FIRMWARE_VERSION` is not helpful imho, can just use `MIN_BITBOX02_MULTI_FIRMWARE_VERSION` wherever you need it.

benma: This should check for minimum supported and minimum not supported, otherwise the library will break with each firmware upgrade, even though most firmware upgrades would be compatible.

I'd do it the same ways as in the app:

https://github.com/digitalbitbox/bitbox02-api-go/blob/master/api/firmware/device.go#L32-L37

It can happen e.g. that the range of supported versions is 6.3.0 to 8.0.0.

TheCharlatan: As discussed I limit the supported range within the lowest supported in the next highest major version. Meaning 4.9.9 is supported if the lowest supported is 4.8.0, but 5.0.0 is no longer supported. Both upper and lower version mismatches now throw their own exception, each with their own unique text.

py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -108,6 +115,19 @@ def device_info(self) -> Dict[str, Any]:
...
+        """
+        Returns False if the bitbox02 version is not what the library expects.
+        The connection should already bail early in case of a version mismatch,
+        so False indicates that some expected functionality might be missing.
+        """
+        # pylint: disable=no-member
+        _, _, edition, _ = self.get_info()
+
+        if edition == BitBox02Edition.MULTI:
+            return bool(self.version == MIN_BITBOX02_MULTI_FIRMWARE_VERSION)
```
NickeZ: Why does the comparisons here use `==`.

NickeZ: It shouldn't be necessary to typecast to `bool`.

TheCharlatan: Mypy seems to think that `==` returns any - not sure why. 
The idea of this check is to establish whether the library will be able to use the full feature set of the firmware, or just a subset of it. The check is only useful, if the minimum firmware version if incremented each time we add a new firmware feature to the library. It can then function as a reminder to the users of the library that the library may be outdated.

py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -108,6 +115,19 @@ def device_info(self) -> Dict[str, Any]:
...
         }
 
+    def check_firmware_version(self) -> bool:
+        """
+        Returns False if the bitbox02 version is not what the library expects.
+        The connection should already bail early in case of a version mismatch,
+        so False indicates that some expected functionality might be missing.
+        """
+        # pylint: disable=no-member
+        _, _, edition, _ = self.get_info()
```
benma: get_info (OP_INFO) was added in v4.2.0 i think, so this will fail for earlier versions. It is also a bit strange to get the edition from `get_info()` but `self.version` from the USB descriptor; both should come from the same source.

Please add `self.edition` alongside `self.version` in __init__, based on the descriptor, same as in the Go lib:

https://github.com/digitalbitbox/bitbox02-api-go/blob/master/api/common/common.go#L45-L57

py/bitbox02/bitbox02/bitbox02/bitbox02.py
benma: This is not what we discussed last time, is it? Bumping firmware minor version should not affect the Electrum user at all, as nothing can break (compatible change by definition).

The user should be forced to upgrade if the version is lower than `MIN_BITBOX02_MULTI_FIRMWARE_VERSION`. 

You proposed to require an upgrade if the version is too low, and tell the user to upgrade the library/Electrum if the version is higher than the next major version (e.g. for the min version 6.1.2, anything >= 7.0.0 would give the warning). That sounded good. What changed?

benma: >  It shouldn't be necessary to typecast to bool.

:+1:

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -407,3 +418,44 @@ def get_info(self) -> Tuple[str, Platform, Union[BitBox02Edition, BitBoxBaseEdit
...
+                self.version
+            )
+        )
+
+    def check_version(self) -> None:
+        """
+        Returns false if the device has an older firmware version than required
+        and the minimum required version.
+        """
+        _, platform, edition, _ = self.get_info()
```
benma: same, set the edition in `__init__` based on the usb descriptor (you can ignore `platform`/base for now, as that needs to come from get_info indeed, but this would need a bigger refactor).

py/bitbox02/bitbox02/bitbox02/bitbox02.py
benma: Now I am confused, why is there a check_firmware_version and a check_version method?

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -407,3 +418,44 @@ def get_info(self) -> Tuple[str, Platform, Union[BitBox02Edition, BitBoxBaseEdit
...
+        if platform == Platform.BITBOX02:
+            if edition == BitBox02Edition.MULTI:
+                if self.version < MIN_BITBOX02_MULTI_FIRMWARE_VERSION:
+                    self.firmware_outdated(MIN_BITBOX02_MULTI_FIRMWARE_VERSION)
```
benma: nit: `raise self.firmware_outdated(...)` would be a bit easier to read, as like this it's not clear that this stops the function.

py/bitbox02/bitbox02/bitbox02/bitbox02.py
TheCharlatan: I can remove this check, if you think it is useless. It's not supposed to force anything, the check in `check_version` does that. It could act as a reminder that a newer version of the python library, that can use more features of the firmware compared to the currently used one might be available, but it's a pretty weak suggestion.

py/bitbox02/bitbox02/bitbox02/bitbox02.py
TheCharlatan: Removed the method here.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
TheCharlatan: Done, moved everything into separate exception classes.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
TheCharlatan: Ok, I removed the check for the BitBoxBase entirely now and use the product string from the device dict to get the edition information.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -156,6 +160,28 @@ class AttestationException(Exception):
     pass
 
 
+class VersionOutdatedException(Exception):
```
benma: unused

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -156,6 +160,28 @@ class AttestationException(Exception):
...
+    pass
+
+
+class FirmwareVersionOutdatedException(Exception):
+    def __init__(self, version: semver.VersionInfo, required_version: semver.VersionInfo):
+        super().__init__(
+            "The BitBox02's firmware is no up to date. Device: {}, Required: {}".format(
```
benma: typo "is not up to date"

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -214,6 +245,9 @@ def __init__(
                     self.close()
                     raise Exception("Unlock process aborted")
 
+        # raises exceptions if either firmware or library are out of date, does not check BitBoxBase
+        self.check_version()
```
benma: This should be moved to above the unlock and attestation check, as you might never come this far if the firmware version is too new and those two calls change (which they will in 7.0.0 :laughing: )

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -190,6 +216,11 @@ def __init__(
         serial_number = device_info["serial_number"]
         self._transport = transport
 
+        if device_info["product_string"] == BITBOX02MULTI:
+            self.edition = BitBox02Edition.MULTI
+        elif device_info["product_string"] == BITBOX02BTC:
+            self.edition = BitBox02Edition.BTCONLY
+
```
benma: else raise exception

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -407,3 +441,27 @@ def get_info(self) -> Tuple[str, Platform, Union[BitBox02Edition, BitBoxBaseEdit
...
+        if self.edition == BitBox02Edition.MULTI:
+            if self.version < MIN_BITBOX02_MULTI_FIRMWARE_VERSION:
+                raise FirmwareVersionOutdatedException(
+                    self.version, MIN_BITBOX02_MULTI_FIRMWARE_VERSION
+                )
+            if self.version >= semver.VersionInfo(
+                MIN_BITBOX02_MULTI_FIRMWARE_VERSION.major + 1, 0, 0
+            ):
+                raise LibraryVersionOutdatedException(self.version)
+        elif self.edition == BitBox02Edition.BTCONLY:
```
benma: else raise exception (think of it like a switch which should always treat all possibilities)

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
TheCharlatan: then I'm raising an exception whenever there is a base?

benma: ahhh good point, nvm then

PR #298, opened 2020-01-20
ugui: Add support for dynamically wrapping title

NickeZ:
This will be needed when we add a lot of tokens to support.


benma:
Imho it would be worth it to wait for https://github.com/digitalbitbox/bitbox02-firmware/pull/286 and then add a unit test that exercises the function for all token names, so the sanitizer can do its work on the actual data.

conte91:
> Imho it would be worth it to wait for #286 and then add a unit test that exercises the function for all token names, so the sanitizer can do its work on the actual data.

I like the idea of having a "basic" test, so we can tackle any bug more easily. We can always add a separate unit test later., so it gets run every time the token list is regenerated.

NickeZ:
@benma could you rereview?

PR #300, opened 2020-01-20
Refactor pkscript

benma:


benma:
Will reopen in a better shape.

PR #303, opened 2020-01-22
apps/btc/sign: keypath_account added to BTCSigninit

benma:
It is the derivation to the account, e.g. m/84'/0'/1'.

It is a prefix to all input and change keypaths.

This subsumes bip44_account (as it is part of it), and will make it
easier to verify the account to be used during signing upfront. Doing
it that way is easier to read/maintain, as we don't need to cache the
result so each input/change don't have to redo the validation.

### DIFF
--- a/src/apps/btc/btc_sign.c
+++ b/src/apps/btc/btc_sign.c
@@ -187,11 +187,12 @@ static app_btc_sign_error_t _sign_input_pass1(
 }
 
 static bool _is_valid_keypath(
+    const uint32_t* keypath_account,
+    size_t keypath_account_count,
     const uint32_t* keypath,
     size_t keypath_count,
     const BTCScriptConfig* script_config,
     uint32_t expected_bip44_coin,
-    uint32_t expected_bip44_account,
     bool must_be_change)
 {
     switch (script_config->which_config) {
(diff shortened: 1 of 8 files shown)


src/apps/btc/btc_sign.c
```diff
@@ -200,15 +201,26 @@ static bool _is_valid_keypath(
...
         }
-        const uint32_t account = keypath[2];
-        const uint32_t change = keypath[3];
-        if (account != expected_bip44_account || (must_be_change && change != 1)) {
+        break;
+    default:
+        return false;
+    }
+
+    // check that keypath_account is a prefix to keypath with two elements left (change, address).
```
TheCharlatan: Are you changing this again for multisig's extra non-hardened derivation level?

benma: which keypath format are you talking about? electrum has bip48 which is an extra hardened derivation level, so the account keypath is four elements; m/48'/coin'/acount'/script_type' (script_type being 2 for p2wsh).

TheCharlatan: Never mind, they are both hardened.

benma:
Uhhm so I accidentally merged this after somehow seeing @conte91's approval, but that must have been a different PR or just my imagination :o

please check the PR anyway, I'll address feedback in a new PR then. Sorry!

src/apps/btc/btc_sign.c
TheCharlatan: No, I was right before, [BIP45](https://github.com/bitcoin/bips/blob/master/bip-0045.mediawiki) uses this weird format : `m/purpose'/cosigner_index/change/address_index`. But we can choose to just not support it.

TheCharlatan: BIP48 does not exist btw, it's all super confusing. It's only documentation is here: https://github.com/bitpay/copay#wallet-export-format , which quotes:

> Since version 1.5, Copay uses the root m/48' for hardware multisignature wallets. This was coordinated with Ledger and Trezor teams. While the derivation path format is still similar to BIP44, the root was in order to indicate that these wallets are not discoverable by scanning addresses for funds. Address generation for multisignature wallets requires the other copayers extended public keys.

...meaning the checks are fine as they are.

benma: > No, I was right before, [BIP45](https://github.com/bitcoin/bips/blob/master/bip-0045.mediawiki) uses this weird format : `m/purpose'/cosigner_index/change/address_index`. But we can choose to just not support it.

yeah, as we've done with the other keypaths, I prefer to be strict at first and whitelist more if the need arises. Thanks for checking the other formats.

PR #304, opened 2020-01-23
device-tests build-system: disable semihosting by default

NickeZ:
You can enable semihosting for the tests with `cmake -DSEMIHOSTING=ON ..`

I've gotten bitten by this a few times. I think it makes sense to have semihosting opt-in for the device tests / demos.


NickeZ:
Since we made a new directory `build-semihosting` for semihosting builds, I definitely think this can be merged.

x1ddos:
> I've gotten bitten by this a few times. I think it makes sense to have semihosting opt-in for the device tests / demos.

Bitten by what?

NickeZ:
Device stops with black screen unless you have gdb connected with semihosting.

x1ddos:
> Device stops with black screen unless you have gdb connected with semihosting.

Ah, good to know!

NickeZ:
The changes in this PR was already partly applied. I still think the part that is left is necessary to apply for it to work correctly. According to [this readme](https://github.com/cbiffle/gcc-arm-embedded/blob/master/readme.txt) for example, the `--specs` flags must be applied to the linking step as well.

benma:
Device-tests don't exist anymore.

PR #305, opened 2020-01-23
apps/btc/common: add btc_common_pkscript_from_multisig

benma:


test/unit-test/test_app_btc_common.c
```diff
@@ -316,6 +319,143 @@ static void _test_btc_common_encode_xpub(void** state)
...
+    const char* xpubs[15];
+    uint32_t keypath_change;
+    uint32_t keypath_address;
+    const char* expected_script_hex;
+} test_case_redeemscript_multisig_p2wsh;
+
+static void _test_btc_common_pkscript_from_multisig(void** state)
+{
+    // clang-format off
+    const test_case_redeemscript_multisig_p2wsh tests[] = {
```
conte91: Add a test having an input with no xpubs (and `.xpubs_count == 0`) maybe

messages/btc.options
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+BTCScriptConfig.Multisig.xpubs max_count:15
+
```
conte91: No empty line I guess?

src/apps/btc/btc_common.c
```diff
@@ -356,3 +356,49 @@ bool btc_common_pkscript_from_outputhash(
...
+        if (bip32_key_from_parent_path(
+                &xpub, keypath, 2, BIP32_FLAG_KEY_PUBLIC, &derived_cosigner_xpub) != WALLY_OK) {
+            return false;
+        }
+        memcpy(
+            &pubkeys[index * EC_PUBLIC_KEY_LEN], derived_cosigner_xpub.pub_key, EC_PUBLIC_KEY_LEN);
+    }
+
+    size_t written;
+    if (wally_scriptpubkey_multisig_from_bytes(
```
conte91: What happens if we receive an input with `xpubs_count == 0`? Does libwally behave correctly with an input of size 0? (Also see below)

benma: Yes it does, although the docstring still says that the multisig params must be validated upfront (another function to do that is coming soon).

messages/btc.options
benma: Removed.

test/unit-test/test_app_btc_common.c
benma: added some negative tests.

PR #307, opened 2020-01-24
apps/btc/common: add btc_common_multisig_is_valid

benma:
Validate a multisig setup.


py/bitbox02/communication/generated/btc_pb2.pyi
```diff
@@ -1,10 +1,15 @@
 # @generated by generate_proto_mypy_stubs.py.  Do not edit!
 import sys
+from .common_pb2 import (
+    XPub as common_pb2___XPub,
```
TheCharlatan: Changes in the protobuf generated files without chanigng the proto files?

src/apps/btc/btc_common.c
```diff
@@ -15,11 +15,15 @@
...
 
 #include "btc_common.h"
+#include <apps/common/bip32.h>
 
 #include <apps/common/bip32.h>
+#include <keystore.h>
 #include <util.h>
 #include <wally_address.h>
 
+#define MULTISIG_P2WSH_MAX_SIGNERS (sizeof(multisig->xpubs) / sizeof(XPub))
```
TheCharlatan: This seems like a complicated way to write 15. Maybe add a comment?

py/bitbox02/communication/generated/btc_pb2.pyi
benma: Yeah it's from the previous PR when Multisig was added, but forgot to `make -C py` there.

src/apps/btc/btc_common.h
```diff
@@ -31,6 +31,7 @@
 
 #define BTC_PURPOSE_P2WPKH_P2SH (49 + BIP32_INITIAL_HARDENED_CHILD)
 #define BTC_PURPOSE_P2WPKH (84 + BIP32_INITIAL_HARDENED_CHILD)
+#define BTC_PURPOSE_MULTISIG (48 + BIP32_INITIAL_HARDENED_CHILD)
```
TheCharlatan: Why do we define some macros in the header and some in the source file? In my opinion they all should be in the source, since I don't see them being used outside of it.

src/apps/btc/btc_common.c
benma: Right, was too complicated. `15` works fine! Changed.

src/apps/btc/btc_common.h
benma: You are totally right, we normally do define them in the source if they are not used elsewhere (like the multisig max signers one ;)). I added a comit to move those. The remaining one will be handled soon.

src/apps/btc/btc_common.h
```diff
@@ -167,4 +160,22 @@ USE_RESULT bool btc_common_pkscript_from_multisig(
...
     size_t* script_out_size);
 
+/**
+ * Validate a m-of-n multisig account. This includes checking that:
+ * - 0 < m <= n <= 15
+ * - the keypath conforms to bip48 for p2wsh: m/48'/coin'/account'/2'
+ * - one of the xpubs is ours (the currenty unlocked keystore).
+ * @param[in] multisig Multisig configuration (threshold, signers). The xpubs are account-level
+ * xpubs.
+ * @param[in] keypath address-level keypath, e.g. m/48'/0'/10'/2'/0/10
```
TheCharlatan: I'm a bit confused here. The doc says here that a full keypath is expected, but then later you check this same keypath in `_is_valid_keypath_account_multisig_p2wsh` and restrict it to a size of four, seemingly without changing its size within `btc_common_multisig_is_valid`. Am I missing something here?

src/apps/common/bip32.c
```diff
@@ -43,3 +45,18 @@ bool apps_common_bip32_xpub_from_protobuf(const XPub* xpub_in, struct ext_key* x
     bip32_key_free(tmp);
     return true;
 }
+
+bool apps_common_bip32_xpubs_equal(const struct ext_key* xpub1, const struct ext_key* xpub2)
```
TheCharlatan: No unit test for this function?

src/apps/btc/btc_common.c
```diff
@@ -403,3 +439,46 @@ bool btc_common_pkscript_from_multisig(
...
+        return false;
+    }
+
+    struct ext_key our_xpub __attribute__((__cleanup__(keystore_zero_xkey))) = {0};
+    if (!keystore_get_xpub(keypath, keypath_len, &our_xpub)) {
+        return false;
+    }
+
+    bool our_xpub_found = false;
+    for (size_t cosigner_idx = 0; cosigner_idx < multisig->xpubs_count; cosigner_idx++) {
```
TheCharlatan: This seems a bit inefficient. How about adding the index of our key to the multisig message and only do the iteration if the memcmp between the key at that index and ours fails?

benma: by key you mean xpub right? key alone is not enough. memcmp doesn't work because of different types (hence `apps_common_bip32_xpubs_equal`).

Index provided by client would make it one comparison instead of looping and comparing all of them, true. Efficiency wise I wouldn't worry, it's a simple operation and isn't noticeable. The client integrations would be slightly more complicated, but on the other hand the rule that at least one has to be ours would be encoded in the message, which is nice.

I think I'll do it :P thanks!

src/apps/btc/btc_common.h
benma: This line in the docs is wrong :O the description a few lines further up is correct. i'll amend.

benma: Done now.

src/apps/common/bip32.c
benma: Added, good catch.

src/apps/btc/btc_common.c
benma: Done now.

benma: Actually this simplified this function and the commits that build on top quite a bit, nice suggestion!

PR #308, opened 2020-01-24
Test warnings

benma:


test/unit-test/test_usart.c
```diff
@@ -264,7 +264,7 @@ static void test_basic(void** state)
         expect_value(__wrap_usb_processing_enqueue, length, msg_len);
         expect_value(__wrap_usb_processing_enqueue, cmd, endpoints[i]);
         will_return(__wrap_usb_processing_enqueue, true);
-        _tx_buffer(encoded[i], encoded_len[i]);
+        _tx_buffer((const uint8_t*)encoded[i], encoded_len[i]);
```
conte91: Can't we directly make `encoded[i]` a `uint8_t*` instead?

benma: True, but not a prio for now :)

PR #309, opened 2020-01-27
eth: Change valid xpub keypath length to 4

NickeZ:
@Tomasvrba

### DIFF
--- a/py/send_message.py
+++ b/py/send_message.py
@@ -185,10 +185,11 @@ class SendMessage:
 
     def _btc_address(self) -> None:
         def address(display: bool) -> str:
+            # pylint: disable=no-member
             return self._device.btc_address(
                 keypath=[84 + HARDENED, 0 + HARDENED, 0 + HARDENED, 0, 0],
                 script_config=bitbox02.btc.BTCScriptConfig(
-                    simple_type=bitbox02.btc.BTCScriptConfig.P2WPKH  # pylint: disable=no-member
+                    simple_type=bitbox02.btc.BTCScriptConfig.P2WPKH
                 ),
                 display=display,
             )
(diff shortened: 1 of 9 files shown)


NickeZ:
We found by lookin at MEW that they use the xpub of depth 4.

Also see the ethereum book:

https://github.com/ethereumbook/ethereumbook/blob/develop/05wallets.asciidoc

> Because BIP-44 was created originally for Bitcoin, it contains a "quirk" that isn’t relevant in the Ethereum world. On the fourth level of the path, change, an HD wallet has two subtrees: one for creating receiving addresses and one for creating change addresses. Only the "receive" path is used in Ethereum, as there is no necessity for a change address like there is in Bitcoin. Note that whereas the previous levels used hardened derivation, this level uses normal derivation. This is to allow the account level of the tree to export extended public keys for use in a nonsecured environment. Usable addresses are derived by the HD wallet as children of the fourth level, making the fifth level of the tree the address_index. For example, the third receiving address for Ethereum payments in the primary account would be M/44&#x27;/60&#x27;/0&#x27;/0/2. BIP-44 HD wallet structure examples shows a few more examples.

Tomasvrba:
tACK from my side

src/apps/eth/eth_common.c
```diff
@@ -25,7 +25,33 @@
 
 #define BIP44_ETH_ACCOUNT_MAX (99) // 100 accounts
 
-bool eth_common_is_valid_keypath(ETHCoin coin, const uint32_t* keypath, size_t keypath_len)
+// Since ethereum doesn't use the "change" path part it is always 0 and have become part of the
```
benma: this should be in the docstring of the function in the header instead of here.

py/send_message.py
```diff
@@ -227,9 +229,7 @@ def _sign_btc_tx(self) -> None:
         ]
         sigs = self._device.btc_sign(
             bitbox02.btc.BTC,
-            bitbox02.btc.BTCScriptConfig(
-                simple_type=bitbox02.btc.BTCScriptConfig.P2WPKH  # pylint: disable=no-member
-            ),
+            bitbox02.btc.BTCScriptConfig(simple_type=bitbox02.btc.BTCScriptConfig.P2WPKH),
```
benma: surprising, the linter should have caught this misformatting before?

PR #311, opened 2020-01-29
commander_btc: Check the return value of snprintf.

conte91:
snprintf writing too much data would make the target
buffer contain a string with no terminating NULL byte. Check
that this is not happening before proceeding.

### DIFF
--- a/src/commander/commander_btc.c
+++ b/src/commander/commander_btc.c
@@ -53,6 +53,7 @@ static commander_error_t _btc_pub_xpub(const BTCPubRequest* request, PubResponse
     }
     if (request->display) {
         char title[100] = {0};
+        int n_written;
         switch (request->output.xpub_type) {
         case BTCPubRequest_XPubType_TPUB:
         case BTCPubRequest_XPubType_XPUB:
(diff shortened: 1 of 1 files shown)


src/commander/commander_btc.c
```diff
@@ -81,6 +82,13 @@ static commander_error_t _btc_pub_xpub(const BTCPubRequest* request, PubResponse
         default:
             return COMMANDER_ERR_GENERIC;
         }
+        if (n_written >= sizeof(title)) {
```
benma: n_written < 0 ||

benma: also move the whole thing to right after `n_written` is set. error handling should ideally be always the next line.

benma:
Also pleae change the commit msg, `snprintf` does null terminate, even if it truncates.

src/commander/commander_btc.c
conte91: Fixed.

src/commander/commander_btc.c
```diff
@@ -62,6 +62,7 @@ static commander_error_t _btc_pub_xpub(const BTCPubRequest* request, PubResponse
             return COMMANDER_ERR_GENERIC;
         }
         char title[100] = {0};
+        int n_written;
```
benma: why not inline? principle of smallest scope.

conte91: You can't allocate stack space within a `switch` statement!

PR #312, opened 2020-01-30
Async workflows

conte91:
This PR introduces a new proposal for async workflows.
A `workflow_t` class can be implemented by all workflows, and will contain methods to process the workflow status (`spin`), to create and destroy each workflow. The workflow currently on top of the stack will be spinned in each iteration of the main loop.


src/workflow/orientation_screen.c
```diff
@@ -0,0 +1,101 @@
...
+#include <ui/components/orientation_arrows.h>
+#include <ui/components/show_logo.h>
+#include <ui/components/waiting.h>
+#include <ui/screen_stack.h>
+#include <ui/workflow_stack.h>
+#if PLATFORM_BITBOXBASE == 1
+#include <usart/usart.h>
+#elif PLATFORM_BITBOX02 == 1
+#include <usb/usb.h>
+#endif
```
benma: remove? maybe double check if the other includes are all needed

src/ui/components/confirm.c
```diff
@@ -24,18 +24,41 @@
...
-            data->confirm_callback = NULL;
-        }
+        _dispath_confirm(component);
+    }
+}
+
+static void _on_confirm(component_t* component)
+{
+    component_t* self = component->parent;
+    _dispath_confirm(self);
```
benma: `dispath` -> `dispatch`?

src/workflow/confirm.c
```diff
@@ -18,48 +18,131 @@
...
+
+workflow_t* workflow_confirm(
+    const confirm_params_t* params,
+    void (*callback)(bool, void*),
+    void* callback_param)
+{
+    workflow_t* result = workflow_allocate(
+        _workflow_confirm_init, _workflow_confirm_cleanup, _workflow_confirm_spin);
+    data_t* data = malloc(sizeof(*data));
+    if (!data) {
```
benma: it's a good idea to memset 0 all malloced structs to avoid accidental undefined data (e.g. when the struct is extended).

then you could also skip the ` = NULL`, `=false` parts.

src/workflow/workflow.c
```diff
@@ -49,43 +38,24 @@ void workflow_confirm_dismiss(const char* title, const char* body)
...
-#if PLATFORM_BITBOX02 == 1
-/**
- * Called when the "select orientation" screen is over.
- * Switch to the main view.
- */
-static void _select_orientation_done(bool upside_down)
-{
-    if (upside_down) {
-        screen_rotate();
+    workflow_t* result = (workflow_t*)malloc(sizeof(*result));
```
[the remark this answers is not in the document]
conte91: This seems to have surprised  everybody, but I've been doing this in PRs since forever :D This way if `result` changes type, there's no risk of crapping out the `malloc`'d size.

src/workflow/confirm.c
conte91: I'm thinking of just moving this inside the `workflow_allocate` function. Every workflow is going to need some private data anyway :|

benma: workflow_allocate doesn't know what the data struct is or how big it is.

conte91: We can just pass `sizeof(data_t)` in `workflow_allocate` as a parameter. The `data` field is `void *` anyway.

src/firmware_main_loop.c
```diff
@@ -14,16 +14,25 @@
...
 void firmware_main_loop(void)
 {
     while (1) {
         screen_process();
-        usb_processing_process(usb_processing_hww());
+        if (usb_is_enabled()) {
+            usb_processing_process(usb_processing_hww());
 #if APP_U2F == 1
-        usb_processing_process(usb_processing_u2f());
+            usb_processing_process(usb_processing_u2f());
```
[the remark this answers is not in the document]
conte91: I don't think it's strictly necessary, but I feel that it's better to process by layers (so all the USB processing first, all application processing afterwards), so that events can propagate from the bottom layer in a single pass of the event loop.

src/workflow/confirm.c
```diff
@@ -103,6 +105,23 @@ workflow_t* workflow_confirm(
     }
     data->done = false;
     data->params = *params;
+    /* Make a copy of the parameter strings. */
+    if (data->params.title) {
+        data->title = strdup(data->params.title);
```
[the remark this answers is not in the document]
conte91: Fixed 0a51bc8

src/workflow/workflow.h
```diff
@@ -35,4 +35,29 @@ void workflow_start(void);
...
+    void (*cleanup)(struct _workflow_t*);
+    /* Private data (depends on the workflow type) */
+    void* data;
+    /* Function to get run at every cycle */
+    void (*spin)(struct _workflow_t*);
+} workflow_t;
+
+typedef void (*workflow_method)(workflow_t*);
+
+workflow_t* workflow_allocate(workflow_method init, workflow_method cleanup, workflow_method spin);
```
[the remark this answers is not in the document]
conte91: b6f6559

src/firmware_main_loop.c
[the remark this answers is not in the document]
conte91: 7471885

src/ui/screen_process.h
```diff
@@ -34,4 +34,10 @@ void screen_process(void);
  */
 #define SCREEN_FRAME_RATE 30
 
+/**
+ * Renders the provided component on the display.
+ *
+ * @param[in] component Screen to draw.
+ */
+void ui_draw(component_t* component);
```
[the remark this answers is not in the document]
conte91: f60fdf9

src/workflow/orientation_screen.c
conte91: Reorganized everything 3a6a731

src/ui/components/confirm.c
conte91: 739798d

src/workflow/confirm.c
conte91: Reorganized this so that `workflow_allocate` handles private data storage automatically - fcca89b

conte91: Also 39b48f4 - Allocated data is initialized to zeros.

PR #315, opened 2020-02-01
Confirm config

benma:


src/rust/bitbox02/src/lib.rs
```diff
@@ -121,14 +123,17 @@ pub fn workflow_confirm(
...
+
     unsafe {
-        bitbox02_sys::workflow_confirm(
-            title_cstr.as_ptr() as *const _,
-            body_cstr.as_ptr() as *const _,
-            font,
-            longtouch,
-            accept_only,
-        )
+        bitbox02_sys::workflow_confirm(&mut params)
```
NickeZ: Doesn't it work to have a non-mutable reference to params?

```
bitbox02_sys::workflow_confirm(&params)
```

benma: Totally does, thanks!

Force pushed.

benma:
Rebased.

PR #317, opened 2020-02-02
apps/btc/confirm_locktime_rbf: fix includes

benma:


src/apps/btc/confirm_locktime_rbf.h
```diff
@@ -15,7 +15,8 @@
 #ifndef _APPS_BTC_CONFIRM_LOCKTIME_RBF_H_
 #define _APPS_BTC_CONFIRM_LOCKTIME_RBF_H_
 
-#include <util.h>
```
TheCharlatan: As a side note, is this the only place where util is imported as a replacement for the other two?

benma: idk, I wasn't aware that this was the purpose. 

grepped through it and found a bunch more useless util.h includes

PR #319, opened 2020-02-02
apps/btc/bip143: remove needless "static" keyword

benma:
Must have been a copy/paste error, ctx can be a stack var instead.

### DIFF
--- a/src/apps/btc/btc_bip143.c
+++ b/src/apps/btc/btc_bip143.c
@@ -32,7 +32,7 @@ void btc_bip143_sighash(
     uint8_t* out // 32 bytes result
 )
 {
-    static sha256_context_t ctx = {0};
+    sha256_context_t ctx = {0};
     sha256_reset(&ctx);
     // https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification
     // 1.


conte91:
> However `static` isn't needless. It can improve performance quite a lot to avoid allocating larger structs on the stack.

It also implies that the initialization will be performed only once at startup - so there will be no memset to zero every time the function is invoked.

From a performance perspective, allocating on the stack is just moving a register down, so does this really make a difference?

NickeZ:
The actually use this here: https://benchmarksgame-team.pages.debian.net/benchmarksgame/program/nbody-gcc-8.html

```
    static alignas(__m128d) double
      position_Deltas[3][ROUNDED_INTERACTIONS_COUNT],
      magnitudes[ROUNDED_INTERACTIONS_COUNT];
```

They still loose against rust though:

https://benchmarksgame-team.pages.debian.net/benchmarksgame/performance/nbody.html

benma:
> From a performance perspective, allocating on the stack is just moving a register down, so does this really make a difference?

Imho it's just easier to reason about as a stack var. No need to think about if it can get into a weird state, or if an interrupt can use the same var in a different call, and so on.

Performance is not a concern for me unless it is noticeable.

conte91:
> > From a performance perspective, allocating on the stack is just moving a register down, so does this really make a difference?
> 
> Imho it's just easier to reason about as a stack var. No need to think about if it can get into a weird state, or if an interrupt can use the same var in a different call, and so on.
> 

100% agree on this - The technical side is still interesting though :P

NickeZ:
Yeah I also agree, using static vars makes your function non-reentrable (not threadsafe) so in general it sucks. I just wanted to point out that it is a known optimization to use static vars when you always write to them.

PR #321, opened 2020-02-03
ETH: Add message signing

NickeZ:
I need a conceptACK of this before continuing.

edit: reference for reference: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign
prusnak reaction: https://github.com/ethereum/go-ethereum/issues/14794
You can verify signature here: https://etherscan.io/verifySig


src/apps/eth/eth_sign_msg.c
```diff
@@ -0,0 +1,88 @@
...
+    }
+    if(request->msg.size > 1024) {
+        Abort("Invalid message size");
+    }
+    const char msg_header[] = "\x19" "Ethereum Signed Message:\n";
+    // sizeof(msg_header) includes null terminator
+    // the maximum length of the signed data is 1024, therefore 4 bytes might be needed as length
+    // prefix.
+    char msg[sizeof(msg_header) + sizeof(request->msg.bytes) + 4] = {0};
+    strcat(msg, msg_header);
```
benma: wouldn't it be possible to add the header as an arg in the snprintf below? then the strcat and null terminator care might be dropped.

src/apps/eth/eth_sign_msg.c
```diff
@@ -0,0 +1,88 @@
...
+    size_t offset = sizeof(msg_header) - 1;
+    snprintf(msg + offset, sizeof(msg) - offset, "%d%.*s", request->msg.size, request->msg.size, request->msg.bytes);
+
+    bool all_ascii = true;
+    for(int i = 0; i<request->msg.size; ++i) {
```
benma: int should be size_t (signed<->unsigned mismatch)

src/apps/eth/eth_sign_msg.c
```diff
@@ -0,0 +1,88 @@
...
+            all_ascii = false;
+        }
+    }
+
+    if (all_ascii) {
+        workflow_confirm_scrollable("Sign ETH Message", &msg[offset], NULL, false);
+    } else {
+        char hex[2048] = {0};
+        util_uint8_to_hex(request->msg.bytes, request->msg.size, hex);
+        workflow_confirm_scrollable("Sign ETH Message", hex, NULL, false);
```
benma: with the two branches technically the user can't tell if they are signing binary data or an ascii string. not sure though how to make that clear except, maybe a `\n(hex)` after the title.

src/apps/eth/eth_sign_msg.c
```diff
@@ -0,0 +1,88 @@
...
+    snprintf(msg + offset, sizeof(msg) - offset, "%d%.*s", request->msg.size, request->msg.size, request->msg.bytes);
+
+    bool all_ascii = true;
+    for(int i = 0; i<request->msg.size; ++i) {
+        if (request->msg.bytes[i] < 20 || request->msg.bytes[i] > 127) {
+            all_ascii = false;
+        }
+    }
+
+    if (all_ascii) {
```
benma: the two confirm calls need to return USER_ABORT, otherwise the signature will come out regardless.

src/apps/eth/eth_sign_msg.c
```diff
@@ -0,0 +1,88 @@
...
+        workflow_confirm_scrollable("Sign ETH Message", &msg[offset], NULL, false);
+    } else {
+        char hex[2048] = {0};
+        util_uint8_to_hex(request->msg.bytes, request->msg.size, hex);
+        workflow_confirm_scrollable("Sign ETH Message", hex, NULL, false);
+    }
+
+    // Sign the transaction
```
benma: :warning: the address on the keypath should also be computed and verified, so the user knows which address the sig is good for. if the user had for example two accounts, it might be bad if the wrong account signed something that might have consequences in some smart contracts.

src/apps/eth/eth_sign_msg.c
NickeZ: done

src/apps/eth/eth_sign_msg.c
NickeZ: done

src/apps/eth/eth_sign_msg.c
NickeZ: done

src/apps/eth/eth_sign_msg.c
NickeZ: done

src/apps/eth/eth_sign_msg.c
NickeZ: done

NickeZ:
@benma rereview bitte

PR #323, opened 2020-02-04
bootloader: Modularize the bootloader code.

conte91:
Split the bootloader code into multiple files to make the code cleaner.

This originally came as a prerequisite  for #322, but grew bigger quickly. Overall, having smaller code units in the bootloader brings nice benefits to the readability of the code.


conte91:
No longer needed for #322

PR #324, opened 2020-02-05
gitignore: Add various files.

conte91:


NickeZ:
Some of these are specific to your setup and should go in your personal gitignore file:

https://git-scm.com/docs/gitignore

conte91:
> Some of these are specific to your setup and should go in your personal gitignore file:
> 

Fair point! Fixed. I still think the merge files (*.orig, *_BACKUP_* etc.) should go in.

PR #326, opened 2020-02-05
apps/btc: move _coin_title to btc_common

benma:
Will be reused by multisig account registration.


benma:
`Abort()` basically crashes the device with a message, and is used for situations which are impossible to enter. Returning errors mostly makes sense if the client can react to the error, but this error would never be returned. It's a shortcoming of C, the compiler could realize that all enum values are checked :/

PR #327, opened 2020-02-05
Py: Add possibility to check noise keys

TheCharlatan:
Add two arguments to the bitbox02 initialization that allow to pass in a private key and an existing set of static device public keys.

### DIFF
--- a/py/bitbox02/bitbox02/bitboxbase/bitboxbase.py
+++ b/py/bitbox02/bitbox02/bitboxbase/bitboxbase.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 """BitBoxBase"""
 
-from typing import Optional, Callable
+from typing import Optional
 
 
 import serial
@@ -37,17 +37,13 @@ class BitBoxBase(communication.BitBoxCommonAPI):
     """Class to communicate with a BitBoxBase"""
 
     # pylint: disable=too-many-public-methods
-
     def __init__(
         self,
         device: communication.TransportLayer,
         device_info: DeviceInfo,
-        show_pairing_callback: Callable[[str], bool],
-        attestation_check_callback: Optional[Callable[[bool], None]] = None,
+        noise_config: communication.BitBoxNoiseConfig,
     ):
-        communication.BitBoxCommonAPI.__init__(
-            self, device, device_info, show_pairing_callback, attestation_check_callback
-        )
+        communication.BitBoxCommonAPI.__init__(self, device, device_info, noise_config)
 
     def _bbb_query(
         self, request: bitboxbase.BitBoxBaseRequest, _expected_response: Optional[str] = None
--- a/py/bitbox02/bitbox02/communication/__init__.py
+++ b/py/bitbox02/bitbox02/communication/__init__.py
@@ -16,6 +16,7 @@
 
 from .communication import PhysicalLayer, TransportLayer
 from .bitbox_api_protocol import (
+    BitBoxNoiseConfig,
     BitBoxCommonAPI,
     Bitbox02Exception,
     UserAbortException,
--- a/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
+++ b/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
@@ -20,7 +20,7 @@ import sys
 import base64
 import binascii
 import hashlib
-from typing import Optional, Callable, List, Dict, Tuple, Union
+from typing import Optional, List, Dict, Tuple, Union
 from typing_extensions import TypedDict
 
 import ecdsa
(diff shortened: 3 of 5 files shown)


py/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -240,6 +246,8 @@ def __init__(
             else:
                 self.close()
                 raise Exception("unexpected response")
+        if remote_static_key not in remote_keys:
+            remote_keys.append(remote_static_key)
```
benma: this does not persist it anywhere. maybe it would be easier to copy how it is done in bitbox02-api-go, e.g. pass an instance with some functions for getting/setting keys?

NickeZ: I think the application (not the communication library) should store any state. So I think this PR is correct.

TheCharlatan:
I re-wrote this change. It now splits up the `__init__` function and adds a class that holds all the required callbacks. Everything is still optional, because I don't want to re-write all the clients using this.
The callbacks should mimic what is going in `bitbox02-api-go`, please double check if the types in the callbacks all make sense.

NickeZ:
I don't like that there are callbacks. That will lead to spaghetti code in the application code. I liked the first version of this PR more. The go api is not a good example of a clean API. Callbacks are a necessary evil sometimes, but they add a lot of mental overhead, so should be avoided.

benma:
@NickeZ 

I agree that callbacks can be not so nice.

> I think the application (not the communication library) should store any state.

The application is still doing the actual storing.

The previous way with the list append has some api disadvantages (which may be overkill to discuss):

- List[] as a type that is changed by reference is not nearly as tight as it could be. The client does not see in the type signature if only one element is appended, if elements can be deleted, and so on. types should ideally allow as little unwanted behavior as possible.
- the private key generation in the previous version was done by the lib client, but it definitely should be done by the library
- Unusual (?) in Python to have an `inout` argument. I was confused there for a second.

I suggested to use a more common Python idioms: inheritance or class mixins (default implementation as class methods, inherit to change them), but then also suggested to just follow the current style, as there are already callbacks. Can convert from one style to another later, but imho consistent style is good to have.

tl;dr: I'd still go with callbacks (agreeing that they are not very nice), followed by going with `self.contians_device_static_key()` etc. and inherit the class at call.

py/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -197,26 +215,43 @@ def __init__(
...
         if self._query(OP_I_CAN_HAS_HANDSHAEK) != RESPONSE_SUCCESS:
             self.close()
             raise Exception("Couldn't kick off handshake")
 
         # init noise channel
         noise = NoiseConnection.from_name(b"Noise_XX_25519_ChaChaPoly_SHA256")
         noise.set_as_initiator()
-        dummy_private_key = os.urandom(32)
-        noise.set_keypair_from_private_bytes(Keypair.STATIC, dummy_private_key)
+        if noise_config is not None:
```
benma: if it is none there is no private_key?

NickeZ:
If you want to keep the config class then I suggest that you write it as a python class. Basically the four functions should be four functions and the application can extend it to provide functionality.

> the private key generation in the previous version was done by the lib client, but it definitely should be done by the library

Why? In that case I think the library should provide a `generate_privkey(): bytes` function so that the application can create a privkey before it calls the library.

> `contains_device_static_pubkey_callback: Callable[[bytes], bool]`

This function is reasonable, but I think it would be more pythonic to have a `Generator` and then call `if pubkey in keys`.

NickeZ:
But I also agree that there shouldn't be any `inout` variables. So I guess we might need a function for `register_pubkey`.

NickeZ:
Other things I noticed is that when these functions are made member functions they don't have to be named `_callback` and to indicate "no value" `None` should be used and not `b""`. `get_app_static_privkey_callback() == b""` Basically the return value should be `Optional[bytes]`.

benma:
:+1: on class methods over callbacks in the config class.

> >    contains_device_static_pubkey_callback: Callable[[bytes], bool]

> This function is reasonable, but I think it would be more pythonic to have a Generator and then call if pubkey in keys.

I think you might have misread the signature. What's passed is ine pubkey `bytes`, result is `bool`. `[bytes]` looks misleading but I think it's just a function with one argument, `bytes`. The whole function already returns if if the pubkey is present.

NickeZ:
> +1 on class methods over callbacks in the config class.
> 
> > > contains_device_static_pubkey_callback: Callable[[bytes], bool]
> 
> > This function is reasonable, but I think it would be more pythonic to have a Generator and then call if pubkey in keys.
> 
> I think you might have misread the signature. What's passed is ine pubkey `bytes`, result is `bool`. `[bytes]` looks misleading but I think it's just a function with one argument, `bytes`. The whole function already returns if if the pubkey is present.

I don't think I misunderstood, and it is fine to have a "contains" method. But what I meant was that the config object has a method `pubkeys(): Generator[bytes]` and that the library can call `if key in config.pubkeys()`. But obviously it is also OK to have every application instead of the API layer implement that logic.

py/bitbox02/communication/bitbox_api_protocol.py
TheCharlatan: Whoops, the private_key is now always generated, but not used if the config provides one.

PR #328, opened 2020-02-05
unit-test: Remove mock_ugui.c.

conte91:
UGUI mocking in the unit test framework seems to be unused
in the tests. Since test_ugui was introduced, the actual UGUI
module is being compiled in libbitbox. This causes (unpredictable)
double definition errors of the UG_* symbols.


conte91:
@NickeZ it looks like this issue came out after #298 was merged. Is there any reason for keeping `mock_ugui.c` in the test code?

test/unit-test/CMakeLists.txt
```diff
@@ -55,7 +55,6 @@ add_library(bitbox
   framework/mock_memory.c
   framework/mock_qtouch.c
   framework/mock_gestures.c
-  framework/mock_ugui.c
```
benma: also delete the actual file!

PR #329, opened 2020-02-05
memory: add functions to get/set multisig account registrations

benma:


src/memory/memory.c
```diff
@@ -98,6 +98,23 @@ typedef union {
     uint8_t bytes[CHUNK_SIZE];
 } chunk_1_t;
 
+typedef struct __attribute__((__packed__)) {
+    // verison fixed at 0xFF for now - can be repurposed to turn this struct into an union to
```
TheCharlatan: s/verison/version/

benma: :+1:

test/unit-test/test_memory_functional.c
```diff
@@ -0,0 +1,109 @@
...
+    assert_string_equal(name, names[1]);
+    // rename
+    const char* name0_renamed = "name 1 renamed";
+    assert_int_equal(
+        MEMORY_MULTISIG_SET_BY_HASH_RESULT_OK,
+        memory_multisig_set_by_hash(hashes[0], name0_renamed));
+    assert_true(memory_multisig_get_by_hash(hashes[0], name));
+    assert_string_equal(name, name0_renamed);
+
+    // rename to a name which already exists fails (duplicate name).
```
NickeZ: Could you also check that the old data still is there after a failed write?

test/unit-test/test_memory_functional.c
```diff
@@ -0,0 +1,109 @@
...
+    assert_int_equal(
+        MEMORY_MULTISIG_SET_BY_HASH_RESULT_ERR_DUPLICATE_NAME,
+        memory_multisig_set_by_hash(hashes[0], names[1]));
+    assert_int_equal(
+        MEMORY_MULTISIG_SET_BY_HASH_RESULT_ERR_DUPLICATE_NAME,
+        memory_multisig_set_by_hash(hashes[1], name0_renamed));
+
+    // Only 5 slots available.
```
NickeZ: I think this test should be in its own function. There should also be some "setup()" function that resets the whole memory to `0xff`.

src/memory/memory.h
```diff
@@ -207,4 +213,37 @@ USE_RESULT bool memory_check_noise_remote_static_pubkey(const uint8_t* pubkey);
...
+typedef enum {
+    // success
+    MEMORY_MULTISIG_SET_BY_HASH_RESULT_OK,
+    // there was no more free slot available.
+    MEMORY_MULTISIG_SET_BY_HASH_RESULT_ERR_FULL,
+    // a different multisig with the same name already exists.
+    MEMORY_MULTISIG_SET_BY_HASH_RESULT_ERR_DUPLICATE_NAME,
+    // memory write error
+    MEMORY_MULTISIG_SET_BY_HASH_RESULT_ERR_UNKNOWN,
+} memory_multisig_set_by_hash_result_t;
```
NickeZ: this type is very long. Couldn't you reuse the error codes within at least multisig? Why are there error codes unique to `SET_BY_HASH` ?

NickeZ: I believe the goal is to have a single error type in the whole project, so at least try to have a single one in each module. So I think `memory_result_t` should be enough.

src/memory/memory.c
```diff
@@ -710,3 +727,77 @@ bool memory_add_noise_remote_static_pubkey(const uint8_t* pubkey)
...
     return _write_chunk(CHUNK_1, chunk.bytes);
 }
+
+memory_multisig_set_by_hash_result_t memory_multisig_set_by_hash(
+    const uint8_t* hash,
+    const char* name)
+{
+    chunk_2_t chunk = {0};
+    CLEANUP_CHUNK(chunk);
+    _read_chunk(CHUNK_2, chunk_bytes);
```
NickeZ: what is `chunk_bytes`?

src/memory/memory.h
```diff
@@ -207,4 +213,37 @@ USE_RESULT bool memory_check_noise_remote_static_pubkey(const uint8_t* pubkey);
...
+ * MEMORY_MULTISIG_NUM_ENTRIES different configs can be stored.
+ * If a name is already stored with this hash, the old name will be overwritten.
+ * It's the callers responsibility to validate the name.
+ * @param[in] hash hash identifying the multisig config.
+ * @param[in] human readable name. Must be at most MEMORY_MULTISIG_NAME_MAX_LEN bytes,
+ * including the null terminator (otherwise the name will be truncated).
+ * @return see memory_multisig_set_by_hash_result_t.
+ */
+USE_RESULT memory_multisig_set_by_hash_result_t
+memory_multisig_set_by_hash(const uint8_t* hash, const char* name);
```
NickeZ: I suspect that you can clear an entry by calling it with hash=`0xff...` and name="". Could you either document that or add a function for removing entries?

src/memory/memory.c
```diff
@@ -710,3 +727,77 @@ bool memory_add_noise_remote_static_pubkey(const uint8_t* pubkey)
...
+        const multisig_configuration_t* multisig = &chunk.fields.multisig_configs[i];
+        if (STREQ(multisig->name, name)) {
+            if (!MEMEQ(multisig->hash, hash, sizeof(multisig->hash))) {
+                return MEMORY_MULTISIG_SET_BY_HASH_RESULT_ERR_DUPLICATE_NAME;
+            }
+            // config already exists (equal), early abort, skipping another write.
+            return MEMORY_MULTISIG_SET_BY_HASH_RESULT_OK;
+        }
+    }
+
```
NickeZ: Could you write a comment that this loop is doing two things. Both find the *first* empty index `empty_index` and find the index of the hash if the hash is present and select that for storing into.

src/memory/memory.c
```diff
@@ -710,3 +727,77 @@ bool memory_add_noise_remote_static_pubkey(const uint8_t* pubkey)
...
+    }
+    if (!found && !empty_found) {
+        return MEMORY_MULTISIG_SET_BY_HASH_RESULT_ERR_FULL;
+    }
+    if (!found) {
+        write_index = empty_index;
+    }
+    multisig_configuration_t* multisig = &chunk.fields.multisig_configs[write_index];
+    memcpy(multisig->hash, hash, sizeof(multisig->hash));
+    memset(multisig->name, 0, sizeof(multisig->name));
```
NickeZ: It's nicer to use '\0' when emptying a string.

test/unit-test/test_memory_functional.c
benma: Done.

test/unit-test/test_memory_functional.c
benma: Done

src/memory/memory.h
benma: Good point, done!

src/memory/memory.c
benma: Same as chunk.bytes, fixed. Will fix it in the other functions in a different PR.

src/memory/memory.h
benma: Good catch. I am now explicitly disallowing 0xff... and empty name, and will add a deletion function in the next PR.

src/memory/memory.c
benma: Done

src/memory/memory.c
benma: Done

PR #330, opened 2020-02-05
CI: Cache external dependencies

NickeZ:
this might speed up CI?


benma:
Cached between when?

Shouldn't there at least be a cache per platform (arm, x86)?

NickeZ:
There is, all arm code is in `build` and all x86 code is in `build-build`.

PR #335, opened 2020-02-07
Multisig

benma:
I will tighten up the the unit tests after all review is finished.

### DIFF
--- a/src/apps/btc/btc_sign.c
+++ b/src/apps/btc/btc_sign.c
@@ -417,7 +417,9 @@ app_btc_sign_error_t app_btc_sign_output(
     // get pubkeyhash or scripthash. If request->ours=true, we compute the hash
     // from the keystore, otherwise it is provided in request->hash.
 
-    BTCSignOutputRequest_hash_t hash = {0};
+    uint8_t hash_bytes[sizeof(request->hash.bytes)] = {0};
+    size_t hash_size;
+
     BTCOutputType output_type;
     if (request->ours) {
         if (!_is_valid_keypath(
@@ -444,19 +446,16 @@ app_btc_sign_error_t app_btc_sign_output(
                 return _error(APP_BTC_SIGN_ERR_UNKNOWN);
             }
 
-            size_t out_size = 0;
             // construct pkScript
             if (!btc_common_outputhash_from_pubkeyhash(
                     _init_request.script_config.config.simple_type,
                     pubkey_hash160,
-                    hash.bytes,
-                    &out_size)) {
+                    hash_bytes,
+                    &hash_size)) {
                 return _error(APP_BTC_SIGN_ERR_UNKNOWN);
             }
             output_type =
                 btc_common_determine_output_type(_init_request.script_config.config.simple_type);
-
-            hash.size = (pb_size_t)out_size;
             break;
         }
         case BTCScriptConfig_multisig_tag:
@@ -464,10 +463,10 @@ app_btc_sign_error_t app_btc_sign_output(
                     &_init_request.script_config.config.multisig,
                     request->keypath[request->keypath_count - 2],
                     request->keypath[request->keypath_count - 1],
-                    hash.bytes)) {
+                    hash_bytes)) {
                 return _error(APP_BTC_SIGN_ERR_UNKNOWN);
             }
-            hash.size = 32;
+            hash_size = 32;
             output_type = BTCOutputType_P2WSH;
             break;
         default:
@@ -475,7 +474,8 @@ app_btc_sign_error_t app_btc_sign_output(
         }
 
     } else {
-        hash = request->hash;
+        hash_size = request->hash.size;
+        memcpy(hash_bytes, request->hash.bytes, hash_size);
         output_type = request->type;
     }
     if (request->value == 0) {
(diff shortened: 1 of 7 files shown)


test/unit-test/test_btc_sign.c
```diff
@@ -731,25 +762,45 @@ static void _sign(const _modification_t* mod)
...
-            "\x95\x09\x43\x09\xa2\xd2\x77\xd3\xa6\x8d\xde\xd3\x3d\x50\xa7\x47\xf2\xee\xfb\x3f\x54"
-            "\x8a\x92\x45\x15\xdb\x62\xbe\x06\xa1\xae\xa4\x56\x92\x91\xe5\x2e\x6f\xea\x95\xf8\xb6"
-            "\x75\x23\xb1\x9b\x35\x9a\x84\x85\xd8\xaa\x3c\xa0\x2d\xb3\x74\x70\x01\x0b\x19\x9b\x0c"
-            "\xe3";
-        assert_memory_equal(next.signature, expected_signature, sizeof(next.signature));
+    if (mod->check_sigs) {
+        switch (mod->script_type) {
+        case BTCScriptConfig_SimpleType_P2WPKH: {
+            const uint8_t expected_signature[64] =
+                "\x95\x09\x43\x09\xa2\xd2\x77\xd3\xa6\x8d\xde\xd3\x3d\x50\xa7\x47\xf2\xee\xfb\x3f"
```
NickeZ: The autoformatter kind of failed here? I guess if you concatenate everything into one string and then do clang-format it will look better.

src/apps/btc/btc_sign.c
```diff
@@ -263,6 +307,22 @@ static app_btc_sign_error_t _sign_input_pass2(
                 return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
             }
             break;
+        case BTCScriptConfig_multisig_tag: {
+            uint8_t sighash_script_tmp[MAX_PK_SCRIPT_SIZE - MAX_VARINT_SIZE] = {0};
```
NickeZ: Can you explain why this is MAX_PK_SCRIPT_SIZE *minus* MAX_VARINT_SIZE?

src/apps/btc/btc_sign.c
```diff
@@ -357,7 +417,9 @@ app_btc_sign_error_t app_btc_sign_output(
     // get pubkeyhash or scripthash. If request->ours=true, we compute the hash
     // from the keystore, otherwise it is provided in request->hash.
 
-    BTCSignOutputRequest_hash_t hash = {0};
+    uint8_t hash_bytes[32] = {0};
```
NickeZ: unnamed 32

src/apps/btc/btc_sign.c
```diff
@@ -384,27 +446,36 @@ app_btc_sign_error_t app_btc_sign_output(
...
+            hash_size = 32;
+            output_type = BTCOutputType_P2WSH;
+            break;
         default:
             return _error(APP_BTC_SIGN_ERR_INVALID_INPUT);
         }
 
     } else {
-        hash = request->hash;
+        memcpy(hash_bytes, request->hash.bytes, sizeof(hash_bytes));
```
NickeZ: `sizeof(has_bytes)` should be `min(sizeof(hash_bytes), request->hash.size)` Or there should be an abort in case they aren't the same.

src/apps/btc/btc_sign.c
benma: I can't, because it's wrong :P nice catch! before it was `sizeof(sighash_script) - MAX_VARINT_SIZE`, but that's just `MAX_PK_SCRIPT_SIZE`.

Fixed.

src/apps/btc/btc_sign.c
benma: Done

src/apps/btc/btc_sign.c
benma: copying only request->hash.size now, as that is either 20 or 32.

for the record, the previous way (copying 32 bytes) works just as well, as both buffers have the same capacity (32 bytes), so copying all of it does not technically hurt.

test/unit-test/test_btc_sign.c
benma: Fixed.

src/apps/btc/btc_sign.c
NickeZ: No, if you don't use `min(..)` or verify that they are the same you will either have an out-of-bounds write or an out-of-bounds read. Both are equally bad.

NickeZ: > copying only request->hash.size now, since min(sizeof(hash_bytes), request->hash.size) is request->hash.size always.

Adding the below code will be optimized away in case they are the same. It is just a sanity check for ourselves.

```
if(request->hash != sizeof(hash_bytes) {
    Abort(...)'
}
```

benma: since the last push `hash_bytes` is initialized with `sizeof(request->hash.bytes)` size, so it is the same.

NickeZ: Ah OK, I'll rereview then

test/unit-test/test_btc_sign.c
NickeZ: It doesn't look fixed, but it is only test code so no need to fix.

benma: Strange, maybe I fixed another instance of it. Anyway, fixed in https://github.com/digitalbitbox/bitbox02-firmware/pull/348

PR #336, opened 2020-02-08
eth: Add support for data field

NickeZ:
Transactions to smart contracts typically use the data field. Add
general support for now.

I'm hoping I can reuse the confirm screen additions in the other ETH PR.

I tested manually with data of different sizes, especially around 64 bytes.


src/apps/eth/eth_verify.c
```diff
@@ -136,13 +137,19 @@ app_eth_sign_error_t app_eth_verify_standard_transaction(const ETHSignRequest* r
     if (params == NULL) {
         return APP_ETH_SIGN_ERR_INVALID_INPUT;
     }
-    if (request->value.size == 0) {
```
benma: unsure, but possibly value==0 and data.size==0 is invalid (must have at least one of both?)

src/apps/eth/eth_verify.c
```diff
@@ -136,13 +137,19 @@ app_eth_sign_error_t app_eth_verify_standard_transaction(const ETHSignRequest* r
...
-        // Must transfer non-zero value.
-        return APP_ETH_SIGN_ERR_INVALID_INPUT;
-    }
     if (request->data.size != 0) {
-        // Standard tx has no data.
-        return APP_ETH_SIGN_ERR_INVALID_INPUT;
+        char hex[sizeof(request->data.bytes) * 2 + 1] = {0};
```
benma: can also do: `hex[BB_HEX_SIZE(request->data.bytes)]`

src/apps/eth/eth_verify.c
NickeZ: OK, added that case

src/apps/eth/eth_verify.c
NickeZ: IMO doesn't really increase legibility, can be done later.

src/apps/eth/eth_verify.c
```diff
@@ -134,15 +137,26 @@ app_eth_sign_error_t app_eth_verify_standard_transaction(const ETHSignRequest* r
...
+        util_uint8_to_hex(request->data.bytes, request->data.size, hex);
+        confirm_params_t cparams = {
+            .title = "ETH TX data\n(hex)",
+            .body = hex,
+            .scrollable = true,
+            .shorten_body = true,
+            .size = request->data.size,
```
benma: rename .size to .display_size, at first i thought it was used to limit the length or the body or simlar.

NickeZ: done

PR #346, opened 2020-02-12
btc,eth: validate names / ascii

benma:
To be able to print names on the screen for confirmation, they need to
be all ascii and not start/end with spaces.

ascii util function added also for bytes as that can be reused in
eth's sign message function.

### DIFF
--- a/src/util/util_string.c
+++ b/src/util/util_string.c
@@ -1,0 +1,61 @@
+// Copyright 2019 Shift Cryptosecurity AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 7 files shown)


src/util/util_string.c
```diff
@@ -0,0 +1,50 @@
...
+    return true;
+}
+
+bool util_string_all_ascii(const char* str)
+{
+    return util_string_all_ascii_bytes((const uint8_t*)str, strlen(str));
+}
+
+bool util_string_validate_name(const char* name, size_t max_len)
+{
```
NickeZ: What do you think about an `Abort(...)` in case `name==NULL`?

benma: :+1: done

benma:
> Great job, could we use this to validate the set device name as well?

I'd move that to the next release, as that needs to happen together with validation in the app.

PR #350, opened 2020-02-13
wip: test another font size

NickeZ:
Some things look bad due to slightly higher font. But the font itself looks great.


benma:
We increased the font size in some PRs derived from this PR a long time ago already, just forgot to close this.

PR #354, opened 2020-02-17
py: Cache keys by default

NickeZ:
I added optional keys caching in our python tools.

@TheCharlatan I split out the two old callbacks again from the noise config object.

### DIFF
--- a/py/bitbox02/bitbox02/util.py
+++ b/py/bitbox02/bitbox02/util.py
@@ -13,9 +13,16 @@
 # limitations under the License.
 """Useful functions"""
 
+from typing import Optional, List, Tuple
+from pathlib import Path
+import json
+import os
+import binascii
+
 import base58
 
 from .bitbox02 import common
+from .communication import bitbox_api_protocol
 
 
 def parse_xpub(xpub: str) -> common.XPub:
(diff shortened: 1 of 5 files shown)


py/bitbox02/bitbox02/util.py
```diff
@@ -39,3 +46,90 @@ def parse_xpub(xpub: str) -> common.XPub:
...
+    def _find_cache_file() -> Path:
+        cachedir_env = os.environ.get("XDG_CACHE_HOME", "")
+        if cachedir_env == "":
+            homedir = os.environ.get("HOME", "")
+            if homedir == "":
+                raise RuntimeError("Can't find cache dir")
+            cachedir = Path(homedir) / ".cache"
+        else:
+            cachedir = Path(cachedir_env)
+        return cachedir / "shift" / "bitbox02.dat"
```
benma: the whole thing should also add an appname, otherwise there is a risk that downstream projects will all use this cache util class and all use the same noise key, but there should be one per application.

py/bitbox02/bitbox02/util.py
```diff
@@ -39,3 +46,90 @@ def parse_xpub(xpub: str) -> common.XPub:
...
+
+    def contains_device_static_pubkey(self, pubkey: bytes) -> bool:
+        data = self._read_cache()
+        if pubkey in data.device_static_pubkeys:
+            return True
+        return False
+
+    def add_device_static_pubkey(self, pubkey: bytes) -> None:
+        data = self._read_cache()
+        data.device_static_pubkeys.append(pubkey)
```
benma: this also needs to check if it's already in there (i guess the four methods should have been documented ;) )

py/bitbox02/bitbox02/util.py
NickeZ: I think the opposite, I want all the apps to use the same key. In my case `send_message` and `load_firmware`. Or maybe I want all of them to use the same list of device keys...

py/bitbox02/bitbox02/util.py
NickeZ: Ack

TheCharlatan:
I don't get why you are bringing back the callbacks again. They are doing the exact same thing as before.

py/bitbox02/bitbox02/util.py
NickeZ: changed my mind. Fixed.

NickeZ:
> I don't get why you are bringing back the callbacks again. They are doing the exact same thing as before.

Because I think the callbacks are unrelated to the key management. Basically most apps can reuse my "cache keys in home directory" helper class. Whereas they probably want other implementations for attestation and pairing callbacks.

Honestly though If I designed the API from scratch, there would be an "unlock" call, and then a "pair" call and no callbacks.

TheCharlatan:
You can still keep the key mamangement class and overload it yet again in the send_message and flash_firmware script.

NickeZ:
Good point, I can revert to that.

NickeZ:
@TheCharlatan I've switched to using subclassing, now I see why I split out the callbacks. The code gets a bit hairy since application logic now is mixed with key logic. WDYT?

py/bitbox02/bitbox02/util.py
```diff
@@ -39,3 +46,97 @@ def parse_xpub(xpub: str) -> common.XPub:
...
+            privkey = binascii.hexlify(self.app_static_privkey).decode("utf-8")
+        return json.dumps({"device_static_pubkeys": pubkeys, "app_static_privkey": privkey})
+
+    @staticmethod
+    def deserialize(raw: str) -> Tuple[Optional[bytes], List[bytes]]:
+        """Deserialize content from disk to struct"""
+        try:
+            data = json.loads(raw)
+            privkey = None
+            if data["app_static_privkey"] is not None:
```
TheCharlatan: Should be `if data.get("app_static_privkey") is not None`

TheCharlatan:
i actually like using a single configuration class - it gives clear instructions on what I need to implement. The app can still decide to overload the user cache instead and then subclass the app logic if it gets too complex. It's usable, so I would not spend too much time on it.

py/bitbox02/bitbox02/util.py
NickeZ: Actually no, the key will always be there, but the value is None in case it hasn't been set yet

py/bitbox02/bitbox02/util.py
```diff
@@ -39,3 +46,97 @@ def parse_xpub(xpub: str) -> common.XPub:
...
+            fileh.write(data.serialize())
+
+    def contains_device_static_pubkey(self, pubkey: bytes) -> bool:
+        data = self._read_cache()
+        if pubkey in data.device_static_pubkeys:
+            return True
+        return False
+
+    def add_device_static_pubkey(self, pubkey: bytes) -> None:
+        if not self.contains_device_static_pubkey(pubkey):
```
TheCharlatan: What's with the empty case?

NickeZ: What do you mean? A `UserCache` will always have the two members `device_static_pubkeys` and `app_static_privkey`.

py/bitbox02/bitbox02/util.py
TheCharlatan: I don't see where you are guaranteed to set the key before, what am I missing?

NickeZ: https://github.com/digitalbitbox/bitbox02-firmware/pull/354/files#diff-5c7d5e4282d5df35e114d9b5af36ed86R70

It is always serialized with a `privkey`. Either it is `None` or it is `bytes(...)`

TheCharlatan: Mmh, but the class initializer calls `deserialize` first, before serialize has a chance to be called.

py/bitbox02/bitbox02/util.py
TheCharlatan: Yes, all good :+1:

py/bitbox02/bitbox02/util.py
NickeZ: Right, but if this: `json.loads(raw)` works, then there was a valid json object in the string. And if there was a valid json object then it has that key. I guess my point is that if anyone else puts a valid json file with the wrong content in that location it would throw an exception here yes. I'll update it to catch `KeyError` which would also resolve that issue.

PR #358, opened 2020-02-18
assertion: BitBox02 firmware (btc & multi)

Stadicus:
Independent verification of reproduciable build for latest BitBox02 firmware v6.0.0


benma:
Thanks a lot!

PR #360, opened 2020-02-19
Python scripts: Cleanup Noise API versioning.

conte91:
Change the various if/else checks on the noise version numbers
to use polymorphism instead. Move all the noise-level operations
to a separate set of classes to sharpen the separation between noise
and application layers.


conte91:
Closing/reopening as CI has gone crazy.

PR #363, opened 2020-02-20
Noise encapsulation

conte91:
Add encapsulation to Noise handshake requests and OP_NOISE_MSG responses. This fixes a long standing bug that makes handshakes fail occasionally, and allows errors to be reported in a coherent way on the HWW stack.

### DIFF
--- a/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
+++ b/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
@@ -235,6 +235,11 @@ class BitBoxProtocol(ABC):
         ...
 
     @abstractmethod
+    def _decode_noise_response(self, encrypted_msg: bytes) -> Tuple[bytes, bytes]:
+        """ De-encapsulate an OP_NOISE_MSG response. """
+        ...
+
+    @abstractmethod
     def _handshake_query(self, req: bytes) -> Tuple[bytes, bytes]:
         """
         Executes a OP_HER_COMEZ_TEH_HANDSHAEK query with the given
@@ -251,6 +256,10 @@ class BitBoxProtocol(ABC):
         encrypted_msg = self._encode_noise_request(encrypted_msg)
 
         response = self._raw_query(encrypted_msg)
+        response_status, response = self._decode_noise_response(response)
+        if response_status != RESPONSE_SUCCESS:
+            raise Exception("Noise communication failed.")
+
         result = self._noise.decrypt(response)
         assert isinstance(result, bytes)
         return result
@@ -337,6 +346,16 @@ class BitBoxProtocolV1(BitBoxProtocol):
     def _encode_noise_request(self, encrypted_msg: bytes) -> bytes:
         return encrypted_msg
 
+    def _decode_noise_response(self, encrypted_msg: bytes) -> Tuple[bytes, bytes]:
+        """
+        Until V7 of the protocol, we don't encapsulate OP_NOISE_MSG responses.
+        Let's assume that if a response is empty, that means it
+        contains an error.
+        """
+        if len(encrypted_msg) == 0:
+            return RESPONSE_FAILURE, b""
+        return RESPONSE_SUCCESS, encrypted_msg
+
     def _handshake_query(self, req: bytes) -> Tuple[bytes, bytes]:
         """
         V1-6 of the BB noise protocol doesn't encapsulate handshake requests, and don't
@@ -379,6 +398,9 @@ class BitBoxProtocolV7(BitBoxProtocolV4):
 
     def _handshake_query(self, req: bytes) -> Tuple[bytes, bytes]:
         return self.query(OP_HER_COMEZ_TEH_HANDSHAEK, req)
+
+    def _decode_noise_response(self, encrypted_msg: bytes) -> Tuple[bytes, bytes]:
+        return encrypted_msg[:1], encrypted_msg[1:]
 
 
 class BitBoxCommonAPI:
--- a/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
+++ b/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
@@ -113,6 +113,7 @@ OP_ATTESTATION = b"a"
 OP_UNLOCK = b"u"
 OP_INFO = b"i"
 OP_I_CAN_HAS_HANDSHAEK = b"h"
+OP_HER_COMEZ_TEH_HANDSHAEK = b"H"
 OP_I_CAN_HAS_PAIRIN_VERIFICASHUN = b"v"
 OP_NOISE_MSG = b"n"
 
@@ -233,6 +234,15 @@ class BitBoxProtocol(ABC):
         """ Encapsulates an OP_NOISE_MSG message. """
         ...
 
+    @abstractmethod
+    def _handshake_query(self, req: bytes) -> Tuple[bytes, bytes]:
+        """
+        Executes a OP_HER_COMEZ_TEH_HANDSHAEK query with the given
+        request data.
+        Returns a pair (response status, response data).
+        """
+        ...
+
     def encrypted_query(self, msg: bytes) -> bytes:
         """
         Sends msg bytes and reads response bytes over an encrypted channel.
@@ -261,13 +271,20 @@ class BitBoxProtocol(ABC):
         noise.set_keypair_from_private_bytes(Keypair.STATIC, private_key)
         noise.set_prologue(b"Noise_XX_25519_ChaChaPoly_SHA256")
         noise.start_handshake()
-        noise.read_message(self._raw_query(noise.write_message()))
+        start_handshake_status, start_handshake_reply = self._handshake_query(noise.write_message())
+        if start_handshake_status != RESPONSE_SUCCESS:
+            self.close()
+            raise Exception("Handshake process request failed.")
+        noise.read_message(start_handshake_reply)
         remote_static_key = noise.noise_protocol.handshake_state.rs.public_bytes
         assert not noise.handshake_finished
         send_msg = noise.write_message()
         assert noise.handshake_finished
         pairing_code = base64.b32encode(noise.get_handshake_hash()).decode("ascii")
-        response = self._raw_query(send_msg)
+        handshake_finish_status, response = self._handshake_query(send_msg)
+        if handshake_finish_status != RESPONSE_SUCCESS:
+            self.close()
+            raise Exception("Handshake conclusion failed.")
 
         # Check if we recognize the device's public key
         pairing_verification_required_by_host = True
@@ -320,6 +337,14 @@ class BitBoxProtocolV1(BitBoxProtocol):
     def _encode_noise_request(self, encrypted_msg: bytes) -> bytes:
         return encrypted_msg
 
+    def _handshake_query(self, req: bytes) -> Tuple[bytes, bytes]:
+        """
+        V1-6 of the BB noise protocol doesn't encapsulate handshake requests, and don't
+        send back a status code in the response.
+        """
+        noise_result = self._raw_query(req)
+        return RESPONSE_SUCCESS, noise_result
+
 
 class BitBoxProtocolV2(BitBoxProtocolV1):
     """ BitBox Protocol from firmware V2.0.0 onwards. """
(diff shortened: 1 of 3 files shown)


py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -270,21 +301,133 @@ def _create_noise_channel(self, noise_config: BitBoxNoiseConfig) -> NoiseConnect
...
+            self._bitbox_protocol = BitBoxProtocolV4(transport)
+        elif self.version >= semver.VersionInfo(3, 0, 0):
+            self._bitbox_protocol = BitBoxProtocolV3(transport)
+        else:
+            self._bitbox_protocol = BitBoxProtocolV1(transport)
+
+        if self.version >= semver.VersionInfo(2, 0, 0):
+            noise_config.attestation_check(self._perform_attestation())
+            unlock_result = self._bitbox_protocol.unlock_query()
+            if unlock_result != RESPONSE_SUCCESS:
```
benma: This is only for version 3.0.0+ according to the previous code. in 2.0.0, the response was empty (not even a status code, see https://github.com/digitalbitbox/bitbox02-firmware/commit/40ea9cb4811fe041413da52234822f657c7c28e8#diff-fb3b76656f7207bf1004582fd62acadaR85

Before this PR this was handled explicitly like this:

```Python
            unlock_result = self._query(OP_UNLOCK)
            if self.version < semver.VersionInfo(3, 0, 0):
                assert unlock_result == b""
            else:
                # since 3.0.0, unlock can fail if cancelled
                if unlock_result == RESPONSE_FAILURE:
                    self.close()
                    raise Exception("Unlock process aborted")
```

The pattern of always returning a status code does not work for all previous versions.

Unless I am reading the code wrong, unlock_result for v2.0.0 will be `b""`, so it will always fail for v2.0.0?

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -270,21 +301,133 @@ def _create_noise_channel(self, noise_config: BitBoxNoiseConfig) -> NoiseConnect
...
+        return RESPONSE_SUCCESS, noise_result
+
+class BitBoxProtocolV2(BitBoxProtocolV1):
+    """ BitBox Protocol from firmware V2.0.0 onwards. """
+
+    def unlock_query(self) -> bytes:
+        unlock_result, unlock_data = self.query(OP_UNLOCK, b"")
+        if len(unlock_data) != 0:
+            raise ValueError(f"OP_UNLOCK (V3) replied with wrong length.")
+        return unlock_result
```
benma: in v2 there is no unlock_result either. I think the signature of `unlock_query` should probably change, as not all versions have a result (v2.0.0 has no result). It could be more readable and correct if each versions `unlock_query` checked for the expected result and raised an exception otherwise, instead of passing it upwards.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -270,21 +301,133 @@ def _create_noise_channel(self, noise_config: BitBoxNoiseConfig) -> NoiseConnect
...
+
+
+class BitBoxProtocolV7(BitBoxProtocolV4):
+    """ Noise Protocol from firmware V7.0.0 onwards. """
+
+    def _handshake_query(self, req: bytes) -> Tuple[bytes, bytes]:
+        return self.query(OP_HER_COMEZ_TEH_HANDSHAEK, req)
+
+    def _decode_noise_response(self, encrypted_msg: bytes) -> Tuple[bytes, bytes]:
+        print("Decoding: " + encrypted_msg.hex())
```
benma: remove

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -270,21 +301,133 @@ def _create_noise_channel(self, noise_config: BitBoxNoiseConfig) -> NoiseConnect
...
+        )
+
+        self._bitbox_protocol: BitBoxProtocol
+        if self.version >= semver.VersionInfo(7, 0, 0):
+            self._bitbox_protocol = BitBoxProtocolV7(transport)
+        elif self.version >= semver.VersionInfo(4, 0, 0):
+            self._bitbox_protocol = BitBoxProtocolV4(transport)
+        elif self.version >= semver.VersionInfo(3, 0, 0):
+            self._bitbox_protocol = BitBoxProtocolV3(transport)
+        else:
```
benma: BitBoxProtocolV2 missing (and if it is there, it runs into the `Unlock process aborted` exception)

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
benma: Otherwise I tested all upgrades from v3.0.0 onwards and those worked. 2.0.0 is granted a low prio as it was a pre-release version, but it's probably good to keep it right anyway.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -270,21 +301,133 @@ def _create_noise_channel(self, noise_config: BitBoxNoiseConfig) -> NoiseConnect
...
+    def _encode_noise_request(self, encrypted_msg: bytes) -> bytes:
+        return encrypted_msg
+
+    def _decode_noise_response(self, encrypted_msg: bytes) -> Tuple[bytes, bytes]:
+        """
+        Until V7 of the protocol, we don't encapsulate OP_NOISE_MSG responses.
+        Let's assume that if a response is only one byte long, that means it
+        contains an error.
+        """
+        if len(encrypted_msg) == 1:
```
benma: before the introduction of `OP_NOISE_MSG`, a noise response could never be 1 byte, could it? it would be zero in case noise decryption failed:

```
if (noise_cipherstate_decrypt(_recv_cipher, &noise_buffer) != NOISE_ERROR_NONE) {
        return false;
    }
```

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -232,13 +256,20 @@ def _create_noise_channel(self, noise_config: BitBoxNoiseConfig) -> NoiseConnect
         noise.set_keypair_from_private_bytes(Keypair.STATIC, private_key)
         noise.set_prologue(b"Noise_XX_25519_ChaChaPoly_SHA256")
         noise.start_handshake()
-        noise.read_message(self._query(noise.write_message()))
+        start_handshake_status, start_handshake_reply = self._handshake_query(noise.write_message())
+        if start_handshake_status != RESPONSE_SUCCESS:
```
benma: similar to the other comment, this code could also be moved to `_handshake_query` of `BitBoxProtocolV7`. Then `handshake_query` does not have to return the status code, and avoid the "fake" response in V1-V6 (`return RESPONSE_SUCCESS`).

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -179,46 +180,69 @@ def set_app_static_privkey(self, privkey: bytes) -> None:
...
+    def encrypted_query(self, msg: bytes) -> bytes:
+        """
+        Sends msg bytes and reads response bytes over an encrypted channel.
+        """
+        encrypted_msg = self._noise.encrypt(msg)
+        encrypted_msg = self._encode_noise_request(encrypted_msg)
+
+        response = self._raw_query(encrypted_msg)
+        response_status, response = self._decode_noise_response(response)
+        if response_status != RESPONSE_SUCCESS:
```
benma: same as below, if this is moved into `_decode_noise_response`, the function does not need to return a fake response in V1-V6.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -270,21 +301,133 @@ def _create_noise_channel(self, noise_config: BitBoxNoiseConfig) -> NoiseConnect
...
+        noise_result = self._raw_query(req)
+        return RESPONSE_SUCCESS, noise_result
+
+class BitBoxProtocolV2(BitBoxProtocolV1):
+    """ BitBox Protocol from firmware V2.0.0 onwards. """
+
+    def unlock_query(self) -> bytes:
+        unlock_result, unlock_data = self.query(OP_UNLOCK, b"")
+        if len(unlock_data) != 0:
+            raise ValueError(f"OP_UNLOCK (V3) replied with wrong length.")
```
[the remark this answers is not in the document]
conte91: This belongs to #362!

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
conte91: Fixed in #362

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
conte91: Fixed in #362

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -179,46 +180,69 @@ def set_app_static_privkey(self, privkey: bytes) -> None:
...
-                if unlock_result == RESPONSE_FAILURE:
-                    self.close()
-                    raise Exception("Unlock process aborted")
+    def query(self, cmd: bytes, msg_data: bytes) -> Tuple[bytes, bytes]:
+        """
+        Encapsulates the given OP_* command and message data into a packet,
+        and unpacks the response status code and data.
+        """
+        response = self._raw_query(cmd + msg_data)
+        return response[0:1], response[1:]
```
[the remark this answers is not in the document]
conte91: Fixed in #362

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
conte91: Fixed in #362

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
conte91: But it can return failure (which we detect if we receive back an empty response) - this avoids throwing the same exception from two points.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
conte91: This allows to throw with the correct error message (as `_handshake_query` doesn't know "which" part of the handshake has failed)

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
conte91: Fixed in cae3727

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
conte91: Fixed in cae3727. Checking for empty message instead.

src/usb/noise.c
```diff
@@ -226,30 +258,50 @@ bool bb_noise_process_msg(
     const size_t max_out_len,
     bb_noise_process_msg_callback process_msg)
 {
+    if (!_check_message_length(in_packet)) {
+        out_packet->len = 1;
+        out_packet->data_addr[0] = OP_STATUS_FAILURE;
+        return false;
+    }
+    uint8_t cmd = in_packet->data_addr[0];
```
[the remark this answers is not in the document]
conte91: a9abcad

src/usb/noise.c
```diff
@@ -226,30 +258,50 @@ bool bb_noise_process_msg(
...
+            out_packet->len = 1;
+        }
+        return handshake_success;
     }
-    if (_recv_cipher == NULL) {
+
+    if (!_recv_cipher) {
         // Not paired yet, abort
         out_packet->len = 1;
         out_packet->data_addr[0] = OP_I_CAN_HAS_HANDSHAEK;
```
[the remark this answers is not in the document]
conte91: Separate PR for this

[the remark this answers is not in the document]
conte91: See #364 once this goes in.

PR #368, opened 2020-02-24
py: explicit max upper version

benma:
We support 6.0.0 and 7.0.0, so failing at min version + 1 does not
work.

Also the min check somehow did the max check, must have been a rebase
error or sth.

### DIFF
--- a/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
+++ b/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
@@ -120,8 +120,10 @@ OP_NOISE_MSG = b"n"
 RESPONSE_SUCCESS = b"\x00"
 RESPONSE_FAILURE = b"\x01"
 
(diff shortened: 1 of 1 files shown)


py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -122,6 +122,8 @@ class AttestationPubkeyInfo(TypedDict):
 
 MIN_BITBOX02_MULTI_FIRMWARE_VERSION = semver.VersionInfo(6, 0, 0)
 MIN_BITBOX02_BTCONLY_FIRMWARE_VERSION = semver.VersionInfo(6, 0, 0)
```
conte91: Maybe rename this to `MIN_SUPPORTED_*` for consistency?

benma: Done

PR #369, opened 2020-02-24
eth: make keypath check less restrictive (allow testnet)

Tomasvrba:
When @NickeZ made the change to allow keypath length of 4 for getting the xpub, he also whitelisted only the `m'/44'/60'/0'/0` path.
To allow getting the xpub and address for our currently supported testnets (ropsten and rinkeby), we also need the `m'/44'/1'/0'/0` path


benma:
Discussed offline, imho we we should not change the firmware to allow mixing testnet/mainnet; the client needs to know what it wants. Temporary workaround would be to change the coin based on the keypath to get what you want.

Tomasvrba:
Will make the check in the api instead to provide the correct coin until we can maybe get MEW to supply HWWs with the network

PR #371, opened 2020-02-27
More async workflows

conte91:
Add asynchronous password entry and unlock workflows. Depends on #312.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <hardfault.h>
 #include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -54,6 +55,15 @@ void util_cleanup_32(uint8_t** buf)
 void util_cleanup_64(uint8_t** buf)
 {
     util_zero(*buf, 64);
+}
+
+char* util_strdup(const char* str)
+{
+    char* result = strdup(str);
+    if (!result) {
+        Abort("malloc failed in util_strdup.");
+    }
+    return result;
 }
 
 bool safe_uint64_add(uint64_t* a, uint64_t b)
--- a/src/workflow/get_mnemonic_passphrase.c
+++ b/src/workflow/get_mnemonic_passphrase.c
@@ -1,0 +1,40 @@
+#include "get_mnemonic_passphrase.h"
+
+#include <string.h>
+
+#include <hardfault.h>
+#include <ui/components/trinary_input_string.h>
+#include <ui/fonts/password_11X12.h>
+
+#include "blocking.h"
+#include "confirm.h"
+#include "password_enter.h"
+#include "status.h"
+
+bool get_mnemonic_passphrase(char* passphrase_out)
+{
+    if (passphrase_out == NULL) {
+        Abort("_get_mnemonic_passphrase");
+    }
+    while (true) {
+        password_enter_blocking("Enter\noptional passphrase", true, passphrase_out);
+        if (strlen(passphrase_out) == 0) {
+            // No need to confirm the empty passphrase.
+            break;
+        }
+        const confirm_params_t params = {
+            .title = "",
+            .body = "You will be asked to\nvisually confirm your\npassphrase now.",
+            .accept_only = true,
+        };
+        if (!workflow_confirm_blocking(&params)) {
+            return false;
+        }
+        if (workflow_confirm_scrollable_longtouch_blocking(
+                "Confirm", passphrase_out, &font_password_11X12)) {
+            break;
+        }
+        workflow_status_blocking("Please try again", false);
+    }
+    return true;
+}
--- a/src/workflow/get_mnemonic_passphrase.c
+++ b/src/workflow/get_mnemonic_passphrase.c
@@ -138,19 +138,3 @@ workflow_t* workflow_get_mnemonic_passphrase(void (*callback)(char*, void*), voi
     data->callback_param = callback_param;
     return result;
 }
-
-static void _unlock_cb(char* passphrase, void* param)
-{
-    char* passphrase_out = (char*)param;
-    int n_written = snprintf(passphrase_out, SET_PASSWORD_MAX_PASSWORD_LENGTH, "%s", passphrase);
-    if (n_written < 0 || n_written >= SET_PASSWORD_MAX_PASSWORD_LENGTH) {
-        Abort("Mnemonic passphrase invalid length.");
-    }
-    workflow_blocking_unblock();
-}
-
-void get_mnemonic_passphrase_blocking(char* passphrase_out)
-{
-    workflow_stack_start_workflow(get_mnemonic_passphrase(_unlock_cb, passphrase_out));
-    workflow_blocking_block();
-}
--- a/src/workflow/get_mnemonic_passphrase.c
+++ b/src/workflow/get_mnemonic_passphrase.c
@@ -5,36 +5,152 @@
 #include <hardfault.h>
 #include <ui/components/trinary_input_string.h>
 #include <ui/fonts/password_11X12.h>
+#include <ui/workflow_stack.h>
 
 #include "blocking.h"
 #include "confirm.h"
 #include "password_enter.h"
 #include "status.h"
 
-bool get_mnemonic_passphrase(char* passphrase_out)
+typedef enum {
+    MNEMONIC_PASSPHRASE_PW_ENTER,
+    MNEMONIC_PASSPHRASE_CONFIRM_MSG,
+    MNEMONIC_PASSPHRASE_CONFIRM_PASSPHRASE,
+    MNEMONIC_PASSPHRASE_TRY_AGAIN,
+    MNEMONIC_PASSPHRASE_FINISHED
+} mnemonic_passphrase_state_t;
+
+typedef struct {
+    mnemonic_passphrase_state_t state;
+    char* passphrase;
+    void (*callback)(char*, void*);
+    void* callback_param;
+} mnemonic_passphrase_data_t;
+
+/**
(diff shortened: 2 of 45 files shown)


src/workflow/unlock.h
```diff
@@ -15,15 +15,31 @@
...
 
+#include "workflow.h"
+
 #include <compiler_util.h>
 #include <keystore.h>
 
+/**
+ * Create a workflow that prompts the user for the password and unlocks the keystore, calls
+ * the provided callback, and exits. The workflow will terminate if either the user enters the
+ * correct password, or the device is reset after too many failed attempts. The callback will
```
benma: > or the device is reset after too many failed attempts.

In this case the device reboots and there won't be any response, right?

What about if HWW_REQ_CANCEL comes in?

src/util.c
```diff
@@ -58,6 +59,17 @@ void util_cleanup_64(uint8_t** buf)
     util_zero(*buf, 64);
 }
 
+char* util_strdup(const char* str)
+{
+    size_t str_size = strlen(str);
+    char* result = malloc(str_size + 1);
```
benma: why not use `strdup` here?

src/workflow/unlock.c
```diff
@@ -122,37 +143,143 @@ keystore_error_t workflow_unlock_and_handle_error(const char* password)
...
+        util_zero(data->password, sizeof(data->password));
+        free(data->password);
+        data->password = NULL;
+    }
+}
+
+/**
+ * Callback invoked when the password entry completes.
+ */
+static void _unlock_password_completed(const char* password, void* param)
```
benma: not about this particular callback, but in general: before i went through a lot of effort to reduce "callback hell", and now it seems we are going callback heavy again.

it might be required, but it would be nice if we could figure out something that is easier to read/follow than lots of ad-hoc state machines and callbacks. 

before, the callbacks were hard to maintain and bug-prone to a point where we did not introduce new workflows because of it.

cc @NickeZ

src/workflow/unlock.h
conte91: If the workflow is stopped, cleanup() will be called. No further action needed (as every active workflow cleans up after itself). Note that atm there is no `abort()` mechanism, so stopping a workflow from "outside" is still out of scope for this PR. That will come with #164. Forcefully aborting a workflow will just destroy (`workflow_stop_workflow()`) every workflow in the stack until the selected one, which should guarantee a proper cleanup no matter the state.

So tl;dr: 
> What about if HWW_REQ_CANCEL comes in?

This is not responsibility of the `workflow` stack :)

src/util.c
conte91: Something something because portability (nah, it's because I'm stupid. Will fix).

src/workflow/unlock.c
conte91: Discussion ongoing offline

src/util.c
conte91: Fixed! 3278d241

test/unit-test/framework/workflows/workflow_status.c
```diff
@@ -0,0 +1,25 @@
...
+void _mock_status_spin(workflow_t* self)
+{
+    data_t* data = (data_t*)self->data;
+    if (data->callback) {
+        data->callback(data->callback_param);
+    }
+    workflow_stack_stop_workflow();
+}
+
+workflow_t* mock_workflow_status(void (*callback)(void*), void* callback_param)
```
NickeZ: why can't you use the real `workflow_status`?

src/workflow/unlock_bip39.c
```diff
@@ -0,0 +1,106 @@
...
+{
+    unlock_bip39_data_t* data = (unlock_bip39_data_t*)self->data;
+    { // animation
+        // Cannot render screens during unlocking (unlocking blocks)
+        // Therefore hardcode a status screen
+        UG_ClearBuffer();
+        image_lock(SCREEN_WIDTH / 2, SCREEN_HEIGHT / 2 - 1, IMAGE_DEFAULT_LOCK_RADIUS);
+        UG_SendBuffer();
+#ifndef TESTING
+        delay_ms(1200);
```
NickeZ: what the heck, do we artificially increase the processing time here? This is bad for u2fhid compatability no?

src/workflow/unlock.c
NickeZ: As discussed in other places, now that we have proper asynchronous workflows we can't have code that look synchronous/sequential since C doesn't have support for async.

I'm totally fine with this PR.

src/workflow/unlock_bip39.c
conte91: Yep, it is, and should be easy  to fix. I guess it's just to have a "nice" animation.

test/unit-test/framework/workflows/workflow_status.c
conte91: This is used by the `workflow_cancel` unit test, so the status workflow is mocked (it already was, but had to mock the async version as well!) to make it a "proper" _unit_ test and avoid testing stuff that doesn't belong to it.

src/workflow/get_mnemonic_passphrase.c
```diff
@@ -0,0 +1,156 @@
...
+{
+    char* passphrase_out = (char*)param;
+    int n_written = snprintf(passphrase_out, SET_PASSWORD_MAX_PASSWORD_LENGTH, "%s", passphrase);
+    if (n_written < 0 || n_written >= SET_PASSWORD_MAX_PASSWORD_LENGTH) {
+        Abort("Mnemonic passphrase invalid length.");
+    }
+    workflow_blocking_unblock();
+}
+
+void get_mnemonic_passphrase_blocking(char* passphrase_out)
```
benma: unused

src/workflow/get_mnemonic_passphrase.c
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * Invoked when the user has finished with the visual
+ * passphrase confirmation.
+ */
+static void _confirm_passphrase_cb(bool result, void* param)
+{
+    mnemonic_passphrase_data_t* data = (mnemonic_passphrase_data_t*)param;
+    if (!result) {
+        /* Confirmation failed. Start over. */
+        _clean_passphrase(data);
```
benma: nit: a little bit brittle, would not be needed if the passphrase was not on the heap but fixed size allocated with `SET_PASSWORD_MAX_PASSWORD_LENGTH`

src/workflow/unlock.c
```diff
@@ -118,37 +73,129 @@ keystore_error_t workflow_unlock_and_handle_error(const char* password)
...
     if (!keystore_is_locked()) {
-        return true;
+        data->state = UNLOCK_STATUS_FINISHED;
+        data->result = true;
+        return;
     }
 
+    /*
+     * FUTURE: maybe remove?
+     * It might be a "security" feature/hack, but it's not
```
benma: it's not a security feature, it's a "failsafe": 

- some screen is launched and on the stack (e.g. the random number call)
- app quits
- app restarts
- app sends unlock command
- old screen is still there, op_unlock pushes on top of it, and after it's popped, the old screen reappears

pop_all() is/was here because a new app/wallet will call op_unlock as one of the first things.

Not sure if we can get rid of it yet.

src/workflow/unlock.c
```diff
@@ -13,86 +13,37 @@
...
-#include <ui/components/ui_images.h>
-#include <ui/fonts/password_11X12.h>
 #include <ui/screen_stack.h>
-#include <ui/ugui/ugui.h>
+#include <ui/workflow_stack.h>
 #include <util.h>
-#ifndef TESTING
-#include <hal_delay.h>
-#endif
+#include <workflow/get_mnemonic_passphrase.h>
```
benma: does not seem used in this file?

src/workflow/get_mnemonic_passphrase.h
```diff
@@ -0,0 +1,11 @@
+#ifndef _GET_MNEMONIC_PASSPHRASE_H
+#define _GET_MNEMONIC_PASSPHRASE_H
+
+#include <stdbool.h>
+
+#include "workflow.h"
+
+workflow_t* get_mnemonic_passphrase(void (*callback)(char*, void*), void* callback_param);
```
benma: should be `workflow_` prefixed

src/workflow/unlock.c
conte91: That can't happen with the current code, can it?
* Blocking workflow (e.g. random number) starts
* App quits
* App restarts
* App sends unlock commands -> USB stack is blocked, command is dropped

Or with the async workflows:
* Async workflow (e.g. random number) starts, locks the USB stack
* App quits
* App restarts
* App sends unlock commands -> USB stack is blocked, command is rejected

And after implementing timers:
* Async workflow (e.g. random number) starts, locks the USB stack
* App quits
* App restarts
* App sends unlock commands -> USB stack is blocked, command is rejected
* Original workflow times out after a bit -> USB stack unlocked
* App retries the unlock command -> Unlock workflow starts.

NickeZ: > pop_all() is/was here because a new app/wallet will call op_unlock as one of the first things.
> 
> Not sure if we can get rid of it yet.

It would be great if we never have to do `pop_all()`.

src/workflow/get_mnemonic_passphrase.c
conte91: Removed 9d82e082

src/workflow/unlock.c
conte91: 133648c0

src/workflow/get_mnemonic_passphrase.h
conte91: 13dac1e2

PR #372, opened 2020-02-27
hww: graduate op_info to the hww framing layer

benma:
OP_INFO's purpose is to query the firmware version if it is not
available via the usb descriptor (bitboxbase, bitboxbridge, etc.), so
querying it should work the same way independently of the firmware
version.

Since the HWW framing layer was introduced, the call to OP_INFO needs
to be framed, which breaks the purpose. The fix is to graduate it to
the top level (treated as a special case).

The get_info method is additionally converted to a function (new name;
query_info), as it does not depend on the instance anymore (no use of
self).


NickeZ:
> OP_INFO's purpose is to query the firmware version if it is not available via the usb descriptor (bitboxbase, bitboxbridge, etc.)

btw, the bridge could provide more usb descriptor info via `api/v1/devices`. Currently it only provides `idProduct`.

PR #373, opened 2020-02-27
apps/eth: sighash bug (big endian vs little endian)

benma:
The length prefix of RLP parts should be big endian, not little
endian.

Previously we had a limit of 100 bytes for the data field, but
extended it to 1024 recently, which needs a 2 bytes length prefix,
revealing this issue.


NickeZ:
We should branch at the current release and do a hot-fix as well right?

Tomasvrba:
tACK

PR #374, opened 2020-02-29
Add endpoint to get the electrum encryption key

TheCharlatan:
This commit adds an endpoint that returns the encryption key electrum uses at keypath m/4541509'/1112098098'.


src/apps/btc/btc.c
```diff
@@ -49,6 +52,22 @@ bool app_btc_xpub(
     return btc_common_encode_xpub(&derived_xpub, xpub_type, out, out_len);
 }
 
+bool app_btc_identifier_xpub(const uint32_t* keypath, size_t keypath_len, char* out, size_t out_len)
+{
+    if (keypath_len != 2) {
+        return false;
+    }
+    if (keypath[0] != BTC_TOP_IDENTIFIER && keypath[1] != BTC_BOTTOM_IDENTIFIER) {
```
benma: this check does not what you want :D (if you check the test coverage report, you will see the bug ;))

src/apps/btc/btc.c
```diff
@@ -26,6 +26,9 @@
 #include <workflow/status.h>
 #include <workflow/verify_pub.h>
 
+#define BTC_TOP_IDENTIFIER (451509 + BIP32_INITIAL_HARDENED_CHILD)
```
benma: maybe it's just me but i find the naming very non-descriptive. please add comments.

src/apps/btc/btc.h
```diff
@@ -49,6 +49,19 @@ USE_RESULT bool app_btc_xpub(
     char* out,
     size_t out_len);
 
+/**
+ * Returns the identifier xpub.
+ * @param[in] keypath this is a dummy keypath reserved for future use.
```
benma: it's not a dummy, it is in use now. pls describe that it must be `m/4541509'/1112098098'`.

src/apps/btc/btc.h
```diff
@@ -49,6 +49,19 @@ USE_RESULT bool app_btc_xpub(
     char* out,
     size_t out_len);
 
+/**
+ * Returns the identifier xpub.
```
benma: "identifer xpub" does not describe anything really. this should mention electrum, the purpose, etc.

src/commander/commander_btc.c
```diff
@@ -168,6 +168,20 @@ commander_error_t commander_btc_pub(const BTCPubRequest* request, PubResponse* r
     }
 }
 
+commander_error_t commander_btc_identifier_xpub(
+    const BTCIdentifierXPubRequest* request,
+    BTCIdentifierXPubResponse* response)
+{
+    if (!app_btc_identifier_xpub(
+            request->keypath,
+            sizeof(request->keypath) / sizeof(uint32_t),
```
benma: this must be `request->keypath_count`. the user can also send a `0` or `1` length keypath.

messages/btc.proto
```diff
@@ -154,3 +154,14 @@ message BTCResponse {
     BTCIsScriptConfigRegisteredResponse is_script_config_registered = 2;
   }
 }
+
+// This function can be used to get an identifying xpub at the keypath m/4541509'/1112098098'"
+// The keypath argument is reserved for future use
+message BTCIdentifierXPubRequest {
```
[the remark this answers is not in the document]
TheCharlatan: It follows the pattern of the BTCPubRequest (since it shares some code with it in the end), should I still move it to BTCResponse?

src/apps/btc/btc.c
TheCharlatan: Ah damn it ^^

src/commander/commander_btc.c
TheCharlatan: Right, I should have asked you about this :/

NickeZ:
Couldn't this just reuse the current api for xpubs if we whitelisted that keypath?

TheCharlatan:
Thanks for the review @benma , I addressed everything besides moving to the BTCRequest/BTCResponse message.
@NickeZ I don't feel comfortable changing the normal xpub logic for this purpose. I'd rather have a separate path for these special cases.

benma:
> Couldn't this just reuse the current api for xpubs if we whitelisted that keypath?

They serve different purposes and behave differently in a number of ways. Using the same api call would break down (or get messy) if we add a user confirmation to export an xpub and showing the account name, for example.

messages/btc.proto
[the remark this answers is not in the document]
benma: Yes please. The rest is different as it came before we started bundling the apps in their own request/response layer. We might move them also to BTCRequest/BTCResponse some day.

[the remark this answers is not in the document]
TheCharlatan: Ok, moved to BTCRequest/BTCResponse.

py/send_message.py
```diff
@@ -459,6 +467,7 @@ def _menu_init(self) -> None:
             ("Retrieve Ethereum address", self._display_eth_address),
             ("Sign Ethereum tx", self._sign_eth_tx),
             ("Sign Ethereum Message", self._sign_eth_message),
+            ("Show Electrum allet encryption xpub", self._get_electrum_wallet_encryption_xpub),
```
benma: allet->wallet

messages/keystore.proto
```diff
@@ -0,0 +1,12 @@
+// This function can be used to get an identifying xpub at the keypath m/4541509'/1112098098'"
+// The keypath argument has to be m/4541509'/1112098098'
+
+syntax = "proto3";
+
+message ElectrumWalletEncryptionXPubRequest {
```
NickeZ: I would really prefer if the API didn't have `xpub` in it because we don't want users assuming it has something to do with xpubs when they read our API documentation that will be automatically generated from the message descriptions.

TheCharlatan: so s/ElectrumWalletEncryptionXPubRequest/ElectrumWalletEncryptionRequest ?

benma: what about `ElectrumEncryptionKeyRequest`

TheCharlatan: Ok, done renaming.

PR #379, opened 2020-03-03
Keystore: Use keystore to get xpub with simple xpub version bytes

TheCharlatan:
The ethereum app had to call code in the bitcoin app to get a serialized xpub with the normal "xpub" version bytes. This breaks the app isolation in the code. The ethereum xpub call just requires the simple xpub, without version tweaks, so this can be handled in a separate function in keystore.c . Also change this for the electrum encryption key function, which now also serializes the xpub directly in keystore.c .

### DIFF
--- a/src/apps/btc/btc.c
+++ b/src/apps/btc/btc.c
@@ -26,9 +26,6 @@
 #include <workflow/status.h>
 #include <workflow/verify_pub.h>
 
-#define ELECTRUM_WALLET_ENCRYPTION_KEYPATH_LEVEL_ONE (4541509 + BIP32_INITIAL_HARDENED_CHILD)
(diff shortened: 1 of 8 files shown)


benma:
Finally done in https://github.com/digitalbitbox/bitbox02-firmware/pull/379, a rebased and more general version.

x1ddos:
> Finally done in #379, a rebased and more general version.

@benma you mean https://github.com/digitalbitbox/bitbox02-firmware/pull/629, right?

benma:
> > Finally done in #379, a rebased and more general version.
> 
> @benma you mean #629, right?

No, my comment was correct ;) was about moving the xpub encoding function out of the btc app.

benma:
> > Finally done in #379, a rebased and more general version.
> 
> @benma you mean #629, right?

Actually I meant this one :P 

https://github.com/digitalbitbox/bitbox02-firmware/pull/626

PR #380, opened 2020-03-03
rust: refactor

NickeZ:
* move `extern "C"` functions to new crate
* allow rust code in more places like firmwares, bootloaders and unit
  tests


src/rust/util/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+/// Guaranteed to wipe the provided buffer
+pub fn zero(dst: &mut [u8]) {
+    for p in dst {
+        unsafe { core::ptr::write_volatile(p, 0) };
+    }
+}
+
+/// Converts binary data to hex representation. Requires that resulting string is `buf.len() * 2`.
+pub fn u8_to_hex(buf: &[u8], res: &mut str) {
+    const DIGITS: &'static [u8; 16] = b"0123456789abcdef";
```
benma: looks like https://docs.rs/hex/0.4.2/hex/trait.ToHex.html#tymethod.encode_hex could be used here instead of handrolled?

NickeZ: I'll just merge this, we can consider adding external dependencies later

PR #382, opened 2020-03-04
test_salt: Remove double definition of memory_get_salt_root.

conte91:
The `memory_get_salt_root` function is defined twice in the test framework (both in the framework code and in `test_salt.c`).


conte91:
Didn't see it had been fixed already... Closing.

PR #383, opened 2020-03-04
Rust hex

benma:


src/rust/bitbox02-rust-c/src/alloc.rs
```diff
@@ -0,0 +1,41 @@
+#[alloc_error_handler]
+#[cfg(not(test))]
+/// Function name is arbitrary.
```
NickeZ: Since you are using three slashes this is the documentation for the function. Use two forward slashes if you want to write documentation that isn't documentation.

src/rust/util/Cargo.toml
```diff
@@ -3,3 +3,6 @@ name = "util"
 version = "0.1.0"
 authors = ["Shift Cryptosecurity AG <support@shiftcrypto.ch>"]
 edition = "2018"
+
+[dependencies]
+hex = { version = "0.4", default-features = false }
```
NickeZ: no newline at end of file

src/rust/util/src/lib.rs
```diff
@@ -11,21 +10,7 @@ pub fn zero(dst: &mut [u8]) {
...
-    // access to the string bytes.
-    unsafe {
-        let res_bytes = res.as_bytes_mut();
-        for i in 0..buf.len() {
-            let c = buf[i];
-            res_bytes[i * 2] = DIGITS[(c >> 4 & 0xF) as usize];
-            res_bytes[i * 2 + 1] = DIGITS[(c & 0xF) as usize];
-        }
-    }
+    hex::encode_to_slice(buf, unsafe { res.as_bytes_mut() }).unwrap()
```
NickeZ: do we even need the `u8_to_hex` function? We can probably use `hex::encode_to_slice` where we need to convert some bytes? Like directly in `rust_util_u8_to_hex`.

src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -17,7 +17,7 @@ pub extern "C" fn rust_util_uint8_to_hex(
     let out = unsafe {
         core::str::from_utf8_unchecked_mut(core::slice::from_raw_parts_mut(out_buf, buf_len * 2))
```
NickeZ: avoid creating a `str` here

src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -17,7 +17,7 @@ pub extern "C" fn rust_util_uint8_to_hex(
...
     };
-    util::u8_to_hex(buf, out);
+    hex::encode_to_slice(buf, unsafe { out.as_bytes_mut() }).unwrap();
```
NickeZ: you won't need unsafe here when `out` is a `&mut [u8]` instead.

PR #389, opened 2020-03-06
All ascii

benma:


src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -11,6 +11,12 @@ pub extern "C" fn rust_util_zero(mut dst: BytesMut) {
     util::zero(dst.as_mut())
 }
 
+#[no_mangle]
+pub extern "C" fn rust_util_all_ascii_bytes(bytes: Bytes) -> bool {
+    assert!(!bytes.buf.is_null());
```
NickeZ: this isn't needed, if the pointer is null and the length != 0 it will hit an assert in the `as_ref()` implementation. If the pointer is null and the length is 0 it should be `true` to call `all_ascii_bytes()`.

src/rust/util/src/ascii.rs
```diff
@@ -0,0 +1,32 @@
...
+    fn test_all_ascii() {
+        // All ascii chars.
+        assert!(all_ascii(ALL_ASCII));
+        // Edge cases: highest and lowest non ascii chars.
+        assert!(!all_ascii(b"\x7f"));
+        assert!(!all_ascii(b"\x19"));
+        assert!(!all_ascii(b"\n"));
+        assert!(!all_ascii(b"\t"));
+        // Works for any AsRef<[u8]>
+        assert!(all_ascii(b"foo" as &dyn AsRef<[u8]>));
```
NickeZ: To really make a point you could do

```rust
let trait_obj: &dyn AsRef<[u8]> = &"abc";
all_ascii(trait_obj)
```

benma: Done

src/rust/bitbox02-rust-c/src/util.rs
benma: Done

src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -147,12 +158,20 @@ pub extern "C" fn rust_util_bytes_mut(buf: *mut c_uchar, len: usize) -> BytesMut
...
-/// * `len` - Length of buffer, `buf_ptr[buf_len-1]` must be a valid dereference
+/// * `buf` - Must be a valid pointer to a null terminated array of bytes.
 #[no_mangle]
-pub extern "C" fn rust_util_cstr(buf: *const c_char, len: usize) -> CStr {
+pub extern "C" fn rust_util_cstr(buf: *const c_char) -> CStr {
+    assert!(!buf.is_null());
+    let mut len: usize = 0;
+    let mut b = buf;
+    unsafe {
+        while b.read() != 0 {
```
benma: @NickeZ is there any difference from this to `*b != 0`? the read doc says something about "not moving", not sure what it means.

src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -147,12 +147,20 @@ pub extern "C" fn rust_util_bytes_mut(buf: *mut c_uchar, len: usize) -> BytesMut
...
+/// Convert buffer to str.
 ///
-/// * `buf` - Must be a valid pointer to an array of bytes
-/// * `len` - Length of buffer, `buf_ptr[buf_len-1]` must be a valid dereference
+/// * `buf` - Must be a valid pointer to a null terminated array of bytes.
 #[no_mangle]
-pub extern "C" fn rust_util_cstr(buf: *const c_char, len: usize) -> CStr {
+pub extern "C" fn rust_util_cstr(buf: *const c_char) -> CStr {
+    assert!(!buf.is_null());
+    let mut len: usize = 0;
```
NickeZ: do you need `usize` here?

src/rust/util/src/ascii.rs
```diff
@@ -0,0 +1,33 @@
+extern "C" {
+    pub fn puts(s: *const u8) -> i32;
```
NickeZ: old debug code?

src/rust/bitbox02-rust-c/src/util.rs
NickeZ: I don't think there is any difference. I would probable have written `*b` out of pure habit. Regarding `move` it doesn't really matter as the type you would move in this case is "plan old data" and would therefore simply be copied.

* https://doc.rust-lang.org/std/ops/trait.Deref.html
* https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-dereference-operator

NickeZ: Another way of writing similar code is:

```rust
let mut i = 0;
let len = loop {
    if *b == 0 {
        break i;
    }
    i += 1;
}
```

NickeZ: Maybe breaking with a value doesn't make sense in this case.. but I wanted you to know about it.

benma: very neat trick! thx

benma: > Another way of writing similar code is:
> 
> ```rust
> let mut i = 0;
> let len = loop {
>     if *b == 0 {
>         break i;
>     }
>     i += 1;
> }
> ```

you are not advancing `b` anywhere

src/rust/bitbox02-rust-c/src/util.rs
benma: no, removed

src/rust/bitbox02-rust-c/src/util.rs
NickeZ: > you are not advancing `b` anywhere

I just wanted to show "break with return" ;) The rest is left as an exercise for the reader!

src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -11,6 +11,17 @@ pub extern "C" fn rust_util_zero(mut dst: BytesMut) {
...
 
+#[no_mangle]
+pub extern "C" fn rust_util_all_ascii_bytes(bytes: Bytes) -> bool {
+    util::ascii::all_ascii(bytes)
+}
+
+#[no_mangle]
+pub extern "C" fn rust_util_all_ascii(cstr: CStr) -> bool {
+    let s: &str = cstr.as_ref();
+    util::ascii::all_ascii(s)
```
NickeZ: I think this could also be written `util::ascii::all_ascii(&cstr)`.

benma: it can be, but i wanted to make the type clear if `s`. With all the traits/generics it can be hard to tell what the concrete type is at times.

PR #390, opened 2020-03-06
rust: simplify CStrMut.AsMut

benma:
Should not be &str but &[u8].

### DIFF
--- a/src/rust/bitbox02-rust-c/src/util.rs
+++ b/src/rust/bitbox02-rust-c/src/util.rs
@@ -29,9 +29,9 @@ pub extern "C" fn rust_util_all_ascii(cstr: CStr) -> bool {
 /// * `out_ptr` - Must be a valid pointer to an array of bytes that is buf_len*2+1 long
 #[no_mangle]
 pub extern "C" fn rust_util_uint8_to_hex(buf: Bytes, mut out: CStrMut) {
-    let out_len = out.as_ref().len();
     // UNSAFE: We promise that we never write non-utf8 valid bytes to the `str`.
-    let out = unsafe { out.as_mut().as_bytes_mut() };
+    let out = out.as_mut();
+    let out_len = out.len();
     hex::encode_to_slice(&buf, &mut out[0..out_len - 1]).unwrap();
     out[out_len - 1] = b'\0';
 }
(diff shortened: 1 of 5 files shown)


src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -125,18 +125,17 @@ impl AsRef<str> for CStrMut {
     }
 }
 
-impl AsMut<str> for CStrMut {
+impl CStrMut {
     /// Create a slice to a buffer. Only allowed for non-null pointers with length or null pointers
     /// with 0 length due to limitation in `core::slice`.
-    fn as_mut(&mut self) -> &mut str {
+    /// The caller must ensure that the string is null terminated.
+    unsafe fn as_bytes_mut(&mut self) -> &mut [u8] {
```
NickeZ: should probably be `pub fn`.

src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -29,9 +29,9 @@ pub extern "C" fn rust_util_all_ascii(cstr: CStr) -> bool {
 /// * `out_ptr` - Must be a valid pointer to an array of bytes that is buf_len*2+1 long
 #[no_mangle]
 pub extern "C" fn rust_util_uint8_to_hex(buf: Bytes, mut out: CStrMut) {
-    let out_len = out.as_ref().len();
```
NickeZ: please update the `UNSAFE` notice, and the function documentation as well (I forgot that earlier).

benma: Done.

src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -233,6 +248,26 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_cstr_mut() {
+        let mut start = String::from("foo\0bar");
+        let mut cstr_mut = rust_util_cstr_mut(start.as_mut_ptr(), start.len());
+        assert_eq!(cstr_mut.len, start.len());
+        assert_eq!(cstr_mut.as_ref(), "foo");
+        let bytes_mut = unsafe { cstr_mut.as_bytes_mut() };
+        bytes_mut[0] = 'g' as u8;
```
NickeZ: `b'g'` is probably more idiomatic than `'g' as u8`, but I guess both are fine.

benma: TIL

src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -121,22 +123,35 @@ impl AsRef<str> for CStrMut {
...
+            while b.read() != 0 {
+                b = b.offset(1);
+                len += 1;
+                if len == self.len {
+                    panic!("CStrMut not null terminated");
+                }
+            }
+        }
+
+        unsafe { core::str::from_utf8_unchecked(core::slice::from_raw_parts(buf, len)) }
```
NickeZ: actually, if we want to panic on non-utf8 bytes we can remove the `unsafe` here and use `core::str::from_utf8()` (which isn't unsafe, because it checks that the string is a valid utf-8 encoded string).

benma: `from_raw_parts` is unsafe though?

PR #391, opened 2020-03-06
Pairing

benma:
This fixes the 'unexpected response' error we've been seeing on the v7.0.0 firmware.

In addition, the show_pairing flow now behaves the same as as in the bitboxapp/MEW:

1) Wallet shows pairing code, confirmation button is disabled/hidden
2) User confirms on device, which enables/shows the OK button on the wallet
3) User confirms on the host.

send_message.py now also implements this as a demo. Accepting is the default for least annoyance ;)

### DIFF
--- a/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
+++ b/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
@@ -329,16 +329,15 @@ class BitBoxProtocol(ABC):
         pairing_verification_required_by_device = response == b"\x01"
         if pairing_verification_required_by_host or pairing_verification_required_by_device:
             cid = self._transport.generate_cid()
-            self._transport.write(OP_I_CAN_HAS_PAIRIN_VERIFICASHUN, HWW_CMD, cid)
             client_response_success = noise_config.show_pairing(
                 "{} {}\n{} {}".format(
                     pairing_code[:5], pairing_code[5:10], pairing_code[10:15], pairing_code[15:20]
                 )
             )
+            pairing_response = self._raw_query(OP_I_CAN_HAS_PAIRIN_VERIFICASHUN)
             if not client_response_success:
                 self.close()
                 raise Exception("pairing rejected by the user on client")
-            pairing_response = self._transport.read(HWW_CMD, cid)
 
             if pairing_response == RESPONSE_SUCCESS:
                 pass
--- a/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
+++ b/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
@@ -21,7 +21,7 @@ import base64
 import binascii
 import hashlib
 import time
-from typing import Optional, List, Dict, Tuple, Union
+from typing import Callable, Optional, List, Dict, Tuple, Union
 from typing_extensions import TypedDict
 
 import ecdsa
@@ -210,8 +210,13 @@ class BitBoxNoiseConfig:
     """ Stores Functions required setup a noise connection """
 
     # pylint: disable=no-self-use,unused-argument
-    def show_pairing(self, code: str) -> bool:
-        return True
+    def show_pairing(self, code: str, device_response: Callable[[], bool]) -> bool:
+        """
+        Returns True if the user confirms the pairing (both device and host).
+        Returns False if the user rejects the pairing (either device or host).
+        This function must call device_response() to invoke the pairing screen on the device.
+        """
+        return device_response()
 
     def attestation_check(self, result: bool) -> None:
         return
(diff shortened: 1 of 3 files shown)


benma:
@NickeZ the api might be a bit funky now, not sure how to make it much easier for the library user though.

@TheCharlatan pls check and if ok try to integrate into electrum, the fix should ideally be out before the release, otherwise the v7.0.0 firmware release will be incompatible due to this bug.

TheCharlatan:
How soon do you want to release the new firmware? The change is not compatible with the current workflow. I'll have to re-write the pairing screen such that it is non-blocking. Since I have little experience with qt in this regard, it would take me quite some time to fix.

benma:
> > the fix should ideally be out before the release

I meant before the Electrum release ;) Since the Python lib you use is already accepting 7.0.0 firmware but has this bug, when people upgrade the firmware, electrum will not show a "needs Electrum upgrade" warning, but a pairing error.

> How soon do you want to release the new firmware?

~2 weeks? Not quite sure.

NickeZ:
* Why does this fix the bug?
* I'm not really happy with using an asynchronous runtime inside an API wrapper.

benma:
> Why does this fix the bug?

See first commit message (the pairing code query only was split into a separate write followed by read, and did not use the query function which behaves differently according to the firmware version):

> > _raw_query behaves correctly according to the firmware
version. Calling _transport.write directly does not and hence the
pairing broke on v7.0.0.

> I'm not really happy with using an asynchronous runtime inside an API wrapper.

I am also unsure of it, but it does make it easy for the library user to implement the pairing flow as it is done in the other apps.

Do you have other ideas?

NickeZ:
> > Why does this fix the bug?
> 
> See first commit message (the pairing code query only was split into a separate write followed by read, and did not use the query function which behaves differently according to the firmware version):

OK, read the message and understand now.

> I am also unsure of it, but it does make it easy for the library user to implement the pairing flow as it is done in the other apps.
> 
> Do you have other ideas? Happy to improve it but don't know how.

I didn't know that asyncio was part of the stdlib (already since 3.4 :O) that makes it more OK. It seems to interoperate with the other possible async runtimes as well.

It seems wrong to me to `close()` the event loop. Maybe I'm misunderstanding something. If you fetch an event loop from the application, can it continue to use it afterwards? The documentation recommends `asyncio.run()` over the event loop interface. Does `run()` not work in our case?

Actually I think async/await syntax solves many problems, it is possible to write asynchronous code sequentially which is much more readable. So I'm not inherently against it.

One idea: async/await is just a specialization of generators. So potentially we could solve it with python generators instead.

benma:
> asyncio.run()

Unfortunately only available since Python 3.8.

> It seems wrong to me to close() the event loop. Maybe I'm misunderstanding something. If you fetch an event loop from the application, can it continue to use it afterwards? 

Good catch! It seems I misread how the eventloop api works. I thought I could just run one for the duration of `noise_connect()`. Great idea about using a generator instead, I'll try that out.

benma:
@NickeZ got two solutions:

`async` without asyncio, but a handrolled "one-shot" eventloop:

```Python

async def show_pairing(device_response):
    if not await device_response():
        return False
    return input("user: ").strip() == 'y'

def pair():
    async def query_device():
        return input("device: ").strip() == 'y'
    try:
        print(show_pairing(query_device).send(None))
    except StopIteration as exc:
        return exc.value

print("Result", pair())
```

```
$ python3 test.py
device: y
user: y
Result True
```

Or generator based:

```

def show_pairing(device_response):
    if not next(device_response()):
        return False
    return input("user: ").strip() == 'y'

def pair():
    def query_device():
        yield input("device: ").strip() == 'y'
    return show_pairing(query_device)

print("Result", pair())
```

As far as i can tell, both to exactly the same thing, so it in the end it's a matter of syntax: use `await device_response` in the `show_pairing` vs using `next(device_response())`.

The `await` one makes it clear that the response is async and one result, while in the generator case, it's not clear without documentation that the generator yields always exactly one value.

Which variant to pick?

benma:
@NickeZ so generators worked, as well as invoking async/await without asyncio, but then I realized that this is kind a roundabout way of making `device_response` in `show_pairing` be a callback. I pushed a commit that removes all the async stuff and just makes `device_response` a callback.

Not sure I am missing something?

NickeZ:
Seems like the solution with a callback is more in line with how we have done things so far. So I like it more

benma:
@TheCharlatan works?

TheCharlatan:
When the pairing is rejected on the device I'm getting a "Noise communication failed." exception with my current implementation. All I'm doing is in send_message.py:
```
show_pairing(self, code, device_response):
       thr = threading.Thread(target=device_response, args=(), kwargs={})
       thr.start()
       reply = input("Accept pairing? [y]/n: ").strip() != "n"                                                                                                   
       thr.join()
       return reply
```
I don't quite get what is going wrong here.

NickeZ:
It seems you are ignoring the result from device_response(). If that function returns false your callback must also return false.

PR #393, opened 2020-03-07
rust/CStrMut: Implement a safe interface

NickeZ:


src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -119,46 +115,36 @@ impl AsRef<str> for CStrMut {
...
+    pub fn unused(&mut self, req: usize) -> &mut [u8] {
+        // Must be room for requested amount of bytes and null terminator.
+        if self.cap - self.len < req + 1 {
+            panic!("Not enough bytes left in buffer");
+        }
+        let len = self.len;
+        self.len += req;
+        unsafe {
+            self.buf.offset((len + req) as isize).write(0);
+            core::slice::from_raw_parts_mut(self.buf.offset(len as isize), req)
```
benma: should be outside of `unsafe{}` so the caller has to mark the section unsafe, as they have to promise not to insert a null terminator, and actually to promise to fill the bytes (they might be all null to start with). otherwise `self.len` is wrong.

Could an alternative be to not return a `&mut [u8]`, but to do sth like this?

`cstrmut.unused(|&bytes| hex::encode_to_slice(&buf, &bytes))`? Then we could run the callback, and afterwards check that there is no null in `bytes`, making it safe.

src/rust/bitbox02-rust-c/src/util.rs
```diff
@@ -119,46 +115,36 @@ impl AsRef<str> for CStrMut {
...
-            buf = core::ptr::NonNull::dangling().as_ptr();
+    /// Create a mutable slice to an unused range of the buffer. Will ensure that byte immediately
+    /// following the range is null.
+    pub fn unused(&mut self, req: usize) -> &mut [u8] {
+        // Must be room for requested amount of bytes and null terminator.
+        if self.cap - self.len < req + 1 {
+            panic!("Not enough bytes left in buffer");
+        }
+        let len = self.len;
+        self.len += req;
```
benma: i'm not sure I see the purpose of being able to call `unused()` multiple times to do a sort of `concat`. Do you have a specific use case for it?

NickeZ: I was trying to come up with a safe interface so that the business logic code doesn't have to do null termination and use unsafe functions

src/rust/bitbox02-rust-c/src/util.rs
NickeZ: You are right, the user could write null into the byte array.. so currently this function would have to be marked unsafe.

We could also have a "write(&[u8])" function that does all the checks. A bit like "FixedCString" I've already implemented.

NickeZ:
Implemented with `std::fmt::Write`. Has one drawback, we need a temporary stack-allocated buffer. Since rust doesn't have `alloca` it is fixed size, which kind of sucks.

NickeZ:
The last commit now implements a callback interface instead which doesn't require a temporary buffer in the case of hex-conversion.

One benefit of implementing the `std::fmt::Write` trait is that one can use the normal `write!()` macro to write formatted strings like `write!(out, "{} {}", "Hello", "world!");` which is kind of nice.

src/rust/bitbox02-rust-c/src/util.rs
benma: not sure how that line is related to it, this is ensured by the checks below. 
is calling `.write()` twice expected to not overwrite, but append? if so it would be good to document.

NickeZ: yes, exactly, if a container is growable it will append by defualt:

```rust
use std::fmt::Write;
fn main() {
    let mut s = String::from("abc");
    let _ = write!(s, "d");
    assert!(&s == "abcd");
}
```

NickeZ: oh, and regarding your first comment. If you ever write over bytes in a string you also have to ensure that the following 3 bytes are valid. (So that you are not eating into a 4 byte utf-8 char by mistake). So anything that is appending is much easier than something that is modifying.

PR #400, opened 2020-03-10
cmake: don't use cmake libraries to create link cli

NickeZ:


benma:
Well done!

A detailed commit message will be highly appreciated :grin:

NickeZ:
I've changed around some things and improved the commit message.

test/device-test/CMakeLists.txt
```diff
@@ -125,16 +129,32 @@ foreach(name ${BB02_TEST_LIST})
...
+      secp256k1
+      bitbox02_rust_c
+      noiseprotocol
+      cryptoauthlib
+      base32
+      bignum # TODO: only eth
+      sha3   # TODO: Only eth
+      ctaes
+      c
+      asf4-drivers-min
```
conte91: These lines seem to be repeated pretty much everywhere - can we store them in a variable? I have the impression we're going to screw up this sequence next time we change the `CMakeLists.txt` and we'll be back to the starting point.

NickeZ: yeah that makes sense, as long as it works with the linking order and so on. The rust-library can't be in the variable for example.

NickeZ: did a variable.

PR #403, opened 2020-03-12
Rust products

benma:


benma:
@NickeZ `target-firmware` (multi edition) can then add features like `app_ethereum` etc.

PR #409, opened 2020-03-13
libwally: enforce optimize for size

NickeZ:
libwally was compiling with `-03`. no bueno.

```
old:
   text    data     bss     dec     hex filename
 452980   22924  208352  684256   a70e0 firmware.elf

new:
   text    data     bss     dec     hex filename
 443532   22932  208352  674816   a4c00 firmware.elf
```

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -120,6 +120,7 @@ ExternalProject_Add(libwally-core
                   ${CMAKE_CURRENT_SOURCE_DIR}/libwally-core/configure
                   ${CONFIGURE_FLAGS}
                   ${LIBWALLY_CONFIGURE_FLAGS}
+  COMMAND         find ${CMAKE_CURRENT_BINARY_DIR}/libwally-core -name Makefile | xargs sed -i "s/-O3/-Os/"
   # TODO: Add darwin hack to libwally repo
   BUILD_COMMAND   ${CMAKE_MAKE_PROGRAM}
   INSTALL_COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}


benma:
libwally is the lib which performs the slow bip39 unlock, and also ships secp256k1. did you check if unlock does not take longer, and that signing a tx works?

external/CMakeLists.txt
```diff
@@ -120,6 +120,7 @@ ExternalProject_Add(libwally-core
                   ${CMAKE_CURRENT_SOURCE_DIR}/libwally-core/configure
                   ${CONFIGURE_FLAGS}
                   ${LIBWALLY_CONFIGURE_FLAGS}
+  COMMAND         find ${CMAKE_CURRENT_BINARY_DIR}/libwally-core -name Makefile | xargs sed -i "s/-O3/-Os/"
```
benma: This is very not nice :D Is there really no way to configure it or pass it as an env var or sth? 

otherwise you can also make a new branch here: https://github.com/shiftdevices/libwally-core

NickeZ:
* I'll have to check that libwally operations don't take significantly longer.
* It isn't possible to change the optimization flags for libwally, they have only 2 options, debug `-Og` and non-debug `-O3`.
* Using `-O3` in general is absolute madness. They should be using `-O2`. `-O3` enables new and exciting optimizations which hasn't been battle tested. It is well known to sometimes pessimize code.

I'll patch our fork of libwally instead of monkeypatching like this though..

benma:
In this case I suggest to fix it upstream, and not do this change for now. Seems a bit premature, we are not strapped for space yet.

NickeZ:
Solved upstream: https://github.com/ElementsProject/libwally-core/pull/182

They agreed that -O3 is wrong and overriding the user is wrong. So now it is possible to set optimization level with `CFLAGS`.

x1ddos:
> Solved upstream: ElementsProject/libwally-core#182

Still need to pull it in and update commit hash in https://github.com/digitalbitbox/bitbox02-firmware/tree/master/external, right.

Also, would be nice if `url` entries in https://github.com/digitalbitbox/bitbox02-firmware/blob/26a8b821aad9814007b9ff5dacf043f8883ced70/.gitmodules were updated to point to github/digitalbitbox/repo instead of shiftdevices.

NickeZ:
Sure, but both those things are unrelated to this PR. Make comments / create issues instead.

PR #410, opened 2020-03-13
New gcc

NickeZ:
didn't push the new container image yet.

This new toolchain increases the size of the targets by about 10%.


benma:
Here is a a later commit that attempts an upgrade (to GCC 11.3): 

https://github.com/benma/bitbox02-firmware/tree/update-gcc

That one increases the binary size by ~15kB, which is a lot.

PR #412, opened 2020-03-15
Hexaddress

benma:


src/rust/apps/ethereum/src/address.rs
```diff
@@ -0,0 +1,119 @@
...
+/// `recipient` - 20 byte tail (last 20 bytes of the pubkeyhash).
+/// `result` out will hold the address string, must have space for 42 chars.
+pub fn from_pubkey<T>(pubkey_uncompressed: &[u8; 65], result: &mut T) -> core::fmt::Result
+where
+    T: core::fmt::Write,
+{
+    let hash = sha3::Keccak256::digest(&pubkey_uncompressed[1..]);
+    let mut last20: [u8; 20] = [0; 20];
+    last20.copy_from_slice(&hash[hash.len() - 20..]);
+    from_pubkey_hash(&last20, result)
```
benma: i don't like that i can't get to a array from a slice directly without copying :/

NickeZ: What you want is `const generics` which we don't have yet. But we have workarounds ;)!, if you add the "arrayref" crate you can write the following:

```rust
    // Would have preferred sha3::Keccak256::output_size(), but it isn't const fn.
    let last20 = arrayref::array_ref!(hash, 32-20, 20);
```

Eventually (probably in a year or two) we will have proper const generics and all these crates will go away.

NickeZ: Hah, actually `hash.len()` is const so the following works:

```rust
    let last20 = arrayref::array_ref!(hash, hash.len()-20, 20);
```

NickeZ:
Could you put all the new crates in it's own commit to simplify reviewing?

src/rust/bitbox02-rust-c/src/app_ethereum.rs
```diff
@@ -17,3 +19,60 @@ pub unsafe extern "C" fn rust_ethereum_keypath_is_valid_address(
...
+    use core::convert::TryFrom;
+    let recipient = <[u8; 20]>::try_from(recipient.as_ref()).unwrap();
+    ethereum::address::from_pubkey_hash(&recipient, &mut out).unwrap();
+}
+
+#[no_mangle]
+pub extern "C" fn rust_ethereum_address_from_pubkey(pubkey: Bytes, mut out: CStrMut) {
+    assert!(pubkey.as_ref().len() == 65);
+    let ptr = pubkey.as_ref().as_ptr() as *const [u8; 65];
+    let pubkey = unsafe { &*ptr };
```
NickeZ: I would recommend using `arrayref` here as well: 
```rust
    let pubkey: &[u8;65] = arrayref::array_ref!(pubkey.as_ref(), 0, 65);
```

benma: Nice, done.

src/rust/apps/ethereum/src/address.rs
benma: excellent! this is much better.

PR #414, opened 2020-03-17
USB: Async workflow timeouts.

conte91:
When the USB stack is waiting for an asynchronous operation
to complete, a timer is started. If no traffic is received by the BB02
within 0.5s, the outstanding operation is forcefully aborted.

### DIFF
--- a/src/u2f/u2f_app.c
+++ b/src/u2f/u2f_app.c
@@ -32,16 +32,18 @@ static const app_t _apps[] = {
     },
 };
 
-struct {
+static struct {
     /** Type of outstanding async operation. */
     enum u2f_app_confirm_t outstanding_confirm;
+    /** Confirmation workflow. */
+    workflow_t* confirm_wf;
     /** App ID of the outstanding async operation. */
     uint8_t app_id[32];
     /** Whether the outstanding async operation has been confirmed. */
     bool confirmed;
     /** Whether the confirmation step has finished. */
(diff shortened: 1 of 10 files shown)


src/u2f/u2f_app.c
```diff
@@ -35,13 +35,15 @@ static const app_t _apps[] = {
...
+    /** Confirmation workflow. */
+    workflow_t* confirm_wf;
     /** App ID of the outstanding async operation. */
     uint8_t app_id[32];
     /** Whether the outstanding async operation has been confirmed. */
     bool confirmed;
     /** Whether the confirmation step has finished. */
     bool confirmed_done;
-} _state;
+} _state = {0};
```
benma: `static` missing?

src/u2f/u2f_app.h
```diff
@@ -48,4 +48,6 @@ async_op_result_t u2f_app_confirm_retry(enum u2f_app_confirm_t type, const uint8
  */
 void u2f_app_confirm_finish(void);
 
+void u2f_app_confirm_abort(void);
```
benma: never forget the docstrings :P

conte91: 2f261fa

src/u2f/u2f_app.c
conte91: 6f20cd8

PR #415, opened 2020-03-17
Fix unknown tokens

NickeZ:
This has two fixes for ethereum

### DIFF
--- a/src/apps/eth/eth.c
+++ b/src/apps/eth/eth.c
@@ -14,35 +14,44 @@
 
 #include "eth.h"
 #include "eth_params.h"
+#include "eth_verify.h"
 #include <apps/btc/btc_common.h>
+#include <workflow/verify_pub.h>
 
 #include <keystore.h>
 #include <rust/rust.h>
+#include <util.h>
 
 #include <hww.pb.h>
 #include <secp256k1.h>
 
-bool app_eth_address(
+static const char* _coin_eth = "Ethereum";
+static const char* _coin_ropsten_eth = "Ropsten";
+static const char* _coin_rinkeby_eth = "Rinkeby";
+
+app_eth_sign_error_t app_eth_address(
     ETHCoin coin,
     ETHPubRequest_OutputType output_type,
     const uint32_t* keypath,
     size_t keypath_len,
     char* out,
-    size_t out_len)
+    size_t out_len,
+    bool display,
+    const uint8_t* contract_address)
 {
     if (coin > _ETHCoin_MAX) {
-        return false;
+        return APP_ETH_SIGN_ERR_INVALID_INPUT;
     }
 
     const app_eth_coin_params_t* params = app_eth_params_get(coin);
     if (params == NULL) {
-        return false;
+        return APP_ETH_SIGN_ERR_INVALID_INPUT;
     }
 
     switch (output_type) {
     case ETHPubRequest_OutputType_ADDRESS: {
         if (!rust_ethereum_keypath_is_valid_address(keypath, keypath_len, params->bip44_coin)) {
-            return false;
+            return APP_ETH_SIGN_ERR_INVALID_INPUT;
         }
         uint8_t pubkey_uncompressed[65];
         if (!keystore_secp256k1_pubkey(
@@ -51,7 +60,7 @@ bool app_eth_address(
                 keypath_len,
                 pubkey_uncompressed,
                 sizeof(pubkey_uncompressed))) {
-            return false;
+            return APP_ETH_SIGN_ERR_INVALID_INPUT;
         }
         rust_ethereum_address_from_pubkey(
             rust_util_bytes(pubkey_uncompressed, sizeof(pubkey_uncompressed)),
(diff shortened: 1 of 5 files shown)


src/apps/eth/eth_sign_msg.c
```diff
@@ -33,9 +41,13 @@ app_eth_sign_error_t app_eth_sign_msg(
...
         return APP_ETH_SIGN_ERR_INVALID_INPUT;
     }
+    // assert keypath length so that we can access the "network" piece
+    if (request->keypath_count < 2) {
+        return APP_ETH_SIGN_ERR_INVALID_INPUT;
+    }
     // Only support main net for now. Otherwise a user could be tricked into signing something for
     // main net even if they believe they are signing for testnet.
-    if (request->coin != ETHCoin_ETH) {
+    if (!_verify_network(request->coin, request->keypath[1])) {
```
benma: The keypath/coin check is already performed by `app_eth_address()`, so it seems unnecessary here (it would also be a duplication of `app_eth_params_get` and `rust_ethereum_keypath_is_valid_address`.

src/apps/eth/eth_sign_msg.c
```diff
@@ -33,9 +41,13 @@ app_eth_sign_error_t app_eth_sign_msg(
     if (request->msg.size > 9999) {
         return APP_ETH_SIGN_ERR_INVALID_INPUT;
     }
+    // assert keypath length so that we can access the "network" piece
+    if (request->keypath_count < 2) {
+        return APP_ETH_SIGN_ERR_INVALID_INPUT;
+    }
     // Only support main net for now. Otherwise a user could be tricked into signing something for
     // main net even if they believe they are signing for testnet.
```
benma: comment out of date.

> Otherwise a user could be tricked into signing something for
>     // main net even if they believe they are signing for testnet.

This concern is still valid; the coin name should be verified as well, maybe like when verifying the address in the first place (https://github.com/digitalbitbox/bitbox02-firmware/blob/594dd433b22d7db60b12d4b68c9cc977b1bee407/src/commander/commander_eth.c#L28-L30)

What about moving the `display` param from the commander `_api_pub` to `app_eth_address()`, then it comes for free?

src/apps/eth/eth_verify.c
```diff
@@ -115,23 +122,39 @@ app_eth_sign_error_t app_eth_verify_erc20_transaction(const ETHSignRequest* requ
...
+        app_eth_sign_error_t result = _verify_recipient(recipient, NULL);
+        if (result != APP_ETH_SIGN_OK) {
+            return result;
+        }
+        const _amount_t total = {
+            .unit = NULL,
+            .decimals = 0,
+            .value = &value_scalar,
+        };
+        result = _verify_total_fee(request, &total, params->unit);
```
benma: pass `NULL` instead of `&total`, as you do not need any of the fields, not even value

src/apps/eth/eth_sign_msg.c
NickeZ: so is it enough to call `rust_ethereum_keypath_is_address` in place of `_verify_network`?

src/apps/eth/eth_sign_msg.c
Tomasvrba: The problem with messages, unlike transactions, is we can't really verify the coin, because the coin will always match (on MEW specifically), since in the case of messages, the coin is derived from the keypath since we don't have the chainId like for a tx.
So until MEW also provides the device with the network, we can't verify this

NickeZ: @benma meant that we should show the network on the device. So the device will print "Ethereum", "Ropsten" or "Rinkeby" depending on which network we intend to sign for. Then the user can abort in case it is wrong.

src/apps/eth/eth.c
```diff
@@ -60,15 +69,54 @@ bool app_eth_address(
...
-            return false;
+            return APP_ETH_SIGN_ERR_INVALID_INPUT;
         }
         struct ext_key derived_xpub __attribute__((__cleanup__(keystore_zero_xkey))) = {0};
         if (!keystore_get_xpub(keypath, keypath_len, &derived_xpub)) {
-            return false;
+            return APP_ETH_SIGN_ERR_INVALID_INPUT;
+        }
+        if (!btc_common_encode_xpub(&derived_xpub, BTCPubRequest_XPubType_XPUB, out, out_len)) {
+            return APP_ETH_SIGN_ERR_INVALID_INPUT;
```
benma: this is more of unknown/generic error

src/apps/eth/eth.c
```diff
@@ -60,15 +69,54 @@ bool app_eth_address(
     }
```
benma: `return true` one line up looks wrong. possibly address verification does not work anymore because of the early abort here.

src/apps/eth/eth.h
```diff
@@ -31,13 +32,17 @@
  * @param[out] out will hold the result.
  * @param[in] out_len must be at least APP_ETH_ADDRESS_HEX_LEN for ADDRESS, and at least 80
  * (BIP32_SERIALIZED_LEN+1) for XPUB.
+ * @param[in] display Wether to ask user to confirm address
+ * @param[in] contract_address pointer to contract address
```
benma: "Must be 20 bytes. If none present, must be 20 zero bytes".

or have the commander do the 20 zero bytes check and pass in NULL in that case (would fit there as it's protobuf input parsing related).

benma:
@NickeZ Can you add two more items to send_message.py please? one to sign an erc20 tx, one to sign an *unknown* erc20 tx. for easy review and testing.

benma:
Cont. in https://github.com/digitalbitbox/bitbox02-firmware/pull/420

PR #423, opened 2020-03-24
hww.c: Turn OP_UNLOCK into an asynchronous operation.

conte91:
Make the OP_UNLOCK hww opcode start the unlock workflow asynchonously.
Support the retry/cancel flow during unlock.

### DIFF
--- a/src/hww.c
+++ b/src/hww.c
@@ -218,6 +218,10 @@ static void _process_packet(const in_buffer_t* in_req, hww_packet_rsp_t* out_rsp
                 out_rsp->buffer.data[0] = OP_STATUS_FAILURE_UNINITIALIZED;
                 out_rsp->buffer.len = 1;
                 return;
+            } else if (!keystore_is_locked()) {
+                out_rsp->buffer.data[0] = OP_STATUS_SUCCESS;
+                out_rsp->buffer.len = 1;
+                return;
             } else {
                 _state.unlock_wf = workflow_unlock(&_unlock_finished_cb, NULL);
                 workflow_stack_start_workflow(_state.unlock_wf);
--- a/src/hww.c
+++ b/src/hww.c
@@ -20,6 +20,7 @@
 #include <keystore.h>
 #include <memory/memory.h>
 #include <platform_config.h>
+#include <ui/workflow_stack.h>
 #include <usb/noise.h>
 #include <usb/usb_packet.h>
 #include <usb/usb_processing.h>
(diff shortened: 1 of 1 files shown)


src/hww.c
```diff
@@ -149,28 +168,62 @@ static size_t _api_info(uint8_t* buf)
...
         case OP_UNLOCK:
             if (!memory_is_initialized()) {
-                out_rsp->data[0] = OP_STATUS_FAILURE_UNINITIALIZED;
+                out_rsp->buffer.data[0] = OP_STATUS_FAILURE_UNINITIALIZED;
+                out_rsp->buffer.len = 1;
+                return;
             } else {
-                out_rsp->data[0] =
-                    workflow_unlock_blocking() ? OP_STATUS_SUCCESS : OP_STATUS_FAILURE;
+                _state.unlock_wf = workflow_unlock(&_unlock_finished_cb, NULL);
```
benma: workflow_unlock finishes immediately if the device is already unlocked. With the current retry logic, the unlock request is delayed by whatever retry-interval the client has, which is noticeable.

How to solve this?

src/hww.c
```diff
@@ -149,28 +168,62 @@ static size_t _api_info(uint8_t* buf)
...
+ */
+static void _unlock(void)
+{
+    _clear_state();
+    if (_state.blocking_op != HWW_BLOCKING_OP_NONE) {
+        usb_processing_unlock();
+    }
+}
+
+static void _unlock_finished_cb(bool result, void* param)
```
benma: is this ever called with `result=false` on cancel? not sure i understand how HWW_REQ_CANCEL leads to the proper (negative) response.

conte91: The `HWW_REQ_CANCEL` is handled by the separate arbitration logic, which workflows know nothing about. The only case the unlock can fail (`result == false`) is if `memory_is_initialized() == false` when the unlock workflow is started.

`HWW_REQ_CANCEL` is handled internally by `hww.c` (see `_process_msg`).

src/hww.c
conte91: I would just add a check in `hww.c` that avoids calling `workflow_unlock` entirely if the device is already unlocked. The only other case in which the workflow will return immediately is the path with `!memory_is_initialized()`, which is rare enough (IMHO) to make a 200ms irrelevant. (Also, if `OP_UNLOCK` is only sent during pairing, I don't think 200ms would make a difference there as well, but that's again IMHO).

src/hww.c
benma: you mean `_cancel_packet`. thx, i see now

src/hww.c
conte91: Fixed in e9a5715

benma:
Thanks for all the amazing async work and teaching. This PR is superseded by https://github.com/digitalbitbox/bitbox02-firmware/pull/442/, which implements almost the same, but in Rust.

PR #427, opened 2020-03-25
rust: use nightly-2020-03-25

benma:
This is to be able to use async/await syntax, which was merged a few
days ago into nightly for no_std:
https://github.com/rust-lang/rust/pull/69033.

This feature will be in beta in April and stable in June, at which
point we can start to use those channels again.

rust-toolchain file is added as a sanity check or for devs/users not
using Docker.


benma:
Travis running here: https://travis-ci.org/github/digitalbitbox/bitbox02-firmware/builds/666704674

Dockerfile
```diff
@@ -132,9 +132,11 @@ ENV PATH /opt/lcov-1.14/bin:$PATH
 # Install rust compiler
 ENV PATH /opt/cargo/bin:$PATH
 ENV RUSTUP_HOME=/opt/rustup
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | CARGO_HOME=/opt/cargo sh -s -- --default-toolchain 1.42.0 -y
+# Also update ./rust-toolchain when changing the default toolchain.
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | CARGO_HOME=/opt/cargo sh -s -- --default-toolchain nightly-2020-03-25 -y
```
x1ddos: The rust-toolchain is very helpful! Could we do something like this?

```
COPY rust-toolchain /tmp/
RUN ... | CARGO_HOME=/opt/cargo sh -s -- --default-toolchain $(cat /tmp/rust-toolchain) -y
```

So to attempt to keep them in sync.

NickeZ:
With this change we technically don't need the `RUSTC_BOOTSTRAP=1` any more. Any thoughts on removing that?

benma:
> With this change we technically don't need the `RUSTC_BOOTSTRAP=1` any more. Any thoughts on removing that?

I'd still keep it with the hope of going beta->stable in the next months. Let's see if we can resist the temptation of nightly :D

Dockerfile
benma: Can be fragile, e.g. if someone slips in a newline in there, it will break the docker build without anyone noticing.

x1ddos: Sure. Someone could also update Dockerfile while forgetting to update rust-toolchain or the other way around. :)
Either way it fine with me!

NickeZ:
I would prefer that you cat the rust-toolchain file before approving this PR as @x1ddos wrote.

benma:
> I would prefer that you cat the rust-toolchain file before approving this PR as @x1ddos wrote.

I'm not for it:

> Can be fragile, e.g. if someone slips in a newline in there, or deletes the file, it will break the docker build without anyone noticing

benma:
Instead of rust-toolchain, i could add an ENV var to the dockerfile and use it in `RUSTUP_TOOLCHAIN`? https://github.com/rust-lang/rustup#override-precedence

x1ddos:
> Can be fragile, e.g. if someone slips in a newline in there, or deletes the file, it will break the docker build without anyone noticing

TBH that's what code reviews and CIs are for :)

If `RUSTUP_TOOLCHAIN` works as an alternative, why not. I'd imagine something like:

```
docker build ... -e RUSTUP_TOOLCHAIN=$(cat rust-toolchain)
```

but it's equivalent to the other solution anyway.

x1ddos:
Ah sorry, forget about `docker build -e...`. I don't know what I was thinking.

x1ddos:
> > Can be fragile, e.g. if someone slips in a newline in there, or deletes the file, it will break the docker build without anyone noticing
> 
> TBH that's what code reviews and CIs are for :)

Plus, if someone submits a change to update `rust-toolchain`, I'd expect them to also run with it manually to verify it works.

NickeZ:
Solve it any way you want, this string: `nightly-2020-03-25` should only be in one place in the project.

benma:
Ok did the `cat` thing

PR #428, opened 2020-03-25
Remove async status workflow

benma:
The status workflow should always be blocking and never async. It is
not supposed to be ever cancellable (the user needs to see the
messages and they are relatively short anyway). This means that the
async machinery for this workflow is overkill and equivalent to a
simple blocking call.

### DIFF
--- a/test/unit-test/test_workflow_status.c
+++ b/test/unit-test/test_workflow_status.c
@@ -26,22 +26,12 @@
 
 const char* _msg = "message foo";
 
-component_t* __real_status_create(
-    const char* text,
-    bool status_success,
-    int delay,
-    void (*callback)(void*),
-    void* callback_param);
-component_t* __wrap_status_create(
-    const char* text,
-    bool status_success,
-    int delay,
-    void (*callback)(void*),
-    void* callback_param)
+component_t* __real_status_create(const char* text, bool status_success);
+component_t* __wrap_status_create(const char* text, bool status_success)
 {
     assert_string_equal(text, _msg);
     check_expected(status_success);
-    return __real_status_create(text, status_success, delay, callback, callback_param);
+    return __real_status_create(text, status_success);
 }
 
 static void _test_workflow_status(void** state)
--- a/test/unit-test/test_workflow_status.c
+++ b/test/unit-test/test_workflow_status.c
@@ -26,7 +26,12 @@
 
 const char* _msg = "message foo";
 
-static component_t _component;
+component_t* __real_status_create(
+    const char* text,
+    bool status_success,
+    int delay,
+    void (*callback)(void*),
+    void* callback_param);
 component_t* __wrap_status_create(
     const char* text,
     bool status_success,
@@ -34,11 +39,9 @@ component_t* __wrap_status_create(
     void (*callback)(void*),
     void* callback_param)
 {
-    (void)callback_param;
     assert_string_equal(text, _msg);
     check_expected(status_success);
-    mock_blocking_set_unblock_func(callback, callback_param);
-    return &_component;
+    return __real_status_create(text, status_success, delay, callback, callback_param);
 }
 
 static void _test_workflow_status(void** state)
@@ -47,10 +50,7 @@ static void _test_workflow_status(void** state)
         const bool status = flags & 1;
 
         expect_value(__wrap_status_create, status_success, status);
-        expect_value(__wrap_ui_screen_stack_push, component, &_component);
         workflow_status_blocking(_msg, status);
-        mock_screen_stack_assert_clean();
-        assert_true(mock_blocking_is_unblocked());
     }
 }
 
--- a/src/reset.c
+++ b/src/reset.c
@@ -23,26 +23,6 @@
 #ifndef TESTING
 #include "securechip/securechip.h"
 #include <driver_init.h>
-#include <hal_delay.h>
-#include <ui/components/status.h>
-#include <ui/ugui/ugui.h>
-#endif
-
-#if !defined(TESTING)
-/*
- * Shows a centered "Device reset" label.
- * Waits for 3000ms, then exit.
- */
-static void _show_reset_label(bool status)
-{
-    const char* msg = "Device reset";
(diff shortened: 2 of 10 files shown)


test/unit-test/test_workflow_status.c
```diff
@@ -26,19 +26,22 @@
 
 const char* _msg = "message foo";
 
-static component_t _component;
+component_t* __real_status_create(
+    const char* text,
+    bool status_success,
+    int delay,
+    void (*callback)(void*),
+    void* callback_param);
```
x1ddos: I'm a bit confused and surely missing somehing. Doesn't `status_create` take only 2 args now and shouldn't this be:

```c
component_t* __real_status_create(const char* text, bool status_success);
```
?

OTOH the CI tests pass. So, this must be working somehow?

x1ddos:
Also, you'll probably want to remove `--wrap=workflow_blocking_block,--wrap=workflow_blocking_unblock` from https://github.com/digitalbitbox/bitbox02-firmware/blob/a5aa077c9738ba54383f82e6d97120d6df24386d/test/unit-test/CMakeLists.txt#L234, right.

test/unit-test/test_workflow_status.c
benma: Wow, that's messed up. I don't know why this compiles. Not even the memory sanitizer complains.

benma:
Addresses both, thanks.

NickeZ:
hmm I'm not sure i agree with the premise of this,. when we go async no code must ever block. does this change mean that you block the usb stack for several seconds?

benma:
Yes, discussed with @conte91. Reasons include:

- cannot ever be cancelled
- cannot have more than one workflows going at the same time
- then there is no reason to be async here.

What could be the downsides? Something about U2F maybe?

test/unit-test/test_workflow_status.c
x1ddos: Right. This wasn't even supposed to compile. Maybe I'll try to figure it out in my local copy.

x1ddos:
> does this change mean that you block the usb stack for several seconds?

I don't know yet how USB stack is designed here but I'd normally expect USB frames to be handled via hardware interrupts which always take priority over software. In other words, isn't `workflow_status_blocking` here means only blocking other software functions, i.e. the regular execution, and hardware comms like USB continue to operate?

NickeZ:
> Yes, discussed with @conte91. Reasons include:
> 
>     * cannot ever be cancelled
>     * cannot have more than one workflow going at the same time
>     * then there is no reason to be async here.
> 
> What could be the downsides? Something about U2F maybe?
> 
> edit: fyi happy to close it, but would be good to understand.

Yeah, it is incompatible with U2F. So we have to avoid calling this function in any u2f handling code. Which I guess is the case today and probably won't change.

The downside is that it is architecturally unsound. This is probably not a problem today, but in the future one could imagine that you would like to be able to handle other sources of asynchronous events such as timers (lock the device after 5 minutes) and maybe user input.

What you are doing is blocking the main loop. if you did this in an android app the operating system will open a popup asking if you want to kill the whole thing as it is unresponsive.

So practically, it probably doesn't matter today. But we should strive for a code base that doesn't use `delay_*` at all, since blocking behavior is incompatible with an event loop.

> > does this change mean that you block the usb stack for several seconds?
> 
> I don't know yet how USB stack is designed here but I'd normally expect USB frames to be handled via hardware interrupts which always take priority over software. In other words, isn't `workflow_status_blocking` here means only blocking other software functions, i.e. the regular execution, and hardware comms like USB continue to operate?

Yeah, an interrupt will trigger and the buffer is read out, a sequence of frames is parsed into a packet. All in interrupt space. But then the main loop takes over and actually does something with the packet.

This PR will introduce a 3 second "sleep" in the main loop, therefore not allowing anything else to happen.

benma:
Nice explanation, thanks.

> Which I guess is the case today and probably won't change.

If you enter the unlock workflow in u2f and enter your password incorrectly, the status screen is invoked, so it is in fact used.

x1ddos:
I still would like to know how Travis managed to compile and run tests with
success!

benma:
> I still would like to know how Travis managed to compile and run tests with success!

yeah, C is f*ked :joy: 

I am assuming that the compiler did not catch it because we do not call the function directly in the test. All the real calls from the production code have the correct number of arguments. 

The declaration of the `__real_` function had too many args, but was called with that number of args, so it's ok for the compiler.

The wrapped functions are inserted by the linker, which is too late for compilation errors.

Now you can figure out if this leads to a memory bug or if it's a harmless mistake ;)

conte91:
My 2p on this. The original idea came out of two points:

>     * cannot ever be cancelled
ATM we only use `workflow_status` to implement small nudges at the end of a workflow (Mostly, for error messages). So in that moment all USB traffic necessary to any operation should be finished.

>     * cannot have more than one workflow going at the same time
> 
This is improperly worded: we cannot have stacked _blocking_ workflows (i.e. a blocking workflow cannot call another blocking workflow as part of its operation). This is because the blocking workflow API is stupid and uses a global `_is_done` parameter. So this is more of a workaround for it - if something blocks on an async workflow and the async workflow spawns `workflow_status()`, the BB02 won't crash. On the other hand, a simple and proper fix would be to have separate `_is_done` flags for each `workflow_blocking_block()`, so that one can have nested blocks.

IMHO we should try to use async everything when possible, but having a "common" implementation of a blocking status window can be useful, instead of manually implementing it as needed.
 
> What could be the downsides? Something about U2F maybe?

A U2F request coming just after another one has finished (so while the U2F nudge is displaying a status) will be dropped/replied to with some delay, which breaks Windows. Depending on the point of view, this could be desireable: one shouldn't start two back-to-back U2F requests. If another request is started too quickly, it should be reopened manually. But this could also end up being very confused to the user.

benma:
The general idea that everything should be async, and exceptions to this could have weird consequences (already has for U2F, there could be other manifestations in the future), sounds right. So I'll close this for the time being. Thanks all.

PR #429, opened 2020-03-25
Async enter password and confirm

benma:
this adds
```
    rust_workflow_async_DEMO();
    screen_print_debug("DONE", 5000);
```

to `firmware_main_loop()` as a demo, which i will remove before merging.

The rest is implementing entering the password as a rust future, based on the trinary input string component. The C callback in the component is converted to a Rust future, which can be polled and used nicely, as seen in `rust_workflow_async_DEMO`.

### DIFF
--- a/src/rust/bitbox02/src/password.rs
+++ b/src/rust/bitbox02/src/password.rs
@@ -1,0 +1,50 @@
+// Copyright 2020 Shift Cryptosecurity AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// C-style including null terminator, as it is used in C only so far.
+/// 150 corresponds to SET_PASSWORD_MAX_PASSWORD_LENGTH.
+/// Does *not* implement Copy, so that we can have a Drop to zero the contents.
+// TODO: use a reusable zero-on-drop buffer type
+pub struct Password([u8; 150]);
+
+impl Password {
+    /// Makes a password buffer filled with 0.
+    pub fn new() -> Password {
+        Password([0; 150])
+    }
+
+    /// Clones the password bytes from `source` without additional allocations.
+    pub fn clone_from(&mut self, source: &Self) {
+        self.0.copy_from_slice(&source.0[..]);
+    }
+}
+
+impl AsRef<[u8; 150]> for Password {
+    fn as_ref(&self) -> &[u8; 150] {
+        &self.0
+    }
+}
+
+impl AsMut<[u8; 150]> for Password {
+    fn as_mut(&mut self) -> &mut [u8; 150] {
+        &mut self.0
+    }
+}
+
+impl Drop for Password {
+    fn drop(&mut self) {
+        util::zero(&mut self.0[..]);
+        crate::screen_print_debug(".", 100);
+    }
+}
--- a/src/rust/bitbox02/src/ui.rs
+++ b/src/rust/bitbox02/src/ui.rs
@@ -1,0 +1,73 @@
+// Copyright 2020 Shift Cryptosecurity AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use util::c_types::{c_char, c_void};
+
+extern crate alloc;
+use crate::password::Password;
+use alloc::boxed::Box;
+use core::pin::Pin;
+
+/// Wraps the C component_t to be used in Rust.
+pub struct Component(*mut bitbox02_sys::component_t);
+
+/// Creates a password input component.
+/// `title` - Shown before any input is entered as the screen title. **Panics** if more than 100 bytes.
+/// `special_chars` - whether to enable the special characters keyboard.
+/// `result` - will be asynchronously set to `Some(<password>)` once the user confirms.
+pub fn trinary_input_string_create_password(
+    title: &str,
+    special_chars: bool,
+    result: Pin<&mut Option<Password>>,
+) -> Component {
+    extern "C" fn on_done_cb(password: *const c_char, param: *mut c_void) {
+        let mut out: Box<Pin<&mut Option<Password>>> = unsafe { Box::from_raw(param as *mut _) };
+        let mut password_out = Password::new();
+        let len = password_out.as_ref().len();
+        password_out
+            .as_mut()
+            .copy_from_slice(unsafe { core::slice::from_raw_parts(password, len) });
+        out.set(Some(password_out));
+    }
+
+    let component = unsafe {
+        bitbox02_sys::trinary_input_string_create_password(
+            crate::str_to_cstr_force!(title, 100).as_ptr(),
+            special_chars,
+            Some(on_done_cb),
+            Box::into_raw(Box::new(result)) as *mut _, // passed to on_done_cb as `param`.
+            None,
+            core::ptr::null_mut(),
+        )
+    };
+    Component(component)
+}
+
+pub fn screen_stack_push(component: &mut Component) {
+    unsafe {
+        bitbox02_sys::ui_screen_stack_push(component.0);
+    }
+}
+
+pub fn screen_stack_pop() {
+    unsafe {
+        bitbox02_sys::ui_screen_stack_pop();
+    }
+}
+
+pub fn screen_process() {
+    unsafe {
+        bitbox02_sys::screen_process();
+    }
+}
--- a/src/rust/bitbox02/src/ui.rs
+++ b/src/rust/bitbox02/src/ui.rs
@@ -54,6 +54,68 @@ pub fn trinary_input_string_create_password(
     Component(component)
 }
 
+#[derive(Default)]
+pub struct ConfirmParams<'a> {
+    /// The confirmation title of the screen. Max 200 chars, otherwise **panic**.
+    pub title: &'a str,
+    /// The confirmation body of the screen. Max 200 chars, otherwise **panic**.
+    pub body: &'a str,
+    /// If true, the body is horizontally scrollable.
+    pub scrollable: bool,
+    /// If true, require the hold gesture to confirm instead of tap.
+    pub longtouch: bool,
+    /// If true, the user can only confirm, not reject.
+    pub accept_only: bool,
+    /// if true, the accept icon is a right arrow instead of a checkmark (indicating going to the
+    /// "next" screen).
+    pub accept_is_nextarrow: bool,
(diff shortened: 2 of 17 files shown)


benma:
@NickeZ 

the many thousand files are just one transitive dep from here, that is not even used:

https://github.com/async-rs/async-task/blob/master/Cargo.toml#L19

Any way to turn that off and not pull it in?

NickeZ:
fork `asyn-task` and make it a `git` dependency instead. I hope that works with `cargo vendor`.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

benma:
> fork `asyn-task` and make it a `git` dependency instead. I hope that works with `cargo vendor`.
> 
> https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

should i add it as a submodule?

benma:
Added async confirmation dialog as well, in the same style. Demo enhanced to show usage:

```Rust
        let params = Params {
            title: "demo",
            body: "Proceed to pw entry?",
            ..Default::default()
        };

        if !confirm(&params).await {
            return;
        }
```

src/rust/bitbox02/src/password.rs
```diff
@@ -0,0 +1,50 @@
...
+pub struct Password([u8; 150]);
+
+impl Password {
+    /// Makes a password buffer filled with 0.
+    pub fn new() -> Password {
+        Password([0; 150])
+    }
+
+    /// Clones the password bytes from `source` without additional allocations.
+    pub fn clone_from(&mut self, source: &Self) {
```
NickeZ: you could probably just `#[derive(Clone)]` on the struct instead.

src/rust/bitbox02/src/password.rs
```diff
@@ -0,0 +1,50 @@
...
+impl AsMut<[u8; 150]> for Password {
+    fn as_mut(&mut self) -> &mut [u8; 150] {
+        &mut self.0
+    }
+}
+
+impl Drop for Password {
+    fn drop(&mut self) {
+        util::zero(&mut self.0[..]);
+        crate::screen_print_debug(".", 100);
```
NickeZ: old debug code?

src/rust/bitbox02/src/ui.rs
```diff
@@ -54,6 +54,68 @@ pub fn trinary_input_string_create_password(
...
+    pub accept_is_nextarrow: bool,
+    /// if true, will only show first and last 32 bytes.
+    pub shorten_body: bool,
+    /// Print the value of this variable in the corner. Will not print when 0
+    pub display_size: usize,
+}
+
+/// Creates a user confirmation dialog screen.
+/// `result` - will be asynchronously set to `Some(bool)` once the user accets or rejects.
+pub fn confirm_create(params: &ConfirmParams, result: Pin<&mut Option<bool>>) -> Component {
```
NickeZ: It feels like you shouldn't need `Pin` here. Why are you using that?

src/rust/bitbox02/src/password.rs
benma: Tried it, the normal clone_from makes an extra allocation though. Maybe not a big deal but also better to have less passwords being copied around.

src/rust/bitbox02/src/ui.rs
benma: The pointer is sent into C land, so it has to stay constant. With some permutation of the code I had hardfaults without pinning.

NickeZ: sounds like a red herring. in case you put it on the heap with box the position should stay the same. if you move the box only the pointer is moved. not the content

benma: Note that confirm_create does not say that the result has to a Box, just a Pin - the caller could have a different pin (I wanted to try stack-pinned).

Ignoring that:

Is there a guarantee that heap objects can't be moved? As far as I can tell there isn't, which is why Box::pin exists. Am I misunderstanding it?

Reading the doc for Box::pin:

> Constructs a new Pin<Box<T>>. If T does not implement Unpin, then x will be pinned in memory and unable to be moved.

Sounds like I also have to somehow make sure Unpin is not implemented? :thinking:

x1ddos:
Would it make sense to merge this in a separate branch, not master, and
run/use a couple bb02 devices for say a month with this?

benma:
> Would it make sense to merge this in a separate branch, not master, and
run/use a couple bb02 devices for say a month with this?

The async functions are there but they are not actually used yet for the real thing. I have to implement a few more async screens and then I'll probably be able to use it for the real thing too. 

Sounds like a good idea to have it in a staging branch for a bit once it's used. Do you already want to put the (unused) functions in this PR in a staging branch as well?

x1ddos:
Yeah, I get that this new rust code is unused but that's why it might make
maintenance of the real thing harder when everything is in one branch imho.

I would merge this pull request into a staging as is, i.e. without
additional code. then keep improving and adding in separate pull requests
to the same staging branch until it's actually possible to build and run
firmware with this new rust async code as the real thing.

then, flash to real devices and use it ourselves for few weeks. when
n confident enough, merge into master.

makes sense?

src/rust/bitbox02/src/ui.rs
benma: Possibly relevant:

https://www.reddit.com/r/rust/comments/b9sydt/trying_to_understand_how_rust_juggles_memoryraw/ek6nu89/

Still trying to wrap my head around it though.

NickeZ: Pinning is meant for types that contains references to itself, e.g.:

```rust
struct A {
    bu:f [u8;150];
    writer_pos: &'self [u8]
}
```

I've made up the `'self` lifetime. Implementing this type requires that the type doesn't move ever. And if you have a pinned pointer to it, there is a guarantee that it didn't move between operations. You can always create a pinned variable, but you can only unpin types that are "Unpin" because they don't have any of these pesky self-references.

NickeZ: I think the problems you are experiencing is that the type gets dropped while you are still in C-land. I'm not sure exactly why though. It looks like you are doing `into_raw()` to consume the box, and then `from_raw` to construct it again.

benma: What about this:

```Rust
use std::marker::PhantomPinned;

#[derive(Debug)]
struct NotUnpinnable(Option<i32>, PhantomPinned);

fn main() {
    // This program compiles and is wrong. Changing Box::new to Box::pin fixes it, as it does not
    // let you compile it.
    let mut a = Box::new(NotUnpinnable(Some(1), PhantomPinned));
    let mut b = Box::new(NotUnpinnable(Some(2), PhantomPinned));

    let pointer_to_a: *mut NotUnpinnable = &mut *a;

    let mut c = a; // move a
    // A dangerous operation prevented by pinning
    std::mem::swap(&mut c, &mut b);

    unsafe {
        // would happen in C.
        *pointer_to_a = NotUnpinnable(None, PhantomPinned);
    }

    // Expect `None`, but contains `Some(2)` due to mem::swap.
    println!("{:#?}", c);
}
```

NickeZ:
wouldn't you have much less C-rust interop if you first ported the `confirm_created(...)`?

src/rust/bitbox02/src/ui.rs
benma: > `confirm_created`

I don't want to reimplement each component in Rust now. The goal is async workflows, not porting/rewriting everything that it depends on. trinary input string for example would be a lot harder to port.

benma:
> wouldn't you have much less C-rust interop if you first ported the `confirm_created(...)`?

> `confirm_created`

I don't want to reimplement each component in Rust now. The goal is async workflows, not porting/rewriting everything that it depends on. trinary input string for example would be a lot harder to port.

src/rust/bitbox02/src/ui.rs
benma: Pretty sure that without unpin, the pointer can become dangling when the value is accidentally moved, and proper pinning prevents that. Self-referential structs are not the only use case.

Could be wrong though, kind of hard to wade through this :)

NickeZ:
The demo worked for me after this changes:

```diff
diff --git a/src/rust/bitbox02-rust/src/workflow/confirm.rs b/src/rust/bitbox02-rust/src/workflow/confirm.rs
index a0eda04..74cb478 100644
--- a/src/rust/bitbox02-rust/src/workflow/confirm.rs
+++ b/src/rust/bitbox02-rust/src/workflow/confirm.rs
@@ -21,10 +21,11 @@ pub use bitbox02::ui::ConfirmParams as Params;
 
 /// Returns true if the user accepts, false if the user rejects.
 pub async fn confirm(params: &Params<'_>) -> bool {
-    let mut result: Pin<Box<Option<bool>>> = Box::pin(None);
+    //let mut result: Pin<Box<Option<bool>>> = Box::pin(None);
+    let mut result: Option<bool> = None;
 
     // The component will set the result when the user accepted/rejected.
-    let mut component = bitbox02::ui::confirm_create(&params, result.as_mut());
+    let mut component = bitbox02::ui::confirm_create(&params, &mut result);
 
     bitbox02::ui::screen_stack_push(&mut component);
     option(&result).await;
diff --git a/src/rust/bitbox02/src/ui.rs b/src/rust/bitbox02/src/ui.rs
index 2de5a8f..e46b1d8 100644
--- a/src/rust/bitbox02/src/ui.rs
+++ b/src/rust/bitbox02/src/ui.rs
@@ -77,7 +77,7 @@ pub struct ConfirmParams<'a> {
 
 /// Creates a user confirmation dialog screen.
 /// `result` - will be asynchronously set to `Some(bool)` once the user accets or rejects.
-pub fn confirm_create(params: &ConfirmParams, result: Pin<&mut Option<bool>>) -> Component {
+pub fn confirm_create(params: &ConfirmParams, result: &mut Option<bool>) -> Component {
     let params = bitbox02_sys::confirm_params_t {
         title: crate::str_to_cstr_force!(params.title, 200).as_ptr(),
         body: crate::str_to_cstr_force!(params.body, 200).as_ptr(),
@@ -90,16 +90,20 @@ pub fn confirm_create(params: &ConfirmParams, result: Pin<&mut Option<bool>>) ->
         display_size: params.display_size as u32,
     };
 
-    extern "C" fn on_accept_cb(param: *mut c_void) {
-        let mut out: Box<Pin<&mut Option<bool>>> = unsafe { Box::from_raw(param as *mut _) };
-        out.set(Some(true));
+    unsafe extern "C" fn on_accept_cb(param: *mut c_void) {
+        let param = (param as *mut Option<bool>).as_mut().expect("null ptr");
+        *param = Some(true);
+        //let mut out: Box<Pin<&mut Option<bool>>> = unsafe { Box::from_raw(param as *mut _) };
+        //out.set(Some(true));
     }
-    extern "C" fn on_reject_cb(param: *mut c_void) {
-        let mut out: Box<Pin<&mut Option<bool>>> = unsafe { Box::from_raw(param as *mut _) };
-        out.set(Some(false));
+    unsafe extern "C" fn on_reject_cb(param: *mut c_void) {
+        //let mut out: Box<Pin<&mut Option<bool>>> = unsafe { Box::from_raw(param as *mut _) };
+        //out.set(Some(false));
+        let param = (param as *mut Option<bool>).as_mut().expect("null ptr");
+        *param = Some(false);
     }
 
-    let result_ptr = Box::into_raw(Box::new(result)) as *mut _; // passed to callbacks `param`.
+    let result_ptr = result as *mut _ as *mut c_void; // passed to callbacks `param`.
     let component = unsafe {
         bitbox02_sys::confirm_create(
             &params,

```

src/rust/bitbox02/src/ui.rs
benma: after long offline discussions, tldr: Pin+!Unpin can prevent data moves, but it can't prevent a premature `drop()` of the result before it is written to asynchronously by C.

Pin is also not needed when using Box::into_raw/from_raw to pass ownership to C. 

As a result, we have to move the workflow/* code to the unsafe component creation, where we have control over everything and can promise the result is not moved/dropped.

benma:
Closing this in favor of #433, which is a PR against a new staging branch.

PR #437, opened 2020-03-30
Fido2 - conte91

conte91:
Anything else to say?


src/CMakeLists.txt
```diff
@@ -211,8 +213,14 @@ set(FIRMWARE-DRIVER-SOURCES
 set(FIRMWARE-DRIVER-SOURCES ${FIRMWARE-DRIVER-SOURCES} PARENT_SCOPE)
 
 set(FIRMWARE-U2F-SOURCES
+  ${CMAKE_SOURCE_DIR}/src/fido2/ctap.c
```
benma, line 216: did you modify those files in any way?

would be great to move them to external/ instead and compile them as a lib, like `ctaes` for example.

src/fido2/extensions.c
```diff
@@ -0,0 +1,175 @@
...
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+#include <stdint.h>
+#include <string.h>
+#include "extensions.h"
+#if 0
```
x1ddos, line 11: these `#if 0` look weird... I wish we had a set of .patch files and could simply apply them to the upstream when updating.

src/fido2/fido2_keys.c
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Automatically generated with the following command:
+ * ['./scripts/create_fido2_attestation_cert']
+ */
+
+const uint8_t FIDO2_ATT_PRIV_KEY[] = {
+        0xd9, 0x6e, 0x5e, 0x56, 0x34, 0x15, 0x98, 0xa6, 0x98, 0xd6,
+        0xd4, 0x9e, 0x5d, 0x15, 0x4f, 0xef, 0x04, 0xe3, 0x45, 0x5c,
+        0x19, 0x51, 0xd6, 0x31, 0x79, 0x00, 0x83, 0x20, 0x52, 0xae,
+        0x03, 0x71
```
x1ddos, line 12: Is this supposed to actually be checked in?

src/memory/memory.c
```diff
@@ -810,3 +832,42 @@ bool memory_multisig_get_by_hash(const uint8_t* hash, char* name_out)
...
+{
+    return CHUNK_3 + key_idx / MEMORY_CTAP_RESIDENT_KEYS_PER_CHUNK;
+}
+
+bool memory_get_ctap_resident_key(int key_idx, ctap_resident_key_t* key_out)
+{
+    fido2_resident_key_chunk_t chunk = {0};
+    CLEANUP_CHUNK(chunk);
+    _read_chunk(_ctap_resident_key_chunk(key_idx), chunk.bytes);
+    /** TODO: simo: implement */
```
x1ddos, line 846: implement what? :)

src/memory/memory.c
```diff
@@ -810,3 +832,42 @@ bool memory_multisig_get_by_hash(const uint8_t* hash, char* name_out)
     }
     return false;
 }
+
+static int _ctap_resident_key_chunk(int key_idx)
+{
+    return CHUNK_3 + key_idx / MEMORY_CTAP_RESIDENT_KEYS_PER_CHUNK;
```
x1ddos, line 838: What if `key_idx` is `46`? Then I suppose `_read_chunk` below tries reading from an invalid chunk 5. Haven't check but can `_read_chunk` handle that?

src/u2f.c
```diff
@@ -76,6 +82,14 @@ typedef struct {
...
      */
     uint8_t last_cmd;
+    /**
+     * last_cmd points to a valid byte,
+     * and incoming requests that match that
+     * command should be allowed. This is true
+     * for blocking U2F requests, but false for CTAP
+     * requests (as CTAP requests are not sent multiple times).
+     */
+    bool allow_cmd_retries;
```
x1ddos, line 92: In other words, true == U2F mode and false is CTAP?

src/usb/usb.c
```diff
@@ -25,13 +25,14 @@
...
+#include <driver_init.h>
 #include <hal_timer.h>
+#include <u2f.h>
 #include <u2f/u2f_packet.h>
 #include <usb/usb_packet.h>
-extern struct timer_descriptor TIMER_0;
 #endif
 
-#define TIMEOUT_TICK_PERIOD_MS 100
+#define TIMEOUT_TICK_PERIOD_MS 50
```
x1ddos, line 35: I wonder how this can impact other non-U2F operations.

benma: Looks like the value might reused here but without using a common definition?

https://github.com/digitalbitbox/bitbox02-firmware/blob/master/src/usb/usb_processing.c#L44

So that might need to be updated too.

benma: And this comment: https://github.com/digitalbitbox/bitbox02-firmware/blob/master/src/usb/usb_processing.c#L119

x1ddos:
> @conte91 should I take it over or you're still on this?

Going to take this over, cherry-picking with updates and rebasing, and sending PRs to [fido2](https://github.com/digitalbitbox/bitbox02-firmware/tree/fido2) branch. When all is done, will merge fido2 with master.

Let's keep this PR open for now, as a reference.

benma:
In Chromium on Linux, cancelling the u2f (not fido2) registration process during the password unlock does not cancel the unlock workflow. It still works on Firefox on Linux. Haven't tested other combinations.

This works on master on both Chromium and Firefox on Linux, so it is a regression of this PR.

benma:
I rebased this PR on current master (2a8c435d) here:

https://github.com/benma/bitbox02-firmware/tree/fido2

All comments of this PR still apply, I haven't changed any semantics (not deliberately, anyway).

PR #439, opened 2020-04-01
components/confirm: one callback instead of two

benma:
Makes most call sites simpler.


NickeZ:
This also changes the behavior of the dismiss screen. Please write a comment about that in the commit message.

benma:
> This also changes the behavior of the dismiss screen. Please write a comment about that in the commit message.

It doesn't actually. It behaves exactly the same way.

NickeZ:
right, now I see it, you added `accept_only` where it used to be `NULL`/`NULL`.

PR #441, opened 2020-04-04
hww: move actual api processing to new file

benma:
Separate the files a bitm so hww.c is doing arbitration (HWW_REQ_*
level) and hww_api.c is doing the actual api processing.


benma:
This will make it easier to turn the message processing in hww.c async (in Rust). 

Rust being able to call hww_api.c makes it possible to have the message processor async and migrate the api step by step (e.g. first unlock, then attestation, then noise, then protobuf).

PR #442, opened 2020-04-04
Rust async hww

benma:


benma:
Timeout on unlock now also works nicely: launch send_message.py, Ctrl-C, and the usb task will kill itself.

src/rust/bitbox02-rust-c/src/async_usb.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/// Polls for a result of an async usb task. If a result is available, it is copied to `out`.
+///
+/// Returns:
+/// `UsbResponseNack` if on ask is running.
+/// `UsbResponseAck` if the result was copied.
+/// `UsbResponseNotReady` if a task is running but not yet complete.
+#[no_mangle]
+pub unsafe extern "C" fn rust_async_usb_copy_response(out: *mut bitbox02::buffer_t) -> UsbResponse {
```
x1ddos, line 34: Kind of weird to see `async` everywhere in the function names which don't return futures. Is there a `rust_*sync*_usb_copy_response`? If not, maybe simply name it `rust_usb_copy_response` here and similarily the others?

benma: the naming is just namespacing, as the `copy_response` is in the `rust` `async_usb` module.

x1ddos: That's exactly my point. Why not cut the `async_` prefix? Then it would become `rust_usb_xxx`, right. In a future pull request, anyway.

src/firmware_main_loop.c
```diff
@@ -39,6 +40,9 @@ void firmware_main_loop(void)
             Abort("NULL workflow spin in main");
         }
         workflow->spin(workflow);
+
+        rust_async_usb_spin();
+
         if (usb_is_enabled()) {
```
x1ddos, line 46: Also a bit confusing to see `rust_async_usb_spin` regardless of `usb_is_enabled`. As if it were a typo.

benma: I see. I guess it could be in that if branch, as indeed no usb ask is spawned without usb :)

src/rust/bitbox02-rust-c/src/async_usb.rs
benma: then also rename the file to `usb.rs` here and in the `bitbox02_rust` crate? I just put the async prefix there because it's a file dealing with task spawning, spinning, cancelling - all async things.

src/firmware_main_loop.c
benma: Fixed.

src/hww.c
```diff
@@ -105,6 +105,63 @@ static size_t _api_info(uint8_t* buf)
...
+        response->status = HWW_RSP_NACK;
+        break;
+    }
+}
+
+/**
+ * Executes the HWW packet.
+ * @param[in] in_packet The incoming HWW packet.
+ * @param[in] out_packet The outgoing HWW packet.
+ * @param[in] max_out_len The maximum number of bytes that the outgoing HWW packet can hold.
```
NickeZ, line 134: fix documentation

src/rust/bitbox02-rust/src/async_usb.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// The task has finished and written the result, so the USB response is available. We are now
+    /// waiting for the client to fetch it (HWW_REQ_RETRY). For short-circuited or non-async api
+    /// calls, the result might be returned immediately in response to HWW_REQ_NEW.
+    ResultAvailable(UsbOut),
+}
+
+/// Executor main state. Currently we only have at most one task at a time (usb api processing
+/// task).
+///
+/// It is `mut` because without it, Rust requires thread safety via
```
NickeZ, line 47: Whats the drawback to `unsafe impl Send for RefCell<UsbTaskState>`? Should be safe because we are single threaded.

benma: It seems to be both prefixing `mut` as well as marking with with `Sync` (this is enough for immutable) are workarounds, not sure which is better.

Sidenote: can't impl foreign impl on foreign trait; it would work if I wrapped it.

`Sync` marks it as thread safe, which it is not. Of course we are single threaded, but it feels bad to mark it as thread safe. 

The whole thing actually also works without a RefCell:

`static mut USB_TASK_STATE: UsbTaskState = UsbTaskState::Nothing;`

The RefCell only is a small dev help so that it's not borrowed mutably twice (e.g. by accidentally calling spin() inside of spawn()  etc.), but in spirit the var is mutable, so I didn't mind the `mut` in front.

Maybe `Sync` is a safer workaround, because then the global cannot actually be mutated (sidestepping RefCell)?

benma: Pushed a commit to use unsafe `Sync`, it does look better -> it reduces the `unsafe` into one place and removes the danger of mutating the state without RefCell.

PR #447, opened 2020-04-06
test/random: test random_32_bytes

benma:


test/unit-test/test_random.c
```diff
@@ -38,7 +38,20 @@ int __wrap_wally_sha256(
     return 0;
 }
 
-static void test_random(void** state)
+static void _test_random_32_bytes_mcu(void** state)
+{
+    uint8_t expected[RANDOM_NUM_SIZE] = {0};
+    uint8_t buf[RANDOM_NUM_SIZE] = "12345678901234567890123456789012";
```
x1ddos, line 44: what would happen if this were, say `buf[1] = "1"`?

x1ddos: Sorry, I mean would it make sense to check the buf's length somewhere in https://github.com/digitalbitbox/bitbox02-firmware/blob/0705df2d704f8133b9b60fca6451c3bdd47fd1c9/src/random.c#L48 ?

benma: there are kind of two styles when dealing with lengths:

either a function also takes `size_t buf_len` as an argument, supplid with `sizeof(buf)`, and returns an error (or aborts) if it is not the expected len, e.g. 32 in this case.

or the function just documents the min length of the buffer (in this case RANDOM_NUM_SIZE and function name/doc).

in our codebase we mostly do the latter, which is probably not the best decision.

x1ddos: in this case it's even more specific imho. the fuction already says what it expects in the name - 32 bytes. so, probably just making sure `len >= 32` in random.c would be sufficient.

you always say: what  if this happens, what if that happens, what it a programmer forgets this or that... :)

benma: yeah by now i also think passing the len and checking it makes sense, but currently random.c does not have the len.

changing it is out of scope imho; there are hundreds of functions that could be changed in the same way. can do it in a separate PR if desired, but low prio for me.

x1ddos: yeah, it's for the future. with rust slowing taking over though, I think it won't be as critical because `out: &mut [u8; 32]`.

PR #448, opened 2020-04-06
rust/bitbox02: wrap random

benma:


benma:
@NickeZ 

any idea why in the unit tests this happens?

`undefined reference to `random_32_bytes_mcu'`

Can't figure out what's special about random.c

benma:
I see! I guess mocking them makes sense.

benma:
@NickeZ i added a new version where i have an implementation for each platform. seems easiest, as i actually do not need to mock this function in tests.

Chose to use  `cfg(not(test))` and `cfg(test)` over the target arch or custom features for now because it's pretty vague what the implications would be, and test vs. not test is the most clear cut right now. Can deal with making the attributes more correct if we ever support other targets for non-test code or can run test code in a bitbox02 emulator.

x1ddos:
Hm... why is `wasi` crate here?

x1ddos:
> Hm... why is `wasi` crate here?

Oh, of course `getrandom` depends on it and pulls in the whole world? Insane.

x1ddos:
OMG... std, libc and everything. This is only for testing, right. Could we simply return `42` instead to avoid 1000s of extra lines and crate deps?

benma:
> OMG... std, libc and everything. This is only for testing, right. Could we simply return `42` instead to avoid 1000s of extra lines and crate deps?

Most tests should probably have not a constant for randomness, e.g. when playing through a noise communication.

Fyi, the deps are actually behind features that we do not enable, so under normal circumstances they would not be pulled and used. But since we vendor using `cargo vendor`, I think that defaults to just getting everything, and then only using what's needed when compiling depending on the features.

Not sure if that can be improved.

x1ddos:
Right. I get that. Just complaining about the amount of dead files it adds to the repo.

> Most tests should probably have not a constant for randomness, e.g. when playing through a noise communication.

Is it actually the case. I mean, what's "playing through a noise communication" in rust tests mean?

Alternatively, just copy&paste the actual 10 lines of `getrandom` func in the test?

benma:
> Is it actually the case. I mean, what's "playing through a noise communication" in rust tests mean?

When porting noise to rust, it will create secret keys with this function. In a unit test I want to exercise this with a unit test that is the noise counterparty.

Could also look into proper mocking by making a Trait instead, but that is kinda over the top.

> Alternatively, just copy&paste the actual 10 lines of getrandom func in the test?

Where are those lines? The lib is huge :P

x1ddos:
> Where are those lines? The lib is huge
I mean this file:
https://github.com/rust-random/getrandom/blob/master/src/linux_android.rs
and skip the test when running on windows or macos.

Another alternative, actually the original question: why not link in C function?

benma:
> Another alternative, actually the original question: why not link in C function?

Gives incentive to move away from C :D 

It's cleaner to reboot the testing in Rust than to hang on to `#ifdef TESTING` stuff from C.

I'll check if I can add the linked code and make it work, should be possible somehow.

Or i'll reconsider making it a trait and mock in in the tests anyway. Too many options to choose from :/

benma:
@x1ddos there, no dep and only 3 lines added ;)

NickeZ:
It's all dead code ;P maybe you should port some rust code to use this new wrapped random function? or is that coming later?

benma:
> It's all dead code ;P maybe you should port some rust code to use this new wrapped random function? or is that coming later?

yeah coming real soon, just wanted approval on how to deal with this C shit :P

benma:
@x1ddos had to change the attribute check, `cfg(test)` does not work because it's only enabled for the current crate! so running a test in crateX calling random.rs will still try to call the non-test random function because `test` is only active in crateX, not in random.rs..

Using target_arch for now like in `c_types.rs` - mabye not ideal, but good enough for now.

x1ddos:
Hm.. interesting. Yeah, looks clean anyway. The best next option, I like it!

@NickeZ do you know what's up with `cfg(test)` and how to enable it across the whole build?

NickeZ:
I can't really come up with a good way to do testing with mocking which bothers me.

Right now I'm thinking that the crate `bitbox02` simply should have two modules. The "normal one" and the mocked one. A bit like what this PR introduces but at the top level.

See for example here: https://github.com/digitalbitbox/bitbox-bridge/blob/8a7106afcc8dc038e106172ace595306342b6886/bitbox-bridge/src/main.rs#L23

Re. dependencies not compiled with "test" set. That is a feature... Not sure if it is the best decision, but this is the way it is.

NickeZ:
Maybe the correct thing is actually to have a c-library that is a mock of the normal stuff and link that in. A bit like unit tests are done in the C unit tests.

x1ddos:
> Re. dependencies not compiled with "test" set. That is a feature... Not sure if it is the best decision, but this is the way it is.

Oh, right: https://github.com/digitalbitbox/bitbox02-firmware/blob/0705df2d704f8133b9b60fca6451c3bdd47fd1c9/src/rust/bitbox02-rust/Cargo.toml#L28-L29

I forgot it's split. In fact, I wonder if the firmware in rust should be just one crate.

NickeZ:
hehe, that c-lib could be implemented in rust :P

NickeZ:
> > Re. dependencies not compiled with "test" set. That is a feature... Not sure if it is the best decision, but this is the way it is.
> 
> Oh, right:
> 
> https://github.com/digitalbitbox/bitbox02-firmware/blob/0705df2d704f8133b9b60fca6451c3bdd47fd1c9/src/rust/bitbox02-rust/Cargo.toml#L28-L29
> 
> I forgot it's split. In fact, I wonder if the firmware in rust should be just one crate.

That would probably be inconvenient for other reasons. Even slower compile times for example.

x1ddos:
> That would probably be inconvenient for other reasons. Even slower compile times for example.

Well, how slow? :)
The question is, would it still be acceptable? I'd think so.

What other downsides?

NickeZ:
> Well, how slow? :)
> The question is, would it still be acceptable? I'd think so.
> 
> What other downsides?

A crate is the unit of compilation. So you would recompile all the code every time. Not sure what that means in practice.

I can't recall any other downsides. I just followed best practice when wrapping C code. This is how all other C-wrapping crates are structured.

x1ddos:
> A crate is the unit of compilation. So you would recompile all the code every time. Not sure what that means in practice.

I believe `cargo build/run/test` would reuse intermediate built artifacts if their sources haven't changed.

https://doc.rust-lang.org/cargo/guide/build-cache.html
https://doc.rust-lang.org/cargo/reference/profiles.html#incremental

x1ddos:
Well, the units of compilation could also be `firmware` and `bootloader`, right.

NickeZ:
> Well, the units of compilation could also be `firmware` and `bootloader`, right.

Most of the code in the bootloader is reused in the firmware.

NickeZ:
What is the goal with this PR actually?

It doesn't make sense to me to write unit-tests that test the c-library bindings unless you have the c-library linked in.

x1ddos:
> > Well, the units of compilation could also be `firmware` and `bootloader`, right.
> 
> Most of the code in the bootloader is reused in the firmware.

Yeah, good point. But I wonder how much of the shared piece that is. I guess it'll be more clear if there's a better layout when more things are ported to Rust.

NickeZ:
> Yeah, good point. But I wonder how much of the shared piece that is. I guess it'll be more clear if there's a better layout when more things are ported to Rust.

The bootloader is a subset of the firmware except for the "jump to the firmware" function. You have driver initialization, some usb communication and some oled driving.

benma:
> What is the goal with this PR actually?
> 
> It doesn't make sense to me to write unit-tests that test the c-library bindings unless you have the c-library linked in.

I wanna do the noise communication in Rust, which needs this random function to generate private keys (same as in C noise).

I also plan on writing some unit tests to run through a noise communication, so that's why this function should also do something in unit tests.

> Re. dependencies not compiled with "test" set. That is a feature... Not sure if it is the best decision, but this is the way it is.

It might be worthwhile to have a TESTING version of the C lib linked with the rust unit tests, as some code should be tested but probably won't be ported to Rust for a long time. This will strain our CMakeLists even more, sounds kind of hard to actually do in practice. But if you want to give it a go, I am all for it :D

benma:
@NickeZ on 2nd thought, many of the C functions I would want to be able to mock in Rust, for example to test the unlock api with the keystore locked, already unlocked, and so on. If it's mocked in Rust (via traits or structs/faux), then no linked C version is needed at all. 

Tricky situation.

NickeZ:
> It might be worthwhile to have a TESTING version of the C lib linked with the rust unit tests, as some code should be tested but probably won't be ported to Rust for a long time. This will strain our CMakeLists even more, sounds kind of hard to actually do in practice. But if you want to give it a go, I am all for it :D

Doesn't this already exist for the C unit tests?

benma:
> Doesn't this already exist for the C unit tests?

yeah, I am still a bit puzzled how to go about it. Let's put it on ice for now and see how it goes, can always port (best option) or mock.

PR #450, opened 2020-04-07
rust/hww: fix unlock err code

benma:
Used the wrong error code when porting to Rust.


benma:
See https://github.com/digitalbitbox/bitbox02-firmware/blob/0705df2d704f8133b9b60fca6451c3bdd47fd1c9/src/hww.c#L167

PR #454, opened 2020-04-08
Relax the python library requirements

TheCharlatan:
The python libraries dependencies now no longer have an upper ceiling. The funcionality of the library was tested with these relaxed requirements with electrum. Comprehensive testing against many other
versions of the dependencies (such as hidapi version 0.9.0) was not done.


benma:
Thx. Rushing it through now, can add saner upper bounds later if needed.

PR #457, opened 2020-04-09
py/bitbox02: namespace all protos with package directive

x1ddos:
Our protobuf definitions have no package specified.
This is intentional. The protos aren't used anywhere except
communications with the device and adding a package namespace makes
type names even longer for too little benefit.

However, the bitbox02 Python package is used elsewhere like the
electrum app where all its py protobufs combined may have name
collisions due to the empty package name.

The generated protos for Go, C and Rust bindings remain as is for now:
package name isn't used on the wire. This is not a breaking change.

Once this is submitted and a new bitbox02 py package is released, this should permanently resolve the issue in https://github.com/spesmilo/electrum/pull/5993#issuecomment-611080550.

### DIFF
--- a/py/bitbox02/Makefile
+++ b/py/bitbox02/Makefile
@@ -21,20 +21,35 @@ TARGETS=$(addprefix ${OUT_DIR}/, $(PROTO_FILES:.proto=_pb2.py))
 
 all: ${TARGETS}
 
+# To avoid protobuf name collisions with other packages used elsewhere like the
+# electrum app, the pb files for Python are generated from a copy
+# of ../../messages where all protos specify a non-empty package directive.
+# The proto files need not be committed to the repo. They are created at build time.
+PY_PROTO_DIR=.proto
+PY_PROTO=$(addprefix ${PY_PROTO_DIR}/,${PROTO_FILES})
+
 # Explicit rules cannot have multiple outputs (it will be interpreted as multiple rules and
 # therefore the same command will be executed multiple times)
 # Instead we create a single implicit rule. Implicit rules allow multiple outputs without executing
 # the same command multiple times when multiple threads are used.
 # https://stackoverflow.com/questions/19822435/multiple-targets-from-one-recipe-and-parallel-execution
 TMP_TARGETS=$(addprefix %/, $(PROTO_FILES:.proto=_pb2.py))
-${TMP_TARGETS}: $(addprefix ../../messages/,${PROTO_FILES})
+${TMP_TARGETS}: ${PY_PROTO}
 	mkdir -p ${OUT_DIR}
-	protoc --proto_path=../../messages --python_out=${OUT_DIR} --mypy_out=${OUT_DIR} $^
+	protoc --proto_path=${PY_PROTO_DIR} --python_out=${OUT_DIR} --mypy_out=${OUT_DIR} $^
 	touch ${@D}/__init__.py
 # The hack below is due to protobuf-python generating absolute paths (which are frowned upon) instead of relative paths.
 # https://github.com/protocolbuffers/protobuf/issues/1491
 	sed -i 's/^from \([^.]*_pb2\)/from .\1/g'  ${OUT_DIR}/*.pyi
 	sed -i 's/^import \([^.]*_pb2\) as \([^.]*__pb2\)/from . import \1 as \2/g' ${OUT_DIR}/*.py
+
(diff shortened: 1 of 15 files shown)


x1ddos:
>  The protos aren't used anywhere except communications with the device and adding a package namespace makes type names even longer for too little benefit

Forgot to add: the big downside if we simply add `package <something>` to protos in [messages/](https://github.com/digitalbitbox/bitbox02-firmware/tree/master/messages/), this spreads out very quickly to Go and C => Rust bindings. 1000s of lines changed for no benefit so far. It was the first thing I tried.

If this becomes a real problem outside of the python package, we can always do it then.

py/bitbox02/Makefile
```diff
@@ -21,21 +21,38 @@ TARGETS=$(addprefix ${OUT_DIR}/, $(PROTO_FILES:.proto=_pb2.py))
...
+	protoc --proto_path=${PY_PROTO_DIR} --python_out=${OUT_DIR} --mypy_out=${OUT_DIR} $^
 	touch ${@D}/__init__.py
 # The hack below is due to protobuf-python generating absolute paths (which are frowned upon) instead of relative paths.
 # https://github.com/protocolbuffers/protobuf/issues/1491
 	sed -i 's/^from \([^.]*_pb2\)/from .\1/g'  ${OUT_DIR}/*.pyi
 	sed -i 's/^import \([^.]*_pb2\) as \([^.]*__pb2\)/from . import \1 as \2/g' ${OUT_DIR}/*.py
 
+# Ideally, the prerequisites would be only those which we need for PROTO_FILES
+# including their optional .options files, but it would make the rules more complicated.
+# So, we just grab all proto and options files, insert package name to protos
```
benma, line 48: you don't need the option files, they are only for nanopb. See 

https://jpa.kapsi.fi/nanopb/docs/concepts.html#modifying-generator-behaviour
https://jpa.kapsi.fi/nanopb/docs/reference.html#defining-the-options-in-a-options-file

I am not 100% sure but I think it only applies to nanopb and is not a general protobuf concept.

benma: They are also not copied to the Go lib, fyi.

x1ddos: let me try without the options and see if it changes anything in pb2.py. never used options in separate files before.

x1ddos: nice find! indeed, no changes in pb2.py files without options.
removed.

x1ddos:
PTAL

py/bitbox02/Makefile
x1ddos: I mean, options is a known feature of protobufs - https://developers.google.com/protocol-buffers/docs/proto3#options - but I've never used it in separate files. My favorite was `deprecated=true`, of course.

x1ddos: So, if say you added `[deprecated=true]`, I'm pretty sure python would generate a pb2.py with a warning or something when that deprecated field is used. 

Not sure why options were split in a separate files and https://github.com/digitalbitbox/bitbox02-firmware/commit/ace1bc4d07fd4599c066bb9116bbb27ab4b72daf doesn't explain it either but we can always merge them back to .proto files if the protobuf usage becomes more complex like deprecating fields or adding custom options.

x1ddos: It's also unclear what happens if you start adding options to both .proto and .options files.
Anyway, I think it's good as is for now.

benma: as i said, the separate file is a nanopb feature (i think ;)).

it is in a separate file because if you put it in the main file, the generators for Go/Python will fail (unrecognized options or sth like that).

x1ddos: oh I thought you meant options in general is a nanopb feature.

> if you put it in the main file, the generators for Go/Python will fail (unrecognized options or sth like that).

I don't it would fail because you then specify which plugin it is, like in nanopb example:
 
```
message MyMessage {
    required string name = 1 [(nanopb).max_size = 40];
    repeated int32 ids = 4   [(nanopb).max_count = 5];
}
```

the `(nanopb)` tells protoc it's for nanopb generator.

x1ddos:
I've manually "tested" it with the `send_message.py`. All seemed to work.

x1ddos:
> The base branch requires all commits to be signed. 

I see. Let me do that.

PR #458, opened 2020-04-09
Pairing

benma:


src/rust/bitbox02-rust/src/workflow/pairing.rs
```diff
@@ -14,14 +14,11 @@
```
benma, line 14: this is a pretty bad diff, easier to have both versions side by side.

src/rust/bitbox02-rust/src/workflow/pairing.rs
```diff
@@ -14,14 +14,11 @@
 
 use arrayvec::ArrayString;
 use binascii;
-use bitbox02;
 use core::fmt::Write;
 
-/// This function takes a slice of bytes that must be at most 60 characters when base32
-/// encoded.
-pub fn create(bytes: &[u8]) -> bool {
+pub async fn confirm(hash: &[u8; 32]) -> bool {
```
benma, line 19: this function will be used soon for the normal bb02 noise pairing screen.

src/rust/bitbox02-rust/src/workflow/pairing.rs
```diff
@@ -14,14 +14,17 @@
 
 use arrayvec::ArrayString;
 use binascii;
-use bitbox02;
 use core::fmt::Write;
 
-/// This function takes a slice of bytes that must be at most 60 characters when base32
-/// encoded.
-pub fn create(bytes: &[u8]) -> bool {
+pub fn extra_hash_create(bytes: &[u8]) -> bool {
```
x1ddos: I know this is existing code so, for the future: sounds very weird to me that a func which seemingly from the name just manipulates some bytes, actually blocks on user interaction. If I didn't see the code, only the signature, I would've never guessed that it blocks on UI.

src/rust/bitbox02-rust/src/async_usb.rs
```diff
@@ -29,20 +29,20 @@ type UsbIn = Vec<u8>;
...
+    Running(Task<'a, UsbOut>),
     /// The task has finished and written the result, so the USB response is available. We are now
     /// waiting for the client to fetch it (HWW_REQ_RETRY). For short-circuited or non-async api
     /// calls, the result might be returned immediately in response to HWW_REQ_NEW.
     ResultAvailable(UsbOut),
 }
 
 /// A safer version of UsbTaskState. RefCell so we cannot accidentally borrow illegally.
-struct SafeUsbTaskState(RefCell<UsbTaskState>);
+struct SafeUsbTaskState(RefCell<UsbTaskState<'static>>);
```
x1ddos, line 45: Why does it have to be static?

src/rust/bitbox02-rust/src/workflow/pairing.rs
```diff
@@ -14,14 +14,17 @@
...
+    let mut buf = [0u8; 32];
+    bitbox02::sha256(bytes, &mut buf).expect("sha256 failed");
+    crate::bb02_async::block_on(confirm(&buf))
+}
+
+pub async fn confirm(hash: &[u8; 32]) -> bool {
     let mut encoded = [0u8; 60];
-    binascii::b32encode(bytes, &mut encoded).unwrap_or_else(|_| panic!("Too short buffer"));
+    let encoded = binascii::b32encode(&hash[..], &mut encoded).unwrap();
```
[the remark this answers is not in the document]
benma: So the function can only fail if the out buffer is too short:

https://docs.rs/binascii/0.1.4/binascii/fn.b32encode.html

So an error here is in the category 'programming error', not 'expected errors that can be handled'. Would you still propagate the error? How to handle it at the top?

So far I have chosen to Abort()/panic for programming errors (stuff that only happens if there is a serious implementation mistake).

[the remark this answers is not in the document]
benma: It is not the same bytes. The one before is the full buffer, the outcoming buffer is the right size ot hold the encoded version. See https://docs.rs/binascii/0.1.4/src/binascii/lib.rs.html#162

src/rust/bitbox02-rust/src/workflow/pairing.rs
benma: The reasoning is/was:

this is workflow/pairing.rs. If all workflow functions are called `create()`, then this follows this naming pattern.

The `_create` naming pattern was carried over from C, where the workflows are/were named like this.

The workflows I ported over didn't follow this and are just named `confirm()` etc.

src/rust/bitbox02-rust/src/async_usb.rs
benma: a few lines down the only use of this is in a static variable. I could make this generic too and have the compiler infer it (I guess!), but that would clutter it more with generic params.

src/rust/bitbox02-rust/src/bb02_async.rs
```diff
@@ -35,13 +38,10 @@ pub struct AsyncOption<'a, O>(&'a RefCell<Option<O>>);
 
 impl<O> core::future::Future for AsyncOption<'_, O> {
     type Output = ();
-    fn poll(
-        self: core::pin::Pin<&mut Self>,
-        _cx: &mut core::task::Context<'_>,
-    ) -> core::task::Poll<Self::Output> {
+    fn poll(self: core::pin::Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
```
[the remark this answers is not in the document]
benma: Fixed

src/rust/bitbox02-rust/src/workflow/pairing.rs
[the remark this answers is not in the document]
benma: In other words, the previous utf8 string below had a bunch of trailing null bytes. They didn't hurt I guess, but it was a double take for me.

src/rust/bitbox02-rust/src/async_usb.rs
x1ddos: ah no, this is good, sorry. I missed the static below. should've unfolded more lines.
:+1:

src/rust/bitbox02-rust/src/workflow/pairing.rs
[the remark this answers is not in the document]
x1ddos: Damn, I did look at the source of b32encode before...
All good :+1: Definitely agree panic here and in other cases like this is good.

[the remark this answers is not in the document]
x1ddos: oh you're right! I should've read more carefully in b32encode:

```rust
    for idx in data_len + 1..total_len {
        output[idx] = b'=';
    }

    Ok(&mut output[..total_len])
```

PR #459, opened 2020-04-11
py/bitbox02: remote typing dependency and bump version

benma:
See
https://github.com/spesmilo/electrum/pull/5993#issuecomment-611719275:

https://pypi.org/project/typing/

    NOTE: in Python 3.5 and later, the typing module lives in the stdlib, and installing this package has NO EFFECT.

### DIFF
--- a/py/bitbox02/bitbox02/bitbox02/__init__.py
+++ b/py/bitbox02/bitbox02/bitbox02/__init__.py
@@ -16,7 +16,7 @@
 from __future__ import print_function
 import sys
 
-__version__ = "2.0.2"
+__version__ = "2.0.3"
 
 if sys.version_info.major != 3 or sys.version_info.minor < 6:
     print(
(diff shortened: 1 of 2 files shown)


x1ddos:
what about typing extensions, is that something extra, not in stdlib?

also, is the comment above it still needed?

benma:
> what about typing extensions, is that something extra, not in stdlib?
> 
> also, is the comment above it still needed?

https://pypi.org/project/typing-extensions/

> The typing module was added to the standard library in Python 3.5 on a provisional basis and will no longer be provisional in Python 3.7. However, this means users of **Python 3.5 - 3.6** who are unable to upgrade will not be able to take advantage of new types

SomberNight:
> also, is the comment above it still needed?

You can actually encode that comment as part of the requirements:
https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies
```
"typing_extensions>=3.7.4 ; python_version<'3.7'",
```

PR #461, opened 2020-04-11
rust/bitbox02: wrap C memory noise getter/setter functions

benma:
Will be needed for noise processing in Rust.


benma:
Been playing with `zeroize` a *lot* these days, it is very ergonomic way to take care of zeroing.

Will probably use it in password.rs and remove the self-made `util::zero()`.

x1ddos:
> Will probably use it in password.rs and remove the self-made `util::zero()`.

Yeah, it seems the crate uses similar methods to `util:zero` like https://docs.rs/zeroize/1.1.0/src/zeroize/lib.rs.html#428.
I would create an issue and add a todo `util:zero`.

src/rust/bitbox02/src/memory.rs
```diff
@@ -42,3 +42,22 @@ pub fn bootloader_hash(out: &mut [u8; 32]) {
...
+        true => Ok(out),
+        false => Err(()),
+    }
+}
+
+pub fn check_noise_remote_static_pubkey(pubkey: &[u8; 32]) -> bool {
+    unsafe { bitbox02_sys::memory_check_noise_remote_static_pubkey(pubkey.as_ptr()) }
+}
+
+pub fn add_noise_remote_static_pubkey(pubkey: &[u8; 32]) -> Result<(), ()> {
```
x1ddos, line 58: Would it make sense for this func to take in the array wrapped in a `Zeroizing`? Similar to what you did in https://github.com/sopium/noise-rust/pull/20/files#diff-912aec4d0b5ac65968f9c18b6d88c19c.

benma: This is a reference, so the caller can use (or not use) Zeroizing and it works the same way.

Since this is a pubkey, zeroizing is not needed, as it is not secret.

x1ddos: Right, that was to exactly to force the caller always use Zeroizing but it is indeed a public key. Not sure what I was thinking. Ignore.

PR #462, opened 2020-04-11
rust/bitbox02: wrap C commander()

benma:
Same style as [process_packet](https://github.com/digitalbitbox/bitbox02-firmware/blob/6b44ed0e5711218efdd875ac308b01b9ee86258c/src/rust/bitbox02/src/hww.rs#L19), which will be deleted soon.

### DIFF
--- a/src/rust/bitbox02/src/commander.rs
+++ b/src/rust/bitbox02/src/commander.rs
@@ -1,0 +1,43 @@
+// Copyright 2020 Shift Cryptosecurity AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


src/rust/bitbox02/src/commander.rs
```diff
@@ -0,0 +1,43 @@
...
+        max_len: output_vec.capacity() as _,
+    };
+    unsafe {
+        // Safety:
+        // input is not NULL and the data length is correct.
+        // output is not NULL and has the correct capacity.
+        bitbox02_sys::commander(&input as *const _, &mut output as *mut _);
+        // Safety: commander is guaranteed to set the number of bytes written
+        // correctly.
+        output_vec.set_len(output.len as _);
```
x1ddos, line 40: or `.truncate(output.len)` outside of `unsafe`? the truncate also calls https://doc.rust-lang.org/beta/core/ptr/fn.drop_in_place.html in the end - a good thing for possibly reducing memory usage?

benma: Thel `.len()` is `0` to start with, the memory is only reserved with `with_capacity()`. `truncate` wouldn't do anything in this case.

benma: > Why deleted soon. In favor of ...?

process_packet currently only does noise and dispatch to commander. I have a PR nearly ready that does the same in Rust, but async.

PR #465, opened 2020-04-12
Rust sha256

benma:
external/noise-c/crypto was used for streaming sha256. After using the Rust version of this, we can remove the noise-c dep.

Actual removal of the submodule will happen in another PR as submodule still looks tricky in 2020.

The first commits are the same as https://github.com/digitalbitbox/bitbox02-firmware/pull/463, as the other one already adds the `sha2` dep.


src/apps/btc/btc_sign.c
```diff
@@ -60,7 +60,8 @@ static uint64_t _outputs_sum_ours = 0;
 static uint64_t _outputs_sum_out = 0;
 
 // used during STATE_INPUTS_PASS1
-static sha256_context_t _hash_prevouts_ctx, _hash_sequence_ctx;
+static void* _hash_prevouts_ctx = NULL;
+static void* _hash_sequence_ctx = NULL;
```
x1ddos, line 64: Why not `rust_sha256_new();` instead of `NULL`? I see these are initialized in `_reset()` and the latter is called from `app_btc_sign_init` but `app_btc_sign_input` is also invoked from the commander when `Request_btc_sign_input_tag` is sent. If the host is misbehaving, could that lead to unnecessary panics/crashes?

src/apps/btc/btc_sign.c
```diff
@@ -69,18 +70,15 @@ static uint8_t _hash_prevouts[32] = {0};
...
 static uint8_t _hash_outputs[32] = {0};
 
 // hashes bytes, puts result into out. bytes and out can overlap.
 static void _sha256(const uint8_t* bytes, size_t bytes_len, uint8_t* out)
 {
-    sha256_context_t ctx;
-    sha256_reset(&ctx);
-    noise_sha256_update(&ctx, bytes, bytes_len);
-    sha256_finish(&ctx, out);
+    rust_sha256(bytes, bytes_len, out);
```
x1ddos, line 81: seems like a wrapper for a single function call now. did you consider removing `_sha256` func altogether and replacing calls with `rust_sha256`?

src/apps/btc/btc_sign.c
```diff
@@ -580,7 +584,7 @@ app_btc_sign_error_t app_btc_sign_output(
             return _error(APP_BTC_SIGN_ERR_USER_ABORT);
         }
 
-        sha256_finish(&_hash_outputs_ctx, _hash_outputs);
+        rust_sha256_finish(&_hash_outputs_ctx, _hash_outputs);
         // hash hash_outputs to produce the final double-hash
         _sha256(_hash_outputs, 32, _hash_outputs);
```
x1ddos, line 589: for instance here, would `rust_sha256(_hash_outputs, 32, _hash_outputs);` work as well?

src/apps/btc/btc_sign.c
benma: It's a static var, so it needs to be constant. 

reset is called at the start of each transaction signing, when it is aborted at any time, or after it finishes normally. I don't see a way for the host to invoke a crash, do you?

`rust_sha256_free()` can be called on NULL without any effect.

src/apps/btc/btc_sign.c
benma: indeed. fixed!

src/rust/bitbox02-rust-c/src/sha2.rs
```diff
@@ -0,0 +1,133 @@
...
+    let out = core::slice::from_raw_parts_mut(out, 32);
+    let hasher = Box::from_raw(*ctx as *mut Sha256); // dropped at the end
+    let hash = hasher.result();
+    out.copy_from_slice(&hash[..]);
+    *ctx = core::ptr::null_mut();
+}
+
+/// Safety: ctx must be a pointer to a valid sha256 context produced by `rust_sha256_new()`, or a
+/// pointer to NULL.
+/// After this, the hasher is dropped and `cx` is set to NULL and must not be used anymore.
```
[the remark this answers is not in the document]
benma: fixed

src/apps/btc/btc_sign.c
x1ddos: Oh right. I missed that `app_btc_sign_input` checks what `_state` it's in on https://github.com/digitalbitbox/bitbox02-firmware/blob/4a2cac951381032ac0fcb511d0ea747c478c95b5/src/apps/btc/btc_sign.c#L375-L377.
Never mind. :+1:

PR #467, opened 2020-04-14
u2f: use Rust unlock/confirm workflows

benma:


benma:
This is a shortcut to get u2f to use the Rust workflows, without rewriting the whole message processor like we did for HWW.

The goal is to be able to delete the C workflows and workflow stack.

unlock is the first one, the only other one is `confirm()`, which can follow in another PR.

benma:
Sth similar will be done in fido2 after https://github.com/digitalbitbox/bitbox02-firmware/pull/437/ is merged.

benma:
Actually pushed another commit to do the same for the confirm workflow. After this PR, u2f does not use any C workflows anymore.

x1ddos:
Nice. I have nothing to add. Let me just try it on a real device.

PR #470, opened 2020-04-21
py/bitbox02: avoid "multiple package definitions" proto error

x1ddos:
Somehow, Python's protobuf gen fails with:

    hww.proto:17:1: Multiple package definitions.
    hww.proto:18:1: Multiple package definitions.
    hww.proto:19:1: Multiple package definitions.
    ...

The only place where this could happen seems to be during the copy
where the cp command doesn't actually replace an existing destination.
It could happen if the host OS has some default cp flags set, like -u.

Instead of copying and then sed-updating in-place, do the package
insert in memory and output the result directly to the dest location
This should make it idempotent.


x1ddos:
(forgot to sign the commit)

PR #472, opened 2020-04-21
More known U2F App IDs

jengo9332:
BitBox02 showing "Unknown site: " on almost every application except Google & Github.
This request adds more U2F names when authenticating.


src/u2f/u2f_app.c
```diff
@@ -29,6 +29,96 @@ static const app_t _apps[] = {
...
+    	// U2F Key for Keeper
+        .app_id = "\x53\xa1\x5b\xa4\x2a\x7c\x03\x25\xb8\xdb\xee\x28\x96\x34\xa4\x8f\x58\xae\xa3\x24"
+                  "\x66\x45\xd5\xff\x41\x8f\x9b\xb8\x81\x98\x85\xa9",
+        .name = "Keeper",
+    },
+    {
+    	// U2F Key for Keeper
+        .app_id = "\xd6\x5f\x00\x5e\xf4\xde\xa9\x32\x0c\x99\x73\x05\x3c\x95\xff\x60\x20\x11\x5d\x5f"
+                  "\xec\x1b\x7f\xee\x41\xa5\x78\xe1\x8d\xf9\xca\x8c",
+        .name = "Keeper",
```
benma, line 103: duplicate name - can you please disambiguate them?

src/u2f/u2f_app.c
```diff
@@ -29,6 +29,96 @@ static const app_t _apps[] = {
...
+    	// U2F Key for Slush Pool
+        .app_id = "\x08\xb2\xa3\xd4\x19\x39\xaa\x31\x66\x84\x93\xcb\x36\xcd\xcc\x4f\x16\xc4\xd9\xb4"
+                  "\xc8\x23\x8b\x73\xc2\xf6\x72\xc0\x33\x00\x71\x97",
+        .name = "Slush Pool",
+    },
+    {
+    	// U2F Key for Slush Pool
+        .app_id = "\x38\x80\x4f\x2e\xff\x74\xf2\x28\xb7\x41\x51\xc2\x01\xaa\x82\xe7\xe8\xee\xfc\xac"
+                  "\xfe\xcf\x23\xfa\x14\x6b\x13\xa3\x76\x66\x31\x4f",
+        .name = "Slush Pool",
```
benma, line 115: same

src/u2f/u2f_app.c
```diff
@@ -29,6 +29,96 @@ static const app_t _apps[] = {
...
+        .app_id = "\x9d\x61\x44\x2f\x5c\xe1\x33\xbd\x46\x54\x4f\xc4\x2f\x0a\x6d\x54\xc0\xde\xb8\x88"
+                  "\x40\xca\xc2\xb6\xae\xfa\x65\x14\xf8\x93\x49\xe9",
+        .name = "Fedora",
+    },
+    {
+        .app_id = "\xa4\xe2\x2d\xca\xfe\xa7\xe9\x0e\x12\x89\x50\x11\x39\x89\xfc\x45\x97\x8d\xc9\xfb"
+                  "\x87\x76\x75\x60\x51\x6c\x1c\x69\xdf\xdf\xd1\x96",
+        .name = "Gandi",
+    },
+    {
```
benma, line 87: delete?

src/u2f/u2f_app.c
```diff
@@ -29,6 +29,96 @@ static const app_t _apps[] = {
                   "\x59\x43\x79\x4e\x88\x4f\x3d\x02\x3a\x82\x29\xfd",
         .name = "Google",
     },
+    {
```
benma, line 32: please add `sha256('...')` to each entry as above, so it's clear which website it is and how the app_id is constructed.

benma: in case you copied the hashes from some source, please share the source, so it's easier to double check the entries.

benma:
Oh and please sign your commits if you can :)

src/u2f/u2f_app.c
jengo9332: Source is from here: https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/webauthn/knownapps.py

jengo9332: I'll also add that I'm clueless when it comes to this.

I'm sorry.

src/u2f/u2f_app.c
jengo9332: https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/webauthn/knownapps.py#L172

src/u2f/u2f_app.c
jengo9332: https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/webauthn/knownapps.py#L212

src/u2f/u2f_app.c
x1ddos: +1 to sha256. @jengo9332 it seems those are auto-generated from https://github.com/trezor/trezor-firmware/tree/8cced84bc47f074b148abbaf595f2510c72ed6fb/common/defs/fido

See more here about App IDs: https://developers.yubico.com/U2F/App_ID.html

You can verify it manually like so:
$ echo -n 'https://www.dropbox.com/u2f-app-id.json' | sha256sum
c50f8a7b708e92f82e7a50e2bdc55d8fd91a22fe6b29c0cdf7805530842af581  -

Here's a couple more known IDs: https://github.com/github/SoftU2F/blob/master/SoftU2FTool/KnownFacets.swift

benma: No worries, thanks. knownapps.py seems to be generated from those sources:

https://github.com/trezor/trezor-firmware/tree/master/common/defs/fido

I'll check again tomorrow.

jengo9332: I see, thanks.
I found an old list with known IDs and added them. However it's lacking Facebook and Fedora.

jengo9332: I cleaned the file and added verified sha256('') to each entry. I also removed webauthn stuff since I just realized that webauthn != u2f.

However I'm still not sure why keeper has two different sha256sum, two different domains or?

src/u2f/u2f_app.c
```diff
@@ -29,6 +29,108 @@ static const app_t _apps[] = {
...
+        // sha256('https://api-9dcf9b83.duosecurity.com')
+        .app_id = "\xf3\xe2\x04\x2f\x94\x60\x7d\xa0\xa9\xc1\xf3\xb9\x5e\x0d\x2f\x2b\xb2\xe0\x69\xc5"
+                  "\xbb\x4f\xa7\x64\xaf\xfa\x64\x7d\x84\x7b\x7e\xd6",
+        .name = "Duo",
+    },
+    {
+        // sha256('facebook.com')
+        .app_id = "\x31\x19\x33\x28\xf8\xe2\x1d\xfb\x6c\x99\xf3\x22\xd2\x2d\x7b\x0b\x50\x87\x78\xe6"
+                  "\x4f\xfb\xba\x86\xe5\x22\x93\x37\x90\x31\xb8\x74",
```
x1ddos, line 71: It seems facebook works only with FIDO2, and trezor also [specifies it](https://github.com/trezor/trezor-firmware/blob/65b0eae89df72eff1561d4de7221c8e34562acd4/common/defs/fido/facebook.json#L3) as webauthn.

In U2F,

> In the Web case, the FacetID MUST be the Web Origin [RFC6454] of the web page triggering the FIDO operation

(from https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-appid-and-facets-v1.2-ps-20170411.html#determining-the-facetid-of-a-calling-application)

I would hold off adding this until https://github.com/digitalbitbox/bitbox02-firmware/pull/437/files is submitted. This won't work otherwise, I believe, and would make things confusing.

src/u2f/u2f_app.c
```diff
@@ -29,6 +29,108 @@ static const app_t _apps[] = {
...
+        .name = "Dashlane",
+    },
+    {
+        // sha256('https://www.dropbox.com/u2f-app-id.json')
+        .app_id = "\xc5\x0f\x8a\x7b\x70\x8e\x92\xf8\x2e\x7a\x50\xe2\xbd\xc5\x5d\x8f\xd9\x1a\x22\xfe"
+                  "\x6b\x29\xc0\xcd\xf7\x80\x55\x30\x84\x2a\xf5\x81",
+        .name = "Dropbox",
+    },
+    {
+        // sha256('https://api-9dcf9b83.duosecurity.com')
```
x1ddos, line 63: This is weird. api-9dcf9b83 feels like a "random" identifier and https://api-9dcf9b83.duosecurity.com responds with 404 not found. Where is it from?

[trezor's fido entries](https://github.com/trezor/trezor-firmware/blob/65b0eae89df72eff1561d4de7221c8e34562acd4/common/defs/fido/duo.json) does list it but I wouldn't add it until we know whether it's actually owned by duosecurity and not some kind of their 3rd party app or a user.

src/u2f/u2f_app.c
x1ddos: (web origin includes the scheme so this would've been `https://facebook.com` if facebook supported it)

src/u2f/u2f_app.c
x1ddos: > FYI - That hostname will be unique for each Duo customer account.

(from https://github.com/github/SoftU2F/pull/32#issuecomment-332979298)

I would definitely remove this.

src/u2f/u2f_app.c
jengo9332: U2F (BitBox02) works with Facebook. When I authenticate it shows me the sha256 provided above and url is "facebook.com".

src/u2f/u2f_app.c
jengo9332: Removed.

src/u2f/u2f_app.c
jengo9332: @x1ddos Something is weird here, can you verify this?

src/u2f/u2f_app.c
```diff
@@ -29,6 +29,102 @@ static const app_t _apps[] = {
...
+        .name = "Slush Pool",
+    },
+    {
+        // sha256('https://dashboard.stripe.com')
+        .app_id = "\x2a\xc6\xad\x09\xa6\xd0\x77\x2c\x44\xda\x73\xa6\x07\x2f\x9d\x24\x0f\xc6\x85\x4a"
+                  "\x70\xd7\x9c\x10\x24\xff\x7c\x75\x59\x59\x32\x92",
+        .name = "Stripe",
+    },
+    {
+        // sha256('https://www.dashlane.com')
```
benma, line 111: dashlane is present twice, please delete one

src/u2f/u2f_app.c
benma: Verified, the facebook app id is correct. Very good.

src/u2f/u2f_app.c
jengo9332: I think I'm going blind. Fixed :)

jengo9332:
Alright, added Tutanota as the last entry. 
I think it all should be good now.

src/u2f/u2f_app.c
x1ddos: I guess I was wrong and facebook's app ID isn't a URL. So, this holds:

> 1. If the AppID is not an HTTPS URL, and matches the FacetID of the caller, no additional processing is necessary and the operation may proceed.

It just so happens that `facebook.com` looks like part of a URL. Although, in this case couldn't I just fake facebok's FacetID? Since it isn't a URL, web origin is not enforced.

x1ddos: They do claim to support U2F though, so I guess it's up to facebook.
https://www.facebook.com/help/401566786855239

jengo9332: Could it be that webauthn is backwards compatible?

x1ddos: > Could it be that webauthn is backwards compatible?

I guess so - ctrl+f "compatibility" on https://www.yubico.com/blog/what-is-fido2/ and https://fidoalliance.org/specifications/ seems to match that as well - but my concern was mostly about whether to allow non-URL facet IDs like `facebook.com` to match U2F app IDs.

We're working on fido2 support in https://github.com/digitalbitbox/bitbox02-firmware/pull/437.

jengo9332: I see, Web facets must use HTTPS when it comes to U2F. 
Should I remove Facebook from the list?

x1ddos: right. web facets start with the scheme and there's also `android:...` and `ios:...` for android and ios apps but facebook is none of those.

If there are currently no users who want to log in to facebook with the bitbox, I'd probably hold off adding this. Not sure. I defer to @benma. :)

jengo9332: I'm using it with facebook atm, however I don't mind waiting.

I guess the right thing would be to not add it yet until FIDO2 support. If there are no downsides when it comes to security I guess it could be added.

Maybe @benma knows 😄

benma: Since we're in doubt, let's remove it for now. Sorry for the trouble.

jengo9332: Alright, removed :)

benma:
> Alright, added Tutanota as the last entry.
> I think it all should be good now.

Thanks. Final step: please squash all commits, so the commit history does not have a lot of small fixup commits.

jengo9332:
> 
> 
> > Alright, added Tutanota as the last entry.
> > I think it all should be good now.
> 
> Thanks. Final step: please squash all commits, so the commit history does not have a lot of small fixup commits.

Ok, done

benma:
Thanks again!

src/u2f/u2f_app.c
x1ddos: To clear the doubts here: no security implications really in the end but some potential troubles for existing users when using U2F on WebAuthn websites like `facebook.com`. We'll continue this in https://github.com/digitalbitbox/bitbox02-firmware/issues/474 - similar case but with `github.com`.

x1ddos: (I still think it was best to remove `facebook.com` from this PR)

PR #475, opened 2020-05-04
Label

benma:


src/ui/components/label.c
```diff
@@ -24,9 +24,13 @@
 #include <ui/ui_util.h>
 #include <util.h>
 
+// Max size of text shown (excl. null terminator). The current size of 640 is chosen to be able to
+// show up to 320 bytes of Ethereum tx data in hex format.
+#define MAX_LABEL_SIZE 640
```
x1ddos, line 29: Would it be possible to expose this as a constant to rust crates? I think it's too easy to update in one place and forget some others in the future.

benma: i don't think so, but maybe there is a way. maybe we could just turn it into an `extern const ...`.

there is a different problem though: even if the const is defined in Rust (`const ...: usize = 640`), I somehow can't use it in the str macro call, so that needs to be fixed first too (this is why the constant is replicated like 4 times in `ui.rs` instead of only once).

x1ddos: ok. I'll create a bug.

NickeZ: I'm pretty sure `--whitelist-var` can be used to expose defines in rust.

PR #477, opened 2020-05-05
apps/btc/sign: replace 'See the BitBoxApp' flicker between outputs

benma:
The device waits for the next api call from the client after
confirming an output. During this time, the 'See the BitBoxApp' is
shown as the default screen if there is nothing else on the screen
stack.

This can result in an ugly flicker if there are multiple outputs to
confirm.

This is an attempt to reduce the flicker by just rendering an empty
screen (black screen) instead the regular waiting screen in this case.

### DIFF
--- a/src/apps/btc/btc_sign.c
+++ b/src/apps/btc/btc_sign.c
@@ -20,8 +20,11 @@
 #include "confirm_multisig.h"
 #include <rust/rust.h>
 
+#include <hardfault.h>
 #include <keystore.h>
 #include <memory/memory.h>
+#include <ui/components/empty.h>
+#include <ui/screen_stack.h>
 #include <util.h>
 #include <workflow/verify_recipient.h>
 #include <workflow/verify_total.h>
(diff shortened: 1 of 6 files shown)


benma:
If this was an async function, doing this kind of thing would be a one liner. Some day!

benma:
Actually I will improve this PR a bit, to push the screen with the first output and pop it with the last output, so that there is no black while processing the inputs.

src/apps/btc/btc_sign.c
```diff
@@ -75,6 +87,18 @@ static void* _hash_outputs_ctx = NULL;
 // https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki step 8.
 static uint8_t _hash_outputs[32] = {0};
 
+static void _maybe_pop_empty_screen(void)
+{
+    if (_empty_component != NULL) {
+        if (ui_screen_stack_top() != _empty_component) {
+            Abort("btc_sign: mismatched screen push/pop");
+        }
+        ui_screen_stack_pop_and_clean();
```
x1ddos, line 96: I wonder why not call the existing `ui_screen_stack_pop` and forget about it? It should be eventually cleaned up in `ui_screen_stack_cleanup`, right. Or is it a special case because of `_reset()`?

benma: On the normal execution with the normal mainloop on the device it works like you suggest, but I didn't want to make the app code rely on having a mainloop environment. The issue shows up in the unit tests of signing, where we don't do the `ui_screen_stack_cleanup()` between runs (we could, but I found this was cleaner).

fyi: `ui_screen_stack_cleanup` is a bad workaround to not run into memory bugs if the component cleans itself during execution of a button-press callback. we will hopefully remove it at some point.

In this case, cleaning it immediately is safe.

WDYT?

x1ddos: Yeah, makes sense. I like this being explicit. In fact, I found it a bit hard to track where pushed components are cleaned up. I actually didn't find an explicit call to `ui_screen_stack_cleanup` at the end of this signing flow, so I assumed it must be somewhere at a "layer" above it or in a loop where it all starts. 

:+1:

PR #478, opened 2020-05-05
u2f: change label 'U2F authenticate' -> 'U2F auth'

benma:
'U2f authenticate' is a bit too long, touching the control buttons on
the top.

### DIFF
--- a/src/u2f/u2f_app.c
+++ b/src/u2f/u2f_app.c
@@ -167,7 +167,7 @@ void u2f_app_confirm_start(enum u2f_app_confirm_t type, const uint8_t* app_id)
         }
         break;
     case U2F_APP_AUTHENTICATE:
-        title = "U2F authenticate";
+        title = "U2F auth";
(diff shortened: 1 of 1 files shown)


src/u2f/u2f_app.c
```diff
@@ -167,7 +167,7 @@ void u2f_app_confirm_start(enum u2f_app_confirm_t type, const uint8_t* app_id)
         }
         break;
     case U2F_APP_AUTHENTICATE:
-        title = "U2F authenticate";
+        title = "U2F login";
```
x1ddos, line 170: Or "confirm" or just "auth" maybe? For example, a stock exchange platform where it asks users to also confirm using a U2F device in order to complete an action.

benma: Makes sense. Changed to `U2F auth`.

x1ddos: :+1:

PR #416, opened 2020-03-18
New font size

NickeZ:

PR #482, opened 2020-05-11
Password improvements

benma:
Part of #416. The other PR accidentally removed the ability to cancel the input (used when restoring from mnemonic). This PR adds a commit that shows a proper cancel button instead.

### DIFF
--- a/src/ui/components/trinary_input_string.c
+++ b/src/ui/components/trinary_input_string.c
@@ -208,7 +208,11 @@ static void _render(component_t* component)
         data->confirm_component->f->render(data->confirm_component);
     }
     if (!confirm_gesture_active) {
-        data->left_arrow_component->f->render(data->left_arrow_component);
+        if (data->string_index != 0 ||
+            trinary_input_char_in_progress(data->trinary_char_component) ||
+            data->cancel_cb != NULL) {
+            data->left_arrow_component->f->render(data->left_arrow_component);
+        }
         if (data->keyboard_switch_component != NULL) {
             data->keyboard_switch_component->f->render(data->keyboard_switch_component);
         }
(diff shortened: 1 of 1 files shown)


src/ui/components/trinary_input_string.c
```diff
@@ -318,9 +325,6 @@ static void _on_event(const event_t* event, component_t* component)
                 data->target_x = SCROLL_RIGHT_LIMIT - string_width;
                 // data->target_x += MIN(SCREEN_WIDTH - SCROLL_RIGHT_LIMIT, string_width);
             }
-        } else if (data->cancel_cb != NULL) {
-            data->cancel_cb(data->cancel_callback_param);
-            data->cancel_cb = NULL;
```
benma, line 323: this line is removed because it was buggy - in restore from mnemonic, when you hit the cancel button and this code is invoked, there is a prompt which asks if you really want to cancel. if you choose `no`, you can't cancel again since the callback was set to NULL.

This is a regression from 2a4627dbf8f6f8700af16c1915c777e242734efa - Rust assumes the callbacks are called at most once. For now it's okay since in Rust we do not use the cancel callback.

x1ddos: Right. I see, `let mut callback = Box::from_raw(...)` then the `callback` goes out of scope and gone.
I guess if a user wants to cancel and changes their minds, they can always unplug as a workaround.

Would be nice to document this. Maybe somewhere in [workflow.h](https://github.com/digitalbitbox/bitbox02-firmware/blob/95abef355581ad3d75b6f7956908fcbc1d206c1f/src/workflow/workflow.h)?

benma: >  I guess if a user wants to cancel and changes their minds, they can always unplug as a workaround.

What?

Everything is working as it should - the cancel callback is unused in Rust.

x1ddos: ah ok. then I got a bit lost, I guess. will look again at the cancelling part in rust.

benma: Maybe it's a terminology overload. This has nothing to do with rust futures or cancelling them. It is just the cancel button at the top right that if you click it, invokes the cancel callback.

PR #483, opened 2020-05-11
font-size: use hollow arrow (not filled), and bigger arrows

benma:
Part of #416. The other PR did not hollow out the right arrow, used in scrolling through the words when showing the mnemonic.

### DIFF
--- a/src/ui/components/confirm_gesture.c
+++ b/src/ui/components/confirm_gesture.c
@@ -50,7 +50,7 @@ bool confirm_gesture_is_active(component_t* component)
  */
 static void _render(component_t* component)
 {
-    uint8_t arrow_height = 5;
+    uint8_t arrow_height = 6;
     int16_t x, y0, y1;
     confirm_data_t* data = (confirm_data_t*)component->data;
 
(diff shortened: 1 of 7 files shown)


benma:
> LGTM. Looks nice on a dev device here. Many literal numbers which I'm not at all familiar with though. Would've appreciated more comments around.

Me too ;) Copied straight from @NickeZ PR. I didn't study the details but just tested that it looks good.

> Are there any docs on the coordinate system and how µGUI + OLED + device orientation fit together? I think I got it by reading the code but would appreciate the docs a lot.

Not that I know of.

> Why are screen dimensions different in ttf2ugui in https://github.com/digitalbitbox/ttf2ugui/blob/d923bac28cd0aafc12b28be692fac1ee59f4f184/ttf2ugui.c#L42-L43? From what I understand, we use the tool to convert fonts.

No clue. @NickeZ might know, he's dealt with the fonts a lot in the past months.

NickeZ:
> Are there any docs on the coordinate system and how µGUI + OLED + device orientation fit together? I think I got it by reading the code but would appreciate the docs a lot.

Not that I'm aware of, maybe there is something on the µGUI website. it is a pretty basic framework that simply calls an implementation of `set_pixel(x,y)`-ish for every pixel you want to draw.

Device orientation is handled by the driver. I rewrote the blitting code so that the buffer could stay the same and we only changed two driver variables instead. You'll find it in the `mirror` or `oled_orientation` code or something like that. Previously the buffer was instead flipped in software (I think).
 
> Why are screen dimensions different in ttf2ugui in https://github.com/digitalbitbox/ttf2ugui/blob/d923bac28cd0aafc12b28be692fac1ee59f4f184/ttf2ugui.c#L42-L43? From what I understand, we use the tool to convert fonts.

Those width/height are a red herring. The tool `ttf2ugui` can display the resulting font in the terminal using asterixes for pixels. I think those defines are only defining how big that "virtual" screen should be. you can try it with the `--show <TESTSTRING>` option or something like that...

x1ddos:
The only thing I could find is a Reference Guide in PDF on this page but it doesn't explain fonts or anything else, unfortunately: http://embeddedlightning.com/ugui/.

NickeZ:
I think, like always, the code is the documentation :P

PR #484, opened 2020-05-11
components/confirm: improved centering

benma:
Fix centering algorithm - before it would center the text in the whole
screen. Now it is centering it in the body part, which is the full
screen without the title bar. This leads to more natural looking
positioning.

### DIFF
--- a/src/ui/components/confirm.c
+++ b/src/ui/components/confirm.c
@@ -15,6 +15,7 @@
 #include "confirm.h"
 #include "../event.h"
 #include "confirm_gesture.h"
+#include "empty.h"
 #include "icon_button.h"
 #include "label.h"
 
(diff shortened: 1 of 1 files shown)


benma:
This puts e.g. the address when confirming an address a bit lower right now.

When we increase the font size in the title and body of confirm dialogs, this will still work nicely. The original PR (https://github.com/digitalbitbox/bitbox02-firmware/pull/416/) was doing more manual work to fix the confirm dialog with bigger sizes, such as manually estimating offsets for titles with and without newlines:

https://github.com/digitalbitbox/bitbox02-firmware/pull/416/files#diff-e2850f41a2f354d9355a3c940009f93aR128-R139

and using custom offsets, reducing font size in some places, etc.

src/ui/components/confirm.c
```diff
@@ -20,6 +20,7 @@
 
 #include <hardfault.h>
 #include <screen.h>
+#include <ui/fonts/arial_fonts.h>
```
x1ddos, line 23: Why is this include needed only now. What happens if you don't include?

benma: oops that's a leftover from when i used a font inside _render(). will remove again

benma:
Added a commit to use the empty container (recently added) instead of basically replicating it.

PR #486, opened 2020-05-12
font-size: bigger input letters in trinary input

benma:


src/ui/components/trinary_input_char.c
```diff
@@ -320,6 +336,8 @@ component_t* trinary_input_char_create(
         Abort("Error: malloc trinary char");
     }
     memset(component, 0, sizeof(component_t));
+
+    data->font = &font_password_9X9;
```
x1ddos, line 340: Wouldn't you want to "start" with the bigger font?

benma: ignoring that it probably does not matter (the font is set every time above anyway), imho 9x9 is more natural to start with, as by default all letters are shown in small font, and the bigger font is only used when you zoom into a group.

x1ddos:
ah you're right of course. it starts with many smaller letters indeed.
makes total sense.

PR #490, opened 2020-05-13
font-size: tweak gesture confirm arrow position

benma:
Slightly to the right to make more space for the body.

Straight from #416.

### DIFF
--- a/src/ui/components/confirm_gesture.c
+++ b/src/ui/components/confirm_gesture.c
@@ -72,7 +72,7 @@ static void _render(component_t* component)
 
     // Draw the top arrow
     const uint16_t padding = 1;
-    x = SCREEN_WIDTH * 15 / 16 - (arrow_height * 2 + 1);
+    x = SCREEN_WIDTH * 15 / 16 - (arrow_height + arrow_height / 2);
     y0 = padding + data->active_count / SCALE;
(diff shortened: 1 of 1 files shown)


x1ddos:
This was kind of expected, right:

```
/bitbox02-firmware/src/ui/components/confirm_gesture.c:75:9: error: result of integer division used
in a floating point context; possible loss of precision [bugprone-integer-division,-warnings-as-errors]

    x = SCREEN_WIDTH * 15 / 16 - (arrow_height * 1.5);

/bitbox02-firmware/src/screen.h:34:22: note: expanded from macro 'SCREEN_WIDTH'

#define SCREEN_WIDTH 128
```

src/ui/components/confirm_gesture.c
```diff
@@ -72,7 +72,7 @@ static void _render(component_t* component)
 
     // Draw the top arrow
     const uint16_t padding = 1;
-    x = SCREEN_WIDTH * 15 / 16 - (arrow_height * 2 + 1);
+    x = SCREEN_WIDTH * 15 / 16 - (arrow_height * 1.5);
```
x1ddos, line 75: how about `arrow_height + arrow_height / 2`?

benma: now i see why niklas did this (this is how he did in in his PR). will do

PR #497, opened 2020-06-06
ci: add github workflow

x1ddos:
This is an alternative to Travis CI.
It runs the same scripts in docker on Ubuntu 18.04.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,19 @@
+# See reference docs at
+# https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions
+name: ci
+on: [push, pull_request]
+
+jobs:
+  linux-docker:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Clone the repo
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: vars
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Clone the repo
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: vars
+        run: echo "---${{github.head_ref}}---"
```
x1ddos, line 15: One thing I'm not sure of is whether `github.head_ref` is the same as `TRAVIS_BRANCH` when a PR is sent from another repo, but I guess we'll find out soon.

x1ddos:
Github was first:

image

.ci/run-container
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
benma, line 1: `run-container` -> would be nice to keep `ci` in the name, `run_container` sounds a bit too generic

x1ddos: ok. renamed to `run-container-ci`.

x1ddos:
:+1: to run them both for a while

PR #498, opened 2020-06-12
commander: split protobuf encoding/decoding into new file

benma:
Easier to call from Rust. (work paused again, but putting in this commit so I don't have to look for it again later).

### DIFF
--- a/src/commander/protobuf.c
+++ b/src/commander/protobuf.c
@@ -1,0 +1,35 @@
+// Copyright 2020 Shift Crypto AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "protobuf.h"
+
+#include <hardfault.h>
+
+#include <pb_decode.h>
+#include <pb_encode.h>
+
+bool protobuf_decode(const in_buffer_t* buf, Request* request_out)
+{
+    pb_istream_t in_stream = pb_istream_from_buffer(buf->data, buf->len);
+    return pb_decode(&in_stream, Request_fields, request_out);
+}
+
+void protobuf_encode(buffer_t* buf_out, const Response* response)
+{
+    pb_ostream_t out_stream = pb_ostream_from_buffer(buf_out->data, buf_out->max_len);
+    if (!pb_encode(&out_stream, Response_fields, response)) {
+        Abort("Abort: pb_encode");
+    }
+    buf_out->len = out_stream.bytes_written;
+}
--- a/src/commander/protobuf.h
+++ b/src/commander/protobuf.h
@@ -1,0 +1,35 @@
+// Copyright 2020 Shift Cryptos AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef _COMMANDER_PROTOBUF_H_
+#define _COMMANDER_PROTOBUF_H_
+
+#include <compiler_util.h>
+#include <util.h>
+
+#include <hww.pb.h>
+
+#include <stdbool.h>
+
+/**
+ * Parses an api protobuf request.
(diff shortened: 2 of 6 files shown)


x1ddos:
Hm... that's a travis script. Doesn't seem to work on github. I'll fix it.

```
+ CONTAINER=shiftcrypto/firmware_v2:16
+ '[' /home/runner/work/bitbox02-firmware/bitbox02-firmware == pull ']'
+ WORKSPACE_DIR=/home/runner/work/bitbox02-firmware/bitbox02-firmware
+ '[' -z /home/runner/work/bitbox02-firmware/bitbox02-firmware ']'
+ TARGET_BRANCH=protobuf
+ '[' '' == true ']'
+ git remote set-branches --add origin protobuf
+ git fetch origin
fatal: couldn't find remote ref refs/heads/protobuf
```

x1ddos:
> Hm... that's a travis script. Doesn't seem to work on github. I'll fix it.
> 
> ```
> + CONTAINER=shiftcrypto/firmware_v2:16
> + '[' /home/runner/work/bitbox02-firmware/bitbox02-firmware == pull ']'
> + WORKSPACE_DIR=/home/runner/work/bitbox02-firmware/bitbox02-firmware
> + '[' -z /home/runner/work/bitbox02-firmware/bitbox02-firmware ']'
> + TARGET_BRANCH=protobuf
> + '[' '' == true ']'
> + git remote set-branches --add origin protobuf
> + git fetch origin
> fatal: couldn't find remote ref refs/heads/protobuf
> ```

Sent https://github.com/digitalbitbox/bitbox02-firmware/pull/499 to fix this.

src/commander/protobuf.h
```diff
@@ -0,0 +1,29 @@
...
+#include <compiler_util.h>
+#include <util.h>
+
+#include <hww.pb.h>
+
+#include <stdbool.h>
+
+USE_RESULT bool protobuf_decode(const in_buffer_t* in_buf, Request* request_out);
+
+void protobuf_encode(const Response* response, buffer_t* buf_out);
```
x1ddos, line 27: Doc comments please? Especially for encode which aborts on error but it doesn't say so anywhere.

src/commander/protobuf.c
```diff
@@ -0,0 +1,35 @@
...
+#include <pb_decode.h>
+#include <pb_encode.h>
+
+bool protobuf_decode(const in_buffer_t* buf, Request* request_out)
+{
+    pb_istream_t in_stream = pb_istream_from_buffer(buf->data, buf->len);
+    return pb_decode(&in_stream, Request_fields, request_out);
+}
+
+void protobuf_encode(const Response* response, buffer_t* buf_out)
```
x1ddos, line 28: Would've inverted the args position, to match pb_encode and somehow it also makes sense to me in general - the first arg is where the output is. Many stdlib functions like `write(2)` are similar.

benma:
> This is very nice but why stop at commander's protobuf. How about src/backup.c and restore.c?

Unsure what you mean? I can't refactor the whole commander in one PR ;)

The mid term goal is to have Rust use those functions to protobuf encode/decode, and then be able to process api calls in an async function, slowly replacing commander.

> Also, I'm seeing some unused `#include <pb_encode.h>` and `#include <pb_decode.h>` around. 

Where?

x1ddos:
> > This is very nice but why stop at commander's protobuf. How about src/backup.c and restore.c?
> 
> Unsure what you mean?

I meant for backup and restore to also somehow use these new `protobuf_decode` and `protobuf_encode` functions. So, they wouldn't be in src/commander but probably in src/.

>> Also, I'm seeing some unused #include <pb_encode.h> and #include <pb_decode.h> around.
>
> Where?

Like here: https://github.com/digitalbitbox/bitbox02-firmware/blob/2f0ab993/src/restore.c#L26. I don't believe restore uses `pb_encode`. It only decodes. Or did I miss it? Similarly https://github.com/digitalbitbox/bitbox02-firmware/blob/2f0ab993/src/backup.c#L23 and some test/unit-tests - grep for "pb_decode" and "pb_encode".

x1ddos:
Hey, can you rebase onto the latest master? I want to see if github CI is happy now.

benma:
> I meant for backup and restore to also somehow use these new `protobuf_decode` and `protobuf_encode` functions.

But those don't match the signature, the current functions use `Request` and `Response`, and are only for the commander.

unused includes: will check

src/commander/protobuf.h
benma: done

x1ddos:
> But those don't match the signature, the current functions use `Request` and `Response`, and are only for the commander.

Right. I imagined something like:

```c
// src/protobuf.h
void protobuf_encode(buffer_t* buf_out, const pb_field_t fields[], const void *src);
```

and in Rust I guess you'd have some stronger types.

x1ddos:
Although, now that I think about it, backup/restore use protobufs for a very different purpose. It just happen to be protobufs too like commander, but they don't have to use the same wire format. Probably best to leave backup/restore as is.

benma:
> > But those don't match the signature, the current functions use `Request` and `Response`, and are only for the commander.
> 
> Right. I imagined something like:
> 
> ```c
> // src/protobuf.h
> void protobuf_encode(buffer_t* buf_out, const pb_field_t fields[], const void *src);
> ```
> 
> and in Rust I guess you'd have some stronger types.

nack, can do this in a later step if needed. this seems easier to use in rust, as you can use the exported struct without weird void* casting things.

one step at a time, i don't think backups will be ported anytime soon to rust, so this will go a long way.

PR #500, opened 2020-06-14
apps/btc/sign: move init validation to a new file

benma:
btc_sign.c is growing too big.

With the next commits we will allow mixed inputs, so the init script
configs validation will be extended.

This commit does not change any semantics yet, just moving code.

### DIFF
--- a/src/apps/btc/btc_sign.c
+++ b/src/apps/btc/btc_sign.c
@@ -1,4 +1,5 @@
 // Copyright 2019 Shift Cryptosecurity AG
+// Copyright 2020 Shift Crypto AG
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


src/apps/btc/btc_sign_validate.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <compiler_util.h>
+#include <hww.pb.h>
+#include <stddef.h>
+/**
+ * When starting a signing workflow with the BTCSignInit request, this function is used to validate
+ * the script configs in the request.
+ *
+ * For multisig, this means confirming with the user that they are spending from the correct
+ * account.
```
benma, line 26: here and in `app_btc_sign_validate_input_change_script_config`, I will explain in detail what is validated after we change to allow mixed inputs.

src/apps/btc/btc_sign.c
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Shift Cryptosecurity AG
+// Copyright 2020 Shift Cryptosecurity AG
```
x1ddos, line 1: I'd keep the line as is and add a new with with 2020 Shift Crypto AG.

src/apps/btc/btc_sign_validate.h
x1ddos: :+1:

src/apps/btc/btc_sign.c
benma: oops of course.

PR #501, opened 2020-06-14
apps/btc/sign: allow multiple script configs in a tx

benma:
Before, one script config (address type) was allowed for all inputs
and changes.

This commit changes the API of the signing workflow to allow
specifying multiple script configurations that can be used in a tx,
instead of only one.

In each output and change, the `script_config_index` now references
which script config to use.

The current commit still only allows only one (see TODO), but
implements the change on the API level. The TODO will be handled in a
subsequent commit.

### DIFF
--- a/messages/btc.options
+++ b/messages/btc.options
@@ -1,4 +1,5 @@
 // Copyright 2019 Shift Cryptosecurity AG
+// Copyright 2020 Shift Crypto AG
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


messages/btc.options
```diff
@@ -14,12 +14,13 @@
 
 BTCScriptConfig.Multisig.xpubs max_count:15
 BTCPubRequest.keypath max_count:10
-BTCSignInitRequest.keypath_account max_count:10
+BTCSignInitRequest.script_configs max_count:3
```
x1ddos, line 17: Why only 3? Is it basically to allow a max of one `P2WPKH_P2SH`, `P2WPKH` and `Multisig` each? What about the other coins? I guess I don't have a clear picture of how the unified accounts will look like. Are they going to be grouped by the type, like BTC, ETH, etc?

benma: Yeah, we currently only have really 2 types that can be used at once (segwit and native segwit), and in the future possibly also taproot, so 2 would actually be enough.

multisig is treated separately.

the number here is kind of arbitrary, it could also be 10. it's just so nanopb knows how much memory to allocate for the array.

benma: when you start a sign workflow, you specify the btc-like coin: currently BTC or LTC.

ETH is completely separate and does not need unified accounts (there is only one eth account now, while there are two to three btc accounts).

x1ddos: Oh, right. Somehow I was thinking about inputs, not accounts. And multisig is in `BTCScriptConfig_multisig_tag`.
Yeah, probably good to keep this max count low.

benma:
>     * // Singing flow:

Done for the first comment. 2nd comment looks like no change is needed?

x1ddos:
> > ```
> > * // Singing flow:
> > ```
> 
> Done for the first comment. 2nd comment looks like no change is needed?

Yeah, just wasn't sure where you wanted to put it. :+1: 
Looks good!

PR #502, opened 2020-06-15
unit-test: more negative tests in test_btc_sign

benma:


test/unit-test/test_btc_sign.c
```diff
@@ -166,6 +166,76 @@ static void _test_btc_sign_init(void** state)
...
+        };
+        invalid.script_configs[1] = sc;
+
+        assert_int_equal(APP_BTC_OK, app_btc_sign_init(&invalid, &next));
+        tst_app_btc_reset();
+        invalid.script_configs[0].keypath[2]++;
+        assert_int_equal(APP_BTC_ERR_INVALID_INPUT, app_btc_sign_init(&invalid, &next));
+    }
+    { // can't mix script configs from different bip44 accounts
+        // (mixing input scripts is allowed, but only from the same bip44 account).
```
x1ddos, line 210: I think this comment refers to the previous test and the other way around: this one is about multisig and the previous is about different accounts. So, swap them.

benma: very good catch, fixed

PR #506, opened 2020-06-18
noise: pop screens on new handshake

benma:
Not doing so is a regression from the rewrite of noise to Rust. This
removes the 'See the BitBoxApp' screen and goes to the default waiting
screen.

See
https://github.com/digitalbitbox/bitbox02-firmware/commit/d90d5e14207fcb1f79a69492c7cee78f71b2ea11#diff-ae8fe1d4fe4e70004f0f0ee3300406e8L189

In the future, we should consider showing the 'See the BitBoxApp'
screen again once the app is closed.

The screen stack pop was moved to the beginning of the handshake as
opposed to the end. This avoids a short flash of the 'See the
BitBoxApp' screen (as there are multiple api calls to finish the
pairing process), and we know the app (or a client) connected when a
handshake request comes in.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -360,6 +360,7 @@ add_custom_target(rust-bindgen
     --whitelist-function screen_print_debug
     --whitelist-function ui_screen_stack_push
     --whitelist-function ui_screen_stack_pop
+    --whitelist-function ui_screen_stack_pop_all
     --whitelist-function screen_process
     --whitelist-function label_create
     --whitelist-function confirm_create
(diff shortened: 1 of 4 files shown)


benma:
Since noise is now async, there is a very short flash of the `See the BitBoxApp` after unlock before we enter the default wait screen. I am looking into how to remove that.

x1ddos:
Forgot to mention. On Windows, the device can be opened by multiple processes at the same time, I believe. For instance, the app and the bridge. Not sure what happens then and I don't think it is actually supported in the firmware.

Either way, it's not new but I figured it's somehow related.

PR #508, opened 2020-06-19
rust/clippy: disable redundant_clone

benma:
Has false positives that can be very subtle, so best to play it
safe.

https://github.com/async-rs/async-task/issues/23

### DIFF
--- a/src/rust/bitbox02-rust-c/src/bitboxbase.rs
+++ b/src/rust/bitbox02-rust-c/src/bitboxbase.rs
@@ -207,7 +207,7 @@ pub extern "C" fn bitboxbase_state_set_not_alive() {
 #[no_mangle]
 pub extern "C" fn bitboxbase_state_get() -> BitBoxBaseBackgroundState {
     let state = unsafe { &STATE };
-    state.state.clone()
+    state.state
 }
 
 /// # Safety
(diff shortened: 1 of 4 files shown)


src/rust/bitbox02-rust-c/src/bitboxbase.rs
```diff
@@ -207,7 +207,7 @@ pub extern "C" fn bitboxbase_state_set_not_alive() {
 #[no_mangle]
 pub extern "C" fn bitboxbase_state_get() -> BitBoxBaseBackgroundState {
     let state = unsafe { &STATE };
-    state.state.clone()
+    state.state
```
benma, line 210: @NickeZ clippy warned this

https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy

Sounds legit, do you remember why there was a clone() here?

x1ddos:
Hm... I'm quite disappointed with clippy at this point. It's always best to leave uncertain things as is and avoid false positives.

benma:
> Hm... I'm quite disappointed with clippy at this point. It's always best to leave uncertain things as is and avoid false positives.

agree, at least the default linters should be 100% false positive free, when it comes to semantics.

src/rust/bitbox02-rust-c/src/bitboxbase.rs
NickeZ: Nope, can't remember. Usually I add `clone()`s as temporary fixes to make the borrowchecker happy. Having many clones is a bit of a code smell.

PR #511, opened 2020-06-21
rust: bind memory_set_device_name

benma:
Easiest/first api call to convert to Rust to make it async, so this is
a prerequisite.

### DIFF
--- a/src/rust/bitbox02/src/memory.rs
+++ b/src/rust/bitbox02/src/memory.rs
@@ -1,4 +1,5 @@
 // Copyright 2020 Shift Cryptosecurity AG
+// Copyright 2020 Shift Crypto AG
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -11,6 +12,22 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
+// deduct one for the null terminator.
+pub const DEVICE_NAME_MAX_LEN: usize = bitbox02_sys::MEMORY_DEVICE_NAME_MAX_LEN as usize - 1;
+
+/// `name.as_bytes()` must be smaller or equal to
+/// `DEVICE_NAME_MAX_LEN`, otherwise this function panics.
+pub fn set_device_name(name: &str) -> Result<(), ()> {
+    match unsafe {
+        bitbox02_sys::memory_set_device_name(
+            crate::str_to_cstr_force!(name, DEVICE_NAME_MAX_LEN).as_ptr(),
+        )
+    } {
+        true => Ok(()),
+        false => Err(()),
+    }
+}
 
 pub fn is_initialized() -> bool {
     unsafe { bitbox02_sys::memory_is_initialized() }
--- a/src/rust/bitbox02/src/ui.rs
+++ b/src/rust/bitbox02/src/ui.rs
@@ -19,6 +19,9 @@ extern crate alloc;
 use crate::password::Password;
 use alloc::boxed::Box;
 
+// Taking the constant straight from C, as it's excluding the null terminator.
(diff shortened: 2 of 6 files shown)


src/rust/bitbox02/src/memory.rs
```diff
@@ -12,6 +13,20 @@
...
 // limitations under the License.
 
+// corresponds to C MEMORY_DEVICE_NAME_MAX_LEN - 1 (no null terminator).
+pub const DEVICE_NAME_MAX_SIZE: usize = 63;
+
+/// `name.as_bytes()` must be smaller or equal to
+/// `DEVICE_NAME_MAX_SIZE`, otherwise this function panics.
+pub fn set_device_name(name: &str) -> Result<(), ()> {
+    match unsafe {
+        bitbox02_sys::memory_set_device_name(crate::str_to_cstr_force!(name, 63).as_ptr())
```
benma, line 23: @NickeZ I always wondered why I can't use constants in this macro:

```
error: no rules expected the token `DEVICE_NAME_MAX_SIZE`
  --> ../src/rust/bitbox02/src/memory.rs:24:45
   |
24 |             crate::str_to_cstr_force!(name, DEVICE_NAME_MAX_SIZE).as_ptr(),
   |                                             ^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
   | 
  ::: bitbox02/src/util.rs:80:1
   |
80 | macro_rules! str_to_cstr_force {
   | ------------------------------ when calling this macro

error: no rules expected the token `DEVICE_NAME_MAX_SIZE`
```

Is this a general issue or can the macro be fixed to accept constants as args?

NickeZ: Easy to fix I think. The macro accepts a "literal". But you can change that to something else like "expr" for expression. The complete list of things you can match against in a macro is here: https://doc.rust-lang.org/reference/macros-by-example.html#metavariables

Macros "1.0" or "macros by example" are like a match statement. So `($input:expr, $len:literal)`, will match `str_to_cstr_force(<any expression>, <any literal>)`. You would want `($input:expr, $len:expr)` I think.

NickeZ: The expression must be constant at compile time though, because that number is later used to set the size of an array: `let mut buf = [0u8; $len + 1];`.

benma: great, that worked perfectly. thanks!

src/rust/bitbox02/src/ui.rs
```diff
@@ -19,6 +19,8 @@ extern crate alloc;
 use crate::password::Password;
 use alloc::boxed::Box;
 
+const MAX_LABEL_SIZE: usize = bitbox02_sys::MAX_LABEL_SIZE as _;
```
x1ddos, line 22: Why not `... as usize - 1` to "deduct one for the null terminator" like with `DEVICE_NAME_MAX_LEN`?

x1ddos: I mean, I see `MAX_LABEL_SIZE` says it excludes null terminator in the doc comments but `MEMORY_DEVICE_NAME_MAX_LEN` doesn't. Seems inconsistent and easy to miss.

benma: Yeah it's inconsistent. Not sure which way is better, without null terminator is better for Rust, but with null terminator is better in C (no need to do `+1` everywhere). Maybe I'll add a 2nd constant for each, `..._WITH_NULLTERMINATOR` :sweat_smile:

x1ddos: Maybe just add a comment here why it doesn't have `-1`.

Either way, I think `_WITH_NULL` would be sufficient. No need for long `TERMINATOR` :)

x1ddos: Longer term though I hope we'll unify null/no null constants.

benma: > Maybe just add a comment here why it doesn't have -1.

Done this for now, as suggested.

PR #512, opened 2020-06-23
Reactor Backup, part I

benma:


src/workflow/restore.c
```diff
@@ -14,11 +14,11 @@
 
 #include <workflow/restore.h>
 
-#include <backup.h>
+#include <backup/backup.h>
```
x1ddos, line 17: I don't believe restore.c actually needs backup.h. Didn't find any reference.
I know you just moved it around but this is a refactoring PR might as well remove.

src/backup/backup.c
```diff
@@ -296,6 +255,7 @@ backup_error_t backup_create(uint32_t backup_create_timestamp, uint32_t seed_bir
         // BACKUP_ERR_CHECK.
         // The caller could try again.
         uint8_t read_content[SD_MAX_FILE_SIZE];
+        CLEANUP_BACKUP_BYTES(read_content);
```
x1ddos, line 260: How about moving this and `_cleanup_backup_bytes` to backup_common.h as well, since it already contains similar `backup_cleanup_backup` and `backup_cleanup_backup_data`?

src/workflow/restore.c
benma: :+1:

src/backup/backup.c
benma: No need, it is only used here, so a static function suffices.

PR #513, opened 2020-06-24
apps/btc: do not verify rbf&locktime if locktime is 0

benma:
RBF verification is not a security feature (as described in the
comment), but merely a sanity check, meant for advanced wallets like
Electrum.

Locktime on the other hand might be security relevant, when the
referenced block is in the future.

Until now, we verified the locktime if it is enabled anyway, also 0,
as it technically is considered. Practically however, a locktime of 0
has no effect, as the current block is always > 0.

We want to be able to opt-in to RBF with transactions made in the
BitBoxApp. Currently if doing so, the user would be prompted to verify

"Transaction is RBF. Locktime on block: XYZ"

Which is gibberisch to an average user. Especially the latter, a user
can not really verify.

We do a workaround and skip verification of RBF and locktime if the
locktime is zero, with the effect that the BitBoxApp can make RBF
transcations without disturbing the user with a complicated screen on
the device. This should have no security downside, and UX is better.

### DIFF
--- a/src/apps/btc/btc_sign.c
+++ b/src/apps/btc/btc_sign.c
@@ -801,17 +801,21 @@ app_btc_result_t app_btc_sign_output(
             }
         }
 
-        // This is not a security feature, a transaction that is not rbf
-        // and has a locktime of 0 will not be verified.
-        if (_locktime_applies || _rbf == CONFIRM_LOCKTIME_RBF_ON) {
-            // The RBF nsequence bytes are often set in conjunction with a locktime,
-            // so verify both simultaneously.
-            // There is no RBF in Litecoin, so make sure it is disabled.
-            if (!_coin_params->rbf_support) {
-                _rbf = CONFIRM_LOCKTIME_RBF_DISABLED;
-            }
-            if (!apps_btc_confirm_locktime_rbf(_init_request.locktime, _rbf)) {
-                return _error(APP_BTC_ERR_USER_ABORT);
+        // A locktime of 0 will also not be verified, as it's certainly in the past and can't do any
+        // harm.
+        if (_init_request.locktime > 0) {
+            // This is not a security feature, a transaction that is not rbf or has a locktime of 0
+            // will not be verified.
+            if (_locktime_applies || _rbf == CONFIRM_LOCKTIME_RBF_ON) {
+                // The RBF nsequence bytes are often set in conjunction with a locktime,
+                // so verify both simultaneously.
+                // There is no RBF in Litecoin, so make sure it is disabled.
+                if (!_coin_params->rbf_support) {
+                    _rbf = CONFIRM_LOCKTIME_RBF_DISABLED;
+                }
+                if (!apps_btc_confirm_locktime_rbf(_init_request.locktime, _rbf)) {
+                    return _error(APP_BTC_ERR_USER_ABORT);
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


benma:
Strange that clang-tidy only complains about this issue in this PR, not before:

https://github.com/digitalbitbox/bitbox02-firmware/pull/513/checks?check_run_id=804193839

```
1169 warnings generated.
/bitbox02-firmware/src/apps/btc/btc_common.h:65:5: error: parameter 'keypath_len' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions [readability-avoid-const-params-in-decls,-warnings-as-errors]
    const size_t keypath_len,
    ^~~~~~
/bitbox02-firmware/src/apps/btc/btc_common.h:66:5: error: parameter 'expected_coin' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions [readability-avoid-const-params-in-decls,-warnings-as-errors]
    const uint32_t expected_coin);
```

totally right, added commit to fix it, but why now? :sweat_smile:

x1ddos:
> Strange that clang-tidy only complains about this issue in this PR, not before
> totally right, added commit to fix it, but why now? sweat_smile

My guess is those top level `const`s were introduced in 5fd5cff which was pushed without creating a PR and check-tidy runs only on changed files in https://github.com/digitalbitbox/bitbox02-firmware/blob/009bdeb/.ci/check-tidy#L39 but the list of files came up empty because `git diff origin/master` compares between the same tip.

So, this was actually the first time clang-tidy checked those files after 5fd5cff was merged.

src/apps/btc/btc_sign.c
```diff
@@ -801,17 +801,21 @@ app_btc_result_t app_btc_sign_output(
...
-            if (!_coin_params->rbf_support) {
-                _rbf = CONFIRM_LOCKTIME_RBF_DISABLED;
-            }
-            if (!apps_btc_confirm_locktime_rbf(_init_request.locktime, _rbf)) {
-                return _error(APP_BTC_ERR_USER_ABORT);
+        // A locktime of 0 will also not be verified, as it's certainly in the past and can't do any
+        // harm.
+        if (_init_request.locktime > 0) {
+            // This is not a security feature, a transaction that is not rbf or has a locktime of 0
+            // will not be verified.
```
x1ddos, line 808: Reading it as written feels a bit scary: "a transaction ... wlil not be verified."
It's more like "the extra locktime/RBF user confirmation is skipped."

benma:
> > Strange that clang-tidy only complains about this issue in this PR, not before
> > totally right, added commit to fix it, but why now? sweat_smile
> 
> My guess is those top level `const`s were introduced in [5fd5cff](https://github.com/digitalbitbox/bitbox02-firmware/commit/5fd5cfffad9d0af3a2b320d4dc4f2b1e5693a8e0) which was pushed without creating a PR and check-tidy runs only on changed files in https://github.com/digitalbitbox/bitbox02-firmware/blob/009bdeb/.ci/check-tidy#L39 but the list of files came up empty because `git diff origin/master` compares between the same tip.
> 
> So, this was actually the first time clang-tidy checked those files after [5fd5cff](https://github.com/digitalbitbox/bitbox02-firmware/commit/5fd5cfffad9d0af3a2b320d4dc4f2b1e5693a8e0) was merged.

Didn't we make a PR from staging to master back then? :shrug: anyway

src/apps/btc/btc_sign.c
benma: true! fixed

x1ddos:
> Didn't we make a PR from staging to master back then? shrug anyway

I think it's one of those which slipped accidentally when you merged:

image

benma:
> I think it's one of those which slipped accidentally when you merged:

Of all things :see_no_evil:

PR #515, opened 2020-06-24
Backup refactor Part II

benma:


src/backup/backup_common.h
```diff
@@ -28,4 +54,18 @@ void backup_cleanup_backup_data(BackupData* backup_data);
...
+/**
+ * Fills the backup structure with backup data.
+ * @param[in] backup_create_timestamp The time at which the backup was created.
+ * @param[out] backup The backup structure filled with data.
+ * @param[out] backup_data The backup data structure filled with data.
+ * @param[out] encode_data Additional data required for encoding/decoding.
+ */
+backup_error_t backup_fill(
+    uint32_t backup_create_timestamp,
+    uint32_t seed_birthdate_timestamp,
```
x1ddos, line 66: Missing param doc for birthdate. I know you just moved it around but since you're refactoring, probably a good time to update.

src/backup/backup.h
```diff
@@ -19,6 +19,15 @@
 
 #include <backup.pb.h>
 
+#include <stddef.h>
+#include <stdint.h>
+
+/**
+ * Encodes the backup and returns the number of bytes written, or 0 if encoding failed.
```
x1ddos, line 26: `@return` already says exactly the same on the next line. This could instead mention that it serializes the protobuf. "Encodes" is pretty generic and it's already clear that it "encodes" from the function name. I know you just copied it but since it's a refactoring...

benma:
> What about test_backup.c? Seems like `_get_file_name` is quite useless:
> 
> ```c
> /**
>  * Returns the file name.
>  * @param[out] file_name The name of the file which includes a timestamp.
>  * @param[in] index The index of the backup.
>  */
> static void _get_file_name(char* file_name, uint8_t index)
> {
>     time_t local_timestamp = (time_t)(_current_timestamp);
>     struct tm* local_time = localtime(&local_timestamp);
>     static char local_timestring[100] = {0};
>     strftime(local_timestring, sizeof(local_timestring), "%a_%Y-%m-%dT%H-%M-%SZ", local_time);
> 
>     snprintf(file_name, 257, "backup_%s_%d.bin", local_timestring, index);
> }
> ```
> 
> All that can be precomputed based on the `_current_timestamp` like you did at the top of test_restore.c. Also, "Returns the file name." - yeah, very useful description. smile
> 
> Probably the same with `_get_directory_name`, and then both funcs can be removed from the codebase.

i found this out in parallel and fixed it already - pushed. also added a whole backup encoding fixture.

src/backup/backup_common.h
benma: fixed

src/backup/backup.h
benma: fixed

x1ddos:
Not sure why it's failing? Patch in https://github.com/digitalbitbox/bitbox02-firmware/pull/516 and it'll let you:

```
[  ERROR   ] --- 0x74 != 0x75
[   LINE   ] --- /home/alex/src/bb/firmware/test/unit-test/test_backup.c:205: error: Failure!
[  FAILED  ] test_backup_fixture
```

PR #517, opened 2020-06-25
backup: provide generator to backup_fill

benma:
It is sort of an identifer for what created the backup. The way it is
now, the fixture test in test_backup.c will fail when we change the
firmware version. Proiding it explicitly means we can test with any
string in the test.

### DIFF
--- a/src/backup/backup_common.c
+++ b/src/backup/backup_common.c
@@ -5,7 +5,6 @@
 #include <memory/memory.h>
 #include <sd.h>
 #include <util.h>
-#include <version.h>
 
 #include <pb_encode.h>
 #include <wally_crypto.h>
@@ -102,6 +101,7 @@ static bool _encode_backup_data(pb_ostream_t* ostream, const pb_field_t* field, 
 }
 
 backup_error_t backup_fill(
+    const char* generator,
     uint32_t backup_create_timestamp,
     uint32_t seed_birthdate_timestamp,
     Backup* backup,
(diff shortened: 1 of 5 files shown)


src/backup/backup_common.c
```diff
@@ -120,8 +121,8 @@ backup_error_t backup_fill(
         util_zero(backup_metadata->name, sizeof(backup_metadata->name));
         memory_get_device_name(backup_metadata->name);
         memset(backup_data, 0, sizeof(BackupData));
-        const char* firmware_v = DIGITAL_BITBOX_VERSION_SHORT;
-        snprintf(backup_data->generator, sizeof(backup_data->generator), "%s", firmware_v);
+
+        snprintf(backup_data->generator, sizeof(backup_data->generator), "%s", generator);
```
x1ddos, line 125: would a `memcpy` also work? no need for printf formatting at all.

benma: snprintf is pretty much always the right choice when it comes to string writing, copying, and so on. The reason is that without it, it is very easy to run into buffer overflows.

with `memcpy`, you need to define how many bytes to copy. `strlen(generator)` might be too long and overflow the target buffer (and also traverses the string twice instead of only once). `strcpy` has the same problem, `strncpy` might not result in a null terminated string, etc. etc.

benma:
> You mean the way it was before this change, right?

yes, fixed

src/backup/backup_common.c
x1ddos: right. I was thinking more like `backup_fill(const char generator[], ...)` but never mind. good points.

PR #520, opened 2020-06-26
nanopb: update to 0.4.1

benma:
Update submodule.

proto3_singular_msgs:true is added. Otherwise, fields that are another
protobuf message, such has Backup.backup_v1, BackupV1.content,
BackupContent.metadata, etc. must manually deal with setting a
corresponding `has_*` field. Converesly, in
BTCScriptConfigWithKeypath.script_config etc., the firmware would need
to validate that `has_script_config = true`, adding code bloat.

To generate the proto files, no global installation of the nanopb
generator is needed. The local one from tools/nanopb is used instead.

### DIFF
--- a/.ci/run-container-ci
+++ b/.ci/run-container-ci
@@ -25,7 +25,7 @@
 set -e
 set -x
 
-CONTAINER=shiftcrypto/firmware_v2:16
+CONTAINER=shiftcrypto/firmware_v2:17
 
 if [ "$1" == "pull" ] ; then
 	docker pull "$CONTAINER"
--- a/messages/CMakeLists.txt
+++ b/messages/CMakeLists.txt
@@ -1,5 +1,3 @@
-find_program(PROTOC_GEN_NANOPB protoc-gen-nanopb)
-
 set(PROTO_FILES
   hww.proto
   backup.proto
@@ -17,13 +15,6 @@ set(PROTO_FILES
 )
 string(REGEX REPLACE "\.proto" ".pb.c" OUTPUT_SOURCES "${PROTO_FILES}")
 string(REGEX REPLACE "\.proto" ".pb.h" OUTPUT_HEADERS "${PROTO_FILES}")
-
-if (PROTOC_GEN_NANOPB STREQUAL "PROTOC_GEN_NANOPB-NOTFOUND")
-  message(FATAL_ERROR
-    "Could not find 'protoc-gen-nanopb'.\n"
-    "Please install it from here: https://jpa.kapsi.fi/nanopb/download/\n"
-    "And add generator-bin/ to your path.")
-endif()
 
 # Create absolute paths to protobuf sources
 foreach(i ${PROTO_FILES})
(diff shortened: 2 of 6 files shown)


messages/CMakeLists.txt
```diff
@@ -55,7 +46,11 @@ add_custom_command(
   OUTPUT ${OUTPUT_SOURCES} ${OUTPUT_HEADERS}
   DEPENDS ${PROTO_FILES} ${PROTO_OPTION_FILES}
   COMMAND
-  ${PROTOC} "--proto_path=${CMAKE_CURRENT_SOURCE_DIR}" "--nanopb_out=-I${CMAKE_CURRENT_SOURCE_DIR}:${CMAKE_CURRENT_BINARY_DIR}" ${PROTO_FILES_ABSOLUTE}
+  ${CMAKE_SOURCE_DIR}/tools/nanopb/generator/protoc
```
x1ddos, line 49: Since you here, would it make sense to move nanopb to `/external` dir? For comparison, FIDO2's tinycbor is also in external.

benma: can you do it after this is merged? submodules are my weak point :innocent:

x1ddos: ah, yes course!

messages/btc.options
```diff
@@ -13,6 +13,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// skip `bool has_*` fields for submessages
+// https://github.com/nanopb/nanopb/blob/1fdc916e984b27cc0a3824427f70a15def567a2e/docs/migration.rst#submessages-now-have-has-field-in-proto3-mode
+* proto3_singular_msgs:true
```
x1ddos, line 18: Wouldn't hww.options also need this? It has `message Request { oneof request { ... message fields } }`. I know the code uses `which_xxx` but still, wouldn't you want to get rid of `has_xxx` in those too?

x1ddos:
Out of curiosity, did you measure before/after bin size?

messages/btc.options
benma: I grepped for `has_` in all generated files, only match now is `has_signature`, but that is our own field.

x1ddos: Ok! Let's see how it goes.

benma:
> Out of curiosity, did you measure before/after bin size?

Before: 433924b
After: 433956b

.ci/run-container-ci
```diff
@@ -25,7 +25,7 @@
 set -e
 set -x
 
-CONTAINER=shiftcrypto/firmware_v2:16
+CONTAINER=shiftcrypto/firmware_v2:17
```
x1ddos, line 28: You might want to update `latest` tag on docker hub:

image

I've just accidentally pulled a pile of gigabytes without specifying a tag, which defaults to `latest`, and later realized it's not the image I wanted.

Alternatively, drop the `latest` tag from docker hub, but I'm afraid some clients rely on its presents to fetch a manifest of images and it might break pulling.

PR #534, opened 2020-07-26
releases readme

benma:


benma:
> LGTM.
> 
> Would it make sense to add something like this next to https://guides.shiftcrypto.ch/bitbox02/advanced/reproducible-builds/?

That link just links back to the README I changed in this PR.

x1ddos:
> > LGTM.
> > Would it make sense to add something like this next to https://guides.shiftcrypto.ch/bitbox02/advanced/reproducible-builds/?
> 
> That link just links back to the README I changed in this PR.

Yeah, I saw it. It just isn't clear to me what goes to the guides and what to the readme.

benma:
> Yeah, I saw it. It just isn't clear to me what goes to the guides and what to the readme.

The goal was just to not have the same content in two places, which is why I asked sterny to link back here.

In general, it goes to the guides, unless there is a good reason otherwise. In case of reproducible builds, it makes more sense to have it close to the source (it's for devs and about the code anyway).

PR #535, opened 2020-07-27
py/bitbox02: Appnoise config

benma:


py/bitbox02/bitbox02/util.py
```diff
@@ -141,3 +144,68 @@ def set_app_static_privkey(self, privkey: bytes) -> None:
         data = self._read_cache()
         data.app_static_privkey = privkey
         self._write_cache(data)
+
+
+class BitBoxAppNoiseConfig(bitbox_api_protocol.BitBoxNoiseConfig):
+    """
+    Noise config that reads and stores the noise keys in the same location as the BitBoxApp.
```
x1ddos, line 151: I think it would be good to clarify the purpose of `NoiseConfigUserCache` in its docstring. If I were an integrating app which uses this package, I wouldn't know for sure which one to use: this or NoiseConfigUserCache or both.

py/bitbox02/bitbox02/util.py
```diff
@@ -141,3 +144,68 @@ def set_app_static_privkey(self, privkey: bytes) -> None:
...
+        data = self._read_cache()
+        if "appNoiseStaticKeypair" not in data:
+            return None
+
+        return base64.b64decode(data["appNoiseStaticKeypair"]["private"])
+
+    def set_app_static_privkey(self, privkey: bytes) -> None:
+        data = self._read_cache()
+        pubkey = KeyPair25519.from_private_bytes(privkey).public_bytes
+        data["appNoiseStaticKeypair"] = {
```
x1ddos, line 207: Make a class attributes for `"appNoiseStaticKeypair"` and `"deviceNoiseStaticPubkeys"`? Would make it easier to proof-read and might avoid potential typos in the future.

py/bitbox02/bitbox02/util.py
```diff
@@ -141,3 +144,68 @@ def set_app_static_privkey(self, privkey: bytes) -> None:
...
+            )
+        elif system == "Darwin":
+            folder = Path(os.environ["HOME"]) / "Library" / "Application Support"
+        elif system == "Windows":
+            folder = Path(os.environ["APPDATA"])
+        else:
+            raise NotImplementedError("Unknown system: {}".format(system))
+        self._filename = folder / "bitbox" / "bitbox02" / "bitbox02.json"
+
+    def _read_cache(self) -> Any:
```
x1ddos, line 172: Not really a cache, is it. How about `_read_config` and `_write_config`?

py/bitbox02/bitbox02/util.py
benma: Done.

Ideally you should use none though, but your custom one ;) (also e.g. to implement the attestation check).

py/bitbox02/bitbox02/util.py
benma: Done (just _read and _write)

py/bitbox02/bitbox02/util.py
benma: What do you mean? In the end it has to be in the dict serialized as json anyway?

py/bitbox02/bitbox02/util.py
x1ddos: Well, I could still subclass `BitBoxAppNoiseConfig`, right, which would at least give me the noise config.

py/bitbox02/bitbox02/util.py
x1ddos: It's just they are repeating. I imagined something like this:

```python
class BitBoxAppNoiseConfig(bitbox_api_protocol.BitBoxNoiseConfig):
    DEVICE_NOISE_STATIC_PUBKEYS = "deviceNoiseStaticPubkeys"
    APP_NOISE_STATIC_KEYPAIR = "appNoiseStaticKeypair"
    ...
    data.get(DEVICE_NOISE_STATIC_PUBKEYS)
    data[APP_NOISE_STATIC_KEYPAIR] = ...
```

x1ddos: Well, I meant with `self.XXX` :)

benma: ah of course. done

x1ddos:
:heavy_check_mark:

PR #537, opened 2020-08-04
workflow: use rust's status workflow

benma:


src/workflow/status.c
```diff
@@ -80,19 +81,9 @@ workflow_t* workflow_status(
...
-    workflow_blocking_unblock();
-}
-
-/**
- * Blocking wrapper for workflow_status().
+ * Blocking status.
  */
 void workflow_status_blocking(const char* msg, bool status_success)
 {
-    workflow_stack_start_workflow(workflow_status(msg, status_success, _unlock_cb, NULL));
```
benma, line 96: I'll do a bunch more such commits, and in the end we can remove all workflow_* things.

test/device-test/src/test_smarteeprom.c
```diff
@@ -87,5 +88,8 @@ int main(void)
         _screen_print_debug("SmartEEPROM enabled.\nRebooting.", 2000);
         _reset_mcu();
     }
+    for (;;) {
+        screen_process();
+    }
```
benma, line 93: without this, there is a `undefined reference to screen_process` linking error. I have no idea why this would happen here.

I just know:
- somehow it is LTO-optmized away
- Rust code calls it, which is why there is a reference error

Unclear to me is:
- Why it is optimized away. This does not happen for other C functions that are solely used in Rust
- Why it does not happen for other device tests, like `test_usb_composed_ep_in.c`, which also does not use any workflow/screen_process stuff.

Would be good to know, but not a high prio, since it is only a device test.

x1ddos: Please add a comment in the code? So, we'll remember to come back to this later maybe. A comment wouldn't hurt anyway.

PR #541, opened 2020-08-04
workflow: use rust's unlock bip39 workflow

benma:


benma:
> Don't need `#include "blocking.h"` anymore... or do you?

Fixed

src/workflow/unlock_bip39.c
```diff
@@ -92,14 +92,7 @@ workflow_t* workflow_unlock_bip39(void (*callback)(void* param), void* callback_
...
 
-static void _unlock_cb(void* param)
-{
-    (void)param;
-    workflow_blocking_unblock();
-}
-
 void workflow_unlock_bip39_blocking(void)
 {
-    workflow_stack_start_workflow(workflow_unlock_bip39(_unlock_cb, NULL));
```
x1ddos, line 103: I'm now looking at where `workflow_unlock_bip39` is used and it seems only unlock.c is calling it but since it's removed in https://github.com/digitalbitbox/bitbox02-firmware/pull/542, how about deleting `workflow_unlock_bip39`?

benma: Yeah it's coming, got a big queue of deletion commits :P

x1ddos: :+1:

PR #542, opened 2020-08-04
unlock: remove dead code and use of workflow_stack

benma:


src/workflow/unlock.h
```diff
@@ -15,25 +15,9 @@
...
- * correct password, or the device is reset after too many failed attempts. The callback will
- * receive information on whether the unlock was successful or not.
- *
- * @param[in] callback The callback that will be invoked when the workflow terminates.
- *                     It will receive a parameter that is true iff the unlock was successful,
- *                     and the user-defined parameter.
- * @param[in] callback_param The user-defined parameter that will be passed to the callback.
- * @return workflow_t object ready to be started.
- */
-workflow_t* workflow_unlock(void (*callback)(bool result, void* param), void* callback_param);
```
x1ddos, line 35: What about https://github.com/digitalbitbox/bitbox02-firmware/blob/4703425b00d1244e48b1fa7fcd9fcb8033905569/test/unit-test/framework/mock_workflow.c. Is it still needed?

benma: Good catch. Aaand it's gone.

test/unit-test/CMakeLists.txt
```diff
@@ -262,9 +261,9 @@ endforeach()
 # They are copied from https://github.com/google/u2f-ref-code/tree/master/u2f-tests/HID
 set(U2F_TESTS
    u2f_standard
-   "-Wl,--wrap=workflow_unlock"
+   ""
    u2f_hid
-   "-Wl,--wrap=workflow_status_create"
+   ""
```
x1ddos, line 266: Since there are no args anymore, might as well remove the empty strings and simplify the loop below?

```
set(U2F_TESTS
   u2f_standard
   u2f_hid
)
foreach(TEST_NAME ${U2F_TESTS})
  set(EXE test_${TEST_NAME})
  target_link_libraries(${EXE} PRIVATE
    ...
    u2f-util
    ${CMOCKA_LIBRARIES}
)
```

All the `math()` and `list()` would be gone. Simpler.

You'll say, but we might need them later, and I'd reply with "we can get them back from git commit history". :)

benma: Done :P

PR #543, opened 2020-08-04
more workflow dead code

benma:


src/workflow/unlock_bip39.c
```diff
@@ -13,84 +13,7 @@
...
-{
-    workflow_t* result = workflow_allocate(
-        _unlock_bip39_init, _unlock_bip39_cleanup, _unlock_bip39_spin, sizeof(unlock_bip39_data_t));
-    unlock_bip39_data_t* data = (unlock_bip39_data_t*)result->data;
-    data->callback = callback;
-    data->callback_param = callback_param;
-    return result;
-}
 
 void workflow_unlock_bip39_blocking(void)
```
x1ddos, line 18: I guess at some point, we can remove unlock_bip39.c and replace `workflow_unlock_bip39_blocking` calls directly with the rust version.

benma: yeah

PR #548, opened 2020-08-05
simplify idle_workflow and remove all workflow_ things

benma:


src/workflow/idle_workflow.c
```diff
@@ -48,36 +46,15 @@ static void _idle_workflow_init_communication(void)
...
-        _idle_workflow_cleanup,
-        _idle_workflow_spin,
-        sizeof(idle_workflow_data_t));
+    component_t* waiting_screen = waiting_create();
+    UG_ClearBuffer();
+    waiting_screen->f->render(waiting_screen);
+    UG_SendBuffer();
+    waiting_screen->f->cleanup(waiting_screen);
+#ifndef TESTING
+    delay_ms(1300);
```
benma, line 57: this is manually tested to feel aoout the same duration as before (showing logo to switching to `See the BitBoxApp`).

the delay was originally added by fabian/djb as a UX/visual thing.

src/workflow/idle_workflow.c
```diff
@@ -48,36 +46,16 @@ static void _idle_workflow_init_communication(void)
...
-}
-
-workflow_t* idle_workflow(void)
-{
-    return workflow_allocate(
-        _idle_workflow_init,
-        _idle_workflow_cleanup,
-        _idle_workflow_spin,
-        sizeof(idle_workflow_data_t));
+    component_t* waiting_screen = waiting_create();
```
x1ddos, line 51: Isn't this already done in screen_process.c in by default?

```
 50 static component_t* _get_ui_top_component(void)
 51 {
 52     component_t* result = ui_screen_stack_top();
 53     if (!result) {
 54         return _get_waiting_screen();
 55     }
 56     return result;
 57 }
 58

 34 static component_t* _get_waiting_screen(void)
 35 {
 36     static component_t* waiting_screen = NULL;
 37     if (waiting_screen == NULL) {
 38         waiting_screen = waiting_create();
 39         if (waiting_screen == NULL) {
 40             Abort("Could not create\nwaiting screen");
 41         }
 42     }
 43     return waiting_screen;
 44 }
```

I'm wondering why it needs to be repeated here.

benma: 1. because the mainloop running screen_process is not running yet.
2. we could use `workflow_blocking_block` to do it here, but then I could not use `delay_ms(...)`, but would have to add a counter and a callback to waiting_create() :+1: 

Btw, here it is done in the same way:

https://github.com/digitalbitbox/bitbox02-firmware/blob/2a8c435d0e0ec225a17846345110eecce5e4a2dd/src/reset.c#L36

x1ddos: In other words, would it still work if you removed `idle_workflow_blocking` from firmware.c in:

```
#if PLATFORM_BITBOX02 == 1
    orientation_screen_blocking();
#endif
    idle_workflow_blocking();
    firmware_main_loop();
    return 0;
}
```

x1ddos: Ah, sorry, we were 20sec apart.

benma: No, because after the delay, we have to move on to starting usb comm. and showing `See the BitBoxApp`.

x1ddos: True. I had a different scenario in mind:

1. start usb comms right away
2. loop with just `process_screen`

but then the logo wouldn't show up.

x1ddos: but that's not how it works at the moment :)

PR #550, opened 2020-08-07
workflow: move password_set to rust

benma:


src/rust/bitbox02-rust-c/src/workflow.rs
```diff
@@ -140,18 +140,13 @@ pub unsafe extern "C" fn rust_workflow_status_unlock_bip39_blocking() {
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn rust_workflow_password_enter_blocking(
-    title: crate::util::CStr,
-    special_chars: bool,
+pub unsafe extern "C" fn rust_workflow_password_enter_twice_blocking(
```
x1ddos, line 143: Ah, nice. So, we don't even need to export password::enter to C anymore because the whole unlock is in Rust. Cool

PR #552, opened 2020-08-11
apps/eth: add more erc20 tokens

benma:


x1ddos:
I wonder, how much does this add to the firmware size?

benma:
> I wonder, how much does this add to the firmware size?

51kB, though I can probably reduce this by ~10kB again with some small optimizations.

PR #553, opened 2020-08-11
apps/eth: reduce binary size

benma:
Repeating the same field everywhere is inefficient. This change saves
~6kB of binary size after the previous commit added ~51kB of token data.


x1ddos:
I didn't actually check all tests but CI did:

```
/bitbox02-firmware/test/device-test/src/test_token_pub.c: In function 'main':
/bitbox02-firmware/test/device-test/src/test_token_pub.c:51:35: error: '_erc20_params' undeclared (first use in this function);
```

PR #554, opened 2020-08-11
CMakeLists: bump version to v9.1.0

benma:
- Added more ERC20 tokens

### DIFF
--- a/test/unit-test/test_restore.c
+++ b/test/unit-test/test_restore.c
@@ -30,6 +30,7 @@
 #include <backup/restore.h>
 #include <sd.h>
 #include <util.h>
+#include <version.h>
 
 #define DEVICE_NAME "TestDeviceName"
 static const uint32_t _current_timestamp = 1553098951;
@@ -94,7 +95,7 @@ static int test_setup(void** state)
 
     assert_int_equal(
         backup_fill(
-            "v9.0.0",
+            DIGITAL_BITBOX_VERSION_SHORT,
             _current_timestamp,
             _mock_seed_birthdate,
             &_backup,
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,9 +89,9 @@ endif()
 #
 # Versions MUST contain three parts and start with lowercase 'v'.
 # Example 'v1.0.0'. They MUST not contain a pre-release label such as '-beta'.
-set(FIRMWARE_VERSION "v9.0.0")
-set(FIRMWARE_BTC_ONLY_VERSION "v9.0.0")
-set(FIRMWARE_BITBOXBASE_VERSION "v9.0.0")
+set(FIRMWARE_VERSION "v9.1.0")
+set(FIRMWARE_BTC_ONLY_VERSION "v9.1.0")
+set(FIRMWARE_BITBOXBASE_VERSION "v9.1.0")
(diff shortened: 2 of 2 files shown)


x1ddos:
Well, backup tests use `v9.0.0` but I kind of assumed it's separate from the actual release version.

x1ddos:
Maybe the quickest is to change https://github.com/digitalbitbox/bitbox02-firmware/blob/112a2378e8cb31bc7c94410eac7c1a1a4c0dba95/test/unit-test/test_restore.c#L97 and use `DIGITAL_BITBOX_VERSION_SHORT` instead of `"v9.0.0"`.

benma:
> Maybe the quickest is to change
> 
> https://github.com/digitalbitbox/bitbox02-firmware/blob/112a2378e8cb31bc7c94410eac7c1a1a4c0dba95/test/unit-test/test_restore.c#L97
> and use `DIGITAL_BITBOX_VERSION_SHORT` instead of `"v9.0.0"`.

it was that before until a4aab382a2817a9919b7f937ecf7a4abba74bb8e, which you reviewed :smile: 

> backup: provide generator to backup_fill
> It is sort of an identifer for what created the backup. The way it was
before, the fixture test in test_backup.c will fail when we change the
firmware version. Proiding it explicitly means we can test with any
string in the test.

x1ddos:
> it was that before until [a4aab38](https://github.com/digitalbitbox/bitbox02-firmware/commit/a4aab382a2817a9919b7f937ecf7a4abba74bb8e), which you reviewed smile
> 
> > backup: provide generator to backup_fill
> > It is sort of an identifer for what created the backup. The way it was
> > before, the fixture test in test_backup.c will fail when we change the
> > firmware version. Proiding it explicitly means we can test with any
> > string in the test.

Yeah, I missed it too, apparently.

> Proiding it explicitly means we can test with any string in the test.

I guess that didn't work. :)

benma:
> I guess that didn't work. :)

seems it was only working for test_backup.c, where v9.0.0 is also hardcoded, against a fixture. 

fixed test_restore.c now

test/unit-test/test_restore.c
```diff
@@ -388,21 +389,21 @@ int main(void)
...
-        cmocka_unit_test_setup_teardown(
-            test_restore_1_corrupt_backup_from_directory, test_setup, test_teardown),
-        cmocka_unit_test_setup_teardown(
-            test_restore_2_corrupt_backups_from_directory, test_setup, test_teardown),
-        cmocka_unit_test_setup_teardown(
-            test_restore_3_corrupt_backups_from_directory, test_setup, test_teardown),
-        cmocka_unit_test_setup_teardown(
-            test_fail_restore_if_majority_vote_fails, test_setup, test_teardown),
-        cmocka_unit_test_setup_teardown(test_restore_seed, test_setup, test_teardown),
+        /* cmocka_unit_test_setup_teardown( */
```
x1ddos, line 392: Wait, why are these commented out now, all except the first test?

benma: oh that's embarrassing :O fixed

x1ddos: Ah cool. For a second I thought they somehow weren't truly working anymore. :)

x1ddos:
:+1:

PR #555, opened 2020-08-12
workflow: rename mnemonic -> recovery words

benma:


src/workflow/restore_from_mnemonic.c
```diff
@@ -143,11 +143,11 @@ bool workflow_restore_from_mnemonic(const RestoreFromMnemonicRequest* request)
     UTIL_CLEANUP_32(seed);
     size_t seed_len = 0;
     if (!keystore_bip39_mnemonic_to_seed(mnemonic, seed, &seed_len)) {
-        workflow_status_blocking("Invalid mnemonic", false);
+        workflow_status_blocking("Recovery words\ninvalid", false);
```
x1ddos, line 146: I guess it doesn't fit on one line?

x1ddos: Or how about "Invalid recovery words" to emphasize **invalid**?

benma: yeah, doesn't fit on one line.

the status screen also has a big cross, so it is very emphasized :)

src/workflow/show_mnemonic.c
```diff
@@ -33,8 +33,8 @@
 
 #define NUM_RANDOM_WORDS 5
 
-static const char* _back_label = "Back to seed phrase";
-static const char* _cancel_confirm_title = "Mnemonic";
+static const char* _back_label = "Back to\nrecovery words";
+static const char* _cancel_confirm_title = "Recovery\nwords";
```
x1ddos, line 37: This could still fit on one line though, diesn't it?

benma: No, it is a bit too close to the `x` and checkmark in the title.

PR #559, opened 2020-08-18
ui: fix regression in confirm() regarding label truncating

benma:
When calling the Rust version of confirm(), the title/body were
assumed to be smaller than MAX_LABEL_SIZE. This was not a correct
assumption, the confirm component can take arbitary strings, and if
longer than MAX_LABEL_SIZE, it will truncate it.

This regression resulted in a panic for very long confirm bodies, such
as when confirming an ETH transaction with a long data field.


src/rust/bitbox02/src/ui.rs
```diff
@@ -150,9 +150,21 @@ where
...
 {
+    // We truncate at a bit higher than MAX_LABEL_SIZE, so the label component will correctly
+    // truncate and append '...'.
+    const TRUNCATE_SIZE: usize = MAX_LABEL_SIZE + 1;
+
     let params = bitbox02_sys::confirm_params_t {
-        title: crate::str_to_cstr_force!(params.title, MAX_LABEL_SIZE).as_ptr(),
-        body: crate::str_to_cstr_force!(params.body, MAX_LABEL_SIZE).as_ptr(),
+        title: crate::str_to_cstr_force!(
+            crate::util::truncate_str(params.title, TRUNCATE_SIZE),
```
x1ddos, line 159: What was the reason for `panic!("str did not fit")` in the first place? Why not always trucate strings, i.e. have `str_to_cstr` always trucate.

x1ddos: benma said: becaue we don't want to trucate important data. very good point.

benma: idk, Niklas added it. To me, implicit truncating is slightly scary, I prefer explicit assumptions and limits, and panics if there were reasoning mistakes.

x1ddos: and then I thought: but the data we're trying to display here, now also trucated, is also important.

PR #563, opened 2020-08-22
src/workflow: remove unused password.h include

benma:


src/workflow/reset.c
```diff
@@ -14,7 +14,6 @@
 
 #include "reset.h"
 #include "confirm.h"
-#include "password.h"
 #include "workflow.h"
 
 #include <reset.h>
```
x1ddos, line 19: Since we're here, `reset.h` seems to be included twice, or is this a different reset from the `"reset.h"` above?

src/workflow/reset.c
```diff
@@ -14,7 +14,6 @@
 
 #include "reset.h"
 #include "confirm.h"
-#include "password.h"
 #include "workflow.h"
```
x1ddos, line 17: Builds also without including workflow.h. I think none of that is used here in reset anymore.

src/workflow/show_mnemonic.c
```diff
@@ -16,7 +16,6 @@
 
 #include "blocking.h"
 #include "cancel.h"
-#include "password.h"
 #include "status.h"
 #include "workflow.h"
```
x1ddos, line 20: Also this one isn't needed anymore.

x1ddos (commented):
Looks like a couple more unused includes. A good opportunity to remove those as well.

src/workflow/reset.c
benma: yeah, src/reset.h contains `reset_reset()`.

benma: done

src/workflow/reset.c
benma: done

src/workflow/show_mnemonic.c
benma: done

src/workflow/reset.c
x1ddos: ah, one if workflow_reset and the other one is reset_reset. this is where include with the path would've made it more clear:

```
#include "reset.h"
#include "workflow/reset.h"
```

(both relative to src/)

but ok, not relevant to this PR.

x1ddos (approved):
LGTM

PR #565, opened 2020-08-24
git: ignore C/C++ LSP caches and local configs

x1ddos:
This adds some well-known file and dir names which are not typically checked into the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,13 @@
 /releases/temp/
 NOTES.md
 .vagrant/
+# LSP caches and local configs
+.clangd
+.cquery
+.ccls
+.ccls-cache
+compile_commands.json
+compile_flags.txt
 
 # gnu global
 /src/GPATH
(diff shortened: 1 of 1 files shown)


benma (approved):
Hmm normally I'd say that non-project related files belong in your private gitignore, not the project's gitignore, but since I added the gnu global files there too I can't say anything :zipper_mouth_face: :joy:

.gitignore
```diff
@@ -3,6 +3,13 @@
 /releases/temp/
 NOTES.md
```
benma, line 4: remove this too? seems like this is totally misplaced.

x1ddos: done

x1ddos:
> Hmm normally I'd say that non-project related files belong in your private gitignore, not the project's gitignore, but since I added the gnu global files there too I can't say anything :zipper_mouth_face: :joy:

:) yeah, true. these files are pretty common though. we even generate compile_commands.json in https://github.com/digitalbitbox/bitbox02-firmware/blob/92686a1/CMakeLists.txt#L59 in every build.

PR #567, opened 2020-08-25
rust/async_usb: do not borrow executor state during poll

benma:
Currently, USB_TASK_STATE, the main executor state, is borrowed for
the whole duration of `spin()`, which includes the polling/execution
of the current task.

This prevents the state to be borrowed again during the execution of a
task. In a normal async runtime, there are things like `spawn(task)`
etc. that need to borrow and modify the executor state though.

The normal way to achieve that is to pop the task from the executor
state (often a task queue), poll it, and put it back in if it is not
finished yet. We do the same, but we have an enum with at most one
task instead of a task queue, so popping just means removing the task
from the enum.

Motivation: in a future PR we plan to add a `next_request()` to the
executor, which can be called from the currently running task. This
future will send a response to the host and wait for the next request
from the host.

This will allow sequential handling of api workflows that involve
multiple api calls from the host, e.g. BTC singing. A big state
machine keeping track of what api call we expect next can be simplied
a lot using `.await` instead.

### DIFF
--- a/src/rust/bitbox02-rust/src/async_usb.rs
+++ b/src/rust/bitbox02-rust/src/async_usb.rs
@@ -33,8 +33,13 @@ enum UsbTaskState<'a> {
     /// Waiting for a new query, nothing to do.
     Nothing,
     /// A query came in which launched a task, which is now running (e.g. user is entering a
-    /// password).
-    Running(Task<'a, UsbOut>),
+    /// password). The option inside is `Some`, but is `None` for the brief period during which the
+    /// task is polled. This is akin to popping off the task from the executor state and pushing it
+    /// back in after it is polled, which is best practice (and often done via a message passing
+    /// queue). This allows for the execeutor state to not be borrowed while the task is being
+    /// executed, which allows the task itself to modify the executor state (otherwise we would have
+    /// an illegal double-borrow of the state).
+    Running(Option<Task<'a, UsbOut>>),
     /// The task has finished and written the result, so the USB response is available. We are now
     /// waiting for the client to fetch it (HWW_REQ_RETRY). For short-circuited or non-async api
     /// calls, the result might be returned immediately in response to HWW_REQ_NEW.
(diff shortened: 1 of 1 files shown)


benma:
@NickeZ taught me that the task has to be popped and re-inserted :nerd_face: Thanks Niklas!

src/rust/bitbox02-rust/src/async_usb.rs
```diff
@@ -77,15 +82,26 @@ where
...
-            let result = spin_task(task);
-            if let Poll::Ready(result) = result {
-                *state = UsbTaskState::ResultAvailable(result);
+    // Pop task before polling, so that USB_TASK_STATE does not stay borrowed during the poll.
+    let mut popped_task = match *USB_TASK_STATE.0.borrow_mut() {
+        // Illegal state, `None` is only valid during the poll.
+        UsbTaskState::Running(None) => panic!("task not found"),
+        // Get the task out, putting `None` in. This allows us to release the mutable borrow on the
+        // state.
+        UsbTaskState::Running(ref mut task @ Some(_)) => core::mem::replace(task, None),
```
benma, line 91: core::mem::replace seems to be the only way were you can move a non-Copy type without overhead :shrug: the other way would be to use a `Rc<Task>`, so one can have an intermediate step where there are two references before removing one again.

NickeZ: I didn't even know that mem::replace was doing something so complicated.. https://stackoverflow.com/questions/27197938/memreplace-in-rust

benma: Clippy really is quite good!

https://rust-lang.github.io/rust-clippy/master/index.html#mem_replace_option_with_none

NickeZ: yeah, they have some cool lints

x1ddos (approved):
Nice. LGTM. Would've appreciated a little test covering a lifetime of a task.

benma:
> Nice. LGTM. Would've appreciated a little test covering a lifetime of a task.

test will come once we have function to modify the state that can be called from the task.

x1ddos:
right. that function could've been defined in the test.

PR #568, opened 2020-08-25
rust: get rid of mod.rs

benma:
See
https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#no-more-modrs.

Since Rust 2018, mod.rs is not required anymore if there are
submodules. Having a unique name is better than a ton of `mod.rs` in
the project. It helps with navigation.


x1ddos (approved):
LGTM

PR #573, opened 2020-08-31
rust/bitbox02/ui: move types to a new file

benma:
Keep only functions in ui.rs, so it is easier to do a module switch
for testing after.


x1ddos (approved):
LGTM

PR #574, opened 2020-08-31
rust/bitbox02: add stubs for testing

benma:
The bitbox02 safely wraps C functions. In Rust tests, C is not linked,
and even if it was, properly mocking the functions would be difficult.

We do a workaround instead: by adding the 'testing' feature and
enabling it only in tests, we can supply stub/mock implementations
that can be used in unit tests, not referencing any C functions.

In future commits, we can add some light mocking infrastructure to be
able to check args and mock the return values.

Alternatives dismissed:
- Add a trait for memory, keystore, ui and mock those.
  This resulted in all the code littered with generic params and lots
  of boilerplate.
- Implement the C funtions in Rust using cdylib and use that:
  Too complicated compared to the testing feature.

### DIFF
--- a/src/rust/bitbox02/src/lib.rs
+++ b/src/rust/bitbox02/src/lib.rs
@@ -22,6 +22,8 @@ extern crate std;
 
 pub mod commander;
 pub mod keystore;
+
+#[cfg_attr(feature = "testing", path = "memory_stub.rs")]
 pub mod memory;
 pub mod password;
 pub mod random;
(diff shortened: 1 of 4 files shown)


NickeZ:
noice

src/rust/bitbox02/src/lib.rs
```diff
@@ -22,6 +22,8 @@ extern crate std;
 
 pub mod commander;
 pub mod keystore;
+
+#[cfg_attr(feature = "testing", path = "memory_test.rs")]
```
x1ddos, line 26: agreed to call it `memory_stub.rs`

x1ddos (approved):
LGTM

PR #576, opened 2020-09-03
Rust test set device name

benma:


src/rust/bitbox02-rust/src/hww/api/set_device_name.rs
```diff
@@ -39,3 +39,60 @@ pub async fn process(
...
+
+    use crate::bb02_async::block_on;
+    use bitbox02::testing::{mock, Data};
+
+    #[test]
+    pub fn test_set_device_name() {
+        let some_name: std::string::String = "foo".into();
+
+        // All good.
+        mock(Data {
```
benma, line 56: @x1ddos if you could stomach a dependency, we could try https://docs.rs/mockall/0.8.0/mockall/ to make mocking more flexible and easy. for now the handrolled thing might work, we'll see

src/rust/bitbox02-rust-c/src/lib.rs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![cfg_attr(not(test), no_std)]
+#![no_std]
 #![feature(alloc_error_handler)] // used in alloc.rs
```
x1ddos, line 16: so, allow error handler is now also used in tests, right? just checking.

src/rust/bitbox02-rust/src/hww/api/set_device_name.rs
x1ddos: doesn't look too bad so far, does it. but sure, if you get tired of mocking why not. it's only a dev dependency after all.

x1ddos: as long as it doesn't pull in a large deps tree :)

src/rust/Cargo.lock
```diff
@@ -200,6 +201,11 @@ name = "keccak"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
+[[package]]
+name = "lazy_static"
```
x1ddos, line 205: I was convinced to have seen lazy static before somewhere in the firmware code but maybe I was dreaming.

src/rust/bitbox02/src/testing.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub ui_confirm_create_body: Option<std::string::String>,
+    pub ui_confirm_create_result: Option<bool>,
+}
+
+lazy_static! {
+    pub static ref TESTDATA: Arc<Mutex<Data>> = Arc::new(Mutex::new(Default::default()));
+}
+
+pub fn mock(data: Data) {
+    *TESTDATA.lock().unwrap() = data;
```
x1ddos, line 33: I'm not entirely sure putting this behind a mutex is a good idea. I mean, by default rust tests run concurrencly, right. So, this makes sense. But now imagine 2 tests where each one updates TESTDATA. The end result is we might have failing tests.

Am I missing something?

src/rust/bitbox02-rust-c/src/lib.rs
benma: No, those are both independent module level declarations. It means this crate is no_std, and the alloc_error_handler feature is used.

src/rust/Cargo.lock
benma: I had it in the PoC PR about this, which I closed.

src/rust/bitbox02/src/testing.rs
benma: No you are right I think. So I need to keep the lock for the duration of the test to avoid a race condition, right? Make it a RW-lock too, as the stub itself also accesses the testdata.

benma: Scratch that, makes no sense, as the read lock cannot be acquired while it is write locked...

x1ddos: So, generally speaking we need to serialize all tests which access `TESTDATA`. There are different ways to do it but I haven't come up with what's best yet.

Alternatively, we could run each test which access `TESTDATA` as a separate binary but it feels like overkill.

x1ddos: And yet another way is to have separate `TESTDATA` for each test which needs it, like `TESTDATA_set_device_name`.

benma: Exactly...

benma: added a new iteartion, PTAL:

1) global mutex to serialize tests
2) must be held when accessing DATA, which is now in a RefCell instead of a Mutex (runtime borrow checking, but no thread safety).

x1ddos (approved):
LGTM

PR #577, opened 2020-09-14
libwally: Avoid working around a quirk that has been fixed in wally

jgriffiths:
bip32_key_init() was added some time ago to allow initialising an ext_key without allocating it.

### DIFF
--- a/src/apps/common/bip32.c
+++ b/src/apps/common/bip32.c
@@ -18,32 +18,26 @@
 
 bool apps_common_bip32_xpub_from_protobuf(const XPub* xpub_in, struct ext_key* xpub_out)
 {
-    struct ext_key* tmp = NULL;
-    // libwally quirk: bip32_key_init_alloc takes 20 bytes for the parent fingerprint but only uses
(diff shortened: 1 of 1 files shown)


src/apps/common/bip32.c
```diff
@@ -18,13 +18,12 @@
 
 bool apps_common_bip32_xpub_from_protobuf(const XPub* xpub_in, struct ext_key* xpub_out)
 {
-    struct ext_key* tmp = NULL;
     // libwally quirk: bip32_key_init_alloc takes 20 bytes for the parent fingerprint but only uses
```
benma, line 21: bip32_key_init_alloc -> bip32_key_init

benma (changes requested):
Very nice, thank's a lot! This has been on my todo list for a long time :heart:

src/apps/common/bip32.c
jgriffiths: Fixed in last push :)

benma:
sorry for the trouble, can you also run `./scripts/format`? (needs clang-format, or run it in docker). Then CI will pass :)

benma (approved):

Issue #578, opened 2020-09-16
Autocomplete of Bip39 words


My1:
The wordlist is designed in a way that all words are unique after 4 letters (in fact several backup solutions including the [Steelwallet](https://guides.shiftcrypto.ch/steelwallet/) only support entering 4 letters), so similar to on a keepkey you could just show the complete word instead for asking the extra letters.

while there are some words that are already unique on the third letter if that involves too much complexity one can ignore that.

deletion was brought up on telegram, one could for example check if the word currently has 4 or more letters to just cut to 3, and then proceed normally with deleting.

benma:
Voilà: https://github.com/digitalbitbox/bitbox02-firmware/pull/583/

This does autocompletion for now, with deletion working the the same as before, only one char at a time. Improving deletion is reserved for a future PR.

My1:
sadly no way for me to try :-(

PR #583, opened 2020-09-23
Autocomplete

benma:
Closes https://github.com/digitalbitbox/bitbox02-firmware/issues/578

### DIFF
--- a/src/ui/components/trinary_input_string.c
+++ b/src/ui/components/trinary_input_string.c
@@ -235,6 +235,31 @@ static void _render(component_t* component)
     }
 }
 
+// if the current input uniquely identifies a word from the wordlist by prefix, we autocomplete the
+// word.
+static void _maybe_autocomplete(component_t* trinary_input_string)
+{
+    data_t* data = (data_t*)trinary_input_string->data;
+    if (data->wordlist == NULL) {
+        return;
+    }
+    // initial value means no word was found yet.
+    size_t found_word_idx = data->wordlist_size;
+    for (size_t word_idx = 0; word_idx < data->wordlist_size; word_idx++) {
+        const char* word = data->wordlist[word_idx];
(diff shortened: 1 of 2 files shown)


src/ui/components/trinary_input_string.c
```diff
@@ -235,6 +235,31 @@ static void _render(component_t* component)
...
+        bool is_prefix = strncmp(data->string, word, data->string_index) == 0;
+        if (is_prefix) {
+            if (found_word_idx != data->wordlist_size) {
+                // Not unique.
+                return;
+            }
+            found_word_idx = word_idx;
+        }
+    }
+    data->string_index =
```
benma, line 259: could think about getting rid of string_index completely, and use `strlen(data->string)` in its place everywhere.

src/ui/components/trinary_input_string.c
```diff
@@ -235,6 +235,31 @@ static void _render(component_t* component)
...
+// word.
+static void _maybe_autocomplete(component_t* trinary_input_string)
+{
+    data_t* data = (data_t*)trinary_input_string->data;
+    if (data->wordlist == NULL) {
+        return;
+    }
+    // initial value means no word was found yet.
+    size_t found_word_idx = data->wordlist_size;
+    for (size_t word_idx = 0; word_idx < data->wordlist_size; word_idx++) {
```
x1ddos, line 248: I think there are ways to optimize this but probably not worth it today.

x1ddos (approved):
Cool. LGTM.

src/ui/components/trinary_input_string.c
benma: today or maybe ever? it's a small word list, even the whole prefix search to create the keyboard after each tap is not noticeable. or did you mean something else?

x1ddos: yeah I actually meant "maybe ever" :)
didn't notice any issues visually.

PR #586, opened 2020-09-24
components/trinary_input: render title on top, always visible

benma:
When the input is used to enter words from a wordlist (bip39
mnemonic), the title should be rendered at the top instead of at the
center, to be always visible, even when during typing.

- The user usually copies the words from paper, and should never be
  disoriented at which word they are.

- This will help with a future change: the user will be able to go
  back one word, in which case the previous word will be shown and the
  user knows at which position they are.

### DIFF
--- a/src/ui/components/trinary_input_string.c
+++ b/src/ui/components/trinary_input_string.c
@@ -89,6 +89,11 @@ typedef struct {
     // Show last character instead of asterisk. Applies only if `hide` is true.
     bool show_last_character;
 
+    // If the title should be drawn at the top of the screen. If true, the title is always
+    // visible. If false, the title is rendered in the center, and hidden as soon as the user starts
+    // typing.
+    bool title_on_top;
+
     component_t* title_component;
     component_t* trinary_char_component;
     component_t* confirm_component;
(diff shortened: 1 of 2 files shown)


src/ui/components/trinary_input_string.c
```diff
@@ -89,6 +89,11 @@ typedef struct {
     // Show last character instead of asterisk. Applies only if `hide` is true.
     bool show_last_character;
 
+    // If the title should be drawn at the top of the screen. In true, the title is always
```
x1ddos, line 92: "if true ..." typo

src/ui/components/trinary_input_string.c
```diff
@@ -161,7 +167,7 @@ static void _render(component_t* component)
         uint8_t string_y = STRING_POS_Y;
 
         if (i == data->string_index &&
-            (confirm_gesture_active || show_title || blink ||
+            (confirm_gesture_active || (!data->title_on_top && show_title) || blink ||
```
x1ddos, line 170: I'm a bit confused. `show_title` includes `data->title_on_top || ...` so this expr results in:

```
!data->title_on_top && (data->title_on_top || ...rest)
```

which is the same as

```
!data->title_on_top && ...rest
```

benma: I was aware, but should I extract `rest` and name it `show_title_if_not_on_top`? :sweat_smile:

benma: Actually I did sth else: remove the `data->title_on_top` from `show_title`.

src/ui/components/trinary_input_string.c
benma: done

x1ddos (approved):
LGTM

PR #588, opened 2020-09-25
restore from mnemonic: allow editing previous word

benma:
It is infuriating to have to restart the whole process and enter again all words in case of mistake. This allows the user to go back and edit the previous word, repeatedly if necessary.

This PR is best looked at commit by commit.

### DIFF
--- a/src/ui/components/trinary_input_string.h
+++ b/src/ui/components/trinary_input_string.h
@@ -54,4 +54,12 @@ component_t* trinary_input_string_create_password(
     void (*cancel_cb)(void* param),
     void* cancel_callback_param);
 
+/**
+ * Only applicable in wordlist-mode.
+ * Sets the current word. The user can then accept it or delete characters.
+ * The word must be in the wordlist as passed to `trinary_input_string_create_wordlist()`, otherwise
+ * this function aborts.
+ */
+void trinary_input_string_set_input(component_t* trinary_input_string, const char* word);
+
 #endif
--- a/src/ui/components/trinary_input_string.h
+++ b/src/ui/components/trinary_input_string.h
@@ -36,6 +36,8 @@
  * @param[in] confirm_cb The callback that is called when the user entered the string. Will be
  * called at most once.
  * @param[in] cancel_cb Called when the user cancels by hitting the back button.
+ * @param[in] cancel_is_backbutton whether the cancel button should be rendered as a back button
+ *            instead of as a cross.
  */
 component_t* trinary_input_string_create_wordlist(
     const char* title,
@@ -44,7 +46,8 @@ component_t* trinary_input_string_create_wordlist(
     void (*confirm_cb)(const char* input, void* param),
     void* confirm_callback_param,
     void (*cancel_cb)(void* param),
-    void* cancel_callback_param);
+    void* cancel_callback_param,
+    bool cancel_is_backbutton);
 
 component_t* trinary_input_string_create_password(
     const char* title,
--- a/src/workflow/trinary_input.c
+++ b/src/workflow/trinary_input.c
@@ -18,12 +18,15 @@
 #include "cancel.h"
 
 #include <hardfault.h>
+#include <ui/components/menu.h>
 #include <ui/components/trinary_input_string.h>
+#include <ui/screen_stack.h>
 #include <util.h>
 
 #include <stdio.h>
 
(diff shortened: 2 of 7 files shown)


src/workflow/trinary_input.c
```diff
@@ -35,12 +38,28 @@ static void _confirm(const char* word, void* param)
     workflow_blocking_unblock();
 }
 
+static void _select(uint8_t idx)
+{
+    ui_screen_stack_pop();
+    if (idx == 0) {
+        _cancel_reason = WORKFLOW_TRINARY_INPUT_RESULT_DELETE;
+        workflow_cancel_force();
+    } else if (idx == 1) {
```
x1ddos, line 47: It feels a bit fragile, referring to indexes of an array in another function.
How about making `words[] = {"Edit prev", "Cancel restore"}` and these indexes constants, outside of the `_select` and `_cancel`?

x1ddos: Or, at the very least, add a comment explaining which array `idx` refers to?

src/workflow/trinary_input.c
```diff
@@ -35,12 +38,28 @@ static void _confirm(const char* word, void* param)
...
+        _cancel_reason = WORKFLOW_TRINARY_INPUT_RESULT_CANCEL;
+        workflow_cancel();
+    }
+}
```
x1ddos, line 51: Blank line to separate two functions? I wonder why clang lint didn't catch it.

src/workflow/trinary_input.c
```diff
@@ -50,9 +69,9 @@ bool workflow_trinary_input_wordlist(
             "Restore",
             trinary_input_string_create_wordlist(
                 title, wordlist, wordlist_size, _confirm, NULL, _cancel, NULL))) {
-        return false;
+        return _cancel_reason;
```
x1ddos, line 81: Why not make `workflow_cancel_run` return `workflow_trinary_input_result_t`? I think it might simplify some parts and remove the need for `_cancel_reason` static var.

src/ui/components/trinary_input_string.h
```diff
@@ -54,4 +54,11 @@ component_t* trinary_input_string_create_password(
     void (*cancel_cb)(void* param),
     void* cancel_callback_param);
 
+/**
+ * Only applicable in wordlist-mode.
+ * Sets the current word. The user can then accept it or delete characters.
+ * The word must be in the wordlistm, otherwise this function aborts.
```
x1ddos, line 63: "wordlist" typo. also, consider being more explicit: "must be in the component's data->wordlist" for people like me.

src/workflow/trinary_input.c
```diff
@@ -60,14 +65,15 @@ static void _cancel(void* param)
 }
 
 workflow_trinary_input_result_t workflow_trinary_input_wordlist(
+    size_t word_idx,
     const char* title,
```
x1ddos, line 69: since you have the word index now, might as well get rid of the `title` arg and create it here, inside `workflow_trinary_input_wordlist`.

x1ddos (commented):
Nice! Works as expected :heart_eyes_cat: 
Added a few comments to possibly simplify the code a bit.

src/workflow/trinary_input.c
benma: `workflow_cancel_run()` can be used in any context, so the return would have to be `void*`, and internally in cancel.c passed via `workflow_cancel_force()`. Then there would be a static var in cancel.c instead of here.

Not sure how nice all of that would be vs. a local static var.

src/ui/components/trinary_input_string.h
benma: done

src/workflow/trinary_input.c
benma: done

src/workflow/trinary_input.c
benma: done

src/workflow/trinary_input.c
benma: Nice idea, done.

benma:
PTAL

src/workflow/trinary_input.c
x1ddos: Oh, I see what you mean now. Need to read a bit more code. :+1:

x1ddos (approved):
LGTM

PR #591, opened 2020-09-26
Unify sdcard components

benma:


benma:
This also saves 352 bytes of binary size :joy:

x1ddos (approved):
LGTM

PR #594, opened 2020-09-27
workflow/backup: remove dead code

benma:
Those three cases are unreachable with the current code. The default
case covers them sufficiently anyway.


benma:
Folded into https://github.com/digitalbitbox/bitbox02-firmware/pull/595

PR #595, opened 2020-09-28
commander: port CheckBackup to Rust

benma:


src/rust/bitbox02/src/backup.rs
```diff
@@ -0,0 +1,44 @@
...
+// limitations under the License.
+
+extern crate alloc;
+use alloc::string::String;
+pub use bitbox02_sys::backup_error_t as Error;
+
+pub struct CheckData {
+    pub id: String,
+    pub name: String,
+    pub birthdate: u32,
```
x1ddos, line 22: please add a `// unix timestamp` comment for people like me :)

src/rust/bitbox02/src/backup.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/// Returns the backup id, name and birthdate of the backup that matches the current keystore seed.
+/// If none matches, `Err()` is returned.
+pub fn check() -> Result<CheckData, Error> {
+    let mut id = [0u8; 65];
```
x1ddos, line 28: `backup_check` in https://github.com/digitalbitbox/bitbox02-firmware/blob/75470b6/src/backup/backup.h#L34 says something odd:

> id_out must have max 256 bytes in size; hww.options BackupInfo.id

but looking at backup.c, it clearly is 32 bytes in hex. I guess the comment in `backup_check` in backup.h is old and needs to be updated.

src/rust/bitbox02/src/backup.rs
x1ddos: also, UCT or local? my guess the former since there's no offset here but it's just my guess.

src/rust/bitbox02-rust/src/hww/api/backup.rs
```diff
@@ -0,0 +1,70 @@
...
+            if !silent {
+                let params = confirm::Params {
+                    title: "Name?",
+                    body: &name,
+                    scrollable: true,
+                    ..Default::default()
+                };
+
+                if !confirm::confirm(&params).await {
+                    return Err(Error::COMMANDER_ERR_GENERIC);
```
x1ddos, line 40: Since you're here, why not change it to `COMMANDER_ERR_USER_ABORT`? https://github.com/digitalbitbox/bitbox02-firmware/blob/75470b6/src/workflow/backup.c#L139

> // TODO Change to error "User Abort"

src/rust/bitbox02/src/backup.rs
x1ddos: ah, never mind. I see it's updated in the last commit 51bb552.

x1ddos (approved):
LGTM

src/rust/bitbox02/src/backup.rs
benma: yeah, UTC. added comment

src/rust/bitbox02-rust/src/hww/api/backup.rs
benma: It would technically be a breaking change and a major version bump, but if you agree, i'll just do the change and not bump the major version :sweat_smile: only the bitboxapp (api-go) and send_message use this api call currently, and the app will work the same without any change.

x1ddos: yeah, agree. there isn't any reasonable benefit in bumping the major version because of this.

x1ddos: don't forget there are two places.

benma: Done

x1ddos: :heavy_check_mark:

PR #596, opened 2020-09-28
util: add util_format_datetime

benma:
Isolate C datetime formatting so it is easier to wrap from Rust.

Rust does not seem to have a good way to format datetimes without the
`chrono` crate, which is huge, also in terms of added binary size.


x1ddos (approved):
LGTM

PR #600, opened 2020-10-01
keystore: do not require caller of get_bip39_mnemonic to free()

benma:
The wally function uses malloc/free. This commit makes this nuisance
internal, so the caller can just use a regular stack var.

This makes it easier to wrap this function in Rust.

### DIFF
--- a/src/workflow/show_mnemonic.c
+++ b/src/workflow/show_mnemonic.c
@@ -132,16 +132,12 @@ static bool _show_words(const char** words, uint8_t words_count)
 }
 
 typedef struct {
-    char* mnemonic;
-    // Keep len as mnemonic is tokenized using strtok, so util_cleanup_str() does not work anymore
-    // to clean the string.
-    size_t len;
+    char mnemonic[300];
 } mnemonic_t;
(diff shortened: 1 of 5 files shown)


src/workflow/show_mnemonic.c
```diff
@@ -132,16 +132,12 @@ static bool _show_words(const char** words, uint8_t words_count)
 }
 
 typedef struct {
-    char* mnemonic;
-    // Keep len as mnemonic is tokenized using strtok, so util_cleanup_str() does not work anymore
-    // to clean the string.
-    size_t len;
+    char mnemonic[300];
```
x1ddos, line 135: Why not 216?

> Should be at least 216 bytes (longest possible 24 word phrase plus null terminator).

x1ddos (approved):
LGTM. Nice.

src/workflow/show_mnemonic.c
benma: No harm in having a safety buffer, in case there is a miscalculation at 216 ;)

PR #601, opened 2020-10-01
Rust menu

benma:


src/rust/bitbox02-rust/src/workflow/menu.rs
```diff
@@ -0,0 +1,34 @@
...
+pub async fn menu() {
+    let result = RefCell::new(None);
+    let mut component = bitbox02::ui::menu_create(
+        &["foo", "bar"],
+        Some("demo title"),
+        Some(Box::new(|idx| {
+            *result.borrow_mut() = Some(idx);
+        })),
+        Some(Box::new(|| crate::print_debug!(1000, "hit continue"))),
+        Some(Box::new(|| crate::print_debug!(1000, "hit cancel"))),
```
x1ddos, line 29: tbh it's not the easiest to read. would it make sense to define a `StatusComponent`, `MenuComponent`, etc. where I could do:

```rust
let mut menu = bitbox02::ui::Menu {
  words: &["foo", "bar"],
  title: Some("title"),
  on_select: Some(Box::New(|i| { *result.borrow_mut() = Some(i); })),
  on_cancel: ...,
  ...
}
menu.screen_stack_push(); // sets everything up and calls bitbox02_sys::menu_create(...)
```

so that it's clear what's what, especially which callback is which.

benma: I made a MenuParams struct similar to ConfirmParams, which gives each param a name. Should be more readable now, ptal.

src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -195,6 +201,98 @@ where
...
+        .map(|title| crate::str_to_cstr_force!(title, MAX_LABEL_SIZE));
+    let component = unsafe {
+        bitbox02_sys::menu_create(
+            c_words.as_ptr(),
+            select_word_cb,
+            select_word_cb_param,
+            words.len() as _,
+            title.map_or_else(
+                || core::ptr::null(),
+                |title| crate::str_to_cstr_force!(title, MAX_LABEL_SIZE).as_ptr(),
```
x1ddos: why not do it all at once above in `let title = params.title.map ...`?

benma: It's again drop-scrope discussion time :joy: 

If you did everything above like so:

```Rust
let title = params.title.map_or_else(
        || core::ptr::null(),
        |title| crate::str_to_cstr_force!(title, MAX_LABEL_SIZE).as_ptr(),
    );
```

The pointer would be dangling, because the cstr made in the closure is dropped as soon as the closure ends.

I have verified this in two ways:

- Demo much like the previous demo, but using a closure: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=891c88e26487977aa63c0d831bc4d250

- I tested the above snippet and indeed the menu title disappeared, presumably due to the memory error.

src/rust/bitbox02/src/ui/types.rs
```diff
@@ -86,3 +89,11 @@ impl<'a> ConfirmParams<'a> {
         }
     }
 }
+
+pub struct MenuParams<'a> {
+    pub words: &'a [&'a str],
+    pub title: Option<&'a str>,
+    pub select_word_cb: Option<Box<dyn FnMut(u8) + 'a>>,
```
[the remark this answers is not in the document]
benma: I don't think you can use traits to absolutely enforce that the Box::into_raw() type matches what you dereference here. You would at best move the potential of typo from here to where you instantiate the trait.

I added a commit that gives the callbacks unique types, so at least the `defined in a single place` part of your comment is fulfilled. wdyt?

src/rust/bitbox02/src/ui/types.rs
```diff
@@ -86,3 +89,35 @@ impl<'a> ConfirmParams<'a> {
         }
     }
 }
+
+pub struct SelectWordCb<'a>(pub Box<dyn FnMut(u8) + 'a>);
```
NickeZ, line 93: Why don't you just use type aliases? https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=be5c51bd5ebc72a6afee5976551ccebe

benma: :thinking: i thought I tried but didn't work, but obviously it does... Cool, let me try again.

benma: Done! Thanks.

src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -14,20 +14,22 @@
...
 use alloc::boxed::Box;
+use alloc::vec::Vec;
 
 use core::marker::PhantomData;
 
 /// Wraps the C component_t to be used in Rust.
 pub struct Component<'a> {
     component: *mut bitbox02_sys::component_t,
     is_pushed: bool,
+    on_drop: Box<dyn FnMut()>,
```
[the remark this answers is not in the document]
x1ddos: So, why not use the option here too? Seems appropriate. Not every component needs it.

src/rust/bitbox02/src/ui/ui.rs
x1ddos: What confuses me is you have str_to_cstr twice:

image

shouldn't this be:

```rust
menu_create(..., title.map_or_else(|| core::ptr::null(), title.as_ptr(), ...)
```

x1ddos (commented):
Yeah, much nicer with type aliases.
A couple comments left.

src/rust/bitbox02/src/ui/ui.rs
benma: :heavy_check_mark:

src/rust/bitbox02/src/ui/ui.rs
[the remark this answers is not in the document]
benma: Done

src/rust/bitbox02-rust/src/workflow/menu.rs
```diff
@@ -0,0 +1,36 @@
...
+use crate::bb02_async::option;
+use bitbox02::ui;
+use core::cell::RefCell;
+
+extern crate alloc;
+use alloc::boxed::Box;
+
+pub async fn menu() {
+    let result = RefCell::new(None);
+    let mut component = ui::menu_create(ui::MenuParams {
```
x1ddos, line 24: When this demo menu is created, there's no "Continue" top right corner. I tap the `>` arrow bottom right and "Continue" appears. No matter if I tap `<` or `>` arrows, the "Continue" is always present now.

I can also tap "Select" bottom middle and the `select_word_cb` gets called without tapping "Continue". Doesn't feel exactly right but I guess it's the underlying implementation in C, nothing to do with this PR.

x1ddos (approved):
LGTM

src/rust/bitbox02-rust/src/workflow/menu.rs
benma: Yeah this is intended. The two uses cases are: scroll through mnemonics, where Continue appears once you are through. There is no `Select` in this.

The other is confirming a mnemonic word, where there is `Select`, but no Continue.

PR #602, opened 2020-10-01
rust: wrap keystore_get_bip39_mnemonic

benma:


src/rust/bitbox02/src/keystore.rs
```diff
@@ -51,3 +54,26 @@ pub fn create_and_store_seed(password: &Password, host_entropy: &[u8; 32]) -> bo
...
+struct ZeroizedMnemonic([u8; 256]);
+impl core::default::Default for ZeroizedMnemonic {
+    fn default() -> Self {
+        ZeroizedMnemonic([0; 256])
+    }
+}
+impl zeroize::DefaultIsZeroes for ZeroizedMnemonic {}
+
+pub fn get_bip39_mnemonic() -> Result<zeroize::Zeroizing<String>, ()> {
+    let mut mnemonic = zeroize::Zeroizing::new(ZeroizedMnemonic([0u8; 256]));
```
x1ddos, line 68: Why not `new(ZeroizedMnemonic(Default::default()))`?

src/rust/bitbox02/src/keystore.rs
```diff
@@ -51,3 +54,26 @@ pub fn create_and_store_seed(password: &Password, host_entropy: &[u8; 32]) -> bo
         bitbox02_sys::keystore_create_and_store_seed(password.as_cstr(), host_entropy.as_ptr())
     }
 }
+
+#[derive(Copy, Clone)]
+struct ZeroizedMnemonic([u8; 256]);
```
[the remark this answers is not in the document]
benma: was playing around hoping that 256 would one of the lengths supported by zeroize out of the box, but wasn't.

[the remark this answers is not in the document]
x1ddos: Ah yeah, I also checked. Seems to be up to 64: https://docs.rs/zeroize/1.1.0/zeroize/trait.Zeroize.html#implementors

src/rust/bitbox02/src/keystore.rs
x1ddos: how about this:

```rust
let mut mnemonic: zeroize::Zeroizing<ZeroizedMnemonic> = zeroize::Zeroizing::new(Default::default());
```

benma: i don't see how that is better?

x1ddos: you've already implemented the default anyway. why do it here again. maybe tomorrow you'll want to change the default to `ZeroizedMnemonic([0; 300])`. just feels a bit weird. but anyway, not a big deal.

x1ddos (approved):
LGTM. Tests?

benma:
> Tests?

How? :O unit test can't call C functions, and also keystore.rs is not even used in unit tests, keystore_stub.rs is...

x1ddos:
> > Tests?
> 
> How? :O unit test can't call C functions, and also keystore.rs is not even used in unit tests, keystore_stub.rs is...

I forgot keystore is actually stubbed in tests. Maybe a better way would've been to stub at the `bitbox02_sys` level. Then we could test deeper. WDYT?

benma:
> Maybe a better way would've been to stub at the `bitbox02_sys` level. Then we could test deeper. WDYT?

In theory that makes a lot of sense. In practice, it would be quite terrible to write mock functions in each unit test (which currently use the _stub.rs functions) at the C-level instead. Each mocked string would need to be turned into a C-string, being sure there are no dangling pointers (like [this one](https://github.com/digitalbitbox/bitbox02-firmware/pull/603#discussion_r498738840)), and so on.

It would be good to be able to do it just to unit test the bitbox02 crate itself somehow, but when unit testing the bitbox02-rust crate, I would keep the current stubs. Maybe it's possible with a tricky combination of features and different invocations of `cargo test` :fearful:

PR #603, opened 2020-10-01
rust/util: improve errors in str_to_cstr

x1ddos:
There are more than one case where making a C str out of Rust may fail.
For example, when Rust str contains non-ASCII characters.

This commit distinguishes between string length and non-ASCII errors
when panicking.

### DIFF
--- a/src/rust/bitbox02/src/lib.rs
+++ b/src/rust/bitbox02/src/lib.rs
@@ -90,11 +90,13 @@ pub use bitbox02_sys::buffer_t;
 #[macro_use]
 pub mod util;
 
+// ug_put_string displays a debug message on the screen for 3 sec.
 pub fn ug_put_string(x: i16, y: i16, input: &str, inverted: bool) {
-    if let Ok(buf) = str_to_cstr!(input, 128) {
-        unsafe { bitbox02_sys::UG_PutString(x, y, buf.as_ptr() as *const _, inverted) }
-    } else {
-        screen_print_debug("string didn't fit", 3000);
+    match str_to_cstr!(input, 128) {
+       Ok(buf) => unsafe {
+           bitbox02_sys::UG_PutString(x, y, buf.as_ptr() as *const _, inverted);
+       },
+       Err(msg) => screen_print_debug(msg, 3000),
     }
 }
 
@@ -140,8 +142,11 @@ pub fn screen_print_debug(msg: &str, duration: i32) {
         Ok(cstr) => unsafe {
             bitbox02_sys::screen_print_debug(cstr.as_ptr() as *const _, duration)
         },
-        Err(cstr) => unsafe {
-            bitbox02_sys::screen_print_debug(cstr.as_ptr() as *const _, duration)
+        Err(errmsg) => unsafe {
+            bitbox02_sys::screen_print_debug(
+                str_to_cstr_force!(errmsg, 200).as_ptr() as *const _,
+                duration,
+            )
         },
     }
 }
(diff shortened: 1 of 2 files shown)


src/rust/bitbox02/src/lib.rs
```diff
@@ -140,8 +141,8 @@ pub fn screen_print_debug(msg: &str, duration: i32) {
         Ok(cstr) => unsafe {
             bitbox02_sys::screen_print_debug(cstr.as_ptr() as *const _, duration)
         },
-        Err(cstr) => unsafe {
-            bitbox02_sys::screen_print_debug(cstr.as_ptr() as *const _, duration)
+        Err(errmsg) => unsafe {
+            bitbox02_sys::screen_print_debug(errmsg.as_ptr() as *const _, duration)
```
benma, line 145: errmsg is a rust string that is not null terminated, so this does not work.

benma (changes requested):

src/rust/bitbox02/src/lib.rs
x1ddos: done

x1ddos:
PTAL

src/rust/bitbox02/src/lib.rs
```diff
@@ -140,8 +142,11 @@ pub fn screen_print_debug(msg: &str, duration: i32) {
         Ok(cstr) => unsafe {
             bitbox02_sys::screen_print_debug(cstr.as_ptr() as *const _, duration)
         },
-        Err(cstr) => unsafe {
-            bitbox02_sys::screen_print_debug(cstr.as_ptr() as *const _, duration)
+        Err(errmsg) => unsafe {
+            bitbox02_sys::screen_print_debug(
+                str_to_cstr_force!(errmsg, 200).as_ptr() as *const _,
```
benma, line 147: i think this is also not memory safe, as the `str_to_cstr_force!(errmsg, 200)` can be dropped immediately before `screen_print_debug()` starts execution, so the ptr returned by `as_pt()` becomes dangling.

fix: put the cstr into a separate var one line above.

unsafe rust is tricky business :see_no_evil:

benma (changes requested):

src/rust/bitbox02/src/lib.rs
benma: Looks like this could actually work safely:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=428df35ef1dd2d35b76c59534820cc14

Though I don't know which, if any of the rules explains that clearly: 

https://doc.rust-lang.org/reference/destructors.html#drop-scopes

To be safe, I would still bind the cstr to a local variable, at least there the rules are much clearer: it will be dropped when leaving the containing block: https://doc.rust-lang.org/reference/destructors.html#scopes-of-local-variables

benma: @NickeZ pointed us to the right section in the docs:

https://doc.rust-lang.org/reference/destructors.html#temporary-scopes

It is indeed dropped the function runs. In this particular case, the temporary is in in the final expression of the function, so it is dropped at the end of the whole match statement. 

So you can leave the code as is :joy:

benma (approved):

src/rust/bitbox02/src/lib.rs
x1ddos: same with:

```rust
let words: Vec<[u8; 101]> = params
        .words
        .iter()
        .map(|word| crate::str_to_cstr_force!(word, 100))
        .collect();
```

in https://github.com/digitalbitbox/bitbox02-firmware/pull/601/files#diff-55bcc1cf6f7436ec1f00d9a58956c7a2R224

benma: Can you be very specific? In the quoted code, there is no temporary, so the same discussion does not apply. Unless you mean something else.

PR #604, opened 2020-10-01
rust: wrap keystore_get_bip39_word

benma:


x1ddos (approved):
LGTM

benma:
Please check again, changed to a `&'static str` return value instead of making a new String.

Previously:

```Rust
pub fn get_bip39_word(idx: u16) -> Result<String, ()> {
    let mut word: *mut u8 = core::ptr::null_mut();
    match unsafe { bitbox02_sys::keystore_get_bip39_word(idx, &mut word) } {
        false => Err(()),
        true => {
            let s = unsafe {
                let len = crate::util::strlen_ptr(word);
                core::slice::from_raw_parts(word, len as _)
            };
            Ok(core::str::from_utf8(&s[..]).unwrap().into())
        }
    }
}

```

benefit: less copies.

benma:
> Hm... static lifetime means it'll live forever. Every time `get_bip39_word` is called, it'll create a new static, right. Is that actually better than copy and eventually drop?

No, you can't create a static dynamically just like that. The wordlist is already static in C.

x1ddos:
Oh, I see. `get_bip39_word` word always returns the same thing now, pointed to somewhere in the C's worlist.
Yeah, that's better. Makes sense.

:+1:

x1ddos (approved):

PR #605, opened 2020-10-01
rust/util/ascii: give all_ascii a more suitable name

x1ddos:
Rust's stdlib has a similar method is_ascii but it's different in that
bb's all_ascii considers only printable characters. This commit makes
the distinction explicit.


benma (approved):

benma:
I took the liberty of merging this :)

PR #606, opened 2020-10-01
rust/api: validate user input in SetDeviceName

x1ddos:
This avoids panic on too long strings and non-ascii characters.
The API now responds with ERR_INVALID_INPUT error, as opposed to
ERR_MEMORY in such cases and results in the bb app reporting back with
"Device name could not be set".

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/set_device_name.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/set_device_name.rs
@@ -22,6 +22,10 @@ use crate::workflow::confirm;
 pub async fn process(
     pb::SetDeviceNameRequest { name }: &pb::SetDeviceNameRequest,
 ) -> Result<Response, Error> {
+    if !util::name::validate(name, bitbox02::memory::DEVICE_NAME_MAX_LEN) {
+        return Err(Error::COMMANDER_ERR_INVALID_INPUT);
+    }
+
     let params = confirm::Params {
         title: "Name",
         body: &name,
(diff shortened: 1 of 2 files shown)


src/rust/bitbox02-rust/src/hww/api/set_device_name.rs
```diff
@@ -22,6 +22,10 @@ use crate::workflow::confirm;
 pub async fn process(
     pb::SetDeviceNameRequest { name }: &pb::SetDeviceNameRequest,
 ) -> Result<Response, Error> {
+    if !util::name::validate(name, bitbox02::memory::DEVICE_NAME_MAX_LEN) {
```
benma, line 25: nice find! was added for the multisig name, but fits here perfectly

src/rust/bitbox02/src/memory.rs
```diff
@@ -16,16 +16,16 @@
 // deduct one for the null terminator.
 pub const DEVICE_NAME_MAX_LEN: usize = bitbox02_sys::MEMORY_DEVICE_NAME_MAX_LEN as usize - 1;
 
-/// `name.as_bytes()` must be smaller or equal to
-/// `DEVICE_NAME_MAX_LEN`, otherwise this function panics.
+/// `name.as_bytes()` must be smaller or equal to `DEVICE_NAME_MAX_LEN`.
```
benma, line 19: not sure this change helps, as now the caller doesn't know if it is a memory or invalid_input error. imho it's fine to document the panic and have the caller check the input, which you do now anyway.

benma (approved):
cool!

src/rust/bitbox02/src/memory.rs
x1ddos: > now the caller doesn't know if it is a memory or invalid_input 

the caller didn't know it before either. it was a panic on invalid input

> it's fine to document the panic

but it doesn't panic anymore.

> have the caller check the input, which you do now anyway.

true. so, I'm thinking to revert this part since the input is now validated in the API. sounds good?

x1ddos: reverted.

x1ddos:
Rebased.

PR #613, opened 2020-10-05
Rust show mnemonic

benma:


src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    let words: Vec<&str> = mnemonic_sentence.split(' ').collect();
+
+    // Part 1) Scroll through words
+    mnemonic::show_mnemonic(&words).await?;
+
+    // Part 2) Confirm words
+    for (word_idx, word) in words.iter().enumerate() {
+        let title = format!("{:02}", word_idx + 1);
+        let (correct_idx, mut choices) = create_random_unique_words(word, NUM_RANDOM_WORDS);
```
NickeZ, line 77: Nothing against vectors, but another option is to pass a `&mut [Option<&str>, 5]` to avoid allocations.

NickeZ: aha, I didn't consider FFI so you can keep it. BUT now I will shock you with my rust-foo. `Option` has null pointer optimizations, and since &str cannot be null, null is used to represent `None`.

So it would be equivalent to send `Option<&str>` to FFI as `&str`.

See https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1321987a22c83a654259e17e6fab73b7

If None needed it's own storage it would be larger than 16.

benma: okay that is indeed amazing rust-foo :D is that guaranteed or an implementation detail?

I forgot that it is not sent over FFI directly as ptr, but all strings are str_to_cstr'd (copied into a new buffer with a null terminator). 

But anyway, the menu_create() takes `&[&str]`. Are you saying I can coerce `&[Option<&str>; 5]` to `&[&str]` safely if I know that all the entries are `Some()`? :scream:

NickeZ: It's practically guaranteed, otherwise `Option<>` would add a lot of overhead to types that don't need it.

No I'm not saying you should do that, you need to unwrap the options if you stay in rust.

But wait a minute, why don't you just do this? `fn create_random_unique_words() -> (u8, [&'static str; NUM_RANDOM_WORDS])`

benma: At the result site I am pushing one more element to the result.

NickeZ: `fn create_random_unique_words() -> (u8, [&'static str; NUM_RANDOM_WORDS+1])`?

benma: breaks modularity? :joy: 

but no I get it, it is definitely possible to do it without allocations. But imho Vec<> is so ergonomic and I see no downside in using it.

src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
```diff
@@ -0,0 +1,91 @@
...
+
+use pb::response::Response;
+
+use crate::workflow::{mnemonic, status, unlock};
+use bitbox02::keystore;
+
+const NUM_RANDOM_WORDS: u8 = 5;
+
+fn create_random_unique_words(word: &str, length: u8) -> (u8, Vec<&str>) {
+    fn rand16() -> u16 {
```
x1ddos, line 29: Would it make sense to move this into a util mod?

x1ddos (approved):
Very nice. LGTML.

src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
benma: Seems to be too specific to be in general util.

src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
NickeZ: 👍 allocations aren't so bad when you have "destructors" :)

PR #614, opened 2020-10-07
apps/btc/sign_msg: stricter input validation

benma:
Future script types should not be automatically accepted.

If a future script type had a different signing algorithm, without
this check, and old firmware would just accept the api call and do the
wrong algorithm.


x1ddos (approved):
Cool. Please add a test case in test/unit-test/test_app_btc_sign_msg.c.
LGTM

benma:
> Cool. Please add a test case in test/unit-test/test_app_btc_sign_msg.c.
> LGTM

Done, also many other invalid inputs.

PR #615, opened 2020-10-07
py/bitbox02: min version requirement and restrict btc_sign_msg

benma:


py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -701,5 +710,12 @@ def _check_max_version(self) -> None:
             if self.version >= MIN_UNSUPPORTED_BITBOX02_BTCONLY_FIRMWARE_VERSION:
                 raise LibraryVersionOutdatedException(self.version)
 
+    def _require_atleast(self, version: semver.VersionInfo) -> None:
+        """
+        Raises UnsupportedException if the current firmware version is not at least the required version.
+        """
+        if self.version < version:
```
x1ddos, line 717: This would make dev work slightly more difficult. Can we relax the restriction and allow dev firmware build at any commit?

benma: Why is it restrictive? The only difference is that we need to bump the firmware version as soon as we add a new feature, not right before release. This is good practice anyway.

x1ddos: Imagine I want to try something based in a previous commit at which the version was older than what's required here. I guess I'd need to also update the version to make this check happy. Feels a bit annoying.

x1ddos: But I guess I can live with it :)

x1ddos (approved):
LGTM

PR #616, opened 2020-10-07
CMakeLists: bump firmware version to 9.2.0

benma:


x1ddos (approved):
For those who follow from home: this will satisfy the min version requirement in py/bitbox02 added in https://github.com/digitalbitbox/bitbox02-firmware/pull/615 first.

Whenever we make a change which bumps major semver, we'll bump the firmware version in master from now on. This is different from earlier where we always updated firmware versions just before a release.

LGTM

PR #619, opened 2020-10-07
rust/password: use heap allocation

benma:
Much better ergonomics and less copies.

### DIFF
--- a/src/rust/bitbox02-rust-c/src/workflow.rs
+++ b/src/rust/bitbox02-rust-c/src/workflow.rs
@@ -20,7 +20,6 @@ extern crate alloc;
 
 use alloc::boxed::Box;
 use alloc::string::String;
-use bitbox02::password::Password;
 use bitbox02_rust::bb02_async::{block_on, spin, Task};
 use bitbox02_rust::workflow::{confirm, password, status, unlock, verify_message};
 use core::fmt::Write;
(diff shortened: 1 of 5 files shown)


x1ddos (commented):
Don't need `clear` fn and `Drop` impl since Box will do it automatically?

Also, in `test_clear` would be better to just drop the `pw` and check that the value was cleared.

benma:
> Don't need `clear` fn and `Drop` impl since Box will do it automatically?

Box will safely zero its contents on drop? I need a reference for that.

We should change it to use zeroize though, but that is a different matter.

benma:
> > Don't need `clear` fn and `Drop` impl since Box will do it automatically?
> 
> Box will safely zero its contents on drop? I need a reference for that.
> 
> We should change it to use zeroize though, but that is a different matter.

Disproven in private chat y example ;)

x1ddos:
> > > Don't need `clear` fn and `Drop` impl since Box will do it automatically?
> > 
> > 
> > Box will safely zero its contents on drop? I need a reference for that.
> > We should change it to use zeroize though, but that is a different matter.
> 
> Disproven in private chat by example ;)

Indeed! From our playground it turned out to clear only a few first elements.

x1ddos (approved):
LGTM

PR #620, opened 2020-10-08
Cancelable password check

benma:
Builds on #619

When showing mnemonic or creating a backup, the password entry to check the password cannot be aborted. This PR fixes this.

### DIFF
--- a/src/rust/bitbox02/src/ui/ui.rs
+++ b/src/rust/bitbox02/src/ui/ui.rs
@@ -68,6 +68,7 @@ pub fn trinary_input_string_create_password<'a, F>(
     title: &str,
     special_chars: bool,
     confirm_callback: F,
+    cancel_callback: Option<ContinueCancelCb<'a>>,
 ) -> Component<'a>
 where
     // Callback must outlive component.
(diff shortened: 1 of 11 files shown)


src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -88,21 +89,39 @@ where
         callback(password_out);
     }
 
+    unsafe extern "C" fn c_cancel_callback(param: *mut c_void) {
+        let callback = param as *mut ContinueCancelCb;
```
x1ddos, line 93: I still think separate `ContinueCb` and `CancelCb` alias types would've been better to read.

x1ddos (approved):
Nice. I like `CanCancel` enum a lot.
LGTM but you'll want to also update py/send_message.py because when you cancel a cancellable password input, it dumps a stacktrace:

raise UserAbortException(response.error.code, response.error.message)
bitbox02.communication.bitbox_api_protocol.UserAbortException: error code: 104, message: aborted by the user

src/rust/bitbox02/src/ui/ui.rs
benma: But in menu_create, it is one C callback handling both?

```
unsafe extern "C" fn c_continue_cancel_cb(param: *mut c_void) {
        let callback = param as *mut ContinueCancelCb;
        (*callback)();
    }
```

How to deal with that?

benma:
> raise UserAbortException(response.error.code, response.error.message)
> bitbox02.communication.bitbox_api_protocol.UserAbortException: error code: 104, message: aborted by the user

Done, and also changed show_mnemonic() to raise all exceptions, so the user can check them in a sane way. This is a breaking change and hence a major version bump.

src/rust/bitbox02/src/ui/ui.rs
NickeZ: have three types :D

PR #621, opened 2020-10-09
add a screensaver

benma:
Currently set to ~1min until activation, see screen_saver.c - ACTIVE_AFTER.

The screensaver is reset with any touch or any screen_stack
change. The latter means there is something new for the user to see.

API calls per se do not reset the screensaver - only if they display
something via screen_stack.

The screensaver currently also appears if the user is in the middle of
a workflow or API call, if they do not interact with the device. It is
unclear if this is desirable, but in general it would work well
together with a general API timeout or automatic device lock.

Finally, some more thinking is required around the three main screens
and when and how to present them:

 1. "See the BitBoxApp" - currently shown until unlocked.
 2. Default waiting screen when nothing is going on. Shows the
      BitBox02 logo.
 3. Screensaver after inactivity. Currently it is an animated
      BitBox02 logo.

Currently it is confusing to leave the animated screensaver just to go
back to the same image logo.


My1:
> * "See the BitBoxApp" - currently shown until unlocked.

not exactly precise after the last few tests I had at least before this PR. if you use U2F you unlock the device and get the see bitboxapp screen after using regardless, so it's not really a good indicator of it being locked.

benma:
> > * "See the BitBoxApp" - currently shown until unlocked.
> 
> not exactly precise after the last few tests I had at least before this PR. if you use U2F you unlock the device and get the see bitboxapp screen after using regardless, so it's not really a good indicator of it being locked.

Right, to be accurate: it is shown until a noise connection is requested by the host, which does not happen via U2F. This is inconsistent, good catch!

My1:
honestly I am not even sure whether that's that useful info to the user, rather than the acutal lock state.
in fact I think personally at least that see see bitboxapp screen makes most sense on a reset device for new users but if you have own a device that you setup the bitbox app should be known.

benma:
> Currently it is confusing to leave the animated screensaver just to go back to the same image logo.

This is pretty much solved with the reduced brightness during the screensaver, which makes it clear that one left the screensaver.

My1:
> > Currently it is confusing to leave the animated screensaver just to go back to the same image logo.
> 
> This is pretty much solved with the reduced brightness during the screensaver, which makes it clear that one left the screensaver.

it helps but if anyone wants or cares, I do have (a prototype of) a bouncing shiftcrypto logo similar to the screensavers of DVD Players back in the day

src/ui/screen_saver.c
```diff
@@ -0,0 +1,63 @@
...
+#include "screen_saver.h"
+
+#include <hardfault.h>
+#include <ui/components/screensaver.h>
+
+#include <stdbool.h>
+#include <stdint.h>
+
+// TODO: use a TIMER interrupt to get a more accurate timer.
+// 27000 is ~1min.
```
My1, line 27: I think this comment is missing a zero

benma: :+1: fixed

My1:
> 3\. Screensaver after inactivity. Currently it is an animated
>     BitBox02 logo.

not anymore :-)

src/ui/components/screensaver.c
```diff
@@ -0,0 +1,106 @@
...
+    // pixels per frame
+    static int8_t x_direction = 1;
+    static int8_t y_direction = 1;
+    // setting relative speed for both axes
+    const int8_t x_slowdown = 1;
+    const int8_t y_slowdown = 1;
+
+    const uint16_t master_slowdown = 6; // for slowing down both x and y in the same amount
+
+    if (counter % (master_slowdown * x_slowdown) == 0) {
```
x1ddos, line 43: I'd simplify a bit and remove master_slowdown. You have both x_ and y_ nicely defined next to each other. It's very easy to just change both.

Could also be thought of as x_speedup and y_speedup :)
So, maybe a neutral x_speed and y_speed?

src/ui/components/screensaver.c
```diff
@@ -0,0 +1,106 @@
...
+        image->position.left += x_direction;
+        // if the screensaver is at the edge (or outside e.g. due to screensaver_reset), and moving
+        // away from the screen, flip the direction so it will always be moving inside or towards
+        // the screen
+        if (((image->position.left + image->dimension.width) >= component->dimension.width &&
+             x_direction > 0) ||
```
x1ddos, line 49: I'd put xxx_direction in front of the conditions. Just easier to reason and no need to even compute if the direction is different from the branch.

For example, I was just reading this whole thing: "image->position.left + ..." up to x_direction > 0 where I realized: oh wait, that's not where I wanted to look.

src/ui/components/screensaver.c
```diff
@@ -0,0 +1,106 @@
...
+#include <screen.h>
+#include <ui/ui_util.h>
+
+#include <string.h>
+
+static void _render(component_t* component)
+{
+    component_t* image = component->sub_components.sub_components[0];
+
+    static uint16_t counter = 0;
```
x1ddos, line 30: shouldn't there be a `#include <stdint.h>` at the top for these types?

src/ui/screen_saver.c
```diff
@@ -0,0 +1,74 @@
...
+#ifndef TESTING
+#include <ui/oled/oled.h>
+#endif
+
+#include <stdbool.h>
+#include <stdint.h>
+
+// TODO: use a TIMER interrupt to get a more accurate timer.
+// 270000 is ~1min.
+#define ACTIVE_AFTER 270000
```
x1ddos, line 28: What unit is this, number of the main loop iterations?

x1ddos (commented):
I forgot, is it going into the next release?

src/ui/components/screensaver.c
benma: agree with the simplification, but the naming is accurate: higher value = slower, so `slowdown` is more natural. with `speed`, you would expect higher speed with higher values.

src/ui/components/screensaver.c
```diff
@@ -0,0 +1,106 @@
...
+    counter++;
```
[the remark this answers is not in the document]
benma: commented

src/ui/components/screensaver.c
benma: done

src/ui/components/screensaver.c
benma: done

src/ui/components/screensaver.c
benma: done

src/ui/screen_saver.c
benma: yes; added comment

benma:
> I forgot, is it going into the next release?

yes ;) 

addressed all comments, PTAL

My1:
what's PTAL?

x1ddos (approved):
LGTM. Nice work you two! :rocket:

x1ddos:
@My1 PTAL = please take another look. LGTM = looks good to me. Just quicker to type.

My1:
to be honest, caging the screensaver in like I did took me a bit lol. basically no matter where the thing is it will fly towards the screen, and when inside (even on just one axis) the bounce logic will activate for that axis or if in screen obviously constrained to that.
I got the Idea from the DVD screensavers btw.

My1:
Shiftcrypto logo
for reference, here the image used for the screensaver

PR #623, opened 2020-10-09
Mnemonic U2F

My1:
## What is this for?

so with U2F the bitbox displays this kinda long, hard to recognize or compare half-SHA256 hash in hex for sites that are not listed, it might be easier to recognize sites when using a mnemonic, similar to what happens on a keepkey when using U2F, in fact since the keepkey's implementation is so simple (convert first 44 bits to words using the BIP39 list and that's literally it) I went with the same approach.

Despite the Idea being 100% taken from the nice people at keepkey, the code is 100% hacked together by me using other code that was already in the bitbox firmware and a few extra changes.

## What this does
1) known sites work as before
2) unknown sites get their half AppID-hash replaced by the 4 BIP-39 words represented by the first 44 bits of that hash (a bit less secure but far more recognizable and easier to remember)
3) if something goes wrong with wally, or the word cutter, go back to the half-hash method

## Example
my Fido sandbox https://my1.dev/wa/my1.php uses an rpID/appID of `my1.dev`, which hashes to `CBFC883093DCA10D8D97F079AAFA4712FA71AF97C0F47A8CE2A2754FA68F6598`, which when used as raw entropy in a bip39 tool results in the phrase `slogan tone around chicken skate man curve wrap keen fit mushroom chalk poem style gallery burden kind grow possible tunnel whip element nose beyond`, the first 4 words being `slogan tone around chicken`, which will be what is displayed on the Bitbox02 like this:
image

## What is changed
1) bring back the `_split_and_save_wordlist()` function from show_mnemonic.c with a few changes to allow it to be used for short mnemonics in an easier manner, as well as load up wally_bip39 for the bip39 conversion
2) add a constant to modify the length of short mnemonics (in case that might be ever needed somewhere else)
3) amend the `_app_string()` function with the bip39 conversion, the cutting into a short array of words, and displaying them instead of the hex
4) adding a fallback to use hex if wally fails or the words generated are less than the displayed 4.

## What could be improved
- on success (aka the fallback not being triggered), we could consider adding a touchpoint to the bottom left to alternate between hex and mnemonic for more user choice, not sure if that can be done that easily (even if, probably not by me or at least not that quickly, agent ransack can only help me to an extent)

## Testing Checklist
- [x] successful build,
- [x] successful flash using dev device
- [x] display of known appid
- [x] display of mnemonic
- [ ] bgus appIDs
- [ ] noninterference with other functions

### DIFF
--- a/src/u2f/u2f_app.c
+++ b/src/u2f/u2f_app.c
@@ -7,6 +7,7 @@
 
 #include <stddef.h>
 #include <stdio.h>
+#include <wally_bip39.h>
 
 #define APPID_BOGUS_CHROMIUM "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
 #define APPID_BOGUS_FIREFOX "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
@@ -15,6 +16,22 @@ typedef struct {
     uint8_t app_id[32];
     const char* name;
 } app_t;
+
+// shorten (and split as array) menonic for displaying hashes in a more recognizable way, at the
+// cost of less entropy, output configurable for other use return success
+static bool _short_mnemonic(char* mnemonic, const char** wordlist_out, uint8_t num_words)
+{
+    char* next_word = strtok(mnemonic, " ");
+    int i = 0;
+    // stop when there are no words or the needed length has been reached
+    while (next_word != NULL && i < num_words) {
+        wordlist_out[i] = next_word;
+        next_word = strtok(NULL, " ");
+        i++;
+    }
+    // if menmonic wasnt long enough, return false
+    return (i == num_words);
+}
 
 static const app_t _apps[] = {
     {
@@ -138,6 +155,27 @@ static void _app_string(const uint8_t* app_id, char* out, size_t out_len)
             snprintf(out, out_len, "%s", app->name);
             return;
         }
+    }
+    char* mnemonic = NULL;
+    // fallback to hex if something goes wrong
+    if (bip39_mnemonic_from_bytes(NULL, app_id, 32, &mnemonic) == WALLY_OK) {
+        const char* words[4];
+        // second fallback, if we get less than 4 words (hardcoded in case the constant gets changed
+        // someday), go back to hex
+        if (_short_mnemonic(mnemonic, words, 4)) {
+            snprintf(
+                out,
+                out_len,
+                "Unknown site:\n%s %s\n%s %s",
+                words[0],
+                words[1],
+                words[2],
+                words[3]);
+            free(mnemonic);
+            return;
+        }
+        // free regardless of whether or not enough words have been created
+        free(mnemonic);
     }
     char appid_hex[32 * 2 + 1] = {0};
     util_uint8_to_hex(app_id, 32, appid_hex);


src/u2f/u2f_app.c
```diff
@@ -7,15 +7,37 @@
...
+
+
 typedef struct {
     uint8_t app_id[32];
     const char* name;
 } app_t;
 
+//short menonic for displaying hashes in a more recognizable, way, at the cost of less entropy
+//return number of actually generated words
+static int short_mnemonic(
```
benma, line 25: we have a convention of prefixing static functions with `_`, i.e. `_short_mnemonic`

src/u2f/u2f_app.c
```diff
@@ -139,6 +161,16 @@ static void _app_string(const uint8_t* app_id, char* out, size_t out_len)
             return;
         }
     }
+    char* mnemonic=NULL;
+    //fallback to hex if something goes wrong
+    if (bip39_mnemonic_from_bytes(NULL, app_id, 32, &mnemonic) == WALLY_OK) {
+        const char* words[SHORT_MNEMONIC_LENGTH];
+        //second fallback
+        if(short_mnemonic(mnemonic, words)>=4) {
```
benma, line 169: use the define here instead of `4`

src/u2f/u2f_app.c
```diff
@@ -7,15 +7,37 @@
...
+#define SHORT_MNEMONIC_LENGTH 4
+
+
 typedef struct {
     uint8_t app_id[32];
     const char* name;
 } app_t;
 
+//short menonic for displaying hashes in a more recognizable, way, at the cost of less entropy
+//return number of actually generated words
```
benma, line 24: it is actually returning one less than the number of generated words.

src/u2f/u2f_app.c
```diff
@@ -7,15 +7,37 @@
 
 #include <stddef.h>
 #include <stdio.h>
+#include <wally_bip39.h>
 
 #define APPID_BOGUS_CHROMIUM "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
 #define APPID_BOGUS_FIREFOX "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
 
+#define SHORT_MNEMONIC_LENGTH 4
```
benma, line 15: I propose a more descriptive name: `SHORT_MNEMONIC_NUM_WORDS`.

the current name is ambiguous, as it could refer to bytes, words, etc.

src/u2f/u2f_app.c
```diff
@@ -139,6 +161,16 @@ static void _app_string(const uint8_t* app_id, char* out, size_t out_len)
             return;
         }
     }
+    char* mnemonic=NULL;
+    //fallback to hex if something goes wrong
+    if (bip39_mnemonic_from_bytes(NULL, app_id, 32, &mnemonic) == WALLY_OK) {
```
benma, line 166: this uses `malloc` to allocate `mnemonic`, but a `free(mnemonic)` is missing, leading to a memory leak.

benma (changes requested):
Your first code contribution, awesome! Keep going.

I left some comments. Please fix the functional/code quality ones. The nitpicks you can leave or take  :smile:

src/u2f/u2f_app.c
```diff
@@ -139,6 +161,16 @@ static void _app_string(const uint8_t* app_id, char* out, size_t out_len)
...
+        const char* words[SHORT_MNEMONIC_LENGTH];
+        //second fallback
+        if(short_mnemonic(mnemonic, words)>=4) {
+            snprintf(out, out_len, "Unknown site:\n%s %s\n%s %s", words[0], words[1],words[2],words[3]);
+            return;
+        }
+    }
     char appid_hex[32 * 2 + 1] = {0};
```
benma, line 174: hmm I think this code is not reachable, right? 32 bytes to mnemonic should always succeed, and so should the wowrd splitting. 

Could replace it with `Abort("u2f app string")`, which halts the device in unexpected situations.

src/u2f/u2f_app.c
My1: okay, I tried to generalize where possible to make the function also possible to be used at other places if the need arises while also making sure that someone changing the constant to fit other needs while forgetting this in here (which is due to the sprintf kinda hardwired to a min of 4) to make sure these 4 are in to prevent it crashing

src/u2f/u2f_app.c
My1: that's no problem

src/u2f/u2f_app.c
benma: My bad, you are right.

src/u2f/u2f_app.c
My1: done.

src/u2f/u2f_app.c
My1: done

My1:
okay half a dozen force pushes later (because c is annoying) finally something that satisfied the checks

src/u2f/u2f_app.c
```diff
@@ -7,15 +7,42 @@
...
 } app_t;
 
+// shorten (and split as array) menonic for displaying hashes in a more recognizable way, at the
+// cost of less entropy, output configurable for other use return success
+static bool _short_mnemonic(char* mnemonic, const char** wordlist_out, uint8_t num_words)
+{
+    // use zero for default
+    if (num_words == 0) {
+        num_words = SHORT_MNEMONIC_NUM_WORDS;
+    }
```
benma, line 29: please delete, this is dead code, along with the then unused `SHORT_MNEMONIC_NUM_WORDS` definition.

benma (changes requested):
almost there :)

src/u2f/u2f_app.c
benma: to elaborate, the definition does not help as you can't change it to anything else and expect it to work. so the magic `4` at the call site is good in this exceptional case ;)

My1: okay...

My1:
done. that was an odyssey -.-

src/u2f/u2f_app.c
```diff
@@ -16,6 +17,26 @@ typedef struct {
     const char* name;
 } app_t;
 
+// shorten (and split as array) menonic for displaying hashes in a more recognizable way, at the
+// cost of less entropy, output configurable for other use return success
+static bool _short_mnemonic(char* mnemonic, const char** wordlist_out, uint8_t num_words)
+{
+    // break out if the target array is too short
+    if (num_words > sizeof(wordlist_out)) {
```
benma, line 25: unfortunately C strikes again: sizeof(wordlist_out) is only the size of a pointer, which is 4 or 8 :see_no_evil:  to make it easy, you can just drop this whole check.

benma (changes requested):

src/u2f/u2f_app.c
My1: c sux. done.

benma (approved):

benma:
:100: thanks a lot!

PR #625, opened 2020-10-10
btc: do not show `bech32` in receive screen

benma:
It is redundant and possibly confusing, especially since the BitBoxApp
has unified accounts (only one Bitcoin account) by default.


benma:
cc @jadzeidan

jadzeidan:
Makes sense. Good catch 👍

x1ddos (approved):
+1 LGTM

PR #627, opened 2020-10-10
Remove verify_pub() and move title wrapping to the confirm params

benma:


src/ui/components/confirm.c
```diff
@@ -109,9 +111,21 @@ component_t* confirm_create(
...
 
     slider_location_t slider_position = top_slider;
+
     // Create labels. We nest them in a body component that covers the screen minus the title bar,
     // so that the CENTER positioning starts below the title bar.
-    component_t* title_component = label_create(params->title, NULL, CENTER_TOP, confirm);
+
+    const UG_FONT* font = &font_font_a_11X10;
+    const char* title = params->title;
+    char wrapped_title[128] = {0};
```
x1ddos, line 120: Use `UG_MAX_LINE_COLS` from src/screen.h?

x1ddos (approved):
LGTM

src/ui/components/confirm.c
benma: That seems unrelated, just happens to be the same value. Here 128 is the total buffer that holds the wrapped text and has nothing to do with a single line.

x1ddos: ah ok. then why 128? a short comment would be nice. I know you just copied it but it's always good to improve docs imho.

benma: I couldn't make a good comment, cause I don't know. Or is a comment like "Arbitrary number that fits all wrapped titles" helpful? :D

Maybe the opposite way works: just document in the function that the wrapped title should not exceed this?

x1ddos: I think both would be great. IMHO even if it says "arbitrary" already helps: it doesn't make it want and search for 128 numbers in `#define` and constants; I'll just know there's nothing to search for :)

PR #628, opened 2020-10-11
api: port ElectrumEncryptionKey to Rust

benma:
Builds on https://github.com/digitalbitbox/bitbox02-firmware/pull/626

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/electrum.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/electrum.rs
@@ -1,0 +1,111 @@
+// Copyright 2020 Shift Crypto AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::pb;
+use super::Error;
+
+use pb::response::Response;
+
+use bitbox02::keystore;
+
+const HARDENED: u32 = 0x80000000;
+const ELECTRUM_WALLET_ENCRYPTION_KEYPATH_LEVEL_ONE: u32 = 4541509 + HARDENED;
+const ELECTRUM_WALLET_ENCRYPTION_KEYPATH_LEVEL_TWO: u32 = 1112098098 + HARDENED;
+
+/// Returns the electrum wallet encryption xpub..
+/// `keypath` currently needs to be m/4541509'/1112098098'
+pub async fn process(
+    pb::ElectrumEncryptionKeyRequest { keypath }: &pb::ElectrumEncryptionKeyRequest,
+) -> Result<Response, Error> {
+    if *keypath
+        != [
+            ELECTRUM_WALLET_ENCRYPTION_KEYPATH_LEVEL_ONE,
+            ELECTRUM_WALLET_ENCRYPTION_KEYPATH_LEVEL_TWO,
+        ]
+    {
+        return Err(Error::InvalidInput);
+    }
+    let xpub = keystore::encode_xpub_at_keypath(keypath, keystore::xpub_type_t::XPUB)
+        .or(Err(Error::InvalidInput))?;
+
+    Ok(Response::ElectrumEncryptionKey(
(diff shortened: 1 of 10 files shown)


src/rust/bitbox02-rust/src/hww/api/electrum.rs
```diff
@@ -0,0 +1,98 @@
...
+use super::pb;
+use super::Error;
+
+use pb::response::Response;
+
+use bitbox02::keystore;
+
+const HARDENED: u32 = 0x80000000;
+const ELECTRUM_WALLET_ENCRYPTION_KEYPATH_LEVEL_ONE: u32 = 4541509 + HARDENED;
+const ELECTRUM_WALLET_ENCRYPTION_KEYPATH_LEVEL_TWO: u32 = 1112098098 + HARDENED;
```
x1ddos, line 24: Why is it called "encryption"? It's just an xpub in the way electrum wants it, right. Nothing else special.

benma: Electrum uses it as an encryption key to encrypt the wallet.

x1ddos: Sure but that's on the Electrum side. Who knows what other uses people will find for xpubs :)
All we do here is formatting an xpub, for what concerns the bb02 API anyway.

x1ddos: Ok, maybe the const names are fine but the protobuf request name is strange.

benma: >  Sure but that's on the Electrum side. Who knows what other uses people will find for xpubs :)
> All we do here is formatting an xpub, for what concerns the bb02 API anyway.

It was made for only this purpose. I don't see it being helpful to make a name more generic and less useful as a result. Xpub is a misnomer here as you should not make addresses from it and send coins to it (Electrum simply abused xpubs for the purpose).

We don't call xpubs `bytes` either cause it could be used for other purposes :D

x1ddos: I see. This one is a very good point:

> you should not make addresses from it and send coins to it (Electrum simply abused xpubs for the purpose)

Would've appreciated if it were somewhere in the docs, maybe for `process` fn.

benma: Rebased and added the comment.

src/rust/bitbox02-rust/src/hww/api/electrum.rs
```diff
@@ -0,0 +1,100 @@
...
+                    key: "<xpub>".into()
+                },
+            ))
+        );
+
+        // Invalid keypath.
+        mock(Default::default());
+        assert_eq!(
+            block_on(process(&pb::ElectrumEncryptionKeyRequest {
+                keypath: vec![ELECTRUM_WALLET_ENCRYPTION_KEYPATH_LEVEL_ONE, 0]
```
x1ddos, line 95: I think it'd be nice to have another test with a keypath len != 2

src/rust/bitbox02/src/keystore.rs
```diff
@@ -111,3 +113,21 @@ pub fn secp256k1_pubkey_uncompressed(
...
+        bitbox02_sys::keystore_encode_xpub_at_keypath(
+            keypath.as_ptr(),
+            keypath.len() as _,
+            xpub_type,
+            xpub.as_mut_ptr(),
+            xpub.len() as _,
+        )
+    } {
+        true => Ok(crate::util::str_from_null_terminated(&xpub[..])
+            .unwrap()
```
x1ddos, line 129: Since the returned value is a `Result`, would be make more sense to propagate an error with `?` instead of panic?

x1ddos (approved):
LGTM

src/rust/bitbox02/src/keystore.rs
benma: What do you think?

In my opinion, panics are appropriate for programming errors that should never happen (independent of user input, IO, hardware environment, etc). In this case, the xpub will always be ascii and null terminated and fit in the `XPUB_ENCODED_LEN` length, and if it doesn't, something is very bad.

src/rust/bitbox02-rust/src/hww/api/electrum.rs
benma: done

src/rust/bitbox02/src/keystore.rs
x1ddos: Yeah sure a valid point but I'm personally more for a gracefull error handling and make users suffer as little as possible, even if it's a programming error, trying to catch as many of those as possible at compile time and in unit tests.

PR #629, opened 2020-10-11
multisig xpub option

benma:


src/apps/btc/btc.c
```diff
@@ -138,7 +138,7 @@ app_btc_result_t app_btc_address_multisig(
     }
 
     const char* title = "Receive to";
-    if (!apps_btc_confirm_multisig(title, coin, multisig_registered_name, multisig, false)) {
+    if (!apps_btc_confirm_multisig(title, coin, multisig_registered_name, multisig, false, 0)) {
```
x1ddos, line 141: use `BTCRegisterScriptConfigRequest_XPubType_AUTO_ELECTRUM` for better readability? otherwise, it isn't clear what 0 means here.

x1ddos: Also in btc_sign_validate.c where it calls `apps_btc_confirm_multisig(..., 0)`

x1ddos (approved):
Thanks! LGTM.

src/apps/btc/btc.c
benma: put it in a var called `ignore`, to indicate that it is ignored, since verify_xpubs is false.

src/apps/btc/btc_sign_validate.c
```diff
@@ -65,8 +65,10 @@ app_btc_result_t app_btc_sign_validate_init_script_configs(
             // Not previously registered -> fail.
             return APP_BTC_ERR_INVALID_INPUT;
         }
+        BTCRegisterScriptConfigRequest_XPubType ignored =
```
x1ddos, line 68: This is perfect. Couldn't have been more clear. :+1: 
Thanks!

PR #630, opened 2020-10-12
Attestation

benma:


x1ddos (approved):
Verified the pubkeys with @benma. All match.
:heavy_check_mark:

PR #633, opened 2020-10-13
oled: fix mirroring

benma:
Resetting the offset param was forgotten. This lead to rendering being
broken when rotating the screen more than once in the bootloader.

### DIFF
--- a/src/ui/oled/oled.c
+++ b/src/ui/oled/oled.c
@@ -257,6 +257,7 @@ void oled_mirror(bool mirror)
     } else {
         _write_cmd(OLED_CMD_SET_SEGMENT_RE_MAP_COL127_SEG0);
         _write_cmd(OLED_CMD_SET_COM_OUTPUT_SCAN_UP);
+        _write_cmd_with_param(OLED_CMD_SET_DISPLAY_OFFSET, DISPLAY_OFFSET);
     }
 #elif PLATFORM_BITBOXBASE == 1
     (void)mirror;


src/ui/oled/oled.c
```diff
@@ -257,6 +257,7 @@ void oled_mirror(bool mirror)
     } else {
         _write_cmd(OLED_CMD_SET_SEGMENT_RE_MAP_COL127_SEG0);
         _write_cmd(OLED_CMD_SET_COM_OUTPUT_SCAN_UP);
+        _write_cmd_with_param(OLED_CMD_SET_DISPLAY_OFFSET, DISPLAY_OFFSET);
```
x1ddos, line 260: Makes sense. This is also what's set in the `oled_init` at the beginning.
While there, I would also make a `#define` for the other offset of `0x20` when mirrored.

benma: Done

x1ddos (approved):
LGTM

src/ui/oled/oled.c
```diff
@@ -160,6 +160,7 @@ static inline void _write_cmd_with_param(uint8_t command, uint8_t value)
 #define ADDRESSING_MODE 0x21
 #define SEGMENT_REMAP OLED_CMD_SET_SEGMENT_RE_MAP_COL127_SEG0
 #define DISPLAY_OFFSET 0x60
+#define DISPLAY_OFFSET_MIRRORED 0x20
```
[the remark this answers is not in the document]
x1ddos: ah, never mind. it isn't used in bbbase :)

PR #635, opened 2020-10-15
rust: bump to 1.47 stable

benma:
And use updated Docker image based on this.

I tried this to fix the reproducible builds issue, but it didn't help. Still a valid PR :P


.ci/run-container-ci
```diff
@@ -25,7 +25,7 @@
 set -e
 set -x
 
-CONTAINER=shiftcrypto/firmware_v2:18
+CONTAINER=shiftcrypto/firmware_v2:19
```
x1ddos, line 28: Do the same in dockerenv? https://github.com/digitalbitbox/bitbox02-firmware/blob/ddcdb0e/scripts/dockerenv.sh#L27.

There's also `docker-compose.yml` in the root of the repo and it references unversioned image. Is that compose config file even used?

benma: dockerenv is deliberately at unversioned, so it works after `make dockerinit` and `docker pull shiftcrypto/firmware_v2`.

I never used docker-compose.yml, so let's delete it.

x1ddos: :+1:

x1ddos (approved):
LGTM

PR #636, opened 2020-10-17
rust: do not patch bytes anymore

benma:
It does not add all the optional deps in vendor/ as it used to.


x1ddos (approved):
For those playing along at home: this is most likely because loom was moved to dev dependencies in https://github.com/tokio-rs/bytes/pull/416.

LGTM

PR #637, opened 2020-10-17
releases/v9.2.0: add note about reproducible builds issue

benma:


releases/firmware-btc-v9.2.0/README.md
```diff
@@ -0,0 +1,4 @@
+**Note**: this release has a reproducible build issue. When building as `root` inside Docker, as
+`build.sh` does, the build is slightly different to when it is built as `dockeruser`. It is still
+reproducible using `scripts/dockerenv.sh`. In any case, `v9.2.1` replaces this release, fixing this
```
x1ddos, line 3: To clarify for those playing along at home: there's no v9.2.1 tag yet but there will be.

The issue was that paths present in the final binary contained the cargo paths for the deps vendored via a git commit checkout, as opposed to a regular crate version. This is resolved by https://github.com/digitalbitbox/bitbox02-firmware/pull/636.

And the root cause of this might be https://github.com/rust-lang/cargo/issues/8792.

x1ddos (approved):
LGTM

releases/firmware-btc-v9.2.0/README.md
x1ddos: @benma would be nice to add something like ^^ to the git commit msg.

benma: damn too late again. maybe you should do 'changes requested' instead of 'approved' :)

My1: this was a crazy ride, and quite some headache tbh

PR #640, opened 2020-10-17
Makefile: make bootloader naming consistent

benma:


x1ddos (approved):
Man, I was just thinking about renaming these myself the other day but then I was afraid you would be mad at me so I never sent a pull request :smile:

benma:
> Man, I was just thinking about renaming these myself the other day but then I was afraid you would be mad at me so I never sent a pull request smile

uuh why would i be mad? you didn't know hate inconsistent naming? :joy:

.ci/ci
```diff
@@ -32,16 +32,16 @@ make coverage
 
 # Bootloader variants
 make -j8 bootloader
-make -j8 bootloader-devdevice
-make -j8 bootloader-devdevice-locked
+make -j8 bootloader-development
+make -j8 bootloader-development-locked
```
My1, line 36: kinda curious what's a locked devdevice supposed to do? I mean isnt the point of a dev device to be unlocked so newer FW versions can be flashed onto it?

benma: Locked does not refer to the signature checks on the firmware, but to MCU settings that prevent overwriting the flash with JLink.

My1: oh okay interesting to know, thanks

PR #646, opened 2020-10-20
rust: build prost from vendored sources

benma:
All normal deps are already built from vendored sources.

Prost is a git source. Git sources are vendored with `cargo vendor`,
but not automatically build from the vendored sources with `cargo
build`. The added section in the .cargo/config is needed for that, as
explained by the `cargo vendor` output.

See also: https://github.com/rust-lang/cargo/issues/8792

### DIFF
--- a/src/rust/.cargo/config
+++ b/src/rust/.cargo/config
@@ -3,3 +3,9 @@ replace-with = "vendored-sources"
 
 [source.vendored-sources]
 directory = "vendor"
+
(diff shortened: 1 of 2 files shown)


x1ddos (approved):
LGTM

What do people usually say. It's not a bug. It's a "feature"? :)

benma:
Yeah, I don't understand it. Very unexpected behaviour in my eyes, but maybe there are good reasons for it? :shrug:

PR #648, opened 2020-10-22
rust/features: fix typo in the bootloader-development-locked target

benma:


x1ddos (approved):
LGTM

It seems this activates only the additional bitbox02.rs mod but the file is empty so nothing really changed. Of course it's good to have this corrected either way.

PR #649, opened 2020-10-22
rust: compile less code for the bootloader

benma:
A lot of Rust code is compiled even when the target is the
bootloader. The code is eliminated at link time, but it is better to
not compile it in the first place, for clarity and compile speed.

There is still some unneeded code compiled in util.rs.

### DIFF
--- a/src/rust/bitbox02-rust-c/src/lib.rs
+++ b/src/rust/bitbox02-rust-c/src/lib.rs
@@ -25,16 +25,21 @@ extern crate std;
 #[macro_use]
 mod alloc;
 
-mod async_usb;
-mod noise;
-mod sha2;
 mod util;
+
+#[cfg(feature = "firmware")]
+mod async_usb;
+#[cfg(feature = "firmware")]
+mod noise;
+#[cfg(feature = "firmware")]
+mod sha2;
+#[cfg(feature = "firmware")]
 mod workflow;
 
-#[cfg(feature = "platform-bitboxbase")]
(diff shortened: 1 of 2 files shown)


src/rust/bitbox02-rust-c/src/lib.rs
```diff
@@ -25,16 +25,21 @@ extern crate std;
...
+#[cfg(feature = "firmware")]
 mod workflow;
 
-#[cfg(feature = "platform-bitboxbase")]
+#[cfg(all(feature = "firmware", feature = "platform-bitboxbase"))]
 pub mod bitboxbase;
 
-#[cfg(feature = "platform-bitbox02")]
+#[cfg(all(feature = "firmware", feature = "platform-bitbox02"))]
 pub mod bitbox02;
```
x1ddos, line 43: This seems strange. Why would it need both firmware and platform-bitbox02. Wouldn't platform-bitbox02 be enough?

benma: there is platform bitbox02 and platform bitboxbase. both have a firmware and a bootloader.

since we don't have any real bootloader rust code, I added the firmware guard in `mod bitboxbase` above, and here too for consistency.

x1ddos: oh, I was looking at it in the wrong way. this is deliberate, of course :)

x1ddos (approved):
LGTM

PR #650, opened 2020-11-09
rust/bitbox02: fix dangling pointers

benma:
The title/body buffers are dropped by the end of `to_c_params()`, but
the pointer survived.

### DIFF
--- a/src/rust/bitbox02/src/ui/types.rs
+++ b/src/rust/bitbox02/src/ui/types.rs
@@ -15,6 +15,8 @@
 extern crate alloc;
 use alloc::boxed::Box;
 
+use util::Survive;
+
 // Taking the constant straight from C, as it's excluding the null terminator.
 #[cfg_attr(feature = "testing", allow(dead_code))]
 pub(crate) const MAX_LABEL_SIZE: usize = bitbox02_sys::MAX_LABEL_SIZE as _;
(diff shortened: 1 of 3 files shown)


x1ddos:
I wonder why not instead do something like this:

```rust
pub fn confirm_create<'a, F>(params: &'a ConfirmParams, result_callback: F) -> Component<'a>
```

benma:
> I wonder why not instead do something like this:
> 
> ```rust
> pub fn confirm_create<'a, F>(params: &'a ConfirmParams, result_callback: F) -> Component<'a>
> ```

The issue is not at this level, but inside the function when with `params.to_c_params()`. `params` itself is not needed and does not need to survive, as it does not hold any buffer used in C.

x1ddos:
> > I wonder why not instead do something like this:
> > ```rust
> > pub fn confirm_create<'a, F>(params: &'a ConfirmParams, result_callback: F) -> Component<'a>
> > ```
> 
> The issue is not at this level, but inside the function with `params.to_c_params()`. `params` itself is not needed and does not need to survive, as it does not hold any buffer used in C.

Yeah I know, I just dislike `title_scatch` and `body_scratch`. Is there a more elegant way to solve this?

benma:
> Yeah I know, I just dislike `title_scatch` and `body_scratch`. Is there a more elegant way to solve this?

I am reasonably sure there is a more elegant way, but I haven't found it yet :O 

I tried adding the scratch to the params struct itself, but then it has to be passed as `&mut` everywhere, which is not nice.

I tried putting it into a resulting struct containing both the C params and the scratch, but haven't figured out yet how to make sure the two struct fields live for the same time. Possibly something with Pin<> or more PhantomData could work, but it seems complicated.

x1ddos:
What if the result of `str_to_cstr` macro were somehow a struct with phantom data instead of a plain buf?

benma:
Then you have to return it, so this applies:

> I tried putting it into a resulting struct containing both the C params and the scratch, but haven't figured out yet how to make sure the two struct fields live for the same time. Possibly something with Pin<> or more PhantomData could work, but it seems complicated.

src/rust/bitbox02/src/ui/types.rs
```diff
@@ -42,6 +42,21 @@ impl Default for Font {
     }
 }
 
+/// Survive forces T to live at least as long as lifetme 'a.
+pub struct Survive<'a, T: 'a> {
```
x1ddos, line 46: Since we're going with this, why not move it to some utils mod? To use in other places.

x1ddos (commented):
> Then you have to return it, so this applies:
> 
> > I tried putting it into a resulting struct containing both the C params and the scratch, but haven't figured out yet how to make sure the two struct fields live for the same time. Possibly something with Pin<> or more PhantomData could work, but it seems complicated.

I'd love to dive more into this but not today.
I actually think we should get rid of `str_to_cstr` macro and use something else, similar to ffi::CString in https://doc.rust-lang.org/std/ffi/struct.CString.html. We already have CStr in bitbox02-rust-c crate. Maybe we could extend it to cover the cases where we need a str_to_cstr.

Anyway, what about these ptrs?

- https://github.com/digitalbitbox/bitbox02-firmware/blob/ddcdb0/src/rust/bitbox02/src/ui/ui.rs#L107
- https://github.com/digitalbitbox/bitbox02-firmware/blob/ddcdb0/src/rust/bitbox02/src/ui/ui.rs#L107

benma:
> Anyway, what about these ptrs?
> 
>     * https://github.com/digitalbitbox/bitbox02-firmware/blob/ddcdb0/src/rust/bitbox02/src/ui/ui.rs#L107
> 
>     * https://github.com/digitalbitbox/bitbox02-firmware/blob/ddcdb0/src/rust/bitbox02/src/ui/ui.rs#L185

As per our previous long study about how long those things live, those are safe: the buffer survives until after the function is finished (and the function itself copies the string in label_create()).

benma:
> I actually think we should get rid of str_to_cstr macro and use something else, similar to ffi::CString in https://doc.rust-lang.org/std/ffi/struct.CString.html. We already have CStr in bitbox02-rust-c crate. Maybe we could extend it to cover the cases where we need a str_to_cstr.

That sounds like a great idea.

src/rust/bitbox02/src/ui/types.rs
benma: Done

x1ddos:
> > Anyway, what about these ptrs?
> > ```
> > * https://github.com/digitalbitbox/bitbox02-firmware/blob/ddcdb0/src/rust/bitbox02/src/ui/ui.rs#L107
> > 
> > * https://github.com/digitalbitbox/bitbox02-firmware/blob/ddcdb0/src/rust/bitbox02/src/ui/ui.rs#L185
> > ```
> 
> As per our previous long study about how long those things live, those are safe: the buffer survives until after the function is finished (and the function itself copies the string in label_create()).

I forgot that it's copied in `label_create`. A very hard to spot but important difference. I wish it was more clear somehow, at least a doc comment.

x1ddos (approved):
LGTM

benma:
> I forgot that it's copied in `label_create`. A very hard to spot but important difference. I wish it was more clear somehow, at least a doc comment.

True, very difficult to spot. I added a short comment everywhere it is copied.

x1ddos:
> > I forgot that it's copied in `label_create`. A very hard to spot but important difference. I wish it was more clear somehow, at least a doc comment.
> 
> True, very difficult to spot. I added a short comment everywhere it is copied.

Thanks!! :pray: 
So small of a change, such a big impact on my happiness :)

PR #651, opened 2020-11-09
Trinary params

benma:
I want to use the trinary input to enter a normal string (to enter the multisig account name), but I don't want to add another trinary_input_create_xyz component and workflow function in C and Rust.

This is a cleanup PR that collapses all the variants into one, using a params struct, like with the Confirm component.


src/rust/bitbox02/src/lib.rs
```diff
@@ -37,8 +37,8 @@ pub mod commander;
 pub mod keystore;
 #[cfg_attr(feature = "testing", path = "memory_stub.rs")]
 pub mod memory;
-pub mod password;
 pub mod random;
+pub mod safeinputstring;
```
x1ddos, line 41: maybe could even be just `input`? then `input::SafeInputString` or `input::SafeString`.

src/rust/bitbox02/src/safeinputstring.rs
```diff
@@ -15,19 +15,20 @@
 extern crate alloc;
 use alloc::boxed::Box;
 
+const INPUT_STRING_MAX_SIZE: usize = bitbox02_sys::INPUT_STRING_MAX_SIZE as _;
+
 /// C-style including null terminator, as it is used in C only so far.
```
x1ddos, line 20: It is also used in many parts in rust these days, like unlock_keystore > password::enter these days, without seemingly switching to C. Of course it does underneath. So, I guess still good to mention it's a null-terminated string but C only - that's probably questionable.

x1ddos (approved):
Haven't tried it myself on a dev device but I assume you did and CI passes so LGTM. All changes make sense.

src/rust/bitbox02/src/lib.rs
benma: Done

src/rust/bitbox02/src/safeinputstring.rs
benma: Right! Removed that part.

PR #652, opened 2020-11-09
delete docker-compose.yml

benma:
Unused.


x1ddos (approved):
LGTM but you'll also want to remove `dockerrun` in https://github.com/digitalbitbox/bitbox02-firmware/blob/ddcdb0e/Makefile#L141-L142.

PR #653, opened 2020-11-10
Multisig name on device

benma:
Can test by showing a multisig receive address in send_message, and when prompted for the name, submit the empty string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,4 @@
+# Changelog
+
+## 9.3.0 [version may change, pending release]
+- Enter multisig account name on the device if the name in BTCRegisterScriptConfigRequest is empty.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,9 +89,9 @@ endif()
 #
 # Versions MUST contain three parts and start with lowercase 'v'.
 # Example 'v1.0.0'. They MUST not contain a pre-release label such as '-beta'.
-set(FIRMWARE_VERSION "v9.2.1")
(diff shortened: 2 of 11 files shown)


benma:
@x1ddos the user can easily enter an invalid name, e.g. an empty name or one that starts with whitespace, or is too long.

I will try to improve on this in another PR.

x1ddos (approved):
LGTM

> empty name or one that starts with whitespace, or is too long

whitespace shouldn't be a problem because `special_chars` is false in the passed params, right.

benma:
> whitespace shouldn't be a problem because `special_chars` is false in the passed params, right.

hmm that was an oversight, meant to enable it. space would be good here right?

x1ddos:
> > whitespace shouldn't be a problem because `special_chars` is false in the passed params, right.
> 
> hmm that was an oversight, meant to enable it. space would be good here right?

but then you would enable all other "special" characters. I'd keep it simple for now and leave as is. you did say you'll want to improve it against an empty name and other things anyway.

x1ddos:
> > > whitespace shouldn't be a problem because `special_chars` is false in the passed params, right.
> > 
> > 
> > hmm that was an oversight, meant to enable it. space would be good here right?
> 
> but then you would enable all other "special" characters. I'd keep it simple for now and leave as is. you did say you'll want to improve it against an empty name and other things anyway.

ah well OTOH who cares what the users types in the name... unless it's propagated to the app and causes some panics :)

benma:
I'll circle back to the special chars question later.

PR #654, opened 2020-11-11
apps/btc: add param to sort xpubs before hashing

benma:
The hash is used to identify a multisig registration.

In multisig, the xpub ordering does not really matter, but our hashes
would change with a different order. This was not a real issue in
practice, as Electrum stores and sends the xpubs in the same order
every time.

In PSBT (at least in the way HWI parses it), xpubs are unordered.

This adds a param to the hashing function to sort the xpubs before
hashing, so that their order does not matter.

### DIFF
--- a/src/apps/btc/btc_common.c
+++ b/src/apps/btc/btc_common.c
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <stdio.h>
+#include <stdlib.h>
 
 #include "btc_common.h"
 
@@ -685,9 +686,31 @@ USE_RESULT bool btc_common_multisig_is_valid(
     return true;
 }
 
-bool btc_common_multisig_hash(
+// serialized_out must be of size BIP32_SERIALIZED_LEN.
+static bool _serialize_xpub(const XPub* xpub, uint8_t* serialized_out)
+{
+    struct ext_key wally_xpub __attribute__((__cleanup__(keystore_zero_xkey))) = {0};
+    if (!apps_common_bip32_xpub_from_protobuf(xpub, &wally_xpub)) {
+        return false;
+    }
+    return bip32_key_serialize(
+               &wally_xpub, BIP32_FLAG_KEY_PUBLIC, serialized_out, BIP32_SERIALIZED_LEN) ==
+           WALLY_OK;
+}
+
+static int _xpubs_sort_comp(const void* elem1, const void* elem2)
+{
+    const uint8_t* xpub1 = (const uint8_t*)elem1;
+    const uint8_t* xpub2 = (const uint8_t*)elem2;
+
+    // Sort by xpub serialization, ignoring the version bytes.
+    return memcmp(xpub1 + 4, xpub2 + 4, BIP32_SERIALIZED_LEN - 4);
+}
+
+static bool _multisig_hash(
     BTCCoin coin,
     const BTCScriptConfig_Multisig* multisig,
+    bool sort_xpubs,
     const uint32_t* keypath,
     size_t keypath_len,
     uint8_t* hash_out)
@@ -738,21 +761,25 @@ bool btc_common_multisig_hash(
         rust_sha256_update(ctx, &num, sizeof(num));
     }
     { // 5. xpubs
+        uint8_t xpubs_serialized[sizeof(multisig->xpubs) / sizeof(*multisig->xpubs)]
+                                [BIP32_SERIALIZED_LEN] = {0};
         for (size_t i = 0; i < multisig->xpubs_count; i++) {
-            uint8_t xpub_serialized[BIP32_SERIALIZED_LEN] = {0};
-            struct ext_key xpub __attribute__((__cleanup__(keystore_zero_xkey))) = {0};
-            if (!apps_common_bip32_xpub_from_protobuf(&multisig->xpubs[i], &xpub)) {
+            if (!_serialize_xpub(&multisig->xpubs[i], xpubs_serialized[i])) {
                 return false;
             }
-            if (bip32_key_serialize(
-                    &xpub, BIP32_FLAG_KEY_PUBLIC, xpub_serialized, sizeof(xpub_serialized)) !=
-                WALLY_OK) {
-                return false;
-            }
+        }
+        if (sort_xpubs) {
+            qsort(
+                xpubs_serialized,
+                multisig->xpubs_count,
+                sizeof(*xpubs_serialized),
+                _xpubs_sort_comp);
+        }
+        for (size_t i = 0; i < multisig->xpubs_count; i++) {
             // Drop the first xpub version, which are the 4 first bytes. They are determined by the
             // above `BIP32_FLAG_KEY_PUBLIC` flag and do not add anything, as the xpub version is
             // chosen ad-hoc depending on the context it is used in.
-            rust_sha256_update(ctx, xpub_serialized + 4, sizeof(xpub_serialized) - 4);
+            rust_sha256_update(ctx, xpubs_serialized[i] + 4, BIP32_SERIALIZED_LEN - 4);
         }
     }
     { // 6. keypath len
@@ -766,4 +793,24 @@ bool btc_common_multisig_hash(
     }
     rust_sha256_finish(&ctx, hash_out);
     return true;
+}
+
+bool btc_common_multisig_hash_unsorted(
+    BTCCoin coin,
+    const BTCScriptConfig_Multisig* multisig,
+    const uint32_t* keypath,
+    size_t keypath_len,
+    uint8_t* hash_out)
+{
+    return _multisig_hash(coin, multisig, false, keypath, keypath_len, hash_out);
+}
+
+bool btc_common_multisig_hash_sorted(
+    BTCCoin coin,
+    const BTCScriptConfig_Multisig* multisig,
+    const uint32_t* keypath,
+    size_t keypath_len,
+    uint8_t* hash_out)
+{
+    return _multisig_hash(coin, multisig, true, keypath, keypath_len, hash_out);
 }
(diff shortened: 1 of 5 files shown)


benma:
In the next PR, I will use the sorted version when registering, and check for both when querying, so it is backwards compatible with existing unsorted registrations.

test/unit-test/test_app_btc_common.c
```diff
@@ -986,35 +986,177 @@ print(hashlib.sha256(b''.join(msg)).hexdigest())
...
         sizeof(hash));
 
-    assert_false(btc_common_multisig_hash(_BTCCoin_MAX + 1, &multisig, keypath, 4, hash));
+    assert_false(btc_common_multisig_hash(_BTCCoin_MAX + 1, &multisig, false, keypath, 4, hash));
+
+    // Test that the hash is correct, and the same for all xpubs permutations if xpubs sort is
+    // enabled.
+
+    // clang-format off
+    const char* permutations[24][4] = {
```
x1ddos, line 1023: Feels like a loop permutating a `const char* xpubs[4]` might've been better. It's just hard to spot where each entry is now and also when updating this in the future, if ever. I think it's easy to make a mistake in the test data which leads to unnecessary debuging: is it my test data or the code I wrote that fails the test?

src/apps/btc/btc_common.h
```diff
@@ -253,6 +254,7 @@ USE_RESULT bool btc_common_multisig_is_valid(
 USE_RESULT bool btc_common_multisig_hash(
     BTCCoin coin,
     const BTCScriptConfig_Multisig* multisig,
+    bool sort_xpubs,
```
x1ddos, line 257: I would've added 2 extra functions, `btc_common_multisig_hash_sorted` and `btc_common_multisig_hash_unsorted` and converted `btc_common_multisig_hash` to a static `_btc_common_multisig_hash`. They would both simply call `_btc_common_multisig_hash` with either `true` or `false`.

It just makes it explicit and more clear to see what's happening in the rest of the code. For example, compare

```c
    if (!btc_common_multisig_hash(coin, multisig, false, keypath, keypath_len, hash)) {
        return APP_BTC_ERR_UNKNOWN;
    }
```

to 

```c
    if (!btc_common_multisig_hash_unsorted(coin, multisig, keypath, keypath_len, hash)) {
        return APP_BTC_ERR_UNKNOWN;
    }
```

assuming you don't remember what the signature of the function is and without looking it up.

src/apps/btc/btc_common.c
```diff
@@ -685,9 +686,42 @@ USE_RESULT bool btc_common_multisig_is_valid(
...
+// serialized_out must be of size BIP32_SERIALIZED_LEN.
+static bool _serialize_xpub(const XPub* xpub, uint8_t* serialized_out)
+{
+    struct ext_key wally_xpub __attribute__((__cleanup__(keystore_zero_xkey))) = {0};
+    if (!apps_common_bip32_xpub_from_protobuf(xpub, &wally_xpub)) {
+        return false;
+    }
+    if (bip32_key_serialize(
+            &wally_xpub, BIP32_FLAG_KEY_PUBLIC, serialized_out, BIP32_SERIALIZED_LEN) != WALLY_OK) {
+        return false;
```
x1ddos, line 698: Or, simplified: `return bip32_key_serialize(...) == WALLY_OK`

src/apps/btc/btc_common.c
```diff
@@ -685,9 +686,42 @@ USE_RESULT bool btc_common_multisig_is_valid(
...
+    uint8_t xpub1_serialized[BIP32_SERIALIZED_LEN] = {0};
+    uint8_t xpub2_serialized[BIP32_SERIALIZED_LEN] = {0};
+    if (!_serialize_xpub(xpub1, xpub1_serialized)) {
+        Abort("serialize_xpub");
+    }
+    if (!_serialize_xpub(xpub2, xpub2_serialized)) {
+        Abort("serialize_xpub");
+    }
+
+    // Sort by xpub serialization, ignoring the version bytes.
```
x1ddos, line 717: I wonder why ignore the version. Shouldn't it always be the same? If it isn't, I would expect some check before this to fail the whole registration flow so it wouldn't even come to sorting.

This is what I see in external/.../wally_bip32.h:

```c
/** Version codes for extended keys */
#define BIP32_VER_MAIN_PUBLIC  0x0488B21E
#define BIP32_VER_MAIN_PRIVATE 0x0488ADE4
#define BIP32_VER_TEST_PUBLIC  0x043587CF
#define BIP32_VER_TEST_PRIVATE 0x04358394
```

src/apps/btc/btc_common.c
```diff
@@ -738,15 +772,16 @@ bool btc_common_multisig_hash(
         rust_sha256_update(ctx, &num, sizeof(num));
     }
     { // 5. xpubs
+        const XPub* xpubs[sizeof(multisig->xpubs)] = {0};
+        for (size_t i = 0; i < multisig->xpubs_count; i++) {
+            xpubs[i] = &multisig->xpubs[i];
+        }
+        if (sort_xpubs) {
+            qsort(xpubs, multisig->xpubs_count, sizeof(*xpubs), _xpubs_sort_comp);
```
x1ddos, line 780: Feels like serializing first, then sorting and finally sha256 would work just as well if not better since you don't need to serialize twice? Assuming sorting would work without ignoring the serialized xpub version.

src/apps/btc/btc_common.c
benma: if it is mainnet or testnet is already encoded in the hash.

xpub versioning is very non-standard (Zpub, etc.), and to the device, it does not matter which is used by the computer wallet. if you use electrum (Zpub), then HWI (xpub), the multisig account is still the same and the user should not register a new one. in other words, different xpub versions should not change the hash.

src/apps/btc/btc_common.c
benma: Yeah I guess. I'll try.

> Assuming sorting would work without ignoring the serialized xpub version.

That question seems independent?

test/unit-test/test_app_btc_common.c
benma: i generated the test data with `itertools.permuate(xpubs, 4)` :P 

didn't feel like spending time writing permutation code in C.

src/apps/btc/btc_common.h
benma: what about `bool sorted = false; btc_common_multisig_hash(coin, multisig, sorted, keypath, keypath_len, hash)`?

src/apps/btc/btc_common.c
benma: Done

src/apps/btc/btc_common.c
benma: Done (serialize, then sort + hash).

test/unit-test/test_app_btc_common.c
x1ddos: hehe. nice. a quick comment for future self and others about itertools.permutate would go a long way imho.

src/apps/btc/btc_common.h
x1ddos: yeah, that would also work, I guess. personally, I'd still use 2 separate functions: just feels eaier to read and harder to make a silly mistake setting the variable to incorrect value or be somehow theorically flipped true/false intentionally or otherwise because variables are mutable whereas functions and the names they point to in a read-only executable memory are not.

but whichever you like best.

src/apps/btc/btc_common.c
x1ddos: :+1:

src/apps/btc/btc_common.c
x1ddos: But those `define`s *are* the first 4 bytes according to https://github.com/digitalbitbox/libwally-core/blob/afd0542/src/bip32.c#L246-L255:

```c
    tmp32 = hdkey->version;
    if (!serialize_private) {
        /* Change version if serializing the public part of a private key */
        if (tmp32 == BIP32_VER_MAIN_PRIVATE)
            tmp32 = BIP32_VER_MAIN_PUBLIC;
        else if (tmp32 == BIP32_VER_TEST_PRIVATE)
            tmp32 = BIP32_VER_TEST_PUBLIC;
    }
    tmp32_be = cpu_to_be32(tmp32);
    out = copy_out(out, &tmp32_be, sizeof(tmp32_be));
```

in other words, `bip32_key_serialize` always encodes it in either `xpub` or `tpub` format?

x1ddos (approved):
LGTM but I hope you'll do something about `btc_common_multisig_hash(..., false, ...)` one way or another.

test/unit-test/test_app_btc_common.c
benma: Done

src/apps/btc/btc_common.h
benma: Added the two functions :+1:

benma:
> LGTM but I hope you'll do something about `btc_common_multisig_hash(..., false, ...)` one way or another.

Done. PTAL :)

src/apps/btc/btc_common.c
benma: First of all "xpub" and "tpub" are only in the base58-check encoded string. The serialization we are looking at is the binary encoding of the xpub, where the first four bytes are one of the defines you posted. In our case hardcoded  to `BIP32_VER_MAIN_PUBLIC`.

But I discard them right after by hasing without the first four bytes.

x1ddos:
Very nice! :heavy_check_mark: :rocket: :pray:

src/apps/btc/btc_common.c
x1ddos: :+1:

PR #656, opened 2020-11-12
apps/btc: use sorted xpubs when registering multisig accounts

benma:
Any order of xpubs identifies the same xpub account, and some
downstream projects pass them in in arbitrary order.

When checking for a registered account, we check with both the ordered
and unordered xpubs for backwards compatibility.

### DIFF
--- a/src/apps/btc/btc_common.h
+++ b/src/apps/btc/btc_common.h
@@ -271,4 +271,25 @@ USE_RESULT bool btc_common_multisig_hash_sorted(
     size_t keypath_len,
     uint8_t* hash_out);
 
+/**
+ * Get the name of a registered multisig account. If `name` is NULL, this serves as a check whether
+ * the account was registered.
+ *
+ * The individual params are not validated, they must be pre-validated!
+ *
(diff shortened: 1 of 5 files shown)


src/apps/btc/btc_common.h
```diff
@@ -271,4 +271,26 @@ USE_RESULT bool btc_common_multisig_hash_sorted(
...
+ * Get the name of a registered multisig account. If `name` is NULL, this serves as a check whether
+ * the account was registered.
+ *
+ * The individual params are not validated, they must be pre-validated!
+ *
+ * @param[in] coin The coin this multisig is used with.
+ * @param[in] multisig The multisig config details.
+ * @param[in] keypath Account-level keypath.
+ * @param[in] keypath_len number of elements in keypath.
+ * @param[out] hash_out resulting hash; must be `SHA256_LEN` bytes.
```
x1ddos, line 284: Seems like a copy&paste extra arg typo.

x1ddos (approved):
LGTM

PR #665, opened 2020-11-23
releases: add benma assertions for v9.2.1

benma:


x1ddos (approved):
LGTM! :)
I remember CI verifies the signatures, so I won't.

PR #668, opened 2020-11-25
py/bitbox02: exception for empty multisig acct names on old firmware 

benma:
Empty account name means the user will be prompted on the device. This
feature is only available since 9.3.0. This patch informs the user to
upgrade if they use this feature, e.g. in HWI.


x1ddos (approved):
LGTM

PR #670, opened 2020-11-26
rust: update sha2 dep to v0.9.2

benma:


benma:
Also updated noise-rust-protocol, which bumped its own sha2 dep, so we only have use version of sha2.

benma:
Still have some more "duplicate" deps, looking into them.

benma:
Done, upgraded all the relevant deps so that we do not have any dep twice with different versions.

benma:
bump

x1ddos (approved):
LGTM. Nice! Sorry for stalling.

PR #672, opened 2020-11-28
qtouch: fix device lockup due to a race condition

benma:
There were occasional non-deterministic device freezes where
processing would stop completely.

Using gdb in semihosting, we saw that it gets stuck in
`qtm_measure_node()`, called from `qtm_samd51_ptc_handler()`.

Making the interrupt call a critical section, making the section
atomic, seems to fix the issue.

It is not 100% clear what was going on, but one theory is that the
interrupt is called while the interrupt is running, stepping on its
own feet. The more USB requests are processed, the more likely a
freeze was, which could mean that the usb requests make the execution
time of the qtouch interrupt longer, increasing the chance of a race
condition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 9.3.1
+- Fix a bug where the device could freeze and become unresponsive.
+
 ## 9.3.0 [tagged 2020-11-23]
 - Enter multisig account name on the device if the name in BTCRegisterScriptConfigRequest is empty.
 - Allow new keypaths: m/48'/coin'/account' for multisig, to enable compatibility with the Nunchuk wallet.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,9 +89,9 @@ endif()
 #
 # Versions MUST contain three parts and start with lowercase 'v'.
 # Example 'v1.0.0'. They MUST not contain a pre-release label such as '-beta'.
-set(FIRMWARE_VERSION "v9.3.0")
-set(FIRMWARE_BTC_ONLY_VERSION "v9.3.0")
-set(FIRMWARE_BITBOXBASE_VERSION "v9.3.0")
+set(FIRMWARE_VERSION "v9.3.1")
+set(FIRMWARE_BTC_ONLY_VERSION "v9.3.1")
+set(FIRMWARE_BITBOXBASE_VERSION "v9.3.1")
 set(BOOTLOADER_VERSION "v1.0.3")
 
 find_package(PythonInterp 3.6 REQUIRED)
--- a/src/qtouch/qtouch.c
+++ b/src/qtouch/qtouch.c
@@ -703,8 +703,10 @@ Notes    :  none
 ============================================================================*/
 void ADC0_1_Handler(void)
 {
+    CRITICAL_SECTION_ENTER()
     ADC0->INTFLAG.reg |= 1U;
     qtm_samd51_ptc_handler();
+    CRITICAL_SECTION_LEAVE()
 }
 
 #endif /* TOUCH_C */


x1ddos:
So, I did some digging and concur this is the best ad-hoc solution but we might
want to reconsider how our main loop is designed longer term.

---

1. Setup. All debugging session is in semihosting environment. The freezes are
observed much faster and more reliable when hammering the device with USB packets.
To that end, I used `send_message.py` with this patch:

```patch
diff --git py/bitbox02/bitbox02/communication/bitbox_api_protocol.py py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
index 8afb3ba9..cdabe741 100644
--- py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
+++ py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
@@ -501,7 +501,7 @@ class BitBoxProtocolV7(BitBoxProtocolV4):
             assert (
                 len(payload) == 0
             ), "Unexpected payload of length {} with RSP_NOT_READY response.".format(len(payload))
-            time.sleep(0.2)
+            time.sleep(0.0001)
             to_send = (
                 HwwRequestCode.REQ_CANCEL if self.cancel_requested else HwwRequestCode.REQ_RETRY
             )
```

Once device is powered up, run send_message.py. It'll force the device into the
unlocking process and, most importantly, starts polling for a response every 0.1ms.

2. When the device freezes, a stacktrace in gdb typically shows something like this:

```
(gdb) bt
#0  0x00046b6e in qtm_measure_node ()
#1  0x00047374 in qtm_same54_ptc_handler ()
#2  <signal handler called>
#3  hri_sercomspi_read_INTFLAG_reg (hw=<optimized out>)
    at external/asf4-drivers/hri/hri_sercom_d51.h:414
#4  SPI_0_write_block (block=block@entry=0x2000b721 <_frame_buffer+144>, size=13 '\r')
    at external/asf4-drivers/hpl/spi/spi_lite.c:267
#5  0x0001f6c8 in _write (interface=interface@entry=INTERFACE_DATA, buf=buf@entry=0x2000b721 <_frame_buffer+144> "",
    buf_len=buf_len@entry=16) at src/ui/oled/oled.c:136
#6  0x0001f78a in _write_data (buf_len=16, buf=0x2000b721 <_frame_buffer+144> "")
    at src/ui/oled/oled.c:233
#7  oled_send_buffer () at src/ui/oled/oled.c:233
#8  0x0001a18a in UG_SendBuffer () at src/ui/ugui/ugui.c:861
#9  0x0001a4e0 in ui_screen_render_component (component=component@entry=0x200219a0)
    at src/ui/screen_process.c:32
#10 0x0001a56e in _screen_draw (component=0x200219a0) at src/ui/screen_process.c:65
#11 screen_process () at src/ui/screen_process.c:92
#12 0x00010306 in firmware_main_loop () at src/firmware_main_loop.c:52
#13 0x0002213e in main () at src/firmware.c:50
```

Notice that oled's SPI write is where it's stuck outside of the qtouch's
ADC interrupt. 

2. Try and disable qtouch acquisition routine.

First, apply a patch like this:

```patch
diff --git src/qtouch/qtouch.c src/qtouch/qtouch.c
index 43c97758..3d1dd3ce 100644
--- src/qtouch/qtouch.c
+++ src/qtouch/qtouch.c
@@ -703,7 +703,9 @@ Notes    :  none
 ============================================================================*/
 void ADC0_1_Handler(void)
 {
     ADC0->INTFLAG.reg |= 1U;
+    static volatile int x = 0;
+    if (x) { return; }
     qtm_samd51_ptc_handler();
 }
```

Reload the elf firmware in gdb, `monitor reset` and `continue` to restart the
device. Once you get to the "See the BitBoxApp" welcome screen, break with ctrl-c
and set a breakpoint at `if (x) { return; }`. Continue, which triggers right away:
`set variable x = 0`, disable the breakpoint and continue.

This effectively disables the qtouch acquisition. Run the patched send_message.py
and observe no matter how long you wait, it always responds to the USB packets as
usual. No freezing. Can use wireshark or anything else to confirm but I ended up
with a dumb method of ctrl-c the send_message.py. If the device isn't responding,
it won't exit while blocked on `self._device.read(USB_REPORT_SIZE, timeout_ms)`.

3. At first, I thought USB processing is what takes longer. So, wanted to measure
`firmware_main_loop`. Here's a poor man's measurments due to my ignorance, quirks
with JLinkGDBServer and lack of time. Anyway, apply this patch:

```patch
diff --git src/firmware_main_loop.c src/firmware_main_loop.c
index bd5e6ca1..922dec22 100644
--- src/firmware_main_loop.c
+++ src/firmware_main_loop.c
@@ -25,10 +25,34 @@
 #include "workflow/workflow.h"
 #include <rust/rust.h>
 
+#include <hal_timer.h>
+extern struct timer_descriptor TIMER_0;
+
+uint32_t total_ticks_count;
+
+static void main_loop_measure_cb(const struct timer_task* const t)
+{
+    (void)t; // make cc happy
+    total_ticks_count++;
+}
+
 void firmware_main_loop(void)
 {
+    total_ticks_count = 0;
+    uint32_t loop_count = 0;
+
+    static struct timer_task main_loop_measure_timer;
+    main_loop_measure_timer.interval = 1; // roughly 1ms
+    main_loop_measure_timer.cb = main_loop_measure_cb;
+    main_loop_measure_timer.mode = TIMER_TASK_REPEAT;
+    timer_add_task(&TIMER_0, &main_loop_measure_timer);
+    timer_start(&TIMER_0);
+
+    static volatile int scr = 1;
     while (1) {
-        screen_process();
+        if (scr) {
+          screen_process();
+        }
         /* And finally, run the high-level event processing. */
 
         rust_workflow_spin();
@@ -50,5 +74,15 @@ void firmware_main_loop(void)
             u2f_process();
 #endif
         }
+
+        loop_count++;
+        if (loop_count % 5000 == 0) {
+            //static volatile float threshold = 2.99; // 4 when no USB load
+            volatile float avg = (double)loop_count / (double)total_ticks_count;
+            traceln("avg: %d", (int)(avg*100));
+            //if (avg < threshold) {
+            //    __BKPT(1);
+            //}
+        }
     }
 }
```

Reload, `montor reset` and `continue`. Let it run and observe the traceln output.
For me this showed 4.4 loops per 1ms aka 0.22ms per loop iteration.

Due to an unknown quirk to me, traceln makes everything freeze on high USB load.
That's what the commented lines are for. When hammering the device with USB packets,
remove traceln and use the `threshold`, adjusting it to a lower value every time
it hits the BKPT.

Under heavy USB load, I came down to about 2 loops per 1ms aka 0.5ms per loop.

4. Let's try and disable `screen_process`.

This is what `if (scr) screen_process();` was for. Very similar to disabling qtouch
acquisition: get to the welcome screen on the device, break, set `scr` to 0 and
continue.

What surprised me is the average loop timing. For me this showed 60+ iterations
per 1ms aka 0.017ms per loop iteration. This means ~92% of the main loop is taken
by the screen all the time regardless of animations.

And guess what, no freezes... but also no screen updates. :)

Just to put this number here, we send 16*64=1k of data in `oled_send_buffer`
via SPI bus every 0.2ms unconditionally.

I'm not sure USB load is what actually causes the loop to slow down from 0.22ms to
0.5ms as measured in the previous step. Rather, I believe it might be due to an
internal bus arbitration. Both USB and OLED data are sent via the chip's SERCOM SPI
and it's just easier to hit qtouch's race condition when the bus is busy.

Even if we restructure the screen process, for example enabling it only
when a user interaction or animation occurs, this qtouch issue can still happen 
during those busy intervals although very much reduced otherwise.

What changes with critical section safeguard is interrupts are disabled during
the qtouch's ADC run which avoids the race condition, especially easy to hit
at busy times.

5. On top of that, qtouch docs actually recommended adding a critical section
at some point, although for older versions of the lib:

image

I couldn't find a more recent version of the doc. Their docs management is far
from ideal, dead and missing links, etc. But the overall idea is pretty clear.
Here's a working link as of now with some PDFs at "random" versions and dates:
https://www.microchip.com/DevelopmentTools/ProductDetails/Atmel%20QTouch%20Library

Bottom line is I'd really like to replace qtouch lib with something else longer term.

x1ddos (approved):
LGTM

PR #673, opened 2020-11-28
qtouch: upgrade to v7.4.1

x1ddos:
After coping the files over, the upgrade was heavily modified to adapt
to our custom changes. The original files can be found here:
https://gist.github.com/x1ddos/acca84d3e1a03df738242c232581bccb

Unlike https://github.com/digitalbitbox/bitbox02-firmware/pull/672, the race condition still exists here.


src/qtouch/qtouch.c
```diff
@@ -524,7 +536,13 @@ void qtouch_timer_config(void)
     Timer_task.cb = qtouch_timer_task_cb;
     Timer_task.mode = TIMER_TASK_REPEAT;
 
+    static uint8_t timer_task_added = 0;
+    if (timer_task_added) {
+        timer_remove_task(&TIMER_0, &Timer_task);
```
x1ddos, line 541: It isn't really needed because `timer_add_task` won't add if it alreayd exists in the list. Not sure why they added it. Maybe compatibility with other chips.

src/qtouch/qtouch.c
```diff
@@ -80,6 +87,12 @@ qtm_control_t qtm_control;
 qtm_control_t* p_qtm_control;
 qtm_state_t qstate;
 
+/* Flag to indicate time for touch measurement */
+volatile uint8_t time_to_measure_touch_flag = 0;
+
+/* postporcess request flag */
```
x1ddos, line 93: Not sure if I should leave their typo as is :)

benma:
I assume we will leave this for 9.4.0? 

> . The original files can be found here:
> https://gist.github.com/x1ddos/acca84d3e1a03df738242c232581bccb

Please also describe where those files came from.

x1ddos:
> I assume we will leave this for 9.4.0?
> 
> > . The original files can be found here:
> > https://gist.github.com/x1ddos/acca84d3e1a03df738242c232581bccb
> 
> Please also describe where those files came from.

Yeah, for sure not in 9.3.1. 
I'll add some more info.

src/qtouch/qtouch.c
```diff
@@ -533,21 +551,49 @@ uint16_t qtouch_get_sensor_node_signal(uint16_t sensor_node)
...
 {
     return (qtlib_key_set1.qtm_touch_key_data[sensor_node].sensor_state);
 }
 
+void qtouch_update_sensor_state(uint16_t sensor_node, uint8_t new_state)
+{
+    qtlib_key_set1.qtm_touch_key_data[sensor_node].sensor_state = new_state;
+}
+
+void qtouch_calibrate_node(uint16_t sensor_node)
```
benma, line 589: is our own `qtouch_force_calibrate()` still doing what it is supposed to do?

benma (commented):
Initial test works well!

src/qtouch/qtouch.c
x1ddos: I tried interacting during the regular functions like unlocking - trinary input. Worked as before, seemingly no change at all.

x1ddos: but I suspect lots of things have changed. it is very possible not only could drop our force_calibrate func with this qtouch upgrade, the custom filtering and scroller funcs might go away too. but it would require some testing and for sure tweaking the sensor values.

benma: For context: if you plug in the device while holding the touch sensors, the initial calibration will be off and you touch is broken (the base level is off, so it wouldn't register any more touches after).

qtouch_force_calibrate() resets the base level and is called during the orientation screen, which fixed that issue.

x1ddos: Yeah, I imagined something like that from [the comment](https://github.com/digitalbitbox/bitbox02-firmware/blob/27a37d3/src/qtouch/qtouch.c#L312-L318). I was actually more concerned with the custom [qtouch_get_sensor_node_signal_filtered](https://github.com/digitalbitbox/bitbox02-firmware/blob/27a37d3/src/qtouch/qtouch.c#L312-L318) and [qtouch_process_scroller_positions](https://github.com/digitalbitbox/bitbox02-firmware/blob/27a37d3/src/qtouch/qtouch.c#L312-L318) - those are much more involved.

They do say something about "scroller bug fixes" in https://microchipdeveloper.com/touch:release-notes but it is obviously unknown what they did exactly. But it could eliminate the need for the custom scroller although I wouldn't do it here. Some time later on, unless we manage to get rid of qtouch :)

benma: i tested that calibration still works

benma (approved):
tACK

x1ddos:
Rebased onto latest master

PR #676, opened 2020-11-30
release: add v9.3.1

benma:


x1ddos (approved):
:heavy_check_mark: all hashes match mine

releases/firmware-btc-v9.3.1/assertion.txt
```diff
@@ -0,0 +1,8 @@
+By signing this file, the signer confirms that the firmware binary built from:
+
+git tag firmware-btc-only/v9.3.0
```
benma, line 3: typo, 9.3.0 should be 9.3.1

x1ddos: :see_no_evil: I looked only at hash values.

PR #678, opened 2020-12-13
external: update libwally-core

benma:
To 46da3615b8e0837964f443d6b113ad500ea89f97.

This is a bit ahead of the official latest 0.8.0 release, to include
this commit, which adds secp256k1-zpk as a submodule instead of as a
source-copy inside libwally-core:

https://github.com/ElementsProject/libwally-core/commit/1da463088e64189d04c21073eb8861be1b717ad1

This will make it a lot easier for us use a custom fork of secp256k1,
which we might want to add anti nonce covert channel protection.

The new release also gets rid of all custom commits to libwally-core
and secp256k1, which are now all merged and released upstream.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       - name: Clone the repo
         uses: actions/checkout@v2
         with:
-          submodules: true
+          submodules: recursive
       - name: Pull CI container image
         run: ./.ci/run-container-ci pull
       - name: Run CI in container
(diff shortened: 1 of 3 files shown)


x1ddos (approved):
LGTM

While browsing through, I noticed we don't check wally_sha256's result in a couple places: u2f.c and backup_common.c. Also wally_hmac_sha256 in backup.c. Maybe there are other functions. Either way, unrelated to this commit.

PR #681, opened 2020-12-21
Port ETHSignMessage to Rust

benma:
Builds on https://github.com/digitalbitbox/bitbox02-firmware/pull/680


src/rust/bitbox02-rust/src/hww/api/ethereum/signmsg.rs
```diff
@@ -0,0 +1,217 @@
...
+        output_type: pb::eth_pub_request::OutputType::Address as _,
+        keypath: request.keypath.clone(),
+        coin: request.coin,
+        display: true,
+        contract_address: Vec::new(),
+    };
+
+    // Verify address. We don't need the actual result, but we have to propagate validation or user
+    // abort errors.
+    let _ = super::pubrequest::process(&pub_request).await?;
```
x1ddos, line 45: I don't think you need `let _ =`. Cleaner imho. The comment obviously stays. It's good.

Also, I personally dislike hooking into other API from other entries points. Ideally in my world there would be a `crate::eth::verify_address(...)` which both `pubrequest::process` and this fn calls.

src/rust/bitbox02-rust/src/hww/api/ethereum/signmsg.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    verify_message::verify(&request.msg).await?;
+
+    let mut msg: Vec<u8> = Vec::new();
+    msg.extend(b"\x19Ethereum Signed Message:\n");
```
x1ddos, line 50: Could you add a comment where the msg format is coming from, like the magic `\x19`.

src/rust/bitbox02-rust/src/hww/api/ethereum/signmsg.rs
```diff
@@ -0,0 +1,217 @@
...
+    let sighash: [u8; 32] = sha3::Keccak256::digest(&msg).as_slice().try_into().unwrap();
+    let sign_result = bitbox02::keystore::secp256k1_sign(&request.keypath, &sighash)?;
+
+    let mut signature: Vec<u8> = sign_result.signature.to_vec();
+    signature.push(sign_result.recid);
+
+    Ok(Response::Sign(pb::EthSignResponse { signature }))
+}
+
+#[cfg(test)]
```
x1ddos, line 63: The C version also had some more tests like a too long msg. Maybe it's not so relevant anymore but I think I would appreciate some tests verifying a correct error is returned if say `keystore::secp256k1_sign` fails.

x1ddos (approved):
LGTM

src/rust/bitbox02-rust/src/hww/api/ethereum/signmsg.rs
benma: done

src/rust/bitbox02-rust/src/hww/api/ethereum/signmsg.rs
benma: done, added tests for both.

src/rust/bitbox02-rust/src/hww/api/ethereum/signmsg.rs
benma: did the first. postponing the second, it is not exactly clear to me how to do it.

PR #682, opened 2020-12-21
apps/btc: use keystore_secp256k1_pubkey_hash160 instead of get_xpub

benma:
More specific function for easier code. Goal: no wally ext_key in the
public api of keystore.

### DIFF
--- a/src/apps/btc/btc.c
+++ b/src/apps/btc/btc.c
@@ -23,6 +23,7 @@
 #include <keystore.h>
 #include <memory/memory.h>
 #include <rust/rust.h>
+#include <util.h>
 #include <workflow/confirm.h>
 #include <workflow/status.h>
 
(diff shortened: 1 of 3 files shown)


src/apps/btc/btc.c
```diff
@@ -97,15 +98,15 @@ bool app_btc_address_simple(
             script_type, keypath, keypath_len, params->bip44_coin)) {
         return false;
     }
-    struct ext_key derived_xpub __attribute__((__cleanup__(keystore_zero_xkey))) = {0};
```
x1ddos, line 100: Been wondering... where's this ext_key coming from. Hm... no wally lib include. Ok, let's look at btc.h - I remember hearing something about 1:1 correspondence between .h and .c files - heh, not in there either. I wonder where it was coming from now :thinking: keystore.h? no... btc_params.h? no... aha, found an include in btc_common.h. That was so obivous, I hope I'll remember next time :)

x1ddos (approved):
LGTM

src/apps/btc/btc.c
x1ddos: Not to mention of course that btc.c gets included wally lib headers via btc_common.h even though it doesn't seem to need it at all now.

benma: Since it is in the keystore interface, "obviously" the include you need is keystore.h to use it :joy: In keystore.h you have `#include <wally_bip32.h>` where you can find the definition.

PR #686, opened 2020-12-26
keystore: fix memory leak in use of keystore_bip39_get_word

benma:
The function malloc's the word, but the Rust FFI layer assumed it was
a static string. It was functionally correct, as the address was
static, but there was a memory leak due to a missing free().

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 extern crate alloc;
+use alloc::string::String;
 use alloc::vec::Vec;
 
 use super::Error;
@@ -25,7 +26,7 @@ use bitbox02::keystore;
 
 const NUM_RANDOM_WORDS: u8 = 5;
 
-fn create_random_unique_words(word: &str, length: u8) -> (u8, Vec<&str>) {
+fn create_random_unique_words(word: &str, length: u8) -> (u8, Vec<zeroize::Zeroizing<String>>) {
     fn rand16() -> u16 {
         let mut rand = [0u8; 32];
         bitbox02::random::mcu_32_bytes(&mut rand);
@@ -38,7 +39,7 @@ fn create_random_unique_words(word: &str, length: u8) -> (u8, Vec<&str>) {
         .map(|i| {
             // The correct word at the right index.
             if i == index_word {
-                return word;
+                return zeroize::Zeroizing::new(word.into());
             }
 
             // A random word everywhere else.
@@ -49,7 +50,7 @@ fn create_random_unique_words(word: &str, length: u8) -> (u8, Vec<&str>) {
                     continue;
                 };
                 let random_word = keystore::get_bip39_word(idx).unwrap();
-                if random_word == word {
+                if random_word.as_str() == word {
                     continue;
                 }
                 picked_indices.push(idx);
@@ -74,7 +75,8 @@ pub async fn process() -> Result<Response, Error> {
     // Part 2) Confirm words
     for (word_idx, word) in words.iter().enumerate() {
         let title = format!("{:02}", word_idx + 1);
-        let (correct_idx, mut choices) = create_random_unique_words(word, NUM_RANDOM_WORDS);
+        let (correct_idx, choices) = create_random_unique_words(word, NUM_RANDOM_WORDS);
+        let mut choices: Vec<&str> = choices.iter().map(|c| c.as_ref()).collect();
         choices.push("Back to\nrecovery words");
         let back_idx = (choices.len() - 1) as u8;
         loop {
--- a/src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
@@ -26,6 +26,10 @@ use bitbox02::keystore;
 
 const NUM_RANDOM_WORDS: u8 = 5;
 
+/// Return 5 words from the BIP39 wordlist, 4 of which are random, and
+/// one of them is provided `word`. Returns the position of `word` in
+/// the list of words, and the lis of words.  This is used to test if
+/// the user wrote down the seed words properly.
 fn create_random_unique_words(word: &str, length: u8) -> (u8, Vec<zeroize::Zeroizing<String>>) {
     fn rand16() -> u16 {
         let mut rand = [0u8; 32];
@@ -62,6 +66,10 @@ fn create_random_unique_words(word: &str, length: u8) -> (u8, Vec<zeroize::Zeroi
     (index_word, result)
 }
 
+/// Handle the ShowMnemonic API call. This shows the seed encoded as
+/// 12/18/24 BIP39 English words. Afterwards, for each word, the user
+/// is asked to pick the right word among 5 words, to check if they
+/// wrote it down correctly.
 pub async fn process() -> Result<Response, Error> {
     unlock::unlock_keystore("Unlock device", unlock::CanCancel::Yes).await?;
 
--- a/src/rust/bitbox02/src/keystore.rs
+++ b/src/rust/bitbox02/src/keystore.rs
@@ -85,16 +85,20 @@ pub fn get_bip39_mnemonic() -> Result<zeroize::Zeroizing<String>, ()> {
 }
 
 /// `idx` must be smaller than BIP39_WORDLIST_LEN.
-pub fn get_bip39_word(idx: u16) -> Result<&'static str, ()> {
-    let mut word: *mut u8 = core::ptr::null_mut();
-    match unsafe { bitbox02_sys::keystore_get_bip39_word(idx, &mut word) } {
+pub fn get_bip39_word(idx: u16) -> Result<zeroize::Zeroizing<String>, ()> {
+    let mut word_ptr: *mut u8 = core::ptr::null_mut();
+    match unsafe { bitbox02_sys::keystore_get_bip39_word(idx, &mut word_ptr) } {
         false => Err(()),
         true => {
-            let s = unsafe {
-                let len = crate::util::strlen_ptr(word);
-                core::slice::from_raw_parts(word, len as _)
+            let word = unsafe {
+                let len = crate::util::strlen_ptr(word_ptr);
+                let slice = core::slice::from_raw_parts(word_ptr, len as _);
+                zeroize::Zeroizing::new(core::str::from_utf8(&slice[..]).unwrap().into())
             };
-            Ok(core::str::from_utf8(&s[..]).unwrap())
+            unsafe {
+                bitbox02_sys::wally_free_string(word_ptr as _);
+            }
+            Ok(word)
         }
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 9.4.0 [version may change, pending release]
 - ETHPubRequest api call now fails if a an invalid contract address is provided also if `display` is
   false.
+- Fix a memory leak (freeing a malloc'd string - no a functional or security issue)
 
 ## 9.3.1 [tagged 2020-12-01]
 - Fix a bug where the device could freeze and become unresponsive.
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -327,6 +327,7 @@ add_custom_target(rust-bindgen
     --use-core
     --ctypes-prefix util::c_types
     --rustified-enum backup_error_t
+    --whitelist-function wally_free_string
     --whitelist-function backup_check
     --whitelist-function backup_create
     --whitelist-function memory_is_initialized
--- a/src/rust/Cargo.lock
+++ b/src/rust/Cargo.lock
@@ -67,6 +67,7 @@ dependencies = [
  "sha2",
  "sha3",
  "util",
+ "zeroize",
 ]
 
 [[package]]
--- a/src/rust/bitbox02-rust/Cargo.toml
+++ b/src/rust/bitbox02-rust/Cargo.toml
@@ -34,6 +34,7 @@ bitbox02-noise = {path = "../bitbox02-noise"}
 hex = { version = "0.4", default-features = false }
 sha2 = { version = "0.9.2", default-features = false }
 sha3 = { version = "0.9.1", default-features = false, optional = true }
+zeroize = "1.1.0"
 
 # For stack allocated strings
 [dependencies.arrayvec]
--- a/src/rust/bitbox02-sys/wrapper.h
+++ b/src/rust/bitbox02-sys/wrapper.h
@@ -45,6 +45,7 @@
 #include <ui/ugui/ugui.h>
 #include <util.h>
 #include <wally_bip39.h>
+#include <wally_core.h>
 #include <wally_crypto.h>
 #include <workflow/confirm.h>
 
--- a/src/rust/bitbox02/src/keystore_stub.rs
+++ b/src/rust/bitbox02/src/keystore_stub.rs
@@ -51,7 +51,7 @@ pub fn get_bip39_mnemonic() -> Result<zeroize::Zeroizing<String>, ()> {
     panic!("not implemented")
 }
 
-pub fn get_bip39_word(_idx: u16) -> Result<&'static str, ()> {
+pub fn get_bip39_word(_idx: u16) -> Result<zeroize::Zeroizing<String>, ()> {
     panic!("not implemented")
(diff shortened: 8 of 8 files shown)


benma:
Will attempt a version that doesn't malloc at all.

benma:
Nevermind, while libwally stores the strings statically, the API does not expose it.

src/rust/bitbox02/src/keystore.rs
```diff
@@ -85,16 +85,20 @@ pub fn get_bip39_mnemonic() -> Result<zeroize::Zeroizing<String>, ()> {
...
-                let len = crate::util::strlen_ptr(word);
-                core::slice::from_raw_parts(word, len as _)
+            let word = unsafe {
+                let len = crate::util::strlen_ptr(word_ptr);
+                let slice = core::slice::from_raw_parts(word_ptr, len as _);
+                core::str::from_utf8(&slice[..]).unwrap().into()
             };
-            Ok(core::str::from_utf8(&s[..]).unwrap())
+            unsafe {
+                bitbox02_sys::free(word_ptr as _);
```
x1ddos, line 99: first time I see `free` call in rust. is this really the case, there are no other instances where bitbox_sys would need to free a C-malloc'ed memory?

src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
```diff
@@ -25,7 +26,7 @@ use bitbox02::keystore;
 
 const NUM_RANDOM_WORDS: u8 = 5;
 
-fn create_random_unique_words(word: &str, length: u8) -> (u8, Vec<&str>) {
+fn create_random_unique_words(word: &str, length: u8) -> (u8, Vec<String>) {
```
x1ddos, line 29: Spent like 5min now re-reading the body of this function to remember what the `word` is and what it all does. Would've appreaciated some doc comments a lot.

src/rust/bitbox02/src/keystore.rs
```diff
@@ -85,16 +85,20 @@ pub fn get_bip39_mnemonic() -> Result<zeroize::Zeroizing<String>, ()> {
...
+    match unsafe { bitbox02_sys::keystore_get_bip39_word(idx, &mut word_ptr) } {
         false => Err(()),
         true => {
-            let s = unsafe {
-                let len = crate::util::strlen_ptr(word);
-                core::slice::from_raw_parts(word, len as _)
+            let word = unsafe {
+                let len = crate::util::strlen_ptr(word_ptr);
+                let slice = core::slice::from_raw_parts(word_ptr, len as _);
+                core::str::from_utf8(&slice[..]).unwrap().into()
```
x1ddos, line 96: Rust uses C's malloc/free via [BB02Allocator](https://github.com/digitalbitbox/bitbox02-firmware/blob/29ae4d7/src/rust/bitbox02-rust-c/src/alloc.rs), right. So, if you instead used `crate::util::str_from_null_terminated` to get the `&str`, wouldn't it be `free`'ed automatically when that `&str` is dropped?

src/rust/bitbox02/src/keystore.rs
benma: Good question. I don't think so. The components are malloc'd, but freed in C after they are popped from the screen stack.

src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
benma: :+1: Added docs.

src/rust/bitbox02/src/keystore.rs
benma: The BB02Allocator only exists for Rust objects on the heap, such as String or Vec (anything from the `alloc` crate). 

The `str_from_null_terminated()` does not do an malloc, the result points to the same memory as the input argument.

Even if it did alloc/free a String, it would be a different object than the thing created by `keystore_get_bip39_word`, which needs to be freed in any case.

I kind of want to roll our own Rust BIP39 library now, libwally is yet again a bit too restrictive in its API (the follow up PR also has an issue where I work around the API :/).

x1ddos: But look, in pseudo-rust:

```rust
let mut word_ptr: *mut u8 = core::ptr::null_mut();
bitbox02_sys::keystore_get_bip39_word(&mut word_ptr);
// word_ptr is pointing somewhere on the heap, alloc'ed in C
// box should free word_ptr when it goes out of scope
Box::from_raw(word_ptr)
```

benma: Box::from_raw() assumes that the object at word_ptr is a Rust `Box` object. It might be possible that your code would work, but I would not rely on it. I'd rather be explicit and not mix Rust's and C's heap management like this.

benma: By the way, it is easy enough to wrap the word in a Rust object that calls C's free() on drop. Like here:

https://github.com/digitalbitbox/bitbox02-firmware/pull/687/commits/73cc23ba5a97110b1968ce6032f6704c2c5f18dc#diff-5508fe9d536dd16cbe1c9c07143abac7183e4942bf65057ebe1f3b57a593320fR119

x1ddos: > mix Rust's and C's heap management like this.

but it's the same in our case:

```rust
struct BB02Allocator;

extern "C" {
    pub fn malloc(size: util::c_types::size_t) -> *mut util::c_types::c_void;
    pub fn free(p: *mut util::c_types::c_void);
}

unsafe impl core::alloc::GlobalAlloc for BB02Allocator {
    unsafe fn alloc(&self, layout: core::alloc::Layout) -> *mut u8 {
        malloc(layout.size()) as _
    }
    unsafe fn dealloc(&self, ptr: *mut u8, _layout: core::alloc::Layout) {
        free(ptr as _)
    }
}

#[global_allocator]
static BB02_ALLOCATOR: BB02Allocator = BB02Allocator;
```

that's our Rust allocator, using C's malloc/free.

Re: Box::from_raw:

> For non-zero-sized values, a Box will use the Global allocator for its allocation. It is valid to convert both ways between a Box and a raw pointer allocated with the Global allocator, given that the Layout used with the allocator is correct for the type

`#[global_allocator]` is what tells Rust which Global allocator to use. The proof that the layout matches is in the fact that we pass callback functions created in Rust, called in C. 

> So long as T: Sized, a Box<T> is guaranteed to be represented as a single pointer and is also ABI-compatible with C pointers (i.e. the C type T*). This means that if you have extern "C" Rust functions that will be called from C, you can define those Rust functions using Box<T> types, and use T* as corresponding type on the C side.

benma: But again, why is it better to do an implicit free with `Box::from_raw()` than an explicit `free()`? One is easy to understand, the other needs all the context you posted. I see that it might work, but again I would not rely on it, at least not without manually testing it.

x1ddos: > But again, why is it better to do an implicit free with Box::from_raw() than an explicit free()? 

hm... that feels like a question similar to "why is C better than Rust"? :smile: more speficially, a hypothetical "why doesn't Rust have a `free` function!???" 

or did I misunderstood?

x1ddos: I mean I'm actually ok calling `free` from Rust explicitly but it just feels like a step backwards.

benma: To be clear:

are you suggesting to replace line 99 (`bitbox02_sys::free(word_ptr as _);`) with `Box::from_raw(word_ptr)`? If so, and if you are right that it does the same thing, then is the same result, but is just implicit instead of explicit, so I don't see the benefit.

benma: Another example of why to not mix these: if we ever adjusted the Rust allocator to do something special (e.g. a smarter heap layout), this would silently break.

x1ddos: Well, better memory management is one big benefit of why port code to Rust. In fact, this very pull request is a testimony to that: forgotten `free` call because it's manual. Were it `drop`'ed automatically, there wouldn't even be this pull request, right.

x1ddos: > Another example of why to not mix these: if we ever adjusted the Rust allocator to do something special (e.g. a smarter heap layout), this would silently break (if you used Box::from_raw() instead of free()).

We don't really want to do that as long as we have C<->Rust interop. If we only had Rust, then we wouldn't need all these compatible allocations and nothing would break because we wouldn't simple need Box::from_raw and other things.

x1ddos: Anyway, we can improve this later on.

x1ddos (approved):
LGTM

src/rust/bitbox02/src/keystore.rs
benma: > We don't really want to do that as long as we have C<->Rust interop

Just saying, we could, even if we won't. We could for example put 50% of the heap to Rust and 50% of the heap to C (silly example but it would be a possible change to the BB02Allocator).

I am still convinced that this is wrong. The BB02Allocator is an implementation detail, you should not rely on it's internals here. Memory allocated in C should be freed in C. Memory allocated in Rust should be freed in Rust. Mixing these is no good.

NickeZ: shouldn't it be freed with `wally_free_string`? https://wally.readthedocs.io/en/latest/bip39/#c.bip39_get_word

benma: > Well, better memory management is one big benefit of why port code to Rust. In fact, this very pull request is a testimony to that: forgotten free call because it's manual.

But you still have to call `Box::from_raw()`, which is the same manual memory management, but just the implicit version of `free()`.

NickeZ: Wouldn't it be nice to have a WallyString struct that frees with the `wally_free` and derefs to &str? should be approximately as nice to use as a String.

benma: > shouldn't it be freed with `wally_free_string`? https://wally.readthedocs.io/en/latest/bip39/#c.bip39_get_word

@NickeZ yes, as it zeroes the string :+1: I had it in the back of my mind. good catch, will fix

x1ddos: > The BB02Allocator is an implementation detail, you should not rely on it's internals here. 

But we are in fact relying on this very heavily right now. Otherwise, we wouldn't be certain that Rust->C and C->Rust interop worked in the firmware.

In other words, we *have* to rely on this. It's in the current design.

x1ddos: > shouldn't it be freed with `wally_free_string`? https://wally.readthedocs.io/en/latest/bip39/#c.bip39_get_word

between all these allocations, I lost track of the rest. :see_no_evil: thanks @NickeZ!

benma: > In other words, we have to rely on this. It's in the current design.

That is incorrect. You can change the BB02Allocator and fully expect everything to work. We don't rely on its internals anywhere. Of course you have to be mindful that you share the heap with C, so the BB02Allocator has to be correct in that sense, but beyond that you can change everything. For example, pre-malloc() 50% of the heap and use a custom allocator for that. There is no code I am aware of that would break due to correctness (it could break by having less memory available of course, but that is not a correctness issue per se).

NickeZ: > But we are in fact relying on this very heavily right now. Otherwise, we wouldn't be certain that Rust->C and C->Rust interop worked in the firmware.

I don't think you rely on that fact as long as you free things in C that you have malloced in C and free things in rust that you have allocated in rust. This allows you to technically have two heaps. Of course it is all the same memory in the end, so you can still share pointers between C and rust.

NickeZ: > edit2: now i remember why i postponed this: we only use the words in a 'non-secret' fashion, so if it is not zeroed, it does not matter. will fix anyway.

For the same reason that I think rust things should be freed in rust and c-things in C. I think that any library that allocates should have a "free" function and that function should be used. So in this case I think it is semantically important to use `wally_free_string`. Because the fact that it uses the "system" heap could be an implementation detail.

x1ddos: Really. What about all the callback functions allocated on Rust, passed to and calle from C? What about param structs and any other heap-allocated objects passed between Rust and C. How would that work with an Rust allocator incompatible with the C?

x1ddos: > For example, pre-malloc() 50% of the heap and use a custom allocator inside that area.

What do you think will happen if you pass an object or a function allocated in there to C?

NickeZ: Even if the allocators are incompatible you can read things in the "other" heap through pointers. In rust you typically do `Box::into_raw` to convert a box into a pointer without destructing it. And then when you know that the pointer isn't being used in C-land anymore you do `Box::from_raw` so that you can free it with the rust allocator.

x1ddos: > you can read things in the "other" heap through pointers

well, that depends on what the pointers are pointing to, right. If it's compatible with C then sure. But what if I have a custom allocator composed of some list of structs where each item represents a chunk of allocated memory with timestamps, size and other fields that C wouldn't know about?

NickeZ: >> For example, pre-malloc() 50% of the heap and use a custom allocator inside that area.
>
>What do you think will happen if you pass an object or a function allocated in there to C?

You would typically pre-allocate that through the linker. So you would say "this range is the stack, this range is c-heap and this is rust-heap".

One thing I know is a "feature" of the rust allocator is that it always aligns objects in memory based on the alighment requirements of the members of the struct. Not sure if the C allocator does that. (In our case, since we simply call "malloc" to allocate rust structs, it would work, but for the general case I think it doesn't work.)

NickeZ: > well, that depends on what the pointers are pointing to, right. If it's compatible with C then sure. But what if I have a custom allocator composed of some list of structs where each item represents a chunk of allocated memory with timestamps, size and other fields that C wouldn't know about?

Yeah, if your allocator is giving out indices or some other type that isn't a pointer it would be impossible to share between the systems. It is kind of like saying a file descriptor doesn't make sense to send over network. STDIN is 0, STDOUT is 1 and STDERR is 2. But those numbers mean something unique to every application.

x1ddos: Are you saying that the implementation details of an allocator do matter when C/Rust interop is required? :)

benma: @x1ddos here is an example of a bb02allocator change that works. the only place where bip39_get_word() is used is at the end of showing the mnemonic. there, it will hard fault if you use Box::from_raw() instead of free():

image

NickeZ: I'm saying that if you have a custom allocator that gives out indices, it cannot be used as a global allocator in rust nor c because the API for a global allocator requires you to give out pointers (see definition of malloc/free and the GlobalAlloc trait in rust).

x1ddos: Sure, this particular example happens to work but I just don't see it's ever used in practice. and of course, this particular example does break Rust, I agree.

Ok, you win. I'm convinced we should call `free` now, but I 100% disagree with:

> That is incorrect. You can change the BB02Allocator and fully expect everything to work

You can't just make a random change to the allocator and expect C/Rust to continue sharing memory without issues, **unless** you make it compatible between the two.

x1ddos: So, bottom line is this PR is good and I was wrong :+1:

NickeZ: > You can't just make a random change to the allocator and expect C/Rust to continue sharing memory without issues, unless you make it compatible between the two.

a "random change" is still quite limited. You *need* to give out valid pointers. And as long as you have a pointer it can be shared between the languages.

benma:
@x1ddos @NickeZ pushed a commit that uses wally_free_string() instead.

> Wouldn't it be nice to have a WallyString struct that frees with the wally_free and derefs to &str? should be approximately as nice to use as a String.

Kind of, but imho it just adds more code, unit tests etc. for very little benefit. It's nicer to stick to zeroize in Rust-land imho.

x1ddos:
> @x1ddos @NickeZ pushed a commit that uses wally_free_string() instead.

:+1: but I would squash the commits before merging

PR #689, opened 2020-12-30
common_main: set bootloader auto_enter=true on failures at startup

benma:
Go into bootloader if there was an error during startup, so a firmware update can be
applied. Otherwise, if there is an Abort() during startup, there would no way to reboot into the
bootloader and the device would be bricked.

### DIFF
--- a/src/common_main.c
+++ b/src/common_main.c
@@ -67,21 +67,43 @@ static bool _setup_wally(void)
     return wally_set_operations(&_ops) == WALLY_OK;
 }
 
+// Go into bootloader if there was an error during startup, so a firmware update can be
+// applied. Otherwise, if there is an Abort() during startup, there would no way to reboot into the
+// bootloader and the device would be bricked.
+static void _bootloader_autoenter(void)
+{
+    auto_enter_t auto_enter = {
+        .value = sectrue_u8,
+    };
+    upside_down_t upside_down = {
+        .value = screen_is_upside_down(),
+    };
+    if (!memory_bootloader_set_flags(auto_enter, upside_down)) {
+        // If this failed, we might not be able to reboot into the bootloader.
+    }
+}
+
 void common_main(void)
 {
     mpu_bitbox02_init();
-    if (!_setup_wally()) {
-        Abort("_setup_wally failed");
-    }
     if (!memory_setup(&_memory_interface_functions)) {
+        // If memory setup failed, this also might fail, but can't hurt to try.
+        _bootloader_autoenter();
         Abort("memory_setup failed");
     }
+
+    if (!_setup_wally()) {
+        _bootloader_autoenter();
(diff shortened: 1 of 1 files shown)


benma:
@x1ddos 'securechip_setup failed' happened in the wild, which prompted this PR.

src/common_main.c
```diff
@@ -67,21 +67,43 @@ static bool _setup_wally(void)
...
+// bootloader and the device would be bricked.
+static void _bootloader_autoenter(void)
+{
+    auto_enter_t auto_enter = {
+        .value = sectrue_u8,
+    };
+    upside_down_t upside_down = {
+        .value = screen_is_upside_down(),
+    };
+    if (!memory_bootloader_set_flags(auto_enter, upside_down)) {
```
x1ddos, line 81: I think the `if` is unnecessary. I would remove it. but the comment of course is good.

src/common_main.c
```diff
@@ -67,21 +67,43 @@ static bool _setup_wally(void)
     return wally_set_operations(&_ops) == WALLY_OK;
 }
 
+// Go into bootloader if there was an error during startup, so a firmware update can be
+// applied. Otherwise, if there is an Abort() during startup, there would no way to reboot into the
+// bootloader and the device would be bricked.
+static void _bootloader_autoenter(void)
+{
+    auto_enter_t auto_enter = {
+        .value = sectrue_u8,
```
x1ddos, line 76: Just a note: this is also set when in factorysetup. But I guess it's ok if it's set twice, although in factorysetup the upside_down is overwritten to `false`.

src/common_main.c
```diff
@@ -67,21 +67,43 @@ static bool _setup_wally(void)
...
+    };
+    upside_down_t upside_down = {
+        .value = screen_is_upside_down(),
```
x1ddos, line 79: For bitboxbase it's always set to `false`. I know it doesn't really matter but it just feels inconsistent with the rest of auto-enter usage.

x1ddos (approved):
LGTM

src/common_main.c
benma: USE_RESULT prevents me from doing that:

https://github.com/digitalbitbox/bitbox02-firmware/blob/70e31a705ecbad519db3aef539029c44980928df/src/memory/memory.h#L200

And I think that is a good thing. Nice to see that there is a result by the syntax only, not relying on a comment.

src/common_main.c
benma: Only on Abort() though. In that case the device is dead to be discarded, or it can be fixed, in which case it will run normally afterwards.

src/common_main.c
benma: True on one hand, and on the other hand it feels bad to add ifdefs here :thinking:

PR #692, opened 2020-12-31
securechip: propagate and report error codes

benma:
Negative values for BB02 generated errors, positive values form
cryptoauthlib generated errors.

`int` is chosen as the result type as an enum type can be coerced into
int8_t, leading to overflow with high ATCA_* values.

### DIFF
--- a/src/common_main.c
+++ b/src/common_main.c
@@ -102,8 +102,15 @@ void common_main(void)
 
     // securechip_setup must come after memory_setup, so the io/auth keys to be
     // used are already initialized.
-    if (!securechip_setup(&_securechip_interface_functions)) {
+    int securechip_result = securechip_setup(&_securechip_interface_functions);
+    if (securechip_result) {
         _bootloader_autoenter();
-        Abort("securechip_setup failed");
+        char errmsg[100] = {0};
+        snprintf(
+            errmsg,
+            sizeof(errmsg),
+            "Securechip setup failed.\nError code: %i\nPlease contact support.",
+            securechip_result);
+        Abort(errmsg);
     }
 }
--- a/src/securechip/securechip.c
+++ b/src/securechip/securechip.c
@@ -215,15 +215,15 @@ static ATCA_STATUS _lock_slot(securechip_slot_t slot)
     return ATCA_SUCCESS;
 }
 
-static bool _factory_setup(void)
+static ATCA_STATUS _factory_setup(void)
 {
     if (_interface_functions == NULL) {
-        return false;
+        return SC_ERR_IFS;
     }
     bool is_config_locked = false;
     ATCA_STATUS result = atcab_is_locked(LOCK_ZONE_CONFIG, &is_config_locked);
     if (result != ATCA_SUCCESS) {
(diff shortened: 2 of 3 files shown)


src/securechip/securechip.h
```diff
@@ -21,6 +21,17 @@
...
+typedef enum {
+    SC_ERR_ZONE_UNLOCKED_CONFIG = -1,
+    SC_ERR_ZONE_UNLOCKED_DATA = -2,
+    SC_ERR_CONFIG_MISMATCH = -3,
+    SC_ERR_SLOT_UNLOCKED_IO = -4,
+    SC_ERR_SLOT_UNLOCKED_AUTH = -5,
+    SC_ERR_SLOT_UNLOCKED_ENC = -6,
+    SC_ERR_IFS = -7,
+    SC_ERR_FACTORYSETUP = -8,
+} securechip_error_t;
```
x1ddos, line 33: I like it but would it make sense to have a global src/status.h where we'd define all possible enums, not just the securechip: fatfs, libwally things, nanopb, usb and so on. When handling an API call, we could then catch exact error enum and propagate it to the client, no matter at which call chain it occured. This of course requires that all errors have unique enum number.

x1ddos: Maybe not this time some but somewhere in the future.

src/securechip/securechip.c
```diff
@@ -336,49 +309,75 @@ static bool _factory_setup(void)
...
     }
     _interface_functions = ifs;
     ATCA_STATUS result = atcab_init(&cfg);
     if (result != ATCA_SUCCESS) {
-        return false;
+        return result;
     }
 
 #if FACTORYSETUP == 1
     if (!_factory_setup()) {
```
x1ddos, line 379: _factory_setup could also propagate all the errors. Might be useful, too.

src/common_main.c
```diff
@@ -102,8 +102,15 @@ void common_main(void)
 
     // securechip_setup must come after memory_setup, so the io/auth keys to be
     // used are already initialized.
-    if (!securechip_setup(&_securechip_interface_functions)) {
+    int securechip_result = securechip_setup(&_securechip_interface_functions);
```
x1ddos, line 105: What about all the others? `securechip_update_keys`, `securechip_kdf` and so on could all propagate the error... although that might be a much bigger change.

benma: They don't lead to Aborts(), especially not at boot.

Of course it would be nice, but that is a long chain of propagation all the way to the host. I'd say out of scope for this PR.

src/securechip/securechip.c
benma: Done, PTAL

src/common_main.c
x1ddos: Yeah, that's what I thought too. Leave it for later.

src/securechip/securechip.c
x1ddos: :pray:

x1ddos (approved):
LGTM

PR #694, opened 2021-01-03
rust/async_usb: add `next_request()` future primitive

benma:
There are some API workflows that require not only one
request<->response pair, but many request<->resonse pairs in
succession. Bitcoin transaction signing is one such workflow, where
the transaction is streamed in many requests (inputs, outputs,
previous transactions, etc.).

Currently those are all dispatched to individual functions, which
manipulate a global state see what should be processed next. That
sounds exactly what async functions are designed to simplify.

`let usb_in = next_request(usb_out).await;` can be called from any hww
task to wait for another request from the host.

This can turn for example btc signing into a linear async function, in
this style (pseudocode):

```
async fn process_tx(req: Request) -> Response {
    for i in 0..req.num_inputs {
       let input = api_next_request(want_input(i)).await;
       // process input
    }
    // ...
}
```

`api_next_request()` will be built on top of the `next_request()`
primitive to also handle noise and protobuf.

### DIFF
--- a/src/rust/bitbox02-rust/src/async_usb.rs
+++ b/src/rust/bitbox02-rust/src/async_usb.rs
@@ -13,11 +13,11 @@
 // limitations under the License.
 
 //! This module provides the executor for tasks that are spawned with an API request and deliver a
-//! USB response.
+//! USB response. Terminology: host = computer, device = BitBox02.
 
 extern crate alloc;
 
-use crate::bb02_async::{spin as spin_task, Task};
+use crate::bb02_async::{option, spin as spin_task, Task};
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::cell::RefCell;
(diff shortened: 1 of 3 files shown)


src/rust/bitbox02-rust/src/async_usb.rs
```diff
@@ -13,11 +13,11 @@
 // limitations under the License.
 
 //! This module provides the executor for tasks that are spawned with an API request and deliver a
-//! USB response.
+//! USB response. Terminology: host = computer, client = BitBox02.
```
x1ddos, line 16: Nice! But how about "device" or "peripheral"? Client sounds ambiguous imho and typical reminology in USB docs I've seen are hosts, hubs, peripherals and devices.

src/rust/bitbox02-rust/src/async_usb.rs
```diff
@@ -129,10 +189,20 @@ pub fn copy_response(dst: &mut [u8]) -> Result<usize, CopyResponseErr> {
...
         UsbTaskState::Nothing => Err(CopyResponseErr::NotRunning),
-        UsbTaskState::Running(_) => Err(CopyResponseErr::NotReady),
+        UsbTaskState::Running(Some(_), ref mut next_request_state) => {
+            if let WaitingForNextRequestState::SendingResponse(ref response) = next_request_state {
+                let len = response.len();
+                dst[..len].copy_from_slice(&response);
+                *next_request_state = WaitingForNextRequestState::AwaitingRequest;
+                Ok(len)
+            } else {
+                Err(CopyResponseErr::NotReady)
```
x1ddos, line 199: The docs mention `Err(false)` and `Err(true)` above. I think they're outdated.

Also, how about defining an `Error` enum with `CopyNotReady` and `CopyNotRunning`. I suspect we minght have other errors in the future, not just during "copy".

x1ddos (commented):
Some first comments. Want to look more at this today.

src/rust/bitbox02-rust/src/async_usb.rs
```diff
@@ -115,7 +175,7 @@ pub enum CopyResponseErr {
     NotReady,
 }
 
-/// To be called in response to the client asking for the result of a
+/// To be called in response to the host asking for the result of a
```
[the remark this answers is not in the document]
benma: It is an async model. The host (computer) polls the client (the device) until it has a response. That's why `UsbTaskState` has the `Running` and `ResultAvailable` states.

src/rust/bitbox02-rust/src/async_usb.rs
benma: Done

src/rust/bitbox02-rust/src/async_usb.rs
benma: Fixed the doc.

The second part: not sure, I usually don't like error enums where not every error is possible, cause then in the error handling you have a `_ => ...` arm that swallows new error conditions (where the compiler could warn you to handle them instead).

Anyway, I'll leave it as is for now, we don't even have more errors at the moment.

src/rust/bitbox02-rust/src/async_usb.rs
```diff
@@ -26,6 +26,34 @@ use core::task::Poll;
...
 
+/// If a task is running (see `UsbTaskState`), this state is active and manages waiting for another
+/// request from the host in an async fashion. This allows to have multi-request workflows in async.
+///
+/// Normal flow: request(A) -> response(A)
+/// Multirequess flow: request(A) -> response(WANT_B) -> request(B) -> response(A)
+enum WaitingForNextRequestState {
+    /// We are not waiting for another request from the host. This is the default state when
+    /// processing a request.
+    NotWaiting,
```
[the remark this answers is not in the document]
benma: Done

benma:
Rebased

x1ddos (approved):
Very nice idea.

PR #695, opened 2021-01-03
hww: add `next_request()` future handling noise and protobuf

benma:
Builds on https://github.com/digitalbitbox/bitbox02-firmware/pull/694

Now, in any async api processing function, one can call
`let request = next_request(Response {...}).await?` to wait for
another request from the host.

### DIFF
--- a/src/rust/bitbox02-rust/src/hww.rs
+++ b/src/rust/bitbox02-rust/src/hww.rs
@@ -25,6 +25,25 @@ const OP_STATUS_SUCCESS: u8 = 0;
 const OP_STATUS_FAILURE: u8 = 1;
 const OP_STATUS_FAILURE_UNINITIALIZED: u8 = 2;
 
+/// Must be called during the execution of a usb task. This sends out the response to the host and
+/// awaits the next request. If the request is not a valid noise encrypted protofbuf api request
+/// message, `Err(Error::InvalidInput)` is returned.
+pub async fn next_request(
+    response: crate::pb::response::Response,
+) -> Result<crate::pb::request::Request, api::error::Error> {
+    let mut out = [OP_STATUS_SUCCESS].to_vec();
+    noise::encrypt(&api::encode(response), &mut out).or(Err(api::error::Error::NoiseEncrypt))?;
+    let request = crate::async_usb::next_request(out).await;
+    match request.split_first() {
+        Some((&noise::OP_NOISE_MSG, encrypted_request)) => {
+            let decrypted_request =
+                noise::decrypt(&encrypted_request).or(Err(api::error::Error::NoiseDecrypt))?;
+            api::decode(&decrypted_request[..])
+        }
+        _ => Err(api::error::Error::InvalidInput),
+    }
+}
+
 /// Process OP_UNLOCK.
 async fn api_unlock() -> Vec<u8> {
     match crate::workflow::unlock::unlock().await {
--- a/src/rust/bitbox02-rust/src/hww/api.rs
+++ b/src/rust/bitbox02-rust/src/hww/api.rs
@@ -14,6 +14,8 @@
 
 use crate::pb;
 
+pub(super) mod error;
+
 #[cfg(feature = "app-ethereum")]
 mod ethereum;
 
@@ -23,7 +25,6 @@ mod bitcoin;
 mod backup;
 mod device_info;
 mod electrum;
-mod error;
 mod reset;
 mod restore;
 mod rootfingerprint;
--- a/src/rust/bitbox02-rust/src/hww/api.rs
+++ b/src/rust/bitbox02-rust/src/hww/api.rs
@@ -41,13 +41,23 @@ use pb::response::Response;
 use prost::Message;
 
 /// Encodes a protobuf Response message.
-fn encode(response: Response) -> Vec<u8> {
+pub fn encode(response: Response) -> Vec<u8> {
     let response = pb::Response {
         response: Some(response),
     };
     let mut out = Vec::<u8>::new();
     response.encode(&mut out).unwrap();
     out
+}
+
+/// Decodes a protofbuf Request message.
+pub fn decode(input: &[u8]) -> Result<Request, Error> {
+    match pb::Request::decode(input) {
+        Ok(pb::Request {
+            request: Some(request),
+        }) => Ok(request),
+        _ => Err(Error::InvalidInput),
+    }
 }
 
 /// Returns the field tag number of the request as defined in the .proto file.  This is needed for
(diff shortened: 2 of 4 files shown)


benma:
Check this for a demo:

https://github.com/digitalbitbox/bitbox02-firmware/commit/7d888055e0f6b8ad2a8102a0b8017bea283f7774#diff-9db64c9b2e2a0d14bd88ab8afc92e70ae4c14e318f32743458c5f4285805d424R25

The old random number request basically just puts a screen onto the screen stack in a non blocking way, which means that other workflows can run on top of it (e.g. signing a tx).

What should have been done was to first return the random number (so the client can display it), then another request to show it on the screen. Without `next_requst()`, this would require a state machine in a global, and also caching the random number. Now it is a simple linear function.

x1ddos:
> Check this for a demo:
> 
> [7d88805#diff-9db64c9b2e2a0d14bd88ab8afc92e70ae4c14e318f32743458c5f4285805d424R25](https://github.com/digitalbitbox/bitbox02-firmware/commit/7d888055e0f6b8ad2a8102a0b8017bea283f7774#diff-9db64c9b2e2a0d14bd88ab8afc92e70ae4c14e318f32743458c5f4285805d424R25)
> 
> The old random number request basically just puts a screen onto the screen stack in a non blocking way, which means that other workflows can run on top of it (e.g. signing a tx).
> 
> What should have been done was to first return the random number (so the client can display it), then another request to show it on the screen. Without `next_requst()`, this would require a state machine in a global, and also caching the random number. Now it is a simple linear function.

So, instead of 1 "generate a random number" request, there are now two requests? That doesn't sound simple.

```rust
pub async fn process() -> Result<Response, Error> {
    let number = bitbox02::random::bytes32().to_vec();
    let number_hex = hex::encode(&number);
    let request =
        crate::hww::next_request(Response::RandomNumber(pb::RandomNumberResponse { number }))
            .await?;
    if !matches!(
        request,
        pb::request::Request::RandomNumber(pb::RandomNumberRequest {})
    ) {
        return Err(Error::InvalidInput);
    }
```

I mean, the second `request` looks more like a hack...

benma:
> > Check this for a demo:
> > [7d88805#diff-9db64c9b2e2a0d14bd88ab8afc92e70ae4c14e318f32743458c5f4285805d424R25](https://github.com/digitalbitbox/bitbox02-firmware/commit/7d888055e0f6b8ad2a8102a0b8017bea283f7774#diff-9db64c9b2e2a0d14bd88ab8afc92e70ae4c14e318f32743458c5f4285805d424R25)
> > The old random number request basically just puts a screen onto the screen stack in a non blocking way, which means that other workflows can run on top of it (e.g. signing a tx).
> > What should have been done was to first return the random number (so the client can display it), then another request to show it on the screen. Without `next_requst()`, this would require a state machine in a global, and also caching the random number. Now it is a simple linear function.
> 
> So, instead of 1 "generate a random number" request, there are now two requests? That doesn't sound simple.
> 
> ```rust
> pub async fn process() -> Result<Response, Error> {
>     let number = bitbox02::random::bytes32().to_vec();
>     let number_hex = hex::encode(&number);
>     let request =
>         crate::hww::next_request(Response::RandomNumber(pb::RandomNumberResponse { number }))
>             .await?;
>     if !matches!(
>         request,
>         pb::request::Request::RandomNumber(pb::RandomNumberRequest {})
>     ) {
>         return Err(Error::InvalidInput);
>     }
> ```
> 
> I mean, the second `request` looks more like a hack...

That's a consequence of making the random api call work properly instead of being broken, not a consequence of `next_request()` ;)

The current way this api endpoint works is a hack. It is the *only* api endpoint where the screen is not cleaned up by the end of the api call. The random number stays on the screen. If you start signing a tx, it will be laid over the random number, and in the end goes back to the random number. If you close the app, the random number will stay there.

It should have been like this all along, and `next_request` makes this easy.

That being said, don't focus on the random number request. DJB even agreed to test removing it altogether (app first, then firmware :tada: ). It was just an example of how it can be used to perform a multi-request api endpoint.

x1ddos:
> The current way this api endpoint works is a hack.

Then maybe the second request should be different, something like 
Request::RandomNumberComplete or Finish or Done.

 > It is the *only* api endpoint where the screen is not cleaned up by 
the end of the api call.

Yeah, I guess you know better. Can't remember all API by heart :)

 > It was just an example

Right. The precursor 
https://github.com/digitalbitbox/bitbox02-firmware/pull/694 looks good 
imho but I just wanted to check how it would be used in practice here, 
in different scenarious. I'll move on to the actual commits in the PR.

benma:
rebased

src/rust/bitbox02-rust/src/hww.rs
```diff
@@ -25,6 +25,25 @@ const OP_STATUS_SUCCESS: u8 = 0;
...
 const OP_STATUS_FAILURE_UNINITIALIZED: u8 = 2;
 
+/// Must be called during the execution of a usb task. This sends out the response to the host and
+/// awaits the next request. If the request is not a valid noise encrypted protofbuf api request
+/// message, `Err(Error::InvalidInput)` is returned.
+pub async fn next_request(
+    response: crate::pb::response::Response,
+) -> Result<crate::pb::request::Request, api::error::Error> {
+    let mut out = [OP_STATUS_SUCCESS].to_vec();
+    noise::encrypt(&api::encode(response), &mut out).or(Err(api::error::Error::Generic))?;
```
x1ddos, line 35: I'd rather make an Error::Noise. It probably won't happen but if it does, easy to identify where it's coming from, especially in the future if there's a bug in the noise bits.

x1ddos: Also below at noise::decrypt. How about `Error:NoiseEncrypt` and `Error::NoiseDecrypt`?

x1ddos (approved):
Very cool. Played around with something similar to the random number patch.

LGTM

src/rust/bitbox02-rust/src/hww.rs
benma: Added, PTAL. See also the extended commit message:

> The new noise errors are likely never delivered to the host
successfully. If encryption fails, the error response itself also will
likely fail to be encrypted. If decryption fails, the host will likely
not be able to decrypt the encrypted error response either.

Any errors here are likely not recoverable and a replug will be needed.

x1ddos: Yeah for sure but I still think it's a good practice. Maybe in the future there will be an abort showing the error on the display or something like that. Helps maintenance and debugging imho.

PR #697, opened 2021-01-05
Anti-Klepto part 1) keystore: add host nonce contribution to the signature nonce 

benma:
Builds on https://github.com/digitalbitbox/bitbox02-firmware/pull/696 to avoid conflicts (bitcoin sign message is changed to accept the host nonce contribution).


benma:
The Anti-Klepto PR has finally been completed upstream at secp256k1-zpk. This is the first in a series of PRs impleneting the Anti-Klepto protocol when signing BTC/LTC transactions.

Anti-Klepto protocol described in the comment docs:

https://github.com/ElementsProject/secp256k1-zkp/blob/673e551f4d1092e2816f497e5abe2c34ece7235e/include/secp256k1_ecdsa_s2c.h#L100

This PR prepares step 4. The next PR will add a client (signer) nonce commitment function and a unit test showing the full protocol.

benma:
Hmm the first commit of this PR somehow (rebase/fixup error?) in this commit, already merged :( Sorry!

Change of `external/libwally-core` submodule:
https://github.com/digitalbitbox/bitbox02-firmware/pull/696/commits/305f78c5b75eb2e3a048aceaba31da4dc7424158

external/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ string(REPLACE "-mfpu=fpv4-sp-d16" "" MODIFIED_C_FLAGS ${MODIFIED_C_FLAGS_TMP})
 # wally-core
 
 # configure flags for secp256k1 bundled in libwally core, to reduce memory consumption
-set(LIBWALLY_SECP256k1_FLAGS --with-ecmult-window=2 --with-ecmult-gen-precision=2 --enable-ecmult-static-precomputation)
+set(LIBWALLY_SECP256k1_FLAGS --with-ecmult-window=2 --with-ecmult-gen-precision=2 --enable-ecmult-static-precomputation --enable-module-ecdsa-s2c)
```
x1ddos, line 17: I'm probably missing something because the CI builds this ok, but it seems like `enable-module-ecdsa-s2c` flag was introduced in https://github.com/ElementsProject/secp256k1-zkp/commit/8e46cac. We're using https://github.com/digitalbitbox/libwally-core/tree/bitbox02-firmware-20201213 in which secp256 lib is at an earlier commit [6370bdd](https://github.com/ElementsProject/secp256k1-zkp/tree/6370bdd537492638e5ae0f7235cc73daab13ff12).

How does it work?

---

aha wait, `git log external/libwally-core` was updated by 305f78c5b75eb2e3a048aceaba31da4dc7424158:

image

 So, we're now actually using a libwally at a new branch `bitbox02-20210104` where libsecp is switched to https://github.com/digitalbitbox/secp256k1-zkp. Just a bit confusing, can't remember this happening.

Is the plan to send https://github.com/digitalbitbox/secp256k1-zkp/commit/5e3bd34a5eb0409f5c5f42f3c3385d7ebfcc15dc upstream and will this be our little custom changes?

x1ddos (approved):
> ... the signatures fixtures in the unit tests change. They are still valid and deterministic signatures.

I think the best way to ensure this is to add a simple test that proves it with a `secp256k1_ecdsa_verify` or whichever func fits best.

LGTM otherwise.

external/CMakeLists.txt
benma: About that .... see my previous comment:

https://github.com/digitalbitbox/bitbox02-firmware/pull/697#issuecomment-754973416

Sorry for that :innocent:

benma:
> > ... the signatures fixtures in the unit tests change. They are still valid and deterministic signatures.
> 
> I think the best way to ensure this is to add a simple test that proves it with a `secp256k1_ecdsa_verify` or whichever func fits best.
> 
> LGTM otherwise.

That already exists:

https://github.com/digitalbitbox/bitbox02-firmware/blob/bffcf653ee18dfb3608a4f6961e72b629fd2e1f1/test/unit-test/test_keystore.c#L248

external/CMakeLists.txt
x1ddos: ah those emacs users... :laughing:

x1ddos:
> > > ... the signatures fixtures in the unit tests change. They are still valid and deterministic signatures.
> > 
> > 
> > I think the best way to ensure this is to add a simple test that proves it with a `secp256k1_ecdsa_verify` or whichever func fits best.
> > LGTM otherwise.
> 
> That already exists:
> 
> https://github.com/digitalbitbox/bitbox02-firmware/blob/bffcf653ee18dfb3608a4f6961e72b629fd2e1f1/test/unit-test/test_keystore.c#L248

Yes but only with NULL

benma:
> Yes but only with NULL

no, that `NULL` is the `recid`. `host_nonce` is provided, see here:

https://github.com/digitalbitbox/bitbox02-firmware/blob/16df9fbe2124ecd14b2f1f4639b8c9c8407091f2/test/unit-test/test_keystore.c#L247-L251

granted, it is 32 zero bytes (not NULL), but it is still mixed in and it changes the signature, and signature validity is tested.

x1ddos:
Yeah sorry, I actually meant {0}, just looked at the wrong arg.

 > granted, it is 32 zero bytes (not NULL), but it is still mixed in and 
it changes the signature, and signature validity is tested.

sure but you never know. it isn't that hard to add a non-zero value, is it.

benma:
> sure but you never know. it isn't that hard to add a non-zero value, is it.

Okay, done ;) https://github.com/digitalbitbox/bitbox02-firmware/pull/699

To me it was clear that zeroes are also mixed in, as that is technically a valid host nonce contribution as well, but you are right that this is not obvious in general.

PR #699, opened 2021-01-06
unit-test: better host_nonce test in keystore_secp256k1_sign

benma:
Zero bytes are also mixed in, but a non-zero value gives more confidence.


x1ddos (approved):
Thanks! :pray:

PR #702, opened 2021-01-08
Rust bignum

benma:


benma:
@x1ddos this unfortunately adds ~10kb to the binary size. 7kb of it is in the invocation of `mul()` inside `rust_ethereum_bigint_mul()`, i.e. bignum multiplication. Thinking if there is a good way to reduce that.

benma:
No need to review this anymore, most of it would be obsolete by porting the signing code completely.

PR #703, opened 2021-01-09
rust-unittests: Allow c functions to be called

NickeZ:
BEHOLD! I won over cargo/cmake/cmocka/gcc/clang/ar :) You can now call c functions in rust unit tests.

Not sure if you should merge this as is. I disabled the sanitizers because I thought that was the easy solution. If everything was compiled with clang maybe it would be easier to have them on.

CMake is an absolutely horrible language. However I managed to write a
custom target to merge the library that is outputted from cargo with all
the object files that is "our" code into a library called
"bitbox_merged". Cargo is then told via a build-script to link to that
library (and our external dependencies) for tests. It seems to work \o/.


NickeZ:
I updated it to use a separate build directory for rust unit tests so that santizers could stay enabled for C.

benma:
This is ridiculous skill!! :100: 

I tested it and it works! Amazing. Post your LN invoice, that is worth *lots* of beers.

One downside is that running rust unit tests takes a lot longer now, as it it is recompiling everything every time even if there was no change. Is that what is happening? Do you see any easy way to reduce that?

benma (approved):

NickeZ:
> One downside is that running rust unit tests takes a lot longer now, as seems to be recompiling everything every time even if there was no change. Is that what is happening? Do you see any easy way to reduce that?

I'm sure it can be done, but CMake is an awful language and in general it isn't good to mix build systems (CMake+Cargo). I don't know from the top of my head what could be wrong and I don't know when I'll have time to look into it. Maybe we aren't specifying the "recipes" in CMake correctly so that it understands when to rerun Cargo, or maybe some file gets "touched" so that cargo thinks it has to rebuild stuff...

NickeZ:
(I thought my commit wasn't committed, but actually I had to teach github my new subkeys... Anyway, that was the reason why I pushed a new commit)

NickeZ:
I fixed the cache invalidation so now reubilds are fast when you rerun unit tests. Hopefully I didn't increase the overall CI time too much...

NickeZ:
You can send donations here in case you were serious: lnbc1p0lejaxpp5zk7een4nnjt7dek6886unsqf0j250zekgajwsw09j3065cnrdntqdqqxqyjw5q9qtzqqqqqq9qsqsp5fej6jzh8q0ys5kfpfu33vge2kh98wy9slat7n4h03z8f53mg0a6qrzjqwryaup9lh50kkranzgcdnn2fgvx390wgj5jd07rwr3vxeje0glclll5u0dxsh6nk5qqqqlgqqqqqeqqjq7p094u0ddyuynttfd0fus88pdumgkhwulrhkusu2qzs5tjm6mwlhn2cx6t7k3g0vdwnfh9uwh3e6m7yc8ee6fgcun2xduccx83mq7qcqtrup3z

benma:
Great, rebuild is fast!

I changed a C file, but when running `make run-rust-unit-tests`, it was not rebuilt, and the Rust unit test ran without the change. Can this be fixed? Not a big deal really if not, time spent changing the C code while writing Rust unit tests is hopefully zero anyway.

NickeZ:
Seemed like it was easy to fix, I tested it quickly, so you can do some more testing if you want.

benma:
Works perfectly, great.

Can you just undo the proof on concept unit test with the version? If this stays, the unit test will break with the next version bump.

Then I think this is ready.

NickeZ:
Done and CI is green :)

benma:
Thanks, this will really help!

PR #706, opened 2021-01-17
rust/workflow/confirm: convert to Result return type

benma:
Easier to propagate errors than checking for a bool.

### DIFF
--- a/src/rust/bitbox02-rust/src/workflow/confirm.rs
+++ b/src/rust/bitbox02-rust/src/workflow/confirm.rs
@@ -16,13 +16,19 @@ use crate::bb02_async::option;
 
 pub use bitbox02::ui::{ConfirmParams as Params, Font};
 
+pub struct UserAbort;
+
 /// Returns true if the user accepts, false if the user rejects.
-pub async fn confirm(params: &Params<'_>) -> bool {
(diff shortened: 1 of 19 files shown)


benma:
I also manually tested each changed line just to be sure.

src/rust/bitbox02-rust/src/workflow/confirm.rs
```diff
@@ -16,13 +16,19 @@ use crate::bb02_async::option;
 
 pub use bitbox02::ui::{ConfirmParams as Params, Font};
 
+pub struct UserAbort;
```
x1ddos, line 19: I wonder if if makes sense to start defining common errors. I imagine something like this in `bitbox02::ui` mod:

```rust
enum ErrorKind {
  UserAbort,
  IncorrectPassword,
  InvalidInput,
}
```

Then can use this error enum in all UI and workflow related modules. I think it would work well for `UserAbort` and reduce code for `core::convert::From` in API conversion, but OTOH I like reading `UnlockError::IncorrectPassword` because it tells me exactly where it's coming from. So, not sure but maybe something to consider in the future.

x1ddos (approved):
Nice.

PR #707, opened 2021-01-18
rust: Always reubuild library

NickeZ:
Make cannot know when to rebuild the rust library since it isn't aware
of the source->output dependencies. Cargo does output a "depfile" (which
can be used to teach Make/Ninja about dependencies) but CMake can only
use depfiles if the "Ninja" backend is used.

For now, we add an output that is never produced which will always
trigger a rebuild of the library. Cargo is clever enough to only rebuild
what is needed anyway.


benma (approved):
tACK, thanks!

PR #708, opened 2021-01-19
workflow/mnemonic/get: confirm 24th word

benma:
The 24th word is selected from 8 valid candidaes in a menu. The arrows
and the 'select' button are close to each other, so a typo could
happen quite easily. A typo at this stage is very annoying, but also
the user might not even realize it, and they would proceed thinking
they got the right word.


x1ddos (approved):
LGTM

PR #713, opened 2021-01-22
util: add decimal formatting utility function

benma:
This is simpler than modular arithmetic and will also work with
bigints to format Ethereum values, avoiding bigint modular arithmetic.

### DIFF
--- a/src/rust/util/src/decimal.rs
+++ b/src/rust/util/src/decimal.rs
@@ -1,0 +1,57 @@
+// Copyright 2021 Shift Crypto AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+extern crate alloc;
+use alloc::string::{String, ToString};
+
+pub trait Format: ToString {}
+
+impl Format for u64 {}
+
+/// Formats integer `value` as `value / 10^decimals`, with up to `decimals` decimal places.
+/// E.g. "123450" with decimals=3: "123.45".
+/// Value must consists only of '0'-'9' digits.
+pub fn format<F: Format>(value: F, decimals: usize) -> String {
+    let mut v: String = format!("{:0>width$}", value.to_string(), width = decimals + 1);
+    v.insert(v.len() - decimals, '.');
+    v.trim_end_matches('0').trim_end_matches('.').into()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_format_int() {
+        assert_eq!(format(0u64, 0), "0");
+        assert_eq!(format(0u64, 2), "0");
+        assert_eq!(format(0u64, 6), "0");
+
+        assert_eq!(format(1u64, 0), "1");
+        assert_eq!(format(1u64, 1), "0.1");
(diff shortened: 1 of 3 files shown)


src/rust/util/src/decimal.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020 Shift Crypto AG
```
x1ddos, line 1: 2021 :)

src/rust/util/src/decimal.rs
```diff
@@ -0,0 +1,51 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+extern crate alloc;
+use alloc::string::String;
+
+/// Formats integer `value` as `value / 10^decimals`, with up to `decimals` decimal places.
+/// E.g. "123450" wit decimals=3: "123.45".
```
x1ddos, line 19: with typo

src/rust/util/src/decimal.rs
```diff
@@ -0,0 +1,51 @@
...
+/// Value must consists only if '0'-'9' digits.
```
x1ddos, line 20: you meant "consists only **of**"?

src/rust/util/src/decimal.rs
```diff
@@ -0,0 +1,51 @@
...
+        assert_eq!(format("1", 6), "0.000001");
+
+        assert_eq!(format("12345", 0), "12345");
+        assert_eq!(format("12345", 1), "1234.5");
+        assert_eq!(format("12345", 2), "123.45");
+        assert_eq!(format("12345", 3), "12.345");
+        assert_eq!(format("12345", 4), "1.2345");
+        assert_eq!(format("12345", 5), "0.12345");
+        assert_eq!(format("12345", 6), "0.012345");
+        assert_eq!(format("12345", 7), "0.0012345");
```
x1ddos, line 49: I know it's nothing but would be nice to have a case with both non-zero and zero, like "123450" and "1234500".

src/rust/util/src/decimal.rs
```diff
@@ -0,0 +1,51 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+extern crate alloc;
+use alloc::string::String;
+
+/// Formats integer `value` as `value / 10^decimals`, with up to `decimals` decimal places.
+/// E.g. "123450" wit decimals=3: "123.45".
+/// Value must consists only if '0'-'9' digits.
+pub fn format(value: &str, decimals: usize) -> String {
```
x1ddos, line 21: I'd actually take u64 here and do `satoshi.to_string()`. It just doesn't feel right - formatting a string with decimals which makes sense only for integers. Then, `format_amount` becomes:

```rust
    let mut s = util::decimal::format(satoshi, 8);
    s.push(' ');
    s.push_str(unit);
    s
```

You could even accept the unit here:

```rust
pub fn format(value: u64, decimals: usize, unit: &str) -> String
```

benma: see the commit message - this will also be used with ETH values that exceed u64.

for more type safety i could make a trait and put it on both u64 and BigUint? :thinking:

benma: I did it with a type constraint that I can also put on BigUint later :+1:

src/rust/util/src/decimal.rs
benma: :+1:

src/rust/util/src/decimal.rs
benma: :+1:

src/rust/util/src/decimal.rs
benma: :+1:

src/rust/util/src/decimal.rs
benma: :+1:

x1ddos (approved):
Nice.

PR #714, opened 2021-01-24
rust: use real functions instead of mocking

benma:


x1ddos (approved):
Nice! I guess I missed all the "rust -> c in tests" party but I'll catch up.

benma:
> Nice! I guess I missed all the "rust -> c in tests" party but I'll catch up.

That happened here: https://github.com/digitalbitbox/bitbox02-firmware/pull/703 - this a really great addition which will allow to transition more smoothly to Rust.

PR #716, opened 2021-01-25
rust: port verify_recipient and verify_total workflows

benma:


src/rust/bitbox02-rust/src/workflow/transaction.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use crate::bb02_async::option;
+use core::cell::RefCell;
+
+extern crate alloc;
+use alloc::boxed::Box;
+
+pub struct UserAbort;
+
+pub async fn verify_recipient(recipient: &str, amount: &str) -> Result<(), UserAbort> {
```
x1ddos, line 23: these don't seem to be used anywhere? did you mean to add another commit to replace bits in C to call these functions?

src/rust/bitbox02-rust/src/workflow/transaction.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright 2020 Shift Crypto AG
```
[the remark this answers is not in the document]
benma: done

src/rust/bitbox02-rust/src/workflow/transaction.rs
benma: Used in this follow up PR:

https://github.com/digitalbitbox/bitbox02-firmware/pull/720

src/rust/bitbox02-rust/src/workflow/transaction.rs
```diff
@@ -0,0 +1,49 @@
...
+    let result = RefCell::new(None as Option<Result<(), UserAbort>>);
+
+    let mut component = bitbox02::ui::confirm_transaction_fee_create(
+        total,
+        fee,
+        Box::new(|ok| {
+            *result.borrow_mut() = Some(if ok { Ok(()) } else { Err(UserAbort) });
+        }),
+    );
+    component.screen_stack_push();
```
x1ddos, line 47: A random thought: a component is now popped from the screen stack on drop.
Here it's ok because we await the result but in general, it doesn't feel right to me because on drop of a specific component we just call `ui_screen_stack_pop()` which may result in popping a different one from what's dropped by rust.

x1ddos: oversimplified example:

```rust
comp1.screen_stack_push();
comp2.screen_stack_push();
drop(comp1); // comp2 is actually popped
```

src/rust/bitbox02-rust/src/workflow/transaction.rs
x1ddos: :+1: thanks. good to know :)

src/rust/bitbox02-rust/src/workflow/transaction.rs
x1ddos: of course it has nothing to do this PR in particular but should definitely do something about it.

x1ddos (approved):
LGTM

src/rust/bitbox02-rust/src/workflow/transaction.rs
benma: Yeah we were aware of this. Currently the workflows just have to be sure to push only once. Popping on drop is a requirement because drop is the mechanism to clean up, and the only way to clean up resources e.g. in an aborted async function call.

Maybe we can move the screen push into the component creation function in Rust, so at least push and pop are more close.

Fyi the same issue was in C. Not sure if this can be solved neatly without the component stack also ported to Rust.

x1ddos:
Ok well maybe a simple array of screen stack could be replaced with a 
linked list, for example?

While there, might as well port ui/screen_stack.c to Rust.

benma:
> Ok well maybe a simple array of screen stack could be replaced with a linked list, for example? While there, might as well port ui/screen_stack.c to Rust.

Yeah I guess this is what I meant. Components and the screen stack in Rust would make it possible to make that safer.

Linked list vs. array does not seem relevant, that just changes the representation of the stack?

x1ddos:
A potential issue with using array is: how do you keep track of which 
component is which after popping an item.

Suppose current stack is [A, B, C]. So, A is at index 0, B is 1 and C is 
2. Now, you pop item B. Suddently, you'll have to update indexes of A 
and C so it all remains consistent.

I guess it's technically possible to implement this using an array too, 
just more overhead imho.

Doesn't have to be a linked list though. There are other constructs 
which allow preserving consistency, easier compared to an array.

I'm also still unsure whether "popping from the middle" makes sense in 
general. The screen stack to me is very simple at the moment: the 
component at the top is what's visible. Once popped, the next one 
becomes visible.

With popping from the middle, I'm not sure what that evem means. Like, 
why would you even want to pop a component currently invisible, i.e. in 
"stand by".

Need to think about it but either way, I'm worried this might result in 
a bug sooner or later...

PR #719, opened 2021-02-01
scripts: improve jlink sequence - halt then load

x1ddos:
I believe loadbin/loadfile commands do not halt the CPU of the target.
The correct way to flash is to first halt the CPU, then download the
bin to the target.

The "r" stands for "reset and halt".
I also believe loadbin/loadfile will reset the CPU once done. So, no
need for additional reset after load.

Docs, although you'll notice it's missing loadbin:
https://wiki.segger.com/J-Link_Commander

Remember my weird white dev device that refused to flash at some point?
It works with this sequence.

### DIFF
--- a/scripts/template-bootloader.jlink
+++ b/scripts/template-bootloader.jlink
@@ -1,3 +1,3 @@
-loadfile $file
 r
+loadfile $file
 q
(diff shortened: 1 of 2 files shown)


benma (approved):
utACK

x1ddos:
I'll hold it until you **T**ack :)
Would prefer a second independent check.

benma:
> I'll hold it until you **T**ack :)
> Would prefer a second independent check.

tACK - I managed to flash using the adjusted script.

PR #720, opened 2021-02-02
apps/eth: port transaction signing to Rust

benma:
Builds on top of https://github.com/digitalbitbox/bitbox02-firmware/pull/716


x1ddos (commented):
Cool. What about the app_eth_sighash tests? Now that verify recipients are in rust, they can be stubbed in tests. Seems like it's the only sole ETH test in C.

x1ddos (approved):
LGTM but not sure what the plan is for app_eth_sighash tests in C.

benma:
> LGTM but not sure what the plan is for app_eth_sighash tests in C.

The sighash impl is still in C, and its tests in C remain the implementation it is also ported to Rust.

x1ddos:
Oh, true. I kind of missed that sighash impl is still in C. Never mind. :)

PR #721, opened 2021-02-04
py/bitbox02: spell out the optional dependency on rlp for ethereum

x1ddos:
To use this locally:

    pip install -e .[ethereum]

Users who want to install a released version can do:

    pip install bitbox02[ethereum]

See https://www.python.org/dev/peps/pep-0508/#environment-markers
for docs on how package dependencies can be specified.


x1ddos:
Just trying to please check-pep8

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 - RestoreFrommnemonic: ported to Rust. Will now return UserAbortError on user abort instead of GenericError.
+- Added [rlp] as an optional dependency of the bitbox02 Python package using an environment marker
```
benma, line 5: maybe there should be  a separate changelog for the python package? after all it does not follow the same release schedule as the firmware, and has a different version, etc.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+  `ethereum`; see PEP-508 for environment markers
+
+[rlp]: https://pypi.org/project/rlp/
```
benma, line 8: what does this do?

benma (approved):

NickeZ:
You probably also need to update some README in the `py` directory.

x1ddos (commented):
> You probably also need to update some README in the `py` directory.

done

made some more changes and added `all` optional extras so PTAL

PR #722, opened 2021-02-11
Mock less keystore functions

benma:
After this, we should also not mock the keystore sign functions, which make the tests for signing and address generation and antiklepto more real.


x1ddos (approved):
LGTM

PR #723, opened 2021-02-15
Rust keystore: don't mock pubkey and signing funcs

benma:


NickeZ (approved):
LGTM, I would've asked you to document the `Data` struct. But I guess that struct is being refactored out now that you can test with the C code linked?

benma:
Yeah, except for maybe functions that are hardware related.

PR #724, opened 2021-02-17
Antiklepto for ETH transactions and BTC/ETH messages

benma:
Anti-klepto was released for BTC/LTC transactions in 9.4.0:

https://shiftcrypto.ch/blog/anti-klepto-explained-protection-against-leaking-private-keys/

This adds the same for ETH transaction signing and BTC/ETH message signing.


benma:
I am not sure how to unit test this. I guess mocking the `antiklepto_get_host_nonce()` funcs somehow would work.

NickeZ (approved):
utACK, LGTM :)

PR #729, opened 2021-03-10
apps/eth/sign: warn about unrecognized data

benma:
Confirming the data field without understanding it can be a security
issue. Ideally, all popular contract invocations would be understood
and displayed in a human readable format - showing the raw data is the
last resort. However, the user must understand the meaning when
confirming it.

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
@@ -139,6 +139,21 @@ async fn verify_standard_transaction(
 
     if !request.data.is_empty() {
         confirm::confirm(&confirm::Params {
+            title: "Unknown\ncontract",
+            body: "You will be shown\nthe raw\ntransaction data.",
+            accept_is_nextarrow: true,
+            ..Default::default()
+        })
+        .await?;
+        confirm::confirm(&confirm::Params {
+            title: "Unknown\ncontract",
+            body: "Only proceed if you\nunderstand exactly\nwhat the data means.",
+            accept_is_nextarrow: true,
+            ..Default::default()
+        })
+        .await?;
+
+        confirm::confirm(&confirm::Params {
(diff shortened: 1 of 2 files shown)


src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -138,6 +138,21 @@ async fn verify_standard_transaction(
     let recipient = parse_recipient(&request.recipient)?;
 
     if !request.data.is_empty() {
+        confirm::confirm(&confirm::Params {
+            title: "Unknown\ncontract",
+            body: "You will be shown\nthe raw data field\n of the transactoin.",
```
x1ddos: "... transact**oi**n" typo.

also there's a space in `\n of the transaction`. wouldn't that display the text misaligned like so:

```
You will be shown
the raw data field
 of the transaction.
```

or make it shorter? not sure anyone cares whether it's a *field*.

```
You will be shown
the raw transaction data.
```

x1ddos (approved):
LGTM

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
benma: Fixed space.

if you want (and you should) verify it, you need to know that it is the data field and not the full raw tx.

x1ddos: > it is the data field and not the full raw tx

and yet the 3rd confirmation screen title says "Transaction data".

x1ddos: :) 

not sure you still want text split in 3 lines. seems like it might fit just on two? but not sure.

benma: does not fit unfortunately

PR #733, opened 2021-03-11
rust: update toolchain to 1.50.0

benma:


x1ddos:
Interesting:

```
 /usr/local/bin/../lib/gcc/arm-none-eabi/8.2.1/../../../../arm-none-eabi/bin/ld: 
  ../bin/bootloader-development.elf section `.relocate' will not fit in region `rom'
/usr/local/bin/../lib/gcc/arm-none-eabi/8.2.1/../../../../arm-none-eabi/bin/ld:
  region `rom' overflowed by 2436 bytes
collect2: error: ld returned 1 exit status
```

benma:
Yeah very! The firmware binary also increased by a few hundred bytes 🤨

x1ddos:
I looked at all release notes between 1.47 and 1.50 on https://github.com/rust-lang/rust/blob/master/RELEASES.md but didn't catch anything related. I either missed it or maybe it's somehow related to nostd.

benma:
1.51.0 is already out - will try that before digging into it, maybe it is already fixed

benma:
Closed in favour of https://github.com/digitalbitbox/bitbox02-firmware/pull/740

NickeZ:
The size difference is likely due to changing versions of llvm. The optimisations passes have a big influence.

PR #734, opened 2021-03-21
add assertion from cstenglein to v9.4.0-btc

cstenglein:
First of all, thank you for all the work you've done on the Bitbox02! :)

I think the reproducible builds feature is really awesome, so I would like to add my assertion as well. :)


benma:
Nice! Please sign your commit as well. 

Fyi, https://github.com/digitalbitbox/bitbox02-firmware/releases/tag/firmware-btc-only%2Fv9.5.0 is also available to be verified :)

cstenglein:
@benma done.

Will try to build 9.5.0 soon. :)

benma (approved):

PR #735, opened 2021-03-21
update BUILD and releases/README formatting; fix links and python ins…

cstenglein:
…tall command

When I was trying out the reproducible build, I had some hiccups, so this PR should make some things clearer for others.

Changes I made:

### BUILD.md

- fixed formatting (space before fenced code, see [markdownlint MD031](https://github.com/DavidAnson/markdownlint/blob/v0.23.1/doc/Rules.md#md031)
- added language to every fenced core for better highlighting
- referred to the *BitBox02-Python-library* heading instead of the py/README page, which is less elaborate.
- fixed `pip install` command since there is no `py/bitbox02/requirements.txt`
- fixed cmocka link [markdownlint MD034](https://github.com/DavidAnson/markdownlint/blob/v0.23.1/doc/Rules.md#md034)

### releases/README.md

- added notice that the bitbox python library needs to be installed for the script
- added shasum command for the btc-only firmware
- fixed `Contributing your signature` header

### Things I couldn't fix:

- Somehow make it clear that you don't need the dependencies in the `BUILD.md` file for the reproducible build just run the build.sh script. 
Maybe I just was too confused and this doesn't really need a change, maybe that's something that could be improved.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -35,7 +35,7 @@ The container will contain all tools needed to build the project but it is still
 the J-Link software to flash the bootloader.  Run the commands below to build the container and
 execute a persistent one.
 
-```
+```sh
 make dockerinit
 make dockerdev
 ```
@@ -48,7 +48,7 @@ it is necessary to flash the device in a terminal not running in docker.
 Make sure you have [Homebrew](https://brew.sh) installed.
 Install the dependencies with:
 
-```bash
+```sh
 brew install hidapi cmake protobuf
 brew install automake libtool # for building some code in the external/ folder
 brew tap osx-cross/arm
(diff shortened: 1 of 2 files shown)


NickeZ (approved):

benma:
Thanks. About the last point, did you see this?

https://github.com/digitalbitbox/bitbox02-firmware/tree/d87e39b1f0986b0536455112646128490572851e/releases#how-to-reproduce 

Maybe in BUILD.md, we can make a note/link.

cstenglein:
> Thanks. About the last point, did you see this?

Yes, that's how I was finally able to build everything. I was confused because I was jumping into the BUILD.md first and trying to install the dependencies first. Then I discovered the docker image. And finally the chapter you posted.

So for me it was not clear at first how to proceed, even though the instructions were there. Maybe that's more of a visibility issue / guiding the user to the correct resource than really a documentation issue IMO. That's why I didn't change anything in my PR regarding this.

benma:
Thanks. Please also sign this commit.

cstenglein:
@benma Thanks! Rebased & signed :)

benma:
Thanks!

PR #739, opened 2021-03-31
Reboot purpose

benma:


NickeZ:
I have a feeling that the reboot functions in rust should be marked "no return", right?

benma:
> I have a feeling that the reboot functions in rust should be marked "no return", right?

I am not sure. Is the reboot guaranteed to happen immediately or can there be a (small) delay?

NickeZ:
Yeah depends on how it is implemented. Would have to investigate with a debugger. But I think the idea is that the reboot function never returns.

benma:
> Yeah depends on how it is implemented. Would have to investigate with a debugger. But I think the idea is that the reboot function never returns.

It seems though that not making it no_return can't do harm, while making it no_return can potentially do harm, so I leave it?

NickeZ:
You could potentially use a function that doesn't return in more places, but it probably doesn't matter currently. The example below typechecks even though the `None` arm doesn't have the same type as `s`.

```rust
let val = match some_thing {
    Some(s) => s,
    None => function_that_doesnt_return_is_valid_here(),
}
```

benma:
@x1ddos 

> j/w what kind of harm?

Was just thinking that not having no_return obviously does no harm, but having it has a (remote) chance that code is compiled with undefined behaviour, should the reboot not happen immediately. In any case, we don't need the no_return, as we don't call the function in a way where there is a type checking problem (see Niklas' example).

src/rust/bitbox02-rust/src/hww/api/system.rs
```diff
@@ -0,0 +1,86 @@
...
+            Some(Purpose::Settings) => "Go to\nstartup settings?",
+            // No error, if new client library sends a purpose that we don't understand,
+            // we reboot anyway.
+            None => "Reboot?",
+        },
+        ..Default::default()
+    })
+    .await?;
+    bitbox02::reboot();
+    Ok(Response::Success(pb::Success {}))
```
x1ddos, line 37: So, if the `bitbox02::reboot` was marked as no return, this line wouldn't be needed. It's unreachable anyway, right.

I'm still unclear on what's this mysterious delay are we talking about? in bitbox02::reboot() or somewhere else? and why would it impact the no return?

NickeZ:
It seems you specify no return by having the return type be !. So I guess it would also typecheck? So you can't have the kind of bugs you mention

src/rust/bitbox02-rust/src/hww/api/system.rs
benma: I just said if hypothetically there was a delay, cause I don't know what is going on in a reboot and how it is performed exactly. If you are sure there is no delay at all, I could mark it no_return.

Still fail to see the benefit to be honest. I'll try doing it anyway, but all this (also marking the FFI wrapper as no return) just to remove this one line...

x1ddos: Personally to me it's not "just to remove this one line". It's also about readability. If I see `!` in the funtion declaration I trust it won't return. But at the moment, I look at this:

```rust
bitbox02::reboot();
Ok(Response::Success(pb::Success {}))
```

and my first thought is: wait a second, does reboot returns? when does it actually reboot, in a few microseconds, a minute... an hour? and what happens meanwhile? 

I think I'm starting to understand now. If by "delay" you mean that the `_reboot` > `__NVIC_SystemReset` call chain can return before the MCU is restarted, I just don't see how. https://github.com/digitalbitbox/bitbox02-firmware/pull/739#issuecomment-815617645

x1ddos: If you really don't trust the code in C, add an infinite noop loop in `bitbox02::reboot` func in rust?

benma: Done. I actually had to add the loop{}, as Rust FFI does not recognize that the wrapped function never returns.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -647,11 +647,13 @@ def _msg_query(
             print(response)
         return response
 
-    def reboot(self) -> bool:
+    def reboot(
+        self, purpose: system.RebootRequest.Purpose = system.RebootRequest.Purpose.UPGRADE
+    ) -> bool:
         """TODO: Document"""
```
x1ddos, line 653: since you're here, replace the doc todo?

x1ddos (approved):
LGTM but haven't tried it with my device - don't want to pull in the firmware_v2:19 docker image but if you rebase, I will.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
benma: done

benma:
@x1ddos rebased. Fyi, this is one of the most critical api endpoints. If it doesn't work, one cannot upgrade the firmware ever again.

x1ddos:
So, we were looking at the never_type at 
https://doc.rust-lang.org/std/primitive.never.html but it turns out to 
be behind experimenal flags.

Shame. Reading a Result<!, Error> in the reboot API fn would've been 
very nice.

benma:
@x1ddos unit test fixed

src/rust/bitbox02-rust/src/hww/api/system.rs
```diff
@@ -0,0 +1,83 @@
...
+
+pub async fn reboot(&pb::RebootRequest { purpose }: &pb::RebootRequest) -> Result<Response, Error> {
+    confirm::confirm(&confirm::Params {
+        title: "",
+        body: match Purpose::from_i32(purpose) {
+            Some(Purpose::Upgrade) => "Proceed to upgrade?",
+            Some(Purpose::Settings) => "Go to\nstartup settings?",
+            // No error, if new client library sends a purpose that we don't understand,
+            // we reboot anyway.
+            None => "Reboot?",
```
x1ddos, line 31: This is what will now be displayed if the host doesn't provide a purpose, right. What I imagine is

- a user sees "firmware upgrade available" in the app
- clicks on "upgrade"
- presented with "reboot?" on the device display
- possibly confused what reboot has to do with upgrade?

How about `None` means "Proceed to upgrade?" to keep the original behaviour and add only `Purpose::Settings`?

src/rust/bitbox02-rust/src/shiftcrypto.bitbox02.rs
```diff
@@ -1,103 +1,99 @@
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct PubResponse {
-    #[prost(string, tag="1")]
+    #[prost(string, tag = "1")]
```
x1ddos, line 3: When I build locally, this files gets reformatted. Did your editor auto-formatted it on save?

x1ddos (commented):
Tried this on my dev device. All works but see the other comments.

src/rust/bitbox02-rust/src/hww/api/system.rs
x1ddos: Ah wait, an earlier commit says "Upgrade device (default, backwards compatible)". Where's the default set?

BTW, I'd squash all or most commits.

x1ddos: Oh, never mind. Somehow I missed that `UPGRADE` is `0`.

src/rust/bitbox02-rust/src/shiftcrypto.bitbox02.rs
benma: somehow i get this reformatting. need to figure out why. will fix

src/rust/bitbox02-rust/src/hww/api/system.rs
benma: Exactly, 0 is default. I simply made unknown values also reboot instead of returning an error, just in case.

src/rust/bitbox02-rust/src/shiftcrypto.bitbox02.rs
benma: done

x1ddos (approved):
LGTM but I would squash commit b035494 into afb1917, and e6218ba says "bump version and update CHANGELOG" but there's no version bump?

PR #740, opened 2021-03-31
Rust 1.51.0

benma:


benma:
The firmware decreases in size a little with the new toolchain, but the bootloader increases significantly (55760 vs 47728 bytes).


[bootloaders.tar.gz](https://github.com/digitalbitbox/bitbox02-firmware/files/6263282/bootloaders.tar.gz)

Here are the .bin and .elf files for both.

benma:
@x1ddos added a commit to workaround the issue until it is fixed in the compiler.

x1ddos:
:+1: pulling the new container image...

src/rust/bitbox02-rust/src/bb02_async.rs
```diff
@@ -48,8 +48,8 @@ impl<O> core::future::Future for AsyncOption<'_, O> {
 
 /// Waits for an option to contain a value and returns that value, leaving `None` in its place.
 /// E.g. `assert_eq!(option(&Some(42)).await, 42)`.
-pub fn option<'a, O>(option: &'a RefCell<Option<O>>) -> AsyncOption<'a, O> {
-    AsyncOption(&option)
+pub fn option<O>(option: &RefCell<Option<O>>) -> AsyncOption<'_, O> {
```
x1ddos, line 51: What's the reason for `'_` in returned value? I tried without it. rust seems to be able to figure it out without issues and clippy doesn't complain.

src/rust/bitbox02-rust/src/bb02_async.rs
```diff
@@ -48,8 +48,8 @@ impl<O> core::future::Future for AsyncOption<'_, O> {
...
+    AsyncOption(option)
```
x1ddos, line 52: just checking: this was not clippy, right.

x1ddos (commented):
Can we compare hashes, just to make sure it's still byte-to-byte reproducible. Mine are:

```
$ git rev-parse HEAD
bf832fc89baabe780478c8bfbfb5293cd53d875e
$ git status
On branch rust-1.51.0
nothing to commit, working tree clean

$ make -j firmware; make -j firmware-btc; make -j bootloader
=============================================
--             - General -
-- Firmware version:       v9.5.0-pre+bf832fc89b
-- Firmware v. (git):      v9.5.0+6
-- Firmware-btc version:   v9.5.0-pre+bf832fc89b
-- Firmware-btc v. (git):  v9.5.0+6
-- Bootloader version:     v1.0.3-pre+bf832fc89b
-- Bootloader v. (git):    v1.0.3+250
...

$ sha256sum build/bin/*.bin
04249f530c39260aef809147fc7cd8c3c60f0be810cababb7aadf360047488c5  build/bin/bootloader.bin
a1f741e6e6dd31f88b664b8d1f4276aca13e22651e2ac953ea0a403697259579  build/bin/firmware-btc.bin
e6527c62105a028b8384079fca8635b76aa5c0369f7389681035fd426245b2b8  build/bin/firmware.bin
```

src/rust/bitbox02-rust/src/bb02_async.rs
benma: I guess not. I really don't like that double reference somehow just works. The previous version should have been a compilation error imho.

src/rust/util/src/ascii.rs
```diff
@@ -29,7 +29,7 @@ pub fn is_printable_ascii<T: AsRef<[u8]>>(bytes: T, charset: Charset) -> bool {
     bytes
         .as_ref()
         .iter()
-        .all(|&b| (b >= 32 && b <= 126) || (charset == Charset::AllNewline && b == b'\n'))
+        .all(|&b| (32..=126).contains(&b) || (charset == Charset::AllNewline && b == b'\n'))
```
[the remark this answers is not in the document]
benma: I find it quite readable, but I reverted it and disabled that linter.

src/rust/bitbox02-rust/src/bb02_async.rs
benma: I might have copied it from the original clippy fix suggestion?

I am surprised it works without it, after all, it is an explicit param in the declaration of AsyncOption...

Anyway, removed.

x1ddos (approved):
LGTM

src/rust/bitbox02-rust/src/bb02_async.rs
NickeZ: you are not alone, however auto-deref makes the language much more ergonomic: https://stackoverflow.com/questions/28519997/what-are-rusts-exact-auto-dereferencing-rules

PR #742, opened 2021-04-06
releases/pubkeys: update benma.asc

benma:
Expired a few days ago, this is updated with a new expiry.


x1ddos (approved):
LGTM. Same key as in  https://github.com/digitalbitbox/bitbox-wallet-app/pull/1227#pullrequestreview-631250203

PR #745, opened 2021-04-13
Securechip model

benma:


src/securechip/securechip.c
```diff
@@ -764,3 +764,13 @@ bool securechip_u2f_counter_inc(uint32_t* counter)
...
 }
 #endif
+
+bool securechip_model(securechip_model_t* model_out)
+{
+    uint8_t revision[4] = {0};
+    if (atcab_info(revision) != ATCA_SUCCESS) {
+        return false;
+    }
+    *model_out = revision[3] >= 0x03 ? SECURECHIP_ATECC608B : SECURECHIP_ATECC608A;
```
x1ddos, line 774: I remember reading about somewhere but can't find where it says that any revision greater than 0x03 is 608B. @benma do you remember? Places I looked:

- http://ww1.microchip.com/downloads/en/Appnotes/Migrating-from-the-ATECC608A-to-the-ATECC608B-DS40002237A.pdf
- https://ww1.microchip.com/downloads/en/DeviceDoc/ATECC608B-CryptoAuthentication-Device-Summary-Data-Sheet-DS40002239A.pdf
- http://ww1.microchip.com/downloads/en/Appnotes/Atmel-8984-CryptoAuth-CryptoAuthLib-ApplicationNote.pdf
- https://microchiptech.github.io/cryptoauthlib/a01516.html#ga6e19ecd60be4c74665d9ec142e460771

Maybe I missed it somewhere in https://www.microchip.com/wwwproducts/en/ATECC608B.

x1ddos: Ok, checked elsewhere offline. This is indeed the case: values starting with 0x03 and on indicate 608B. :+1:

messages/bitbox02_system.proto
```diff
@@ -26,11 +26,13 @@ message DeviceInfoRequest {
...
-    string version = 3;
-    bool mnemonic_passphrase_enabled = 4;
-    uint32 monotonic_increments_remaining = 5;
+  string name = 1;
+  bool initialized = 2;
+  string version = 3;
+  bool mnemonic_passphrase_enabled = 4;
+  uint32 monotonic_increments_remaining = 5;
+  // From v9.6.0: "ATECC608A" or "ATECC608B".
+  string securechip_model = 6;
```
x1ddos, line 35: I wonder if an enum would've worked better here?

benma: I wonder the same. If it is an enum, extending it means all client libraries have to be updated with human readable strings matching the enum types.

Like this, the string can be shown as it, and no further protobuf message and client library changes are needed should the other models be added.

x1ddos: Yeah, I guess if the main usecase is to show it in a UI, which it is, string is probably better :+1: 
I think the field comment with exact strings is enough if anyone will want to base logic on a specific model.

x1ddos (approved):
LGTM

benma:
> LGTM

Then approve :pray: :)

PR #750, opened 2021-05-13
memory: increase max number of allowed multisig registrations

benma:
The previous limit of 10 was not enough for a user. While we want keep
as much space reserved for future use as possible, this change should
be okay and still leaves plenty of space. One multisig registration
takes 64 bytes, so we go from 0.625kB to 1.525kB of occupied memory,
out of 8kB in the chunk, with more chunks available for future use.

We can bump the limit easily now since the the space after the multisigs
is unused, so a migration is not necessary.

An alternative to bumping the limit is to provide a way to delete
previous registrations, which would force a re-registration and
re-verification when that multisig account would be used again. This
is also not ideal - having more space to fit accounts is better
and a simpler solution for now. We can observe if there will be users
than require an even greater number, in which case we might offer a
way to delete entries, or suggest to purchase more devices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add securechip_model to DeviceInfo: ATECCC608A or ATECC608B.
 - Added reboot purpose for clearer UX: "Proceed to upgrade?" vs. "Go to startup settings?"
 - Allow creation of 128 bit seeds (12 BIP39 recovery words)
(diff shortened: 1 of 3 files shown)


benma:
Previous bump from 5 to 10: https://github.com/digitalbitbox/bitbox02-firmware/commit/fc522fe6a569c24e09ff9d7151ea12ce240b6fe6

x1ddos (approved):
LGTM!

PR #757, opened 2021-06-07
Display granular error codes when unlock fails unexpectedly

benma:


CMakeLists.txt
```diff
@@ -89,9 +89,9 @@ endif()
 #
 # Versions MUST contain three parts and start with lowercase 'v'.
 # Example 'v1.0.0'. They MUST not contain a pre-release label such as '-beta'.
-set(FIRMWARE_VERSION "v9.6.0")
-set(FIRMWARE_BTC_ONLY_VERSION "v9.6.0")
-set(FIRMWARE_BITBOXBASE_VERSION "v9.6.0")
+set(FIRMWARE_VERSION "v9.6.1")
```
x1ddos, line 92: Didn't we say semver here doesn't really make sense? Why not v9.7.0. But if you insist, showing error codes would be a new feature anyway.

src/keystore.c
```diff
@@ -108,8 +108,19 @@ static uint8_t* _get_bip39_seed(void)
     return _retained_bip39_seed;
 }
 
-static keystore_error_t _stretch_password(const char* password, uint8_t* kdf_out)
+/**
+ * Stretch the user password using the securechip, putting the result in `kdf_out`, which must be 32
+ * bytes. `securechip_result_out`, if not NULL, will contain the error code from `securechip_kdf()`
```
x1ddos, line 113: I wonder if another approach would be better in general:
pack errors in a single return value with bit shifts. For example,

```c
uint32_t keystore_func() {
  res = secchip_func();
  if (res) {
    return (res << 8) | KEYSTORE_SOME_ERR;
  }
  return KEYSTORE_OK;
}

uint32_t secchip_func() {
  ATCA_STATUS res = atcah_io_decrypt(...);
  if (res) {
    return (res << 8) | SECCHIP_SOME_ERR;
  }
  return SECCHIP_OK;
}

res = keystore_func();
if ((res & 0xff) == KEYSTORE_SOME_ERR) {
  // is it a secure chip error?
  if ((res >> 8) & 0xff == SECCHIP_SOME_ERR) {
    // which one?
    ((res >> 16) & 0xff) == ATCA_CONFIG_ZONE_LOCKED;
  }
}
```

This avoid spreading extra arguments for inner errors everywhere and simplifies the logic: no need for "if not NULL, then ...". It just returns the errors and the caller decides what to do about it.

Assuming each set of error types does not exceed 1 byte, can pack up
to 4 levels of errors in 32 bits. Even ATCA_STATUS fits in a single byte:
https://github.com/digitalbitbox/cryptoauthlib/blob/63f4ad4/lib/atca_status.h

With a couple of `#define`s for bit shifts, it would be clear and easy to read the code, too.

src/rust/bitbox02-rust/src/workflow/unlock.rs
```diff
@@ -82,9 +82,13 @@ pub async fn unlock_keystore(
                 n => format!("Wrong password\n{} tries remain", n),
             };
             status(&msg, false).await;
-            Err(UnlockError::IncorrectPassword)
```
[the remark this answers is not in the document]
benma: But why maintain an extra error type if it is dropped everywhere and it is not handled anywhere?

Sure I can do it, but pls explain.

src/rust/bitbox02/src/keystore.rs
```diff
@@ -31,10 +31,17 @@ pub fn is_locked() -> bool {
     unsafe { bitbox02_sys::keystore_is_locked() }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Copy, Clone)]
```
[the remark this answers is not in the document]
benma: Oh that is a leftover from when I attempted to propagate the error, which required Copy (which requires Clone). Will remove.

CMakeLists.txt
benma: > Didn't we say semver here doesn't really make sense?

I don't recall that. We had planned to replace this and make semvers for subsystems, but we don't have that, so this is a semver for the whole firmware API. I bumped  the patch version because no API changed. If in a future PR the API changes, the minor would be bumped.

src/keystore.c
benma: I had considered this, but I dislike bitmasks a bit, I find them harder to read and understand quickly than the extra out arg. Also I dislike worrying if the error codes will fit into the sizes I'd allocate (now and what if ATCA_STATUS gets more errors in the future, etc). All in all, more mental strain.

If you don't mind, I'd keep it like this.

CMakeLists.txt
benma: To elaborate, semver is strictly about API, not about feature-set.

https://webcache.googleusercontent.com/search?q=cache:2-JHjIsz8NQJ:https://semver.org/+&cd=1&hl=en&ct=clnk&gl=ch

(cache since https://semver.org/ is down for me right now)

x1ddos: Oh, it's only about protobuf APIs. I wish it were documented somewhere, in the readme for example.

src/keystore.c
x1ddos: So, you prefer adding and removing extra args everywhere, like in https://github.com/digitalbitbox/bitbox02-firmware/pull/757/commits/f88472e7fd9cc0cc9a8952e393cd1ac7f1eb81a1? I find it a very unsatisfying reason compared to clear advantages with bit shifts. Sure, you make changes here most of the time, so whichever you prefer. But I really don't see how passing around extra args scattered around is easier to maintain.

> harder to read and understand quickly

I did mention a couple `#define`s and simple macros in rust would make it easy to read.

> worrying if the error codes will fit into the sizes I'd allocate

it would always be `uint32_t` in return type and max `uint8_t` for each error set, so 255 possible errors per category: secure chip, keystore, display, qtouch, memory, mmc. I checked most of them and all their errors fit in 1 byte.

> what if ATCA_STATUS gets more errors in the future

It would still be 1 byte. If more, it would be a major breaking change IMHO - they want the lib to be compatible with smaller chips. Making the size larger would complicate things. Very unlikely. If even, it's easy to catch with a simple unit test.

x1ddos (approved):

src/keystore.c
x1ddos: I realize this might be a change bigger than this PR to cover everything but I'm just surprized you don't like the idea.

CMakeLists.txt
x1ddos: Yeah, I got that. Communication. So, changing public APIs like `keystore_unlock` and extending its error enum is not covered.

src/keystore.c
x1ddos: Yeah, an alternative is to have some union struct. One major downside of bit shifts that I would accept is that we don't have a call stack enforcement. So, for example if a func of type A calls B > C > D > A, then it's unclear the octets offset where each error type resides at. A struct can definitely help with that.

CMakeLists.txt
benma: I see, didn't even consider it. But yeah I think it makes sense to treat the communication as the public API, not the rust and C libs that we use to make it.

PR #759, opened 2021-06-14
bootloader: make screen messages more user friendly

x1ddos:
Some consider "bootloader" text on the screen confusing, so the
bootloader now invites users to follow instructions on the
BitBoxApp screen, consistent with the regular mode once a firmware is
booted.

In the case where app flash is empty, it also displays a "let's get
started" message, as well as rendering chunk writes progress with
"INSTALLING" label. BitBox02 is shipped with empty app flash. It can
also be erased manually via the bootloader protocol.


x1ddos:
ok, I reverted changes in `_api_firmware_erase`. @benma convinced me it was mostly useless.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 - Display granular error codes when unlock fails unexpectedly
+- Make bootloader mode more user-friendly
```
benma, line 5: Bootloader releases are separate from firmware releases (different versions and cycle) - maybe it's best to add a separate changelog or separate section for the bootloader in this changelog?

src/bootloader/bootloader.c
```diff
@@ -160,6 +160,26 @@ typedef union {
 
 static uint8_t _loading_ready = 0;
 static uint8_t _firmware_num_chunks = 0;
+// Indicates whether the whole app flash contains only 0xFF.
+// This controls bootloader text messages on the screen.
+// The value is computed at bootloader enter and at OP_ERASE behind a condition.
+static uint8_t _is_app_flash_empty = 0;
```
benma, line 166: we usually use `bool` for bools :P

benma (commented):
nice! utACK, but I want to test it first

src/bootloader/bootloader.c
x1ddos: that's not what I'm seeing in the code:

- `_loading_read` is treated as bool
- some vars use `sectrue_u8` and `secfalse_u8`
- others do use false

wanted to be consistent with the other static vars.

benma:
Please bump the bootloader patch version as well.

x1ddos:
> Please bump the bootloader patch version as well.

done

x1ddos:
PTAL

src/bootloader/bootloader.c
benma: Great you found the only other example :D ok fine - but consider changing both to `bool` :)

benma (approved):
tACK.

from a UI perspective I have a feeling putting the text and logo into the corners is not good design. Centering them and having some padding would look more normal to me. 

Not in this PR of course, but maybe sth for @jadzeidan to consider.

jadzeidan:
> tACK.
> 
> from a UI perspective I have a feeling putting the text and logo into the corners is not good design. Centering them and having some padding would look more normal to me.
> 
> Not in this PR of course, but maybe sth for @jadzeidan to consider.

The corners seem fine to me, but i'll have a better look once I load the new bootloader onto my device.

x1ddos:
Oh, forgot to attach a photo. Here's how it looks like on empty firmware. I like it. 
image

src/bootloader/bootloader.c
x1ddos: :) ok, will send you a separate PR

thisconnect:
Looks great 👍

src/bootloader/bootloader.c
x1ddos: https://github.com/digitalbitbox/bitbox02-firmware/pull/760

PR #760, opened 2021-06-17
bootloader: change vars type from uint8_t to bool where makes sense

x1ddos:
Some vars with uint8_t type were actually treated as booleans.
Makes sense to change those to bool so it's easier to read.

Note that this affects qtouch and gestures.

This is a follow up from https://github.com/digitalbitbox/bitbox02-firmware/pull/759#discussion_r653545158.

I haven't actually tried it on a real device yet. Just sending it here so I won't forget.


benma (approved):
utACK

benma:
@x1ddos merge this?

x1ddos:
missed stdbool.h include in a qtouch mock test. look ok now but going to flash the bootloader one more time before merging, to be sure.

PR #761, opened 2021-06-28
py/bitbox02: add offset of 27 when signing an ETH message

benma:
The use of this offset for uncompressed pubkeys started in Bitcoin,
and in Ethereum, usage is inconsistent.

MyEtherWallet supports both when verifying:

https://github.com/MyEtherWallet/MyEtherWallet/blob/052f697f3b269db4378f331dbb89d24b406bb7c7/src/components/VerifyMessageInput/VerifyMessageInput.vue#L118

Trezor adds this offset as well.

The testing.aopp.group AOPP demo site requires the offset.

### DIFF
--- a/py/bitbox02/bitbox02/bitbox02/bitbox02.py
+++ b/py/bitbox02/bitbox02/bitbox02/bitbox02.py
@@ -774,8 +774,16 @@ class BitBox02(BitBoxCommonAPI):
     ) -> bytes:
         """
         Signs message, the msg will be prefixed with "\x19Ethereum message\n" + len(msg) in the
-        hardware
+        hardware. 27 is added to the recID to denote an uncompressed pubkey.
         """
+
+        def format_as_uncompressed(sig: bytes) -> bytes:
+            # 27 is the magic constant to add to the recoverable ID to denote an uncompressed
+            # pubkey.
+            s = list(sig)
+            s[64] += 27
+            return bytes(s)
+
         request = eth.ETHRequest()
         # pylint: disable=no-member
         request.sign_msg.CopyFrom(eth.ETHSignMessageRequest(coin=coin, keypath=keypath, msg=msg))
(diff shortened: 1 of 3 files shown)


py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -776,6 +776,14 @@ def eth_sign_msg(
         Signs message, the msg will be prefixed with "\x19Ethereum message\n" + len(msg) in the
         hardware
```
x1ddos, line 777: Can you also add a comment here like in https://github.com/digitalbitbox/bitbox02-api-go/pull/55 for public docs:

> 27 is added to the recID to denote an uncompressed pubkey.

py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -800,9 +808,10 @@ def eth_sign_msg(
             if self.debug:
                 print(f"Antiklepto nonce verification PASSED")
 
-            return signature
+            return modify_recid(signature)
 
-        return self._eth_msg_query(request, expected_response="sign").sign.signature
+        signature = self._eth_msg_query(request, expected_response="sign").sign.signature
+        return modify_recid(signature)
```
x1ddos, line 814: when reading here without knowing the body of `modify_recid`, it's unclear what it does. consider renaming to something like `format_as_uncompressed`.

x1ddos (approved):
LGTM

py/bitbox02/bitbox02/bitbox02/bitbox02.py
benma: done

py/bitbox02/bitbox02/bitbox02/bitbox02.py
benma: done

PR #765, opened 2021-07-29
ui/oled: Add support for new display SSD1312

NickeZ:


benma:
We noticed that the screensaver is black - possibly because the brightness is modified while the screensaver is running. Can you check?

NickeZ:
@benma the valid range for the screen brightness of the new screen is 0x01-0xff. The screensaver sets it to 0x00, so that is out of range. My suggestion is that you try different values for brightness until you find a value that you like with both displays. Alternatively, we could have a set of `_dim()` functions that set different "dimmed" brightnesses on different displays.

benma:
@NickeZ thanks for this, it all seems to work well.

I am continuing this in https://github.com/digitalbitbox/bitbox02-firmware/pull/895, making the choice between drivers dynamic instead of compile-time.

PR #766, opened 2021-08-03
bootloader: remove BitBoxBase

benma:
```
make -j bootloader-production
make -j bootloader-btc-production
```

Should result in the same hash before and after this PR. 

This is to tell that this change does not affect the BitBox02 bootloaders.

To avoid version pollution, I use this local unstaged change to force set the version to an arbitrary number:

image

Giving me:

```
29c0e1a2fab7c7ddd07b33176462dd7c16d42cb7a5302df10d9624674fe8ef5c  build/bin/bootloader-production.bin
a271f6c7cee83c7dd86420db494558ec963ad23bb7627716d1a32a0e8f8b712e  build/bin/bootloader-btc-production.bin
```


NickeZ (approved):
utACK

PR #768, opened 2021-08-05
platform/config: remove unused RUST_ defines

benma:
If needed, it is better to define it in bitbox02-cbindgen.toml,
like RUST_APP_ETHEREUM.


NickeZ (approved):
utACK

PR #771, opened 2021-08-26
remove BitBoxBase

benma:
It is a high maintenance effort and not actively in use in any product.

Some keywords related to the BitBoxBase in the codebase that I looked through:

- PLATFORM_BITBOXBASE
- BitBoxBase, bitbox-base, bitbox_bas
- usart (communication)
- led
- button (touch)


NickeZ (approved):
utACK

PR #774, opened 2021-08-29
rust/bitcoin: unit test xpubs

benma:
And remove mocking of `keystore_encode_xpub_at_keypath()`, as we can
now call C code from Rust unit tests directly.


NickeZ (approved):
utACK!

PR #777, opened 2021-08-30
py/bitbox02: lower version number that was erroneously bumped

benma:
The last tagged and released version is 5.3.0. We accidentally bumped
the major version twice. The next version is going to be 6.0.0.

### DIFF
--- a/py/bitbox02/bitbox02/bitbox02/__init__.py
+++ b/py/bitbox02/bitbox02/bitbox02/__init__.py
@@ -16,7 +16,7 @@
 from __future__ import print_function
 import sys
 
-__version__ = "7.0.0"
+__version__ = "6.0.0"
 
 if sys.version_info.major != 3 or sys.version_info.minor < 6:
     print(


x1ddos:
Hm... why "accidentally"? c0a3df117fc618f960cb74e7d1fbd9570cb6027c says:

> For the Python library, we can bump the major version without a
problem.

Is it because 6fa00d50d8dd06eed2836426ffa55b73a08dd34d already changed it from 5.3.0 to 6.0.0? Then maybe let's do version bump only when making a new release, not before?

benma:
> Hm... why "accidentally"? [c0a3df1](https://github.com/digitalbitbox/bitbox02-firmware/commit/c0a3df117fc618f960cb74e7d1fbd9570cb6027c) says:
> 
> > For the Python library, we can bump the major version without a
> > problem.
> 
> Is it because [6fa00d5](https://github.com/digitalbitbox/bitbox02-firmware/commit/6fa00d50d8dd06eed2836426ffa55b73a08dd34d) already changed it from 5.3.0 to 6.0.0? 

Yes because there was already a major bump.

> Then maybe let's do version bump only when making a new release, not before?

Could do, but then it is a bit difficult to figure out in hindsight if it should be a major or minor or patch bump? When you make a change it sounds easier to just bump the appropriate number according to what the change does (if not already done by a previous change).

x1ddos:
I see it differently. When you're about to cut a new release, you look back at all the changes that are to be released, between now and last release, and decide whether its minor or major.

Otherwise, you have to always check: did I or someone else already bump the version? Every time you'll have to look git log for past commits down to the last release.

x1ddos (approved):

benma:
Maybe you got a point :innocent: :sweat: https://github.com/digitalbitbox/bitbox02-api-js/pull/57

> When you're about to cut a new release, you look back at all the changes that are to be released, between now and last release, and decide whether its minor or major.

If you have a large number of commits, then how to decide? It can be difficult to know whether a commit introduces a breaking change or not. Look through all commit messages?

Maybe a solution is to have a tentative next version number in the changelog next to `[unreleased]`?

PR #780, opened 2021-08-31
Remove some unused sd functions

benma:


benma:
sd_write() never even worked as the `bool replace` param is fed into f_open as the `uint8_t mode` with a wrong value

src/sd.c
```diff
@@ -162,7 +162,7 @@ bool sd_write_bin(
     const uint16_t length,
     bool replace)
 {
-    if (!length || length > SD_MAX_FILE_SIZE) {
+    if (!strlens(fn) || data == NULL || !length || length > SD_MAX_FILE_SIZE) {
```
NickeZ, line 165: Seems wrong to me to return false if length is zero. But that is the current behavior, so we can keep it.

NickeZ (changes requested):
Could you also update the docs for `sd_write_bin`?

src/sd.c
benma: Done, PTAL

NickeZ (approved):
utACK

PR #782, opened 2021-09-02
sd: use right types for length_out

benma:
The out length param defined in f_read is of type UINT (typedef of
unsigned int).

The sd_load_bin function uses size_t for the out length param.

On the BitBox02 they are both 4 bytes so it works out. On x86, size_t
is 8 bytes while unsigned int is only 4 bytes, which leads to a
corrupt value due to the invalid cast.


NickeZ (approved):
utACK

PR #783, opened 2021-09-02
external: update FatFs to 0.14b and remove submodule

benma:
http://elm-chan.org/fsw/ff/arc/ff14b.zip

Linked from: http://elm-chan.org/fsw/ff/00index_e.html

Previously we manually patched ff.c to include sdmmc_diskio.h, but
this is not necessary and not advisable: the diskio.h included in the
project is what FatFs expects, and deviations could lead to bugs.

Changelog: http://elm-chan.org/fsw/ff/updates.txt

The submodule was pointed to
https://github.com/digitalbitbox/FatFs/. There is no reason to keep
these files in a separate repo, it just adds inconvenience.


benma:
Please PTAL again

NickeZ (approved):
utACK

PR #786, opened 2021-09-07
keystore: allow zero seeds

benma:
Some users test the wallet using these recovery words:

```
abandon abandon abandon abandon abandon abandon abandon abandon
abandon abandon abandon about
```

Which translates to 16 zero bytes. This is a valid seed just like any
other. The BIP39 seed (the stretched seed using the passphrase) which
is used as the root of BIP32 keys is not zero.

Before this patch, users testing with this seed would get BIP39 unlock
errors due to the wrong sanity check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ## Firmware
 
 ### [Unreleased]
+- Allow recovery words that convert to a zero seed, such as the 12 words `abandon abandon .... about`.
 
 ### 9.7.0 [released 2021-09-06]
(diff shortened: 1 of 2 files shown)


benma:
ping @x1ddos

x1ddos:
Sorry I missed this. The code LGTM but why was this check in there in the first place?

benma:
> Sorry I missed this. The code LGTM but why was this check in there in the first place?

```
static uint8_t _retained_seed[KEYSTORE_MAX_SEED_LENGTH] = {0};
```

I figured since the default value of the seed is zero, it would be good to check it is not zero after unlock, so that there is no chance a bug leaves users accidentally in a wallet derived from zero. what I had not anticipated is that the zero seed is a valid seed and sometimes used by users to test the wallet.

x1ddos (approved):
sorry!

PR #787, opened 2021-09-08
rust: move ethereum crate functions to bitbox02-rust/hww/api/ethereum

benma:
No need for a separate crate - make it more consistent with how
bitcoin code is organized.


NickeZ (approved):
utACK

PR #789, opened 2021-09-10
rust: add better error messages

benma:
The protobuf Error response has a message field, but it is hardcoded
to generic strings depending on the error type, e.g. "generic",
"invalid input", etc. When a user/developer reports this error, better
information is useful.

This patch adds error messages based on the error condition.

Unfortunately, for now we still have to stick to static strings with
no runtime information, e.g. "invalid keypath" over "keypath invalid:
{}" cotaining the actual keypath. I experimented with dynamic strings,
but this immediatelly added another ~7kB in binary bloat due the usage
of String and formatting. Only changing the type from `&'static str`
to `String` adds a few kB.

Alternatives considered:

- use String+format!() to be able to return runtime info (e.g. actual
  keypaths used), as well as chain context strings together, but that
  resulted in too many kilobytes of additional binary bloat.
- error deps like anyhow, snafu, etc. They all add a lot of binary and
code bloat.
- using only numeric error codes instead of static strings to save
binary space. Can still do in the future if needed.
- using `ufmt` crate hoping it produces smaller binaries than
`format!()` for dynamic strings. I tried it and it was about the same.


benma:
Best to look at the unit tests in errors.rs first to see how it is used. The rest of the changes are simply adding some error strings to failure paths.

PR #793, opened 2021-09-20
Mount a FAT32 filesystem in RAM for unit tests

benma:


NickeZ (approved):
utACK

external/fatfs/source/ffconf.h
```diff
@@ -30,7 +30,7 @@
 /  f_findnext(). (0:Disable, 1:Enable 2:Enable with matching altname[] too) */
 
 
-#define FF_USE_MKFS		0
+#define FF_USE_MKFS		1
```
x1ddos, line 33: Should this be set to 1 only when for tests? At least until the real firmware can actually format an SD card in some potential future.

NickeZ: It's ifdeffed on testing in "our" internal api so I think this is fine. I don't think "our" defines should leak into external libs. I would be more OK with setting the variable from cmake (as a compiler command line argument) instead of a header file and have the logic in cmake. But I think the current way is fine.

benma: We also have good dead code elimination.

PR #794, opened 2021-09-20
memory,securechip: make setup param const

benma:
And add missing `static` qualifier in memory.c.

### DIFF
--- a/src/memory/memory.c
+++ b/src/memory/memory.c
@@ -230,7 +230,7 @@ static void _read_shared_bootdata(uint8_t* chunk_out)
 #endif
 }
 
-memory_interface_functions_t* _interface_functions = NULL;
+static const memory_interface_functions_t* _interface_functions = NULL;
 
 /********* Exposed functions ****************/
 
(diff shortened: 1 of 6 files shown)


src/memory/memory.c
```diff
@@ -282,7 +282,7 @@ void memory_get_seed_birthdate(uint32_t* timestamp_out)
     }
 }
 
-bool memory_setup(memory_interface_functions_t* ifs)
+bool memory_setup(const memory_interface_functions_t* ifs)
```
x1ddos, line 285: since you're here, there's also test_memory.c: https://github.com/benma/bitbox02-firmware/blob/5715aeb2/test/unit-test/test_memory.c#L110

src/securechip/securechip.c
```diff
@@ -78,7 +78,7 @@ typedef union {
 
 #pragma GCC diagnostic pop
 
-static securechip_interface_functions_t* _interface_functions = NULL;
+static const securechip_interface_functions_t* _interface_functions = NULL;
```
x1ddos, line 81: there's also common_main.c: https://github.com/benma/bitbox02-firmware/blob/5715aeb2/src/common_main.c#L42-L53
not sure why - the static structs in common_main seems unnecessary but maybe I'm missing somehitng. slightly confused.

x1ddos (approved):
LGTM

src/securechip/securechip.c
benma: They provide the actual functions to be used by memory/securechip in the final firmware (not unit test).

I made them const as well :+1:

PR #795, opened 2021-09-20
memory: introduce address base to address memory areas

benma:
This will make it easy to mock flash memory and smarteeprom memory
transparently in RAM.

### DIFF
--- a/src/memory/memory.c
+++ b/src/memory/memory.c
@@ -152,11 +152,17 @@ static void _clean_chunk(uint8_t** chunk_bytes)
 
 // chunk must have size CHUNK_SIZE. if chunk is NULL, the chunk is erased,
 // i.e. filled with 0xFF.
-static bool _write_to_address(uint32_t addr, uint8_t* chunk)
+//
+// `offset` is a relative address (offset from where the memory starts at `base`) and starts at `0`.
+static bool _write_to_address(uint32_t base, uint32_t offset, uint8_t* chunk)
 {
 #ifdef TESTING
-    return memory_write_to_address_mock(addr, chunk);
+    return memory_write_to_address_mock(base, offset, chunk);
 #else
+    uint32_t addr = base + offset;
+    if (addr < base) {
+        Abort("uint32_t overflow");
+    }
     // Sanity check that the address is correctly aligned,
     // so the erase actually erases only one block.
     if (addr != (addr & ~(CHUNK_SIZE - 1))) {
@@ -205,8 +211,8 @@ static bool _write_chunk(uint32_t chunk_num, uint8_t* chunk)
 #ifdef TESTING
     return memory_write_chunk_mock(chunk_num, chunk);
 #else
-    uint32_t addr = FLASH_APPDATA_START + chunk_num * CHUNK_SIZE;
-    return _write_to_address(addr, chunk);
+    uint32_t offset = chunk_num * CHUNK_SIZE;
+    return _write_to_address(FLASH_APPDATA_START, offset, chunk);
 #endif
 }
 
@@ -324,7 +330,7 @@ bool memory_setup(memory_interface_functions_t* ifs)
     _interface_functions->random_32_bytes(chunk.fields.encryption_key);
     _interface_functions->random_32_bytes(shared_chunk.fields.encryption_key_split);
 
-    if (!_write_to_address(FLASH_SHARED_DATA_START, shared_chunk.bytes)) {
+    if (!_write_to_address(FLASH_SHARED_DATA_START, 0, shared_chunk.bytes)) {
         return false;
     }
 
@@ -338,8 +344,8 @@ bool memory_cleanup_smarteeprom(void)
 {
     // Erase all SmartEEPROM data chunks.
     for (size_t i = 0; i < SMARTEEPROM_ALLOCATED_BLOCKS; ++i) {
-        uint32_t w_addr = FLASH_SMARTEEPROM_START + i * CHUNK_SIZE;
-        if (!_write_to_address(w_addr, NULL)) {
+        uint32_t w_offset = i * CHUNK_SIZE;
+        if (!_write_to_address(FLASH_SMARTEEPROM_START, w_offset, NULL)) {
             return false;
         }
     }
@@ -639,7 +645,7 @@ bool memory_bootloader_set_flags(auto_enter_t auto_enter, upside_down_t upside_d
     chunk.fields.upside_down = upside_down.value ? 1 : 0;
     // As this operation is quite important to succeed, we try it multiple times.
     for (int i = 0; i < 10; i++) {
-        if (_write_to_address(FLASH_SHARED_DATA_START, chunk.bytes)) {
+        if (_write_to_address(FLASH_SHARED_DATA_START, 0, chunk.bytes)) {
             return true;
         }
 #ifndef TESTING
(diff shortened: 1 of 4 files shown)


src/memory/memory.c
```diff
@@ -150,13 +150,30 @@ static void _clean_chunk(uint8_t** chunk_bytes)
...
+#endif
+
 // chunk must have size CHUNK_SIZE. if chunk is NULL, the chunk is erased,
 // i.e. filled with 0xFF.
-static bool _write_to_address(uint32_t addr, uint8_t* chunk)
+static bool _write_to_address(memory_area_t area, uint32_t addr, uint8_t* chunk)
 {
 #ifdef TESTING
-    return memory_write_to_address_mock(addr, chunk);
+    return memory_write_to_address_mock(area, addr, chunk);
```
benma, line 174: this function can then simply switch on the area to write to different static vars in RAM instead of trying to infer it from the actual address.

benma:
This touches `memory_setup()`, which only fully runs once during factorysetup. The change is trivial so it should be okay, but I'll test this manually.

src/memory/memory.c
```diff
@@ -150,13 +150,30 @@ static void _clean_chunk(uint8_t** chunk_bytes)
...
 #else
+    addr += _address_offset(area);
```
x1ddos, line 176: forgot the comment: makes sense check for some addr overflow here? would want to avoid writing to some potentially sensitive memory chunks, even in theory.

src/memory/memory.h
```diff
@@ -46,6 +46,12 @@ typedef enum {
     MEMORY_ERR_UNKNOWN,
 } memory_result_t;
 
+typedef enum {
+    MEMORY_AREA_SHARED_DATA,
+    MEMORY_AREA_APP_DATA,
+    MEMORY_AREA_SMARTEEPROM,
+} memory_area_t;
```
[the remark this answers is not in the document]
benma: it is also used in mock_memory.h/c.

Will add comments

src/memory/memory.c
benma: Which kind of overflow? You mean overflowing uint32?

x1ddos: yeah, something like this: https://onlinegdb.com/EmYJiQDHa

benma: I renamed `addr` to `offset` and added an overflow check, PTAL. Note that unsigned overflow is well defined in C, so the check can be done after the addition.

src/memory/memory.h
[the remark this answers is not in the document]
x1ddos: ah, right. and forward decl. of enums is forbidden in ISO C. that's a shame. 
hold on. why is this an enum, why not do something like this:

```c
// area must be FLASH_SHARED_DATA_START, FLASH_APPDATA_START or eeprom.
static bool _write_to_address(uint32_t area, uint32_t addr, uint8_t* chunk)
```

it would still be easy to branch logic in tests and no need to declare unused types in headers.

src/memory/memory.c
x1ddos: let's hope ARM's GCC adheres to that definition. fingers crossed.
j/k :) but here's a relevant real story: https://kqueue.org/blog/2013/09/17/cltq/

src/memory/memory.c
```diff
@@ -150,13 +150,37 @@ static void _clean_chunk(uint8_t** chunk_bytes)
...
+}
+#endif
+
 // chunk must have size CHUNK_SIZE. if chunk is NULL, the chunk is erased,
 // i.e. filled with 0xFF.
-static bool _write_to_address(uint32_t addr, uint8_t* chunk)
+//
+// `offset` is a relative address (offset from where the memory is given by `area`) and starts at
+// `0`.
+static bool _write_to_address(memory_area_t area, uint32_t offset, uint8_t* chunk)
```
x1ddos, line 174: So, why not this:

```c
// base must be FLASH_SHARED_DATA_START, FLASH_APPDATA_START or FLASH_SMARTEEPROM_START.
static bool _write_to_address(uint32_t base, uint32_t offset, uint8_t* chunk);

// elsewhere - like it was before:
_write_to_address(FLASH_APPDATA_START, offset, chunk);
```

src/memory/memory.c
```diff
@@ -150,13 +150,30 @@ static void _clean_chunk(uint8_t** chunk_bytes)
...
+    default:
+        Abort("unknown memory area");
+    }
+}
+#endif
+
 // chunk must have size CHUNK_SIZE. if chunk is NULL, the chunk is erased,
 // i.e. filled with 0xFF.
-static bool _write_to_address(uint32_t addr, uint8_t* chunk)
+static bool _write_to_address(memory_area_t area, uint32_t addr, uint8_t* chunk)
```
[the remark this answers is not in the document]
benma: done

src/memory/memory.h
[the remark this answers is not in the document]
benma: well they are not unused lol.

i felt like switching on an enum was clearer to read than switching on an actual address value :shrug: 

i can also remove the enum and switch on actual the address values in the tests. let me try that..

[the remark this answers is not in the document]
benma: i removed the enum and use only the base address now, PTAL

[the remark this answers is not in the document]
benma: (had a rebase error, now should be good)

src/memory/memory.c
benma: I ran into undefined behavior with implicit numeric casts myself in the past :facepalm: grateful that Rust is here and we can abandon C forever :joy:

src/memory/memory.c
```diff
@@ -152,11 +152,17 @@ static void _clean_chunk(uint8_t** chunk_bytes)
...
+// `offset` is a relative address (offset from where the memory starts at `base`) and starts at `0`.
+static bool _write_to_address(uint32_t base, uint32_t offset, uint8_t* chunk)
 {
 #ifdef TESTING
-    return memory_write_to_address_mock(addr, chunk);
+    return memory_write_to_address_mock(base, offset, chunk);
 #else
+    uint32_t addr = base + offset;
+    if (addr < base) {
+        Abort("uint32_t overflow");
```
x1ddos, line 164: Not sure how much it makes sense but here's a stricter version:

```c
// Compute target memory address from base and offset.
// Aborts if base is unrecognized or the computation results in uint32 overflow.
static inline uint32_t _validate_memory_address(uint32_t base, uint32_t offset) {
  switch (base) {
  case FLASH_SHARED_DATA_START:
  case FLASH_APPDATA_START:
  case FLASH_SMARTEEPROM_START:
    uint32_t addr = base + offset;
    if (addr < base) {
      Abort("_validate_memory_address: overflow");
    }
    return addr;
  }
  Abort("_validate_memory_address: unknown base");
}

static bool _write_to_address(uint32_t base, uint32_t offset, uint8_t* chunk) {
#ifdef TESTING
  return memory_write_to_address_mock(base, offset, chunk);
#endif
  uint32_t addr = _validate_memory_address(base, offset);
  ...
}
```

src/memory/memory.h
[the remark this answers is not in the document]
x1ddos: :+1: seems cleaner to me and less code.
I think you might be subconsciously trying to make it look more like rust but C is not rust :)

x1ddos (approved):
LGTM but see `_validate_memory_address` proposal.

PR #796, opened 2021-09-20
sha3: remove external/sha3, use existing Rust impl

benma:
sha3.rs is a copy of sha2.rs and works the same way.

Remove a dependcency and reduce binary size (~3kb):

Before:

du -sb build/bin/firmware*.bin
569868	build/bin/firmware.bin
457544	build/bin/firmware-btc.bin

After:

du -sb build/bin/firmware*.bin
566940	build/bin/firmware.bin
457544	build/bin/firmware-btc.bin


src/rust/bitbox02-rust-c/src/sha3.rs
```diff
@@ -0,0 +1,83 @@
...
+            rust_keccak256_update(ctx, data.as_ptr() as *const _, data.len());
+        }
+        let result = unsafe {
+            let mut result = [0u8; 32];
+            rust_keccak256_finish(&mut ctx, result.as_mut_ptr());
+            result
+        };
+        assert_eq!(
+            result,
+            *b"\xf2\x80\x80\x9c\x57\x2c\xc7\x15\x8c\x09\xc4\x2b\xf1\x2a\xac\x3a\xe3\x7f\x5a\x9b\x1b\x45\xf0\x45\x73\x61\x66\x4a\x76\xa1\x9f\x54"
```
x1ddos, line 79: Confirmed with an alternative impl:

```go
$ cat go.mod
module keccak

go 1.16

require golang.org/x/crypto v0.0.0-20210920023735-84f357641f63

$ cat main.go
package main

import (
	"fmt"

	"golang.org/x/crypto/sha3"
)

func main() {
	h := sha3.NewLegacyKeccak256()
	h.Write([]byte("foo"))
	h.Write([]byte(" abc def xyz"))
	h.Write([]byte(" bar"))
	fmt.Printf("%x\n", h.Sum(nil))
}

$ go run main.go
f280809c572cc7158c09c42bf12aac3ae37f5a9b1b45f0457361664a76a19f54
```

x1ddos (approved):
Nice. LGTM.

PR #798, opened 2021-09-21
Port backup restore to Rust

benma:
Breaking change in python bitbox02: `restore_backup` now raises a `Bitbox02Exception` on failure. Callers can catch a more specific `UserAbortException` when the operation is aborted by user.


py/bitbox02/bitbox02/bitbox02/__init__.py
```diff
@@ -16,7 +16,7 @@
 from __future__ import print_function
 import sys
 
-__version__ = "6.0.0"
+__version__ = "7.0.0"
```
benma, line 19: sorry I did it again before any release. i really don't know how to ensure we will not miss bumping the version correctly if a release is weeks or months away without spending too much time.

to revisit the general concept as I think it makes sense:

- if you introduce a change, bump the relevant part of the version if it is not already bumped since the last release
- in case of the firmware version: bumping early also helps so that the Py and Go libraries can do version checks (e.g `if version > 9.X.0 then use new feature`) before the actual release, so a change can already be integrated and tested into the BitBoxApp

x1ddos: > bump the relevant part of the version if it is not already bumped since the last release

so you'll have to check this every time. very annoying imho

> so that the Py and Go libraries can do version checks 

true but you can do that manually, before building a pre-release. besides, I hope there are automated tests for things like "if version > x then y" in the py and go libraries so there's actually no need to bump firmware version just to try this check out manually.

> how to ensure we will not miss bumping the version correctly if a release is weeks or months away

I still stand by https://github.com/digitalbitbox/bitbox02-firmware/pull/777#issuecomment-908226811: right before a release, check git commit messages and decide on the next version. for example,

```
$ git log $(git describe --tags --abbrev=0)..HEAD --oneline | cut -d' ' -f 2-
sha3: remove external/sha3, use existing Rust impl
Merge remote-tracking branch 'benma/sd-unused'
Merge remote-tracking branch 'benma/sd-lfn'
sd: improve docstring
sd: remove more unused functions
sd: remove unused sd_write and sd_load functions
Merge remote-tracking branch 'benma/inline-fatfs'
Merge branch 'pr/790'
releases: add assertion for v9.7.0-btc-only from cstenglein
Merge remote-tracking branch 'benma/rembase-proto'
Merge remote-tracking branch 'benma/sd_size_t'
Merge remote-tracking branch 'benma/drop-eth-crate'
rust: move ethereum crate functions to bitbox02-rust/hww/api/ethereum
Merge branch 'release'
releases: add v9.7.0
external: rename ff back to fatfs
external: update FatFs to 0.14b and remove submodule
unit-test: include ff.h based on the include directory, not full path
sd: remove _USE_LFN sections
sd: use right types for length_out
py: remove BitBoxBase functionality
messages: remove BitBoxBase
```

Merge commits are annoying. So,

```
$ git log --no-merges $(git describe --tags --abbrev=0)..HEAD --oneline | cut -d' ' -f 2-
sha3: remove external/sha3, use existing Rust impl
sd: improve docstring
sd: remove more unused functions
sd: remove unused sd_write and sd_load functions
releases: add assertion for v9.7.0-btc-only from cstenglein
rust: move ethereum crate functions to bitbox02-rust/hww/api/ethereum
releases: add v9.7.0
external: rename ff back to fatfs
external: update FatFs to 0.14b and remove submodule
unit-test: include ff.h based on the include directory, not full path
sd: remove _USE_LFN sections
sd: use right types for length_out
py: remove BitBoxBase functionality
messages: remove BitBoxBase
```

There can be too many commits, yes. So, one way many projects improve this is to add some kind of a marker to a commit msg and filter on those when generating "release notes". For example, `sd: improve docstring` needs no such marker, it's irrelevant to a next version bump. But, "external: update FatFs to 0.14b" may be relevant, so you add something like `[RELEASE NOTES]` or could even add the actual message for changelog: `[CHANGELOG]: upgrade FatFs from x to 0.14b` which you can easily decide on the spot when creating a commit but no need to think about which version it is.

Of course then you'd have a small and handy shell script to generate changelog messages at any given point between last tag and current HEAD, and decide on the next version bump. Such script can even auto-generate links to relevant commits or just include their hashes at the end/begging of a changelog line.

x1ddos: We came to a compromise offline: bump the version at commit time, unless it already was, and make a CI script that checks for "too many bumps".

x1ddos: BTW here's another example of the same concept but a slightly different scheme: https://www.conventionalcommits.org/en/v1.0.0/ and a tool which can generate changelog and *automatically* decide the next version bump: https://github.com/oknozor/cocogitto

x1ddos: I don't personally like the `feat:` and `fix:` prefixes though.

src/rust/bitbox02/src/backup.rs
```diff
@@ -51,3 +53,50 @@ pub fn check() -> Result<CheckData, Error> {
         err => Err(err),
     }
 }
+
+#[derive(Copy, Clone, Default)]
+struct Backup(bitbox02_sys::Backup);
```
x1ddos, line 58: shame those structs mix protobuf in the fields

src/rust/bitbox02/src/backup.rs
```diff
@@ -51,3 +53,50 @@ pub fn check() -> Result<CheckData, Error> {
...
+struct BackupData(bitbox02_sys::BackupData);
+
+impl zeroize::DefaultIsZeroes for Backup {}
+impl zeroize::DefaultIsZeroes for BackupData {}
+
+pub struct RestoreData {
+    pub seed: zeroize::Zeroizing<Vec<u8>>,
+    // unix timestamp, UTC.
+    pub birthdate: u32,
+    // unix timestamp, UTC.
```
x1ddos, line 70: as you know, unix timestamps do not depend on a timezone. a "unix timestamp, CET" would be very weird :)

src/rust/bitbox02/src/backup.rs
```diff
@@ -51,3 +53,50 @@ pub fn check() -> Result<CheckData, Error> {
...
+    pub timestamp: u32,
+    pub name: String,
+}
+
+pub fn restore_from_directory(dir: &str) -> Result<RestoreData, RestoreError> {
+    let mut backup = zeroize::Zeroizing::new(Backup {
+        ..Default::default()
+    });
+    let mut backup_data = zeroize::Zeroizing::new(BackupData {
```
x1ddos, line 79: nice. I was going to say, isn't `backup_data.0.seed[..backup_data.0.seed_length as _].to_vec()` below leaves a copy behind somewhere in stack memory but this zeroizing erases it on drop. is that correct?

py/bitbox02/bitbox02/bitbox02/__init__.py
x1ddos: btw I'm still not sure how to identify *quickly* when this was last changed. the last py- tag is 5.3.0:

image

```
$ git log --graph --oneline --all -- py/bitbox02/bitbox02/bitbox02/__init__.py
* 9da684dd py/bitbox02: handle non-generic errors in restore_backup
* cc024f9c py/bitbox02: lower version number that was erroneously bumped
* c0a3df11 commander: remove RandomNumber API request
* 6fa00d50 py/bitbox02: add offset of 27 when signing an ETH message
* a6bbc161 protobuf: add reboot purpose
* 1d09c7d4 py/bitbox02: bump version to 5.3.0
* be7dc18e py/bitbox02: bump version to 5.2.0
* 814cb7e2 py/bitbox02: bump version to 5.1.0
* 05c1f711 py/bitbox02: have show_mnemonic raise all exceptions
* da925631 py/bitbox02: bump version to 4.2.0
* ed5e9778 py/bitbox02: bump version to 4.1.0
* 091f5f11 py/bitbox02: bump version to 4.0.1
* b3c80ea6 py/bitbox02: bump min supported firmware version
* b87614d7 py/bitbox02: bump version to 3.0.0
* 9045b585 py: fix up some lint issues and formatting
* d6e96e37 py/bitbox02: remote typing dependency and bump version
* 8e4852c2 (tag: py-bitbox02-2.0.2) py/bitbox02: namespace all protos with package directive
...
```

x1ddos (approved):
LGTM

py/bitbox02/bitbox02/bitbox02/__init__.py
benma: > btw I'm still not sure how to identify quickly when this was last changed. the last py- tag is 5.3.0:

It's because I embarrassed myself again and bumped the version without following my own rule :exploding_head: 

The last release indeed was 5.3.0, so the next major bump is to 6.0.0.... 

When CI check? :D

src/rust/bitbox02/src/backup.rs
benma: I think so.

It is not obvious if `backup_data.0.seed[..backup_data.0.seed_length as _].to_vec()` internally makes copies or if it goes straight to the vector in heap though. I guess I could first make the Vector and then copy into it, making it very explicit.

PR #802, opened 2021-09-27
keystore: fix unit test race condition

benma:
The keystore might have been unlocked from some other test - all are
working on global state.


x1ddos (approved):
LGTM

PR #803, opened 2021-09-27
rust: port ListBackups api call to Rust

benma:
- wrap sd_list_subdir
- use it to list and return backups
- remove C code that becomes unused


src/rust/bitbox02/src/util.rs
```diff
@@ -35,6 +36,19 @@ pub fn str_from_null_terminated(input: &[u8]) -> Result<&str, ()> {
     core::str::from_utf8(&input[0..len]).or(Err(()))
 }
 
+/// Parses a utf-8 string out of a null terminated buffer starting at `ptr`. Returns `Err(())` if
+/// the bytes before the null terminator is invalid UTF8.
+///
+/// # Safety
+/// `ptr` must be not null and be a null terminated string.
+pub unsafe fn str_from_null_terminated_ptr(ptr: *const u8) -> Result<String, ()> {
```
NickeZ, line 44: I think it would be better if we avoided allocation in the util function and deferred that to the caller.

This could return `Result<&str, ()>`, and used like this:

```rust
str_from_null_terminated_ptr(...).map(String::from)
```

Otherwise I suggest we change the name to `string_`.

NickeZ (changes requested):
ut but looks good, left a comment

src/rust/bitbox02/src/util.rs
benma: The problem with this would be that the resulting `&str` is only valid as long as the memory pointed to by `ptr` is untouched, but it is hard (impossible?) to ensure that in the type system. The result could become undefined or garbage by accident without any compiler checks/warnings.

This is why I opted to simply allocate a new string.

NickeZ: I would still opt for this function to return &str. The caller should know if they have to allocate or not. This function is unsafe so it is fine that there are requirements beyond what can be proven with the type system.

benma: Ok, done!

NickeZ (approved):
utACK!

PR #806, opened 2021-10-03
keystore: add keystore_get_ed25519_seed

benma:
For Cardano, there are several BIP32 root key derivation styles:

https://github.com/cardano-foundation/CIPs/blob/6c249ef48f8f5b32efc0ec768fadf4321f3173f2/CIP-0003/CIP-0003.md

- Icarus: deviates from BIP39
- Ledger: derives a new root from the BIP39 seed

We implement Ledger-style derivation, as being incompatible with BIP39
means there would be a separate unlock workflow just for
Cardano (unlock animation of twice the duration after the regular
unlock). Being able to derive all keys in the hardware wallet after
unlocking once is better UX. The downside is that the seed is not
compatible with Cardano hot wallets such as Daedalus, which use
Icarus-style derivation. The upside is that it is compatible with
Ledger devices.

Ledger's reference implementation for comparison and cross-checking:

https://github.com/LedgerHQ/orakolo/blob/master/src/python/orakolo/HDEd25519.py


x1ddos (approved):
LGTM

PR #811, opened 2021-10-14
ci: lint all default python files to avoid errors

benma:
If e.g. only send_message.py is linted without the bitbox02 package
files, import errors can happen.

We still only start linting if any relevant Python file has been
touched, but then we lint all of them.

The reason if sometimes worked before is that the bitbox02 package
installed in the system was used (installed in Docker during Docker
build), but that was linting the wrong package.

### DIFF
--- a/scripts/lint-python
+++ b/scripts/lint-python
@@ -14,13 +14,8 @@ command -v ${PYLINT} >/dev/null 2>&1 || { echo >&2 "${PYLINT} is missing"; exit 
 MYPY=${MYPY:-mypy}
 command -v ${MYPY} >/dev/null 2>&1 || { echo >&2 "${MYPY} is missing"; exit 1; }
 
-# Store arguments as array in ARGS
-ARGS=("$@")
-
-if [ $# -eq 0 ] ; then
-	# Store default files as array in ARGS
-	ARGS=($(find py releases -name '*.py' | grep -v -e generated -e old))
-fi
+# Store files as array in ARGS
+ARGS=($(find py releases -name '*.py' | grep -v -e generated -e old))
 
 # implicit-reexport: `from foo import bar` re-exports (normal Python3 behavior), otherwise mypy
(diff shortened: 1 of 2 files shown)


scripts/lint-python
```diff
@@ -14,13 +14,8 @@ command -v ${PYLINT} >/dev/null 2>&1 || { echo >&2 "${PYLINT} is missing"; exit
...
 
-# Store arguments as array in ARGS
-ARGS=("$@")
-
-if [ $# -eq 0 ] ; then
-	# Store default files as array in ARGS
-	ARGS=($(find py releases -name '*.py' | grep -v -e generated -e old))
-fi
+# Store default files as array in ARGS
+ARGS=($(find py releases -name '*.py' | grep -v -e generated -e old))
```
x1ddos, line 18: > Store default files as array in ARGS

these are not "default" files anymore

x1ddos (approved):

scripts/lint-python
NickeZ: I'm pretty sure the change you made in check-pep8: `./scripts/lint-python ${FILES}` -> `./scripts/lint-python` was enough for CI to start linting all files. `lint-python` supported checking a list of files or all files when no list was given.

benma: Ah yes, that is right. I figured there was no reasonable subset you could want to lint, as you need to supply all the dep files as well within the project. E.g. just providing py/send_message.py to lint-python is invalid, as the bitbox02 dep files are missing.

Could undo it though, Should I? :thinking:

NickeZ: I think my usecase was to run it on individual files while developing, but maybe it doesn't matter if it is quick enough. Probably isn't worthwhile to undo.

benma: I think that worked 'by accident' as it would take the bitbox02 files from the system installation from the Docker image and lints those instead of the current version in py/bitbox02.

PR #812, opened 2021-10-14
main: autoenter into bootloader when Aborting during boot

benma:
To not brick a device permanently with a panic/Abort during boot, we
set the device to boot into the bootloader during the next boot.

We missed one instance: bitbox02_smarteeprom_init() can also Abort()
with a critical error.

### DIFF
--- a/src/firmware.c
+++ b/src/firmware.c
@@ -38,7 +38,15 @@ int main(void)
     screen_splash();
     qtouch_init();
     common_main();
-    bitbox02_smarteeprom_init();
+    if (!bitbox02_smarteeprom_init()) {
+        /*
+         * Incorrect version!
+         * Something has gone terribly wrong.
+         * Maybe reset the whole device?
+         */
+        common_main_bootloader_autoenter();
+        Abort("Unrecognized SmartEEPROM version.");
+    }
     traceln("%s", "Device initialized");
     orientation_screen_blocking();
     idle_workflow_blocking();
--- a/src/common_main.c
+++ b/src/common_main.c
@@ -67,10 +67,7 @@ static bool _setup_wally(void)
     return wally_set_operations(&_ops) == WALLY_OK;
 }
 
-// Go into bootloader if there was an error during startup, so a firmware update can be
-// applied. Otherwise, if there is an Abort() during startup, there would no way to reboot into the
-// bootloader and the device would be bricked.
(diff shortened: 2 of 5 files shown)


src/firmware.c
```diff
@@ -38,7 +38,15 @@ int main(void)
...
     common_main();
-    bitbox02_smarteeprom_init();
+    if (!bitbox02_smarteeprom_init()) {
+        /*
+         * Incorrect version!
+         * Something has gone terribly wrong.
+         * Maybe reset the whole device?
+         */
+        common_main_bootloader_autoenter();
+        Abort("Unrecognized SmartEEPROM version.");
```
x1ddos, line 48: would it be useful for support to show the current version in the message? `Unrecognized version: 0x123.`

benma:
> on a general note, what if we make two `Abort` definitions: one for bootloader, which always calls autoenter before halting, and the other one is normal for when running the main firmware after jump from bootloader?
> 
> the first one is used when compiling bootloader, the normal is for firmware.

Even the one that calls autoenter is called from the firmware. Two definitions could be helpful anyway. Maybe in another PR.

benma:
> Even the one that calls autoenter is called from the firmware. yeah, could also have something like Abort and AbortAutoenter, but i would make only the latter available to the bootloader code. in other words, make it so that the bootloader can never call the regular Abort without autoenter and it should be enforced at compile time.
> […](#)
> On 10/14/21 15:55, benma wrote: > on a general note, what if we make two `Abort` definitions: one for bootloader, which always calls autoenter before halting, and the other one is normal for when running the main firmware after jump from bootloader? > > the first one is used when compiling bootloader, the normal is for firmware. Even the one that calls autoenter is called from the firmware. Two definitions could be helpful anyway. Maybe in another PR.

I don't see when this would be helpful. If the bootloader itself Aborts, it will likely do so again, i.e. bricked device. If autoenter is not set, it will attempt to boot the firmware in the end, but if there was an Abort before, it will not reach that point. If it does and the firmware boots, there was no Abort.

x1ddos (approved):

src/firmware.c
benma: I actually had to go the AbortAutoenter route to make this happen, which seems cleaner anyway. Please check the new version here:

https://github.com/digitalbitbox/bitbox02-firmware/pull/816

PR #819, opened 2021-10-21
Rust Analyzer and clippy without cmake

NickeZ:
Works on my machine, but needs some testing on other developers setups.

If you can run `cmake build-build && cd build-build && make rust-bindgen-includes` it should work. (this means you have nanopb and bindgen working on your development machine).


benma:
Wow I think it is working :heart_eyes: 

Is it possible to skip the tempdir dependency? Pulling in 3k+ files int the vendor lib is bit heavy, maybe there is a simple alternative that does not need that (cc @x1ddos )

src/rust/bitbox02-sys/build.rs
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+        let bitbox02_sys_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+        let cmake_dir = format!("{}/../../../", bitbox02_sys_dir);
+
+        // generate list of includes using CMake
+        let tempdir = TempDir::with_prefix("bitbox02").unwrap();
+        let tempdir = tempdir.path().to_str().unwrap();
+        let _ = Command::new("cmake").arg(&cmake_dir).current_dir(&tempdir).output().unwrap();
+        let _ = Command::new("make").arg("rust-bindgen-includes").current_dir(&tempdir).output().unwrap();
+        let mut includes_file = File::open(format!("{}/src/rust-bindgen.flags", tempdir)).unwrap();
```
benma, line 19: I panic here if I go for example to `src/rust/bitbox02-rust/src/hww/api/set_device_name.rs` and run `cargo test`. Any idea?

NickeZ: I'm not sure exactly what you see, but I see a thing that think is wrong. When you run "cargo test" on a crate it will not compile "test versions" of the dependencies. So basically the tests in "bitbox02-rust" should not depend on a feature called "testing" in the "bitbox02" crate.

NickeZ: One workaround seems to be to use `cargo test --all-features`. But the "testing" feature isn't additive, so it doesn't compile.

https://doc.rust-lang.org/cargo/reference/features.html

> A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features.

benma: Good point, I forgot to enable the testing feature, but as you noticed, it still does not work.

Rust-analyzer seems to be working with this PR though, and also clippy seems to work. Just `cargo test` seems to fail at this line. The error is:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', bitbox02-sys/build.rs:19:91
```

NickeZ: Yeah cargo test is more complicated than clippy and RA since it has to do the final step of linking a working executable together. I'll see how hard it would be to support it

NickeZ:
The reason this depends on #829 or #831 is that this PR crates a cmake build directory in the rust workspace. When running cargo in the rust workspace it gets configured to only use vendored code. So Either prost-build must be vendored or it must be pre-compiled for the cmake build directory to be able to use it when the cmake build directory is in the target dir of the rust workspace.

NickeZ:
rebased on master

benma:
> Run like this: cargo test --features=testing

This feature is not defined in the bitbox02-rust crate.

If I add `testing = ["bitbox02/testing"]` to this crate, and run the above in e.g. `signmsg.rs`, I get:

```
error: failed to run custom build command for `bitbox02-sys v0.1.0 (src/rust/bitbox02-sys)`

Caused by:
  process didn't exit successfully: `src/rust/target/debug/build/bitbox02-sys-9071871369c1a980/build-script-build` (exit status: 101)
  --- stdout
  [ 50%] Generating hww.pb.c, backup.pb.c, common.pb.c, backup_commands.pb.c, bitbox02_system.pb.c, btc.pb.c, cardano.pb.c, eth.pb.c, mnemonic.pb.c, system.pb.c, perform_attestation.pb.c, keystore.pb.c, antiklepto.pb.c, hww.pb.h, backup.pb.h, common.pb.h, backup_commands.pb.h, bitbox02_system.pb.h, btc.pb.h, cardano.pb.h, eth.pb.h, mnemonic.pb.h, system.pb.h, perform_attestation.pb.h, keystore.pb.h, antiklepto.pb.h
  messages/CMakeFiles/generate-protobufs.dir/build.make:110: recipe for target 'messages/hww.pb.c' failed
  CMakeFiles/Makefile2:1217: recipe for target 'messages/CMakeFiles/generate-protobufs.dir/all' failed
  CMakeFiles/Makefile2:1343: recipe for target 'src/CMakeFiles/rust-bindgen-includes.dir/rule' failed
  Makefile:610: recipe for target 'rust-bindgen-includes' failed


           *************************************************************
           *** Could not import the Google protobuf Python libraries ***
           *** Try installing package 'python-protobuf' or similar.  ***
           *************************************************************
      
  Traceback (most recent call last):
    File "../../../../../../../../../tools/nanopb/generator/protoc", line 4, in <module>
      from nanopb_generator import invoke_protoc
    File "tools/nanopb/generator/nanopb_generator.py", line 27, in <module>
      import google.protobuf.text_format as text_format
  ImportError: No module named google.protobuf.text_format
  make[3]: *** [messages/hww.pb.c] Error 1
  make[2]: *** [messages/CMakeFiles/generate-protobufs.dir/all] Error 2
  make[1]: *** [src/CMakeFiles/rust-bindgen-includes.dir/rule] Error 2
  make: *** [rust-bindgen-includes] Error 2


  --- stderr
  thread 'main' panicked at 'explicit panic', bitbox02-sys/build.rs:34:13

```

Rust-analyzer also seems to not work anymore in this crate, while it worked in a previous iteration of this PR :thinking: 

Any idea what is going on?

NickeZ:
Can you run this outside of docker? `cmake build-build && cd build-build && make rust-bindgen-includes`

I added the testing feature to the bitbox02-rust crate and now it works for me.

PR #826, opened 2021-11-03
lockscreen: show device name

benma:
It is useful to see the name of the device before you unlock it. For
example, if you have multiple BitBoxes, it is useful to know which one
is inserted before entering long passwords or passphrases.

The device name is not a secret, and can already be queried in the API
at any time, also before unlock.

Users that dislike this can change the device name in the settings to
something generic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ## Firmware
 
 ### [Unreleased]
+- Display device name on screen before unlock
 - Bitcoin: add support for payment requests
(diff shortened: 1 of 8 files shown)


benma:
Truncation here does not work well because the font is not monospace. I tried a monospace font - it wastes a lot of space and looks a bit ugly. It would be best to either autoscroll the name if it is too long, or to truncate based on actual size, not based on the number of chars. I'll try that later.

benma:
I changed the code to take the actual rendered size of the name into account.

benma:
Rebased.

asi345 (approved):
works as expected. nice addition! tACK

PR #830, opened 2021-11-11
docker: spsed up docker build

NickeZ:
Don't send files to the docker daemon that aren't used in the docker
builds.


benma (approved):
nice!

PR #833, opened 2021-11-17
btc: rename output `hash` to `payload`

benma:
When confirming a tx, the recipient address is reconstructed from the
payload provided in the protobuf `BTCSignOutputRequest` type and
displayed for confirmation.

We previously named the payload `hash` as it always was a hash: either
a pubkey hash in case of P2WPKH (pay to witness pubkey hash), or a
script hash in case of P2SH (pay to scripthash) or P2WSH (pay to
witness script hash).

With the Taproot softfork that recently activated on Bitcoin, the
payload is no longer a hash of something but a 32 byte public key. See
https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki ("Why
is the public key directly included in the output?").

"Payload" as a term is also used in other Bitcoin libraries for the
decoded address input, e.g. in rust-bitcoin. We apply the same rename
to the relevant helper functions in btc_common.h.

This is not a breaking change for the BitBox02 API, as the protobuf
names are not present in the binary serialization, so existing
downstream integrations are not affected.

Projects that update to the changed protobuf file will have to rename
`hash` to `payload` in the code generated from the protobuf files.

Alternatives considered:

- Keep `hash` as is, add another field `pubkey`. This puts as much
  burden on the client libraries as renaming to payload, but adds
  complexity to the firmware code.
- Add `encoded_address`, and let the firmware decode it into the
  payload and validate it. This would be a new way of showing
  recipient addresses, but we'd need to keep supporting the old way
  for backwards compatibility. In some inputs, e.g. PSBT, there is no
  address string to provide, so it does not even make it easier for
  downstream projects in all cases.

It seems that the rename was the easiest solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ### [Unreleased]
 - ListBackups: ported to Rust
 - Cardano: allow transactions with a zero TTL value
+- Protobuf: rename BTCSignOutputRequest.hash to BTCSignOutputRequest.payload
 
 ### 9.8.0 [released 2021-10-21]
 - Multi edition: add Cardano support.
--- a/messages/btc.options
+++ b/messages/btc.options
@@ -26,7 +26,7 @@ BTCSignInputRequest.prevOutHash fixed_length:true max_size:32
 BTCSignInputRequest.keypath max_count:10
 // provide the `has_host_nonce_commitment` field.
 BTCSignInputRequest.host_nonce_commitment proto3_singular_msgs:false
-BTCSignOutputRequest.hash max_size:32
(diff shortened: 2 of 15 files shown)


benma:
CI problem fixed in https://github.com/digitalbitbox/bitbox02-firmware/pull/834

benma:
More Python linting incompatibility problems: https://github.com/digitalbitbox/bitbox02-firmware/pull/838

x1ddos (approved):
LGTM

x1ddos:
sorry, forgot: "... PW2WSH (pay to witness script hash)." in commit msg - did you mean P2WSH?

benma:
> sorry, forgot: "... PW2WSH (pay to witness script hash)." in commit msg - did you mean P2WSH?

yes, fixed!

PR #837, opened 2021-11-18
wally upgrade fixes

benma:
The libwally dep was upgraded in 98822a3db. There was a breaking
change that the previous commit missed.

https://github.com/ElementsProject/libwally-core/blob/e4b62491b6fcc19f53f4cef4919ec11db62f518d/CHANGES.md#version-082

> struct wally_operations has changed to hold the size of the struct and has an additional member to allow overriding the lib secp context used internally. Users must recompile their applications against this version as a result (re-linking or simply upgrading the shared library is insufficient).

This commit updates this to be compatible with the updated libwally.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -14,7 +14,7 @@ string(REPLACE "-mfpu=fpv4-sp-d16" "" MODIFIED_C_FLAGS ${MODIFIED_C_FLAGS_TMP})
 # wally-core
 
 # configure flags for secp256k1 bundled in libwally core, to reduce memory consumption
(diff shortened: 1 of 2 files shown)


external/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ string(REPLACE "-mfpu=fpv4-sp-d16" "" MODIFIED_C_FLAGS ${MODIFIED_C_FLAGS_TMP})
 # wally-core
 
 # configure flags for secp256k1 bundled in libwally core, to reduce memory consumption
-set(LIBWALLY_SECP256k1_FLAGS --with-ecmult-window=2 --with-ecmult-gen-precision=2 --enable-ecmult-static-precomputation --enable-module-ecdsa-s2c)
```
x1ddos, line 17: what about this one? i believe keystore needs it. or is it now enabled by default?

benma: Yeah `enable-module-ecdsa-s2c` is dropped because libwally now enables it

PR #838, opened 2021-11-18
Dockerfile: update Python linting, formatting and protobuf tools

benma:
Linting and mypy compatible anymore since a recent Docker image
rebuild installed a newer Python version.

Updating mypy-protobuf also updates the Python protobuf code generator
to fix typing issues.

Black is updated for good measure.

The changes that follow from this (linting fixes, updated protobuf
generated code, etc.) will come in a future commit which will use the
updated Docker image for CI.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -83,14 +83,12 @@ RUN python3 -m pip install --upgrade --requirement /tmp/requirements.txt
 RUN rm /tmp/requirements.txt
 
 # Python modules for CI
-# TODO: This emits a warning about astroid that is safe to ignore?
-#       https://github.com/PyCQA/astroid/issues/699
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -86,11 +86,11 @@ RUN rm /tmp/requirements.txt
 # TODO: This emits a warning about astroid that is safe to ignore?
 #       https://github.com/PyCQA/astroid/issues/699
```
x1ddos, line 87: is the todo still relevant?

x1ddos (approved):
LGTM

Dockerfile
benma: Didn't see any such warning! Will remove.

PR #845, opened 2021-11-30
util,pukcc: remove dead code

benma:


src/pukcc/pukcc.c
```diff
@@ -194,33 +194,6 @@ static void pukcc_self_test(void)
     }
 }
 
-#if 0
-// Returns 0 on success
-uint8_t pukcc_ecdsa_sign(const uint8_t *private_key,
```
x1ddos, line 199: seems to still be present in the header:

```
$ rg pukcc_ecdsa_sign
src/pukcc/pukcc.h
43:uint8_t pukcc_ecdsa_sign(
```

x1ddos: and also, `pukcc_ecdsa_sign` is called from commented code starting on https://github.com/benma/bitbox02-firmware/blob/3316ce53/src/pukcc/pukcc.c#L230:

```c
#if 0
#include <screen.h>
// Simple unit test from the ref code
// Results displayed on screen
...
```

the example would no longer work.
no idea how useful the example is but anyway, either move the removed `pukcc_ecdsa_sign` into the commented example or delete the example?

x1ddos (approved):
LGTM but see comments

src/pukcc/pukcc.c
benma: removed all :fire:

PR #846, opened 2021-12-02
Send to taproot addresses

benma:


test/unit-test/test_btc_sign.c
```diff
@@ -1452,7 +1482,13 @@ static void _test_antiklepto(void** state)
     valid.antikepto = true;
     _sign(&valid);
 }
-
+static void _test_p2tr_output(void** state)
+{
+    _modification_t invalid = _valid;
+    invalid.p2tr_output = true;
+    invalid.check_sigs = true;
+    _sign(&invalid);
```
x1ddos, line 1490: slightly confused. what does it make initial the `_valid` invalid?

test/unit-test/test_btc_sign.c
```diff
@@ -1104,12 +1122,21 @@ static void _sign(const _modification_t* mod)
...
+                    "\x1f\xc0\xcd\x59\xdb\x40\xef\x41\xd8\xe8\x86\xe5\xcb\x10\x25\xb7\xf5\x92\x95"
+                    "\x84\x34\x2b\xc1\x2f\xb6\x31\xc2\xee\xe7\xf5\x58\x09\x22\x15\x2e\x3d\x18\x43"
+                    "\x9e\x64\x01\x93\xdb\x3e\x2e";
+                assert_memory_equal(next.signature, expected_signature, sizeof(next.signature));
+            } else {
+                const uint8_t expected_signature[64] =
+                    "\x64\xbd\xe8\xf2\xa1\xbf\xaf\x54\x0a\x76\x01\x76\x89\xe9\x51\xa2\xf7\xa5\x7d"
+                    "\x8a\xa7\xd7\x52\xb7\xd8\x82\xc9\x78\x48\x79\xa4\x3c\x7f\x48\xe5\x67\xf2\xbd"
+                    "\x3b\xea\x6b\xf3\x1c\xa9\x9b\xa9\x97\x7c\x0e\x72\xbe\xc0\xe3\x43\x24\x18\x22"
+                    "\xe6\xc9\xd8\xaf\x52\x17\xc8";
```
x1ddos, line 1137: been wondering how to compute these independently, given only the mock seed, inputs, etc?

src/apps/btc/btc_common.h
```diff
@@ -131,6 +131,7 @@ USE_RESULT bool btc_common_address_from_payload(
  * @param[inout] pk_script_len: size of pk_script IN, size of the resulting pk_script OUT.
  */
 USE_RESULT bool btc_common_pkscript_from_payload(
+    const app_btc_coin_params_t* params,
     BTCOutputType output_type,
     const uint8_t* payload,
```
x1ddos, line 136: the doc comment says "payload pubkey hash or script hash" but isn't it just a pubkey for P2TR?

test/unit-test/test_btc_sign.c
benma: one way would be to recreate the transaction with another btc library, but it will be quite time consuming.

i mostly put the signature check here to catch regressions in the code. correctness is checked by reviewing the code and signing testnet/mainnet transactions and broadcasting them successfully.

x1ddos: > broadcasting them successfully

sure, that's a good manual "try it out" test but it's not scalable 

> to catch regressions in the code

but how do you know this literal actually represents correct signature?
i mean, a single bit change in the test input data will produce a different signature which may or may not be valid, depending on other code changes. what i'm actually comparing this at the moment is equivalent to "the sig is not all 0s so it must be good"

src/apps/btc/btc_common.h
benma: yes, true! fixed

test/unit-test/test_btc_sign.c
benma: fixed in a new commit, renaming also all other such vars.

test/unit-test/test_btc_sign.c
benma: > but how do you know this literal actually represents correct signature?

i got no other answer than my previous one:

> correctness is checked by reviewing the code and signing testnet/mainnet transactions and broadcasting them successfully.

Do you have an idea of how to do it differently that is feasible? The other idea I posted, comparing against what a different BTC library produces, might work but it's far from easy.

x1ddos (approved):

PR #850, opened 2021-12-15
keystore: add some `const`s

benma:


src/keystore.c
```diff
@@ -551,7 +551,7 @@ bool keystore_get_bip39_word(uint16_t idx, char** word_out)
 // uncompressed_out must be 65 bytes.
 static bool _compressed_to_uncompressed(const uint8_t* pubkey_bytes, uint8_t* uncompressed_out)
 {
-    secp256k1_context* ctx = wally_get_secp_context();
+    const secp256k1_context* ctx = wally_get_secp_context();
```
x1ddos, line 554: since you're here, also add them in tests, for example https://github.com/digitalbitbox/bitbox02-firmware/blob/0c5f6fca/test/unit-test/test_keystore.c#L203 ?

benma: good catch. done

PR #851, opened 2021-12-15
keystore: add seed_len to mock_state()

benma:
So we can test with seeds of 16 bytes too (12 words), not just 32
bytes (24 words).

### DIFF
--- a/src/keystore.c
+++ b/src/keystore.c
@@ -44,16 +44,16 @@ static bool _is_unlocked_bip39 = false;
 static uint8_t _retained_bip39_seed[64] = {0};
 
 #ifdef TESTING
-void mock_state(const uint8_t* retained_seed, const uint8_t* retained_bip39_seed)
+void mock_state(const uint8_t* seed, size_t seed_len, const uint8_t* bip39_seed)
 {
-    _is_unlocked_device = retained_seed != NULL;
-    if (retained_seed != NULL) {
-        _seed_length = KEYSTORE_MAX_SEED_LENGTH;
-        memcpy(_retained_seed, retained_seed, sizeof(_retained_seed));
+    _is_unlocked_device = seed != NULL;
+    if (seed != NULL) {
+        _seed_length = seed_len;
+        memcpy(_retained_seed, seed, seed_len);
     }
-    _is_unlocked_bip39 = retained_bip39_seed != NULL;
-    if (retained_bip39_seed != NULL) {
-        memcpy(_retained_bip39_seed, retained_bip39_seed, sizeof(_retained_bip39_seed));
+    _is_unlocked_bip39 = bip39_seed != NULL;
+    if (bip39_seed != NULL) {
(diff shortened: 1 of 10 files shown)


src/keystore.c
```diff
@@ -44,16 +44,16 @@ static bool _is_unlocked_bip39 = false;
 static uint8_t _retained_bip39_seed[64] = {0};
 
 #ifdef TESTING
-void mock_state(const uint8_t* retained_seed, const uint8_t* retained_bip39_seed)
+void mock_state(const uint8_t* seed, size_t seed_len, const uint8_t* bip39_seed)
```
x1ddos, line 47: how about renaming this to a `mock_keystore_state`, or better yet a `mock_keystore_unlocked_state`. last time i was browsing tests, it was quite disorienting due to naming.

src/keystore.c
```diff
@@ -44,16 +44,16 @@ static bool _is_unlocked_bip39 = false;
...
 {
-    _is_unlocked_device = retained_seed != NULL;
-    if (retained_seed != NULL) {
-        _seed_length = KEYSTORE_MAX_SEED_LENGTH;
-        memcpy(_retained_seed, retained_seed, sizeof(_retained_seed));
+    _is_unlocked_device = seed != NULL;
+    if (seed != NULL) {
+        _seed_length = seed_len;
```
x1ddos, line 51: i guess i missed it in a previous review but the doc comment for `_seed_length` says:

```c
// seed length defaults to 32 bytes, but we could accept smaller seeds in the future.
static uint8_t _seed_length = KEYSTORE_MAX_SEED_LENGTH;
```

the future is now :)

test/unit-test/test_btc_sign.c
```diff
@@ -417,7 +417,8 @@ static bool _stream_prevtx(
 static void _sign(const _modification_t* mod)
 {
     // Need keystore to derive change and input scripts
-    mock_state(mod->seeded ? _mock_seed : NULL, mod->seeded ? _mock_bip39_seed : NULL);
+    mock_state(
+        mod->seeded ? _mock_seed : NULL, sizeof(_mock_seed), mod->seeded ? _mock_bip39_seed : NULL);
```
x1ddos, line 421: maybe a more readable version would be something like this:

```c
mock_state(
    mod->seeded ? _mock_seed : NULL,
    sizeof(_mock_seed),
    mod->seeded ? _mock_bip39_seed : NULL)
```

so it visually separates the args - was already harder than it could be to read before imho

x1ddos (approved):
LGTM but see comments

src/keystore.c
benma: Done in a separate commit

src/keystore.c
benma: Done - updated the comment and defaulting to 0 because why put any other meaningless value there.

test/unit-test/test_btc_sign.c
benma: clang-format decided how to format. disabling clang-format for these lines is not worth it imo.

PR #854, opened 2021-12-24
keystore: add keystore_secp256k1_schnorr_bip86_[pubkey|sign] funcs

benma:
Needed to to generate taproot receive addresses and sign taproot
inputs in transactions.

A taproot output contains a 32 byte x-only pubkey (see BIP-340), which
may or may not be tweaked with the root of a merkle tree containing
execution scripts as leafs.

In a normal single-sig with no extra scripts, BIP-86 specifies:

> BIP 341 states: "If the spending conditions do not require a script
path, the output key should commit to an unspendable script path
instead of having no script path. This can be achieved by computing
the output key point as Q = P + int(hashTapTweak(bytes(P)))G.".

https://github.com/bitcoin/bips/blob/4c6389f8431f677847b115538a47ce8c826c6be8/bip-0086.mediawiki

This commit adds keystore functions to get such tweaked
pubkey (original pubkey tweaked with the hash of the original pubkey)
and to sign for it.

We put this functionality directly into keystore.c instead of
performing the tweaking outside of it so we only have to derive the
pubkey at the keypath once. The alternative would have been to add a
32 bytes tweak argument to the pubkey/sign functions and have the
caller compute the correct tweak, or to perform the tweaking at the
callsite altogether.

### DIFF
--- a/src/keystore.c
+++ b/src/keystore.c
@@ -25,7 +25,10 @@
 #include "securechip/securechip.h"
 #include "util.h"
 
+#include <rust/rust.h>
 #include <secp256k1_ecdsa_s2c.h>
+#include <secp256k1_extrakeys.h>
+#include <secp256k1_schnorrsig.h>
 
 // This number of KDF iterations on the 2nd kdf slot when stretching the device
 // password.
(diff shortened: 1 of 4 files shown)


src/keystore.c
```diff
@@ -794,3 +797,83 @@ USE_RESULT bool keystore_encode_xpub_at_keypath(
...
+    uint8_t* sig64_out)
+{
+    secp256k1_keypair __attribute__((__cleanup__(_cleanup_keypair))) keypair = {0};
+    if (!_schnorr_bip86_keypair(keypath, keypath_len, &keypair)) {
+        return false;
+    }
+    const secp256k1_context* ctx = wally_get_secp_context();
+    uint8_t aux_rand[32] = {0};
+    random_32_bytes(aux_rand);
+    return secp256k1_schnorrsig_sign(ctx, sig64_out, msg32, &keypair, aux_rand) == 1;
```
x1ddos, line 878: i would check with `secp256k1_schnorrsig_verify` before returning.
i saw it's verified in `_test_keystore_secp256k1_schnorr_bip86_sign` but tests is one thing, runtime is different: bit flips, glitching, etc.

x1ddos (approved):

src/keystore.c
benma: :+1: done

PR #855, opened 2021-12-28
unit-test/test_btc_sign: use a real bip39 seed with known preimage

benma:
The test used a fake bip39 seed that can't be derived from any
mnemonic. This commit changes it to use a seed from a real mnemonic.

When we run the signing code from Rust, it will make it easier to
verify that the Rust version produces the same results (signatures).

The signatures were simply updated with the results of the test run
itself, which is okay because no production code changed.


x1ddos (approved):
LGTM

PR #857, opened 2021-12-30
app/btc/sign: move btc tx sign state machine to Rust async funcs

benma:
The Bitcoin signing algorithm in btc_sign.c involves keeping track of
a big state machine, processing the transaction in a streaming
fashion.

The `next_request()` async primitive is perfect to stream the
transaction from the host in an async function, offloading the state
machine to the Rust compiler, improving readability.

Instead of migrating the whole implementation to Rust in one shot, at
first we only move streaming the transaction to Rust, and call the
existing signing functions of C. `_wrapper()` functions are added for
each of them so they can be called from Rust.

The C implementation is largely unchanged in this commit. In the
future, we can get rid of all the state handling in the C
implementation.

In btc_sign.c, the only changed needed was make _sign_input_pass1()
and _sign_input_pass2() work indepently without relying on
`app_btc_sign_input(). `app_btc_sign_input()` is now only used in C,
while in Rust the two passes are called individually.

Over time, these functions can move to Rust as well.

To ensure that everything works, we port some tests from
test_btc_sign.c:

- _test_btc_sign_init()
- `_sign()` with the default `_modification_t`, so the unmodified
  transaction. We compare the produced signature at the end, which is
  a copy paste from test_btc_sign.c.


benma:
Pushed another commit to delete most of the state handling in btc_sign.c

src/rust/bitbox02-rust/Cargo.toml
```diff
@@ -41,6 +41,7 @@ blake2 = { version = "0.9.2", default-features = false, optional = true }
 minicbor = { version = "0.11.3", default-features = false, features = ["alloc"], optional = true }
 ed25519-dalek = { version = "1.0.1", default-features = false, features = ["u32_backend"], optional = true }
 digest = "0.9.0"
+lazy_static = { version = "1.4.0", optional = true }
```
benma, line 44: should go to devdependencies as it is only used in tests

src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -431,3 +431,31 @@ pub fn screen_stack_pop_all() {
         bitbox02_sys::ui_screen_stack_pop_all();
     }
 }
+
+pub fn progress_create<'a>(title: &str) -> Component<'a> {
+    let component = unsafe {
+        bitbox02_sys::progress_create(
+            crate::str_to_cstr_force!(title, MAX_LABEL_SIZE).as_ptr(), // copied in C,
```
x1ddos, line 438: trailing comma typo in the comment

x1ddos (approved):
LGTM. Nice work!

PR #858, opened 2021-12-30
eth/sign: increase size limit for tx data

### LINKED ISSUE #843: Make Opensea transactions work
The size of the contract data in Opensea transactions is larger than the BitBox02 accepts, leading to an 'Invalid Input' error.

The size limit should be increased to accommodate opensea transactions.

Ideally, Opensea transactions would be parsed and displayed in a human readable format instead. The current way to display the raw contract data is only safe for expert users that know how to verify it, but that is not feasible for huge contract data fields like with Opensea.

benma:
Fixes #843

We previously limited the max. size of contract data to 1024 bytes.

Some smart contract invocations require much larger data,
e.g. transactions involving the Opensea contract. Example tx with a
large data field (~2400 bytes):

https://etherscan.io/tx/0x2db260bf407a246e3bbf57692168238b55b8518ef72c8288e68547e418cf0e08

Ideally, Opensea transactions would be parsed and displayed in a human
readable format instead. The current way to display the raw contract
data is only safe for expert users that know how to verify it, but
that is not feasible for huge contract data fields like with Opensea.

For now, we don't change the data display logic but simply increase
the size limit so transcations with larger data fields don't fail. The
data field is shown in raw hex format, truncated at 320
bytes (`MAX_LABEL_SIZE`) as before this commit.

Our USB packet size is 7609 bytes (`USB_DATA_MAX_LEN`). We bump the
max data limit to a new arbitrary large number, but below the max
packet size so a transaction can still be transmitted to the
device. The other ETH transaction fields combined are less than 200
bytes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 - Cardano: add support for sending tokens
 - Protobuf: rename BTCSignOutputRequest.hash to BTCSignOutputRequest.payload
 - Ethereum: add 807 ERC-20 tokens
+- Ethereum: increase size limit for transaction data to accommodate Opensea transactions
 
 ### 9.8.0 [released 2021-10-21]
 - Multi edition: add Cardano support.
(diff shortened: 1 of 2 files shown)


x1ddos:
> Our USB packet size is 7609 bytes

i was about to ask, what about msg signing, but it looks ok and actually allows for more than USB max :)

```rust
/// Process a sign message request.
///
/// The result contains a 65 byte signature. The first 64 bytes are the secp256k1 signature in
/// compact format (R and S values), and the last byte is the recoverable id (recid).
pub async fn process(request: &pb::EthSignMessageRequest) -> Result<Response, Error> {
    if request.msg.len() > 9999 {
        return Err(Error::InvalidInput);
    }
```

x1ddos (approved):
LGTM although i didn't try this manually. passing tests look sufficient.

PR #859, opened 2022-01-05
api/restore: confirm restore and backup name/ID

benma:
The ID is shown to disambiguate in case there are multiple backups
with the same name.

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/restore.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/restore.rs
@@ -20,17 +20,13 @@ use pb::response::Response;
 use crate::workflow::{confirm, mnemonic, password, status, unlock};
 
 pub async fn from_file(request: &pb::RestoreBackupRequest) -> Result<Response, Error> {
-    #[cfg(feature = "app-u2f")]
-    {
-        let datetime_string =
-            bitbox02::format_datetime(request.timestamp, request.timezone_offset, false);
-        let params = confirm::Params {
-            title: "Is now?",
-            body: &datetime_string,
-            ..Default::default()
-        };
-        confirm::confirm(&params).await?;
-    }
+    // This is a separate screen because 'Restore backup?' does not fit in the title field.
+    confirm::confirm(&confirm::Params {
+        body: "Restore backup?",
+        accept_is_nextarrow: true,
+        ..Default::default()
+    })
+    .await?;
 
     let data = match bitbox02::backup::restore_from_directory(&request.id) {
         Ok(data) => data,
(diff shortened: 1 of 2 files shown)


src/rust/bitbox02-rust/src/hww/api/restore.rs
```diff
@@ -20,25 +20,42 @@ use pb::response::Response;
...
+    let data = match bitbox02::backup::restore_from_directory(&request.id) {
+        Ok(data) => data,
+        Err(_) => {
+            status::status("Could not\nrestore backup", false).await;
+            return Err(Error::Generic);
+        }
+    };
+
+    confirm::confirm(&confirm::Params {
+        body: &format!("Name: {}. ID: {}", &data.name, &request.id),
```
x1ddos, line 39: i wonder why not combine this with the first one?

```rust
let data = match bitbox02::backup::restore_from_directory(&request.id) { ... }
confirm::confirm(&confirm::Params {
  title: "Restore backup?",
  body: &format!("Name: {}\nID: {}", &data.name, &request.id),
  ...
})
.await?;
```

it just seems like a faster and possibly less confusing UX. less confusing because if i see a blank question "restore backup?" i might assume that's it, there's no more confirmations and it'll start restoring.

src/rust/bitbox02-rust/src/hww/api/restore.rs
```diff
@@ -20,25 +20,42 @@ use pb::response::Response;
...
 
 pub async fn from_file(request: &pb::RestoreBackupRequest) -> Result<Response, Error> {
+    confirm::confirm(&confirm::Params {
+        body: "Restore backup?",
+        accept_is_nextarrow: true,
+        ..Default::default()
+    })
+    .await?;
+
+    let data = match bitbox02::backup::restore_from_directory(&request.id) {
```
x1ddos, line 30: this seems like a confusing naming. doesn't it simply reads from SD card and optionally tries to recover a corrupt backup? if so a better name is something like "read backup from directory".

also, not sure what the comment at https://github.com/digitalbitbox/bitbox02-firmware/blob/badbe0269/src/backup/restore.c#L245-L247 refers to:

```c
    // if we found a file that matches the checksum, overwrite corrupt files.
    // do majority-vote error correction if all files are corrupt. Read file bit by bit and update
    // the bits that are diverging.
    return RESTORE_OK;
```

x1ddos: i meant, i know what it referes to - to the loop above it - but it's just a bit confusing to read the comment below what it's referring to. i was trying to figure out what `restore_from_directory` does exactly.

benma: out of scope for this PR.

src/rust/bitbox02-rust/src/hww/api/restore.rs
x1ddos: anyway, wouldn't it be nicer to insert a newline? might as well remove `Name:`. i've just tried, looks better imho:

```rust
body: &format!("{}\nID: {}", &data.name, &request.id),
```

x1ddos (approved):
LGTM

benma:
> anyway, wouldn't it be nicer to insert a newline? might as well remove Name:. i've just tried, looks better imho:

It is scrollable, as the name can be long. Scrolling two lines at once would be possible, but also something we've not done before :thinking:

PR #861, opened 2022-01-10
releases: add v9.9.0

benma:


releases/firmware-btc-v9.9.0/assertion.txt
```diff
@@ -0,0 +1,8 @@
+By signing this file, the signer confirms that the firmware binary built from:
+
+git tag firmware-btc-only/v9.9.0
+git commit hash ea520f5156a64d3fc57eee8cffae505e674f646e
```
x1ddos, line 4: this is what i meant with release tags. if i `git clone -b firmware-btc-only/v9.9.0`, i'll end up at this state in the changelog:

image

not ideal imho

x1ddos (approved):
:+1: but let's try tagging a commit like this one with a release tag next time

PR #865, opened 2022-01-24
releases: add v9.9.0 assertion confirmations from cstenglein

cstenglein:
Thank you @benma for the fix. Build was working now :)


benma (approved):
thanks a lot for your contribution!

PR #869, opened 2022-02-04
cardano/signtx: disallow duplicate token keys in an output

benma:
Token key being: (policy id, asset name).

A duplicate token in an output would be confirmed multiple times by
the user. The resulting tx is valid at the protocol level, but only
the last token with the same key is used when there are duplicates.


x1ddos (approved):
LGTM

PR #884, opened 2022-03-02
eth/params: add arbitrum and optimism

benma:
cc @Tomasvrba


Tomasvrba (approved):
tACK
Tested Optimism, Arbitrum, Fantom and Avalanche with Rabby

PR #887, opened 2022-03-07
dockerenv.sh: set non-unique GID flag in groupadd

douglasbakkum:
This is a bugfix for running the Docker script on Mac. The script was aborting with a "groupadd: GID '20' already exists" error message even though the script comment specifically says it will use an already existing GID.

Solved by adding the `-o` option to `groupadd`. 
```
-o, --non-unique
    This option permits to add a group with a non-unique GID. 
```

### DIFF
--- a/scripts/dockerenv.sh
+++ b/scripts/dockerenv.sh
@@ -58,7 +58,7 @@ dockerdev () {
 
     # Use same user/group id as on the host, so that files are not created as root in the mounted
     # volume.
-    docker exec -it "$CONTAINER_NAME" groupadd -g "$(id -g)" dockergroup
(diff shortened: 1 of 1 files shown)


benma (approved):

benma:
> even though the script comment specifically says it will use an already existing GID.

As on the host - you are seeing an error because the group already exists in the Docker container.

What does `id -g` output for you on your host?

benma:
Regardless, I think the fix is okay, since we refer to this group by name later on, and it doesn't hurt if the group already exists.

douglasbakkum:
On the host `id -g` gives `20`

PR #895, opened 2022-03-16
ui/oled: Add support for new display SSD1312

benma:


NickeZ (approved):
utACK, this obviously needs a lot of testing, both of the bootloader and firmware.

PR #903, opened 2022-04-26
api/rootfingerprint: add ability to display fingerprint

benma:
Let's the user verify the root fingerprint on the device. Advanced
users can use this to verify that the wallet is loaded/unlocked that
they expect.


benma:
I'm considering to abandon this and instead show descriptors on the device, see https://github.com/digitalbitbox/bitbox-wallet-app/pull/1692#discussion_r862725886

PR #917, opened 2022-05-28
Dockerfile: update Python linting deps and rust toolchain to 1.61.0

benma:


benma:
@NickeZ after updating to Rust 1.61.0, there are errors like

```
../../lib/libc-unit-tests_rust_c.a(bitbox02_rust_c-d1aed3e812268234.bitbox02_rust_c.2b72ae46-cgu.0.rcgu.o):(.data.DW.ref.rust_eh_personality[DW.ref.rust_eh_personality]+0x0): undefined reference to `rust_eh_personality'
```

But only when running `make unit-test` (build the firmware etc seems to work fine).

The compatibility notes seem relevant here:

https://github.com/rust-lang/rust/blob/19abca1172ae10d5c084b6c3013d92680f92dd8b/RELEASES.md#compatibility-notes

image

Any idea where the [+whole-archive](https://doc.rust-lang.org/stable/rustc/command-line-arguments.html#linking-modifiers-whole-archive) attribute should be set? Should it even be set?

PR #918, opened 2022-05-29
Compile rust src locally

NickeZ:
Another way to solve the eh_personality thing is to compile core and alloc with panic_abort. This ensures that there is no unwinding code in those libraries either. Reduces code size by about 11k if I interpret the data correctly:

```
old:
   text	   data	    bss	    dec	    hex	filename
 695056	  19052	 210800	 924908	  e1cec	firmware.elf

new:
  text	   data	    bss	    dec	    hex	filename
 686408	  19052	 210800	 916260	  dfb24	firmware.elf
```

One caveat is that `cargo vendor`, the tool that fetches dependencies and vendors them cannot fetch rust-src depedencies yet. So I've added a script to manually fetch those dependencies, and they should only change when we bump rust compiler version.


benma:
In the vendor folder, there is now cfg-if (v1.0.0) and cfg_if (v0.1.10).

Is that potentially a problem? I thought dash and underscore was treated the same by Rust in folder names.

NickeZ:
Interesting, I guess that is what happens when you try to vendor things manually..

benma:
Closing in favor of https://github.com/digitalbitbox/bitbox02-firmware/pull/919 - according to the [workgroup](https://github.com/rust-lang/wg-cargo-std-aware/tree/81765f0eb744b9c47840c16f43a32c9f61fd7f0c), this feature is discouraged from use in production. We can hopefully use it once it becomes (more) stable.

PR #927, opened 2022-06-12
external: update the ctaes dependency

benma:
Before, the submodule pointed to
https://github.com/digitalbitbox/ctaes/tree/cbc, which was a re-host
of https://github.com/bitcoin-core/ctaes/pull/14 before it was merged.

It has since been
merged. https://github.com/digitalbitbox/ctaes/tree/master is now at
8012b062ea4931f10cc2fd2075fddc3782a57ee4, which is up to date with
upstream https://github.com/bitcoin-core/ctaes. The submodule is
updated to this commit revision.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -243,6 +243,5 @@ target_include_directories(base32 SYSTEM PUBLIC base32)
 
 add_library(ctaes
   ctaes/ctaes.c
-  ctaes/ctaes-cbc.c
 )
 target_include_directories(ctaes SYSTEM PUBLIC ctaes)
--- a/src/cipher/cipher.c
+++ b/src/cipher/cipher.c
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #include "cipher.h"
-#include <ctaes-cbc.h>
+#include <ctaes.h>
 
 #include <stdlib.h>
 #include <string.h>


benma:
@Beerosagos the exhaustive tests here make me confident that this version bump does not introduce any bugs:

https://github.com/digitalbitbox/bitbox02-firmware/blob/6e80882a691ba5835aadcd3fcde1305068517cd1/test/unit-test/test_cipher.c#L38

Beerosagos (approved):
Tested with:
- unit tests
- rust unit test
- valgrind on unit  tests

Also flashed the compiled firmware on a dev bb02 and everything seems to work smoothly. I think this is ok!

Beerosagos:
Btw, I think that the `run-valgrind-on-unit-tests` target in the Makefile could be out of date. The tests seem to be compiled in `build-build` folder, instead of `build` and there are some mismatches in the name of the tests.

Should we update it or am I missing something?

benma:
> Tested with:
> 
>     * unit tests
> 
>     * rust unit test

Thanks. Fyi those also run in the CI, so no need to manually run them if the CI is green.

> Btw, I think that the run-valgrind-on-unit-tests target in the Makefile could be out of date. The tests seem to be compiled in build-build folder, instead of build and there are some mismatches in the name of the tests.

> Should we update it or am I missing something?

Good question. We run an address sanitizer on the unit tests, which seems to be some sort of alternative. 

You said you ran valgrind, so you fixed up the target and it works? If so, you could make a PR and also add it to CI.

Beerosagos:
> You said you ran valgrind, so you fixed up the target and it works? If so, you could make a PR and also add it to CI.
Yes, I basically run a `find -exec` on `build-build` folder to run each test found under valgrind. The tests passed smoothly. I'll make a dedicated PR

PR #932, opened 2022-06-15
releases: add v9.12.0

benma:


benma:
Closing as it is not reproducible anymore using `./build.sh firmware/v9.12.0 "make -j firmware"`, which re-builds the Docker image (that is only two weeks old).

benma:
Reopening as the reproducible build works with this fix to the build script. https://github.com/digitalbitbox/bitbox02-firmware/pull/933/files

There was nothing wrong with the original build.

PR #933, opened 2022-06-15
releases/build.sh: fix reproducible builds that failed due to git

benma:
A breaking change happened in git which disallows git commands in a
repo owned by someone else:

```
ailed to run `['git', 'tag', '--list', '--sort=taggerdate', '--points-at', 'HEAD', '--format=%(objecttype) %(if)%(contents:signature)%(then)Y%(else)N%(end) %(refname:strip=3)', 'firmware-btc-only/*']`
stderr: fatal: unsafe repository ('/bb02' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /bb02
```

Related to this fix:

https://github.blog/2022-04-12-git-security-vulnerability-announced/

We need git commands to get the firmware version, which is part of the
binary. With the error above, the version wat not included in the
binary properly, resuling in a wrong build.

We implement the fix suggested in the error msg. The alternative is
run the docker image with the same UID as the host UID, but I am not
sure if this works as well on all host systems.

### DIFF
--- a/releases/build.sh
+++ b/releases/build.sh
@@ -16,8 +16,11 @@ docker build --pull --force-rm --no-cache -t bitbox02-firmware .
 
 # Build the firmware. The inlined python package install can be
 # removed after v4.1.0, but is necessary for v4.1.0 as it is missing
-# in the Dockerfile of that release.
-docker run -it --rm --volume `pwd`:/bb02 bitbox02-firmware bash -c "apt-get update && apt-get install -y python && cd /bb02 && $2"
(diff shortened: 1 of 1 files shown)


Beerosagos (approved):
tACK!

benma:
Merged manually with an added comment, see https://github.com/digitalbitbox/bitbox02-firmware/commit/3c0c1dd9eb81f0ada8bb06c00407f1d7a0e83030

PR #934, opened 2022-06-20
cmake: error if get_version fails

benma:
It would have prevented an issue where reproducible builds failed due
to the version falling back to being 'v0.0.0' instead of exiting with
an error (see e17cfccef30c54cd2646613041b630f823786c87).


Beerosagos (approved):
tACK!

PR #936, opened 2022-06-25
Add pubkey and assertion for btc-v9.12.0 from sutterseba

sutterseba:
I thought I'd leave my signature here

Keep up the great work! ❤️


benma (approved):
Thanks a lot, we appreciate it!

PR #937, opened 2022-06-28
bitcoin/signtx: restore comment lost in Rust conversion

benma:
A comment was not copied over from C from this commit, when porting
the locktime/rbf checks to Rust:
082dd1ae523fbf868dd178b3e857bf4d15ab70be

The dialog is not shown if locktime is 0, as a beginner user should
not be confrontend with RBF, which is enabled by default in the
BitBoxApp.


Beerosagos (approved):
utACK

PR #941, opened 2022-07-04
backup: initialize backup stack vars to zero

benma:
backup_check and backup_create compute the directoy name of the backup
based on the seed field in the backup struct, using
`_get_directory_name`, always using 32 bytes, even if the seed is
shorter (e.g. 16 bytes).

Without explicitly zero-intializing these structs and the seed is
shorter than 32 bytes, it could contain garbage instead of zero as
padding. If that happened, backup creation and restore would work, but
the backup check could fail, as the directory could not be
found (directory name computed based on a seed with garbage padding).

Somehow it seems that the structs are zero initialized in
practice (confirmed by manual testing), but that seems to be a lucky
coincidence. This commit makes this explicit.

### DIFF
--- a/src/backup/backup.c
+++ b/src/backup/backup.c
@@ -85,8 +85,8 @@ static void _get_file_name(uint32_t backup_create_timestamp, char* file_name, ui
  */
 static backup_error_t _check_backup(uint8_t* output, size_t output_length, const Backup* backup)
 {
-    Backup __attribute__((__cleanup__(backup_cleanup_backup))) backup_check;
-    BackupData __attribute__((__cleanup__(backup_cleanup_backup_data))) backup_data_check;
+    Backup __attribute__((__cleanup__(backup_cleanup_backup))) backup_check = {0};
+    BackupData __attribute__((__cleanup__(backup_cleanup_backup_data))) backup_data_check = {0};
     restore_error_t res =
         restore_from_buffer(output, output_length, &backup_check, &backup_data_check);
     if (res != RESTORE_OK) {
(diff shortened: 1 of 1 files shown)


src/backup/backup.c
```diff
@@ -103,8 +103,8 @@ static backup_error_t _check_backup(uint8_t* output, size_t output_length, const
  */
 backup_error_t backup_create(uint32_t backup_create_timestamp, uint32_t seed_birthdate_timestamp)
 {
-    Backup __attribute__((__cleanup__(backup_cleanup_backup))) backup;
-    BackupData __attribute__((__cleanup__(backup_cleanup_backup_data))) backup_data;
+    Backup __attribute__((__cleanup__(backup_cleanup_backup))) backup = {0};
+    BackupData __attribute__((__cleanup__(backup_cleanup_backup_data))) backup_data = {0};
```
benma, line 107: @Beerosagos any idea why thse are zero initialized even without the ` = {0}`? Afaik it should be undefined and contain garbage, but it seems it's always zero anyway.

If it contained garbage, then backup_fill() would only fill the first e.g. 16 bytes of the 32 bytes-sized `seed` field if the created seed is 16 bytes (e.g. if you restore from a 12 word mnemonic phrase and then create an backup onto SD), but it seems that didn't happen.

benma:
> tACK!
> 
> I am not sure if the initialization of backup related structures is actually needed, since it is zeroed in `backup_fill()`, but I don't see any drawback in merging this PR. @benma wdyt?

You just answered this:

> Somehow it seems that the structs are zero initialized in practice

In this case there is actually no potential bug and I can close this PR. Thanks!

PR #945, opened 2022-07-05
api: allow ShowMnemonic for initialization (allows skipping SD backup)

benma:
The normal initialization flow is unseeded -> seeded -> initialized:

SetPassword api call to create the seed and set the password -> seeded.

CreateBackup to create the backup onto a microSD card -> initialized.

ShowMnemonic was only allowed in the initalized state, i.e. after a
microSD card backup was made.

This commit allows skipping microSD card backups in favor of a
mnemonic backup, by allowing the ShowMnemonic API call also in the
seeded (not yet initialized) state.

The api call is modified to set the initialized state, and to not ask
for the user password when not initialized, which is the same way the
CreateBackup call works (which is also allowed after the
initialization).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ## Firmware
 
 ### [Unreleased]
+- Allow skipping the microSD card backup in favor of backing up using the recovery words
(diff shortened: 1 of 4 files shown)


benma:
Unit tests and a functional test to follow.

PR #948, opened 2022-07-07
py/requirements: remove pyserial

benma:
Added in 9bcae185cc1054bad068d2cc73c56a0d2d2ba063 for the BitBoxBase,
which was removed from the Python libs in
e66818cb2f6a64fce216b7698e8a9443778e27cb. Forgot to remove this dep
here.


Beerosagos (approved):
tACK!

PR #950, opened 2022-07-12
backup: fix backup length field to 0 

benma:
This field is kind of dead code in that it is not used or checked or
interpreted by the firmware when loading/checking a backup. It is part
of the backup checksum though, so we can't get rid of it completely,
otherwise loading existing backups would fail checksum verification.

Having a non-zero value here is a maintenance burden (i.e. when
porting this code to Rust), so we just stop using it.

The `test_fixture` unit test in hww/api//backup.rs ensures that
backups created before this change continue to load correctly.


Beerosagos (approved):
tACK!

PR #951, opened 2022-07-13
py/bitbox02/communication: update protobuf generated files

Beerosagos:
Recent HWI update requires protobuf files generated with protoc =>
3.19.0. This update regenerates the files with protoc 3.21.1


benma:
The protobuf files are generated by executing `cd py && make` (inside Docker), and this should continue working so it's easy to generate the files when changing the protobuf messages.

Please update the Dockerfile accordingly :)

Beerosagos:
> The protobuf files are generated by executing `cd py && make` (inside Docker), and this should continue working so it's easy to generate the files when changing the protobuf messages.
> 
> Please update the Dockerfile accordingly :)

Didn't think about that! :| Now should be ok, thanks.

benma:
> Recent HWI update requires protobuf files generated with protoc

Are you sure it's HWI? The issue originally was this:

```
f this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 ```

So it was about the protobuf package being newer than 3.20. [HWI's dep list uses 3.19.1](https://github.com/bitcoin-core/HWI/blob/0e6934e9fe05d5bb35528dd65dd8746c97daac38/poetry.lock#L344). So maybe it was something else that upgraded your protobuf package :thinking: 

Anyway, would be good to include the above in the commit msg :)

**Comment by benma** [comment]
We have a problem maybe. When I now run send_message.py, I get:

```
from google.protobuf.internal import builder as _builder
```
My local protobuf installation is at 3.19.4:

image

Upgrading it fixes it, but that still means it's broken for [protobuf <3.20](https://stackoverflow.com/questions/71759248/importerror-cannot-import-name-builder-from-google-protobuf-internal).

At the very least we need to then bump https://github.com/digitalbitbox/bitbox02-firmware/blob/57d62a98cf52fd4d700551083e925019e5eb16f3/py/bitbox02/setup.py#L78. Maybe that's sufficient :shrug:

**Comment by benma** [comment]
Please also update py/bitbox02/CHANGELOG.md under `6.1.0` (to be released).

**Comment by Beerosagos** [comment]
> We have a problem maybe. When I now run send_message.py, I get:
> 
> ```
> from google.protobuf.internal import builder as _builder
> ```
> 
> My local protobuf installation is at 3.19.4:
> 
> image
> 
> Upgrading it fixes it, but that still means it's broken for [protobuf <3.20](https://stackoverflow.com/questions/71759248/importerror-cannot-import-name-builder-from-google-protobuf-internal).
> 
> At the very least we need to then bump
> 
> https://github.com/digitalbitbox/bitbox02-firmware/blob/57d62a98cf52fd4d700551083e925019e5eb16f3/py/bitbox02/setup.py#L78
> . Maybe that's sufficient shrug

Thanks for pointing this out, didn't notice it! Yes, updating `setup.py` seems to be sufficient, and a good way to avoid compatibility issues.

**Comment by Beerosagos** [comment]
I am not really understanding why the CI failed. Made some tests on my local, but can't reproduce the error.

**Comment by benma** [comment]
> I am not really understanding why the CI failed. Made some tests on my local, but can't reproduce the error.

I think pylint is just very bad with the member checks with protobuf, they are probably false positives. That's why you see so many `# pylint: disable=no-member` scattered :( I don't know why more of them are triggered now... Maybe there is a way to disable that warning altogether, otherwise try putting `# pylint: disable=no-member` and seeing if that helps.

You can lint it locally with `./scripts/docker_exec.sh ./scripts/lint-python `.

**Comment by Beerosagos** [comment]
> > I am not really understanding why the CI failed. Made some tests on my local, but can't reproduce the error.
> 
> I think pylint is just very bad with the member checks with protobuf, they are probably false positives. That's why you see so many `# pylint: disable=no-member` scattered :( I don't know why more of them are triggered now... Maybe there is a way to disable that warning altogether, otherwise try putting `# pylint: disable=no-member` and seeing if that helps.
> 
> You can lint it locally with `./scripts/docker_exec.sh ./scripts/lint-python `.

Thanks for the explanation. I disabled no-member checks for protobuf generated modules.

**Review by benma — APPROVED** [review]
Could you update https://github.com/digitalbitbox/bitbox02-firmware/blob//.ci/run-container-ci#L28 to version 26? I just pushed the updated Docker image with the new protoc there.

I also accidentally overwrote the `25` tag on Docker hub - will re-build and re-push that one :see_no_evil:

File: .pylintrc [hunk 1]
```diff
@@ -9,4 +9,8 @@ docstring-min-length=10
 [MESSAGES CONTROL]
 # Black/pylint mismatch on code formatting
 disable=bad-continuation, fixme, too-few-public-methods, duplicate-code, line-too-long, consider-using-f-string, too-many-locals, raise-missing-from, super-init-not-called, too-many-lines
-good-names=i,j,tx
\ No newline at end of file
+good-names=i,j,tx
+
+[TYPECHECK]
+# Ignore protobuf generated modules for no-member checks
+ignored-modules=bitbox02.communication.generated.btc_pb2,bitbox02.communication.generated.cardano_pb2,bitbox02.communication.generated.common_pb2,bitbox02.communication.generated.eth_pb2,bitbox02.communication.generated.hww_pb2
```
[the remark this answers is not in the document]
benma: Please update the commit msg to explain this and that mypy does the member checks instead.

py/bitbox02/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 6.1.0
 - Add `eth_sign_typed_msg()`
+- Update brotobuf dependency to >= 3.21
```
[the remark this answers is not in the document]
Beerosagos: LOL :sweat_smile:

.pylintrc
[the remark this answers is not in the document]
Beerosagos: Done! :+1:

py/bitbox02/setup.py
```diff
@@ -75,7 +75,7 @@ def find_version() -> str:
     install_requires=[
         "hidapi>=0.7.99.post21",
         "noiseprotocol>=0.3",
-        "protobuf>=3.7",
+        "protobuf>=3.21",
```
SomberNight, line 78: AFAICT protobuf 3.20.x is able to parse both the old and the new format
(while <3.20 can only parse the old format, and >=3.21 can only parse the new format)

Hence, would it not be better to set this to `"protobuf>=3.20"` ?

benma: Interesting, how did you figure this out? The [release notes](https://github.com/protocolbuffers/protobuf/releases/) don't exactly give it away.

Anyway, we're happy to make a patch release that relaxes the constraint.

Fyi, protobuf>=3.20 requires Python 3.7, see also: https://github.com/bitcoin-core/HWI/issues/626

SomberNight: > Interesting, how did you figure this out? The [release notes](https://github.com/protocolbuffers/protobuf/releases/) don't exactly give it away.

When you have new protobuf and old generated files, the runtime error says
` 1. Downgrade the protobuf package to 3.20.x or lower.`.
I manually tested a bunch of protobuf versions from PyPI around  that boundary, hoping some can parse both and to learn the behaviour.
I had tried reading the release notes first but could not make sense of them either :)

> Fyi, protobuf>=3.20 requires Python 3.7, see also: https://github.com/bitcoin-core/HWI/issues/626

That's fine. (in Electrum, we already bumped the minimum to Python 3.8)

benma: 6.1.1 is out, please give it a try.

PR #953, opened 2022-07-18
Makefile: fix valgrind execution of unit tests which was broken

Beerosagos:
Execution of target `run-valgrind-on-unit-tests` looked for tests in the wrong
folder and needed to be wrapped in a single script.


benma (approved):
Also run it in CI? Would be good to see if it catches any errors. Also without it it is bound to break again in the future.

PR #963, opened 2022-08-08
Add rust-analyzer to dev dockerfile

NickeZ:
This setup works for me, unless I'm being confused. I configure vim to run `ra.sh` and I start vim in the `src/rust` directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -130,8 +130,10 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | CARGO_HOME=/opt/
 RUN rustup target add thumbv7em-none-eabi
 RUN rustup component add rustfmt
 RUN rustup component add clippy
+RUN rustup component add rust-src
 RUN CARGO_HOME=/opt/cargo cargo install cbindgen --version 0.23.0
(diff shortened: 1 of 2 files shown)


benma:
Even in the bitbox02-rust crate, which depends on the C stuff?

I've been meaning to look into https://github.com/digitalbitbox/bitbox02-firmware/pull/819 again - if that one works, is this one still needed?

benma:
A bit confused, why does rust-analyzer need to run inside Docker, is there anything special inside the Docker container? Could it also run on the host?

Asking because adding it to the Dockerfile has at least these downsides:

- building the image takes a lot longer, detrimental to reproducible builds
- updating rust-analyzer should be done quite frequently I imagine

NickeZ:
rust-analyzer is technically a compiler, so it requires almost everything that you have in the docker container. The only way to make it work outside the container is to install most of those dependencies on your dev machine.

benma:
Most of the deps means just the Rust toolchain or more? Cause installing the Rust stuff is not a big barrier imo. 

I tried getting this PR to work in my environment, no luck yet. Will try again.

NickeZ:
I don't remember exactly, but you need to be able to generate C headers, rust bindings. If you look at the other PR you can see that there are some changes so that the build scripts can detect if they are run under cmake or not.

PR #969, opened 2022-08-16
port commander_states.c to Rust

benma:
Another reboot unit test was called to test rebooting in the seeded
state. The two existing tests `test_reboot_when_unitialized` and
`test_reboot_when_initialized` covered two out of the three states.

The reboot command is the most important by far, as it is necessary
for firmware upgrades.


Beerosagos (approved):
tACK :+1:

PR #977, opened 2022-08-18
i2c_ecc,random: remove duplicate definitions

benma:
These symbols are already defined in driver_init.c.

The current GCC version doesn't complain, I discovered this when
trying to build with an updated
GCC (arm-gnu-toolchain-11.3.rel1-x86_64) and getting errors about
these redefinitions.

The reason that the current GCC version (GCC 8) didn't complain and
the updated one (GCC 11) did complain is that the default for the
-fcommon flag changed from on to off in GCC 10.

See:
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85678
- https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html (-fcommon)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,6 +320,10 @@ string(APPEND CMAKE_C_FLAGS " -Wno-cast-function-type")
 # Hardening
 string(APPEND CMAKE_C_FLAGS " -fstack-protector-all")
 
+# Disallow duplicate definitions, which is the default since GCC
+# 10. It was not default in gcc-arm-none-eabi-8-2018-q4.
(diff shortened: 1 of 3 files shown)


benma:
@NickeZ any idea why the old compiler would not complain about this but the new would? duplicate definition should not work on the old compiler (the one currently in use) :thinking:

benma:
> @NickeZ any idea why the old compiler would not complain about this but the new would? duplicate definition should not work on the old compiler (the one currently in use) thinking

Found the reason and documented it in the commit msg.

Beerosagos (approved):
tACK!

PR #989, opened 2022-09-07
api/setpassword,restore: return UserAbort on user abort

benma:
Before, Generic would be returned no matter if the user did not enter the same password twice or if they aborted on the `Your password has fewer than 4 characters.\nContinue?` screen.

Now that we handle the user abort, we can also allow to abort during password entry itself.


Beerosagos (approved):
tACK!

PR #991, opened 2022-09-29
api: display error type on screen if create/restore fails

benma:
If there was a securechip or memory error, creating or restoring a
wallet would show and return a generic looking error. If e.g. the
securechip is the problem, we want to see that, so the user can better
report the problem.

We might also return a specific error code to the host, but for now
displaying the error will do (same as do when unlocking a device fails).

### DIFF
--- a/src/keystore.c
+++ b/src/keystore.c
@@ -140,7 +140,7 @@ static keystore_error_t _stretch_password(
         if (securechip_result_out != NULL) {
             *securechip_result_out = securechip_result;
         }
-        return KEYSTORE_ERR_SC_KDF;
+        return KEYSTORE_ERR_SECURECHIP;
     }
     // Second KDF does not use the counter and we call it multiple times.
     for (int i = 0; i < KDF_NUM_ITERATIONS; i++) {
@@ -150,7 +150,7 @@ static keystore_error_t _stretch_password(
             if (securechip_result_out != NULL) {
                 *securechip_result_out = securechip_result;
             }
-            return KEYSTORE_ERR_SC_KDF;
+            return KEYSTORE_ERR_SECURECHIP;
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/keystore.c
```diff
@@ -234,58 +234,59 @@ static bool _verify_seed(
...
     const char* password)
 {
     if (memory_is_initialized()) {
-        return false;
+        return KEYSTORE_ERR_MEMORY;
     }
     keystore_lock();
     if (!_validate_seed_length(seed_length)) {
-        return false;
+        return KEYSTORE_ERR_MEMORY;
```
Beerosagos, line 247: Shouldn't this be `KEYSTORE_ERR_SEEDSIZE`?

src/keystore.c
```diff
@@ -234,58 +234,59 @@ static bool _verify_seed(
...
     }
     if (encrypted_seed_len > 255) { // sanity check, can't happen
         Abort("keystore_encrypt_and_store_seed");
     }
     if (!memory_set_encrypted_seed_and_hmac(encrypted_seed, encrypted_seed_len)) {
-        return false;
+        return KEYSTORE_ERR_MEMORY;
     }
     if (!_verify_seed(password, seed, seed_length)) {
         if (!memory_reset_hww()) {
```
Beerosagos, line 275: The `if` statement can be removed?

src/rust/bitbox02/src/keystore.rs
```diff
@@ -39,12 +39,28 @@ pub enum Error {
...
     Salt,
     Hash,
     SeedSize,
-    Unknown,
+    Encrypt,
+}
+
+impl core::convert::From<keystore_error_t> for Error {
+    fn from(error: keystore_error_t) -> Self {
+        match error {
```
Beerosagos, line 52: `KEYSTORE_ERR_INCORRECT_PASSWORD` missing?

src/keystore.c
benma: Good catch, fixed!

PR #992, opened 2022-10-01
dev: reduce docker layers & combine install commands

cstenglein:
This reduces the layers of the DockerFile from _41_ to _28_.

First, I merged multiple `apt-get` & `pip` installs to one. Currently there are two `apt-get` installs (one with basic tools like `wget` & one with the rest).

Second, I merged multiple `ENV` & `RUN` statements into one but was conservative in that regard to not merge all statements, but just adjacent ones.

This should reduce the install time for the image a little bit & reduce the image size.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,83 +17,16 @@
 FROM ubuntu:22.04
 ENV DEBIAN_FRONTEND noninteractive
 
-RUN apt-get update && apt-get upgrade -y && apt-get install -y wget nano rsync curl gnupg2 jq unzip bzip2
-
-# for clang-*-15, see https://apt.llvm.org/
-RUN echo "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-15 main" >> /etc/apt/sources.list && \
-    echo "deb-src http://apt.llvm.org/jammy/ llvm-toolchain-jammy-15 main" >> /etc/apt/sources.list && \
-    wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
-
-# Install gcc8-arm-none-eabi
-RUN mkdir ~/Downloads &&\
-    cd ~/Downloads &&\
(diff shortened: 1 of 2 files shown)


benma:
Thanks.

> This should reduce the install time for the image a little bit & reduce the image size.


Can you quantify that? What's the difference in size and build time roughly?

cstenglein:
Hi @benma 

> Can you quantify that? What's the difference in size and build time roughly?

Sure, but it's not perfect since a big part of the image build time are the downloads.

Used `time` with 2 builds on my local machine to build the current master vs this branch. Note that this does only the parts of the `build.sh` script until `docker build` since `docker run` won't work without a firmware version.

master old
=> real 19m18.606s

Size 4.8GB

master
=> real 17m13.468s

Size 4.79GB

So size decrease is negligible, but build time decreased a bit (~10%).

benma:
Have you considered moving ,most steps into a installation script.sh, hence collapsing most of the layers into one?

the BitBoxApp does that:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/Dockerfile#L21-L22

cstenglein:
Hi @benma 
Sry for the late answer!

> Have you considered moving most steps into a installation script, hence collapsing most of the layers into one?

Cool idea, will try that in the coming days, thanks!

benma:
No worries, and thanks! Note that there were a lot of changes to the Dockerfile that were merged already. When you try, make sure to use the latest Dockerfile from `master`.

cstenglein:
@benma should work now please test :)

squashed the commits as well.

benma (approved):
Great work!

Could you add a small rational in the commit message body for why this is useful? As you mentioned before, a bit of reduced space and ~10% faster build time.

cstenglein:
Hi @benma 

Thank you for the review! Currently I'm very limited in my time so I'm converting this to a draft and reviewing it at a later time. Hope that's fine.

benma:
@cstenglein I had actually already approved your PR, the rest were just small improvements. If you want we can merge it as is and I add these small changes on top. Let me know.

PR #998, opened 2022-10-25
bitcoin/signtx: allow displaying BTC values as sats

benma:
This allows the BitBoxApp and other wallets which are able to display BTC values as satoshis to direct the BitBox02 to also display them as satoshis during signing.


Beerosagos (approved):
tACK :+1:

PR #1006, opened 2022-11-02
bitcoin: warn if fee is higher than 10%

benma:
This is to prevent accidents with fees. False positives can happen, but the information might still be useful and the user can confirm the high fee.


thisconnect:
please mention in the changelog 😇

benma:
> please mention in the changelog innocent

how could i forget :) done!

Beerosagos (approved):
LGTM!

PR #1007, opened 2022-11-02
cardano: warn if fee is higher than 10%

benma:


src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction.rs
```diff
@@ -257,10 +257,11 @@ async fn _process(request: &pb::CardanoSignTransactionRequest) -> Result<Respons
         })
         .await?;
     } else {
+        let fee_percentage: f64 = 100. * (request.fee as f64) / (total as f64);
```
benma, line 260: need to check for null division

benma: nevermind, it is already in the total!=0 else branch

Beerosagos (approved):
tACK!

PR #1008, opened 2022-11-02
README: update securechip model and other small fixes

benma:


thisconnect (changes requested):
👍 

I noticed that the svg file is a little bit terrible

could you use this cleaned up (14.7 KB ->  2.5 KB) version:

bb02-firmware-arch-new

thisconnect (approved):
LGTM

PR #1010, opened 2022-11-03
eth-testnets: add goerli and remove ropsten and rinkeby

Beerosagos:
Ropsten and Rinkeby ETH testnets are currently deprecated. This commit removes them from the code and introduces support to the Goerly testnet.

Since the erc20 test token has been removed from the wallet-app in 4347b1, it is now removed here too.

### DIFF
--- a/src/apps/eth/eth_params.c
+++ b/src/apps/eth/eth_params.c
@@ -16,15 +16,6 @@
 
 #include <util.h>
 
-static const app_eth_erc20_params_t _ropsten_erc20_params[] = {
-    {
-        .unit = "TEST",
-        .contract_address =
-            "\x2f\x45\xb6\xfb\x2f\x28\xa7\x3f\x11\x04\x00\x38\x6d\xa3\x10\x44\xb2\xe9\x53\xd4",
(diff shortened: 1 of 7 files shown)


src/apps/eth/eth_params.c
```diff
@@ -13904,22 +13895,13 @@ const app_eth_erc20_params_t* app_eth_erc20_params_get(
...
-        break;
-    case 3:
-        erc20_params = _ropsten_erc20_params;
-        len = sizeof(_ropsten_erc20_params) / sizeof(app_eth_erc20_params_t);
-        break;
-    default:
+    if (chain_id != 1) {
         return NULL;
     }
+    len = sizeof(_ethereum_erc20_params) / sizeof(app_eth_erc20_params_t);
```
benma, line 13902: `size_t len = ...`, no need to declare the var separately above.

Could also just inline it in the for-statement.

benma (approved):

src/apps/eth/eth_params.c
Beerosagos: Good point, thx! Done

PR #1012, opened 2022-11-07
ethereum: warn if fee is higher than 10%

benma:
Only for standard transactions, where the fee and amount are of the
same unit. In ERC20-txs, the fees are in ETH while the token is of a
different unit.


Beerosagos (approved):
tACK with one question: the `num-bigint` update is required by the feature or could be split in a separate PR?

benma:
> tACK with one question: the `num-bigint` update is required by the feature or could be split in a separate PR?

Could actually be split.

benma:
Accidentally merged a rebased version of this, which means the commits are in mater, but the PR was not closed. Closing manually.

PR #1015, opened 2022-11-08
restore/mnemonic: fix cancel dialog

benma:
d9029fff23ceb490f412f5d800840810d4336493 introduced a regression - it added the 'Do you really want to cancel?' dialog to all trinary input dialogs. The purpose was that the user could cancel the password entry when setting a password. The cancel dialog was accidentally introduced also when cancelling the 'restore from mnemonic' workflow, which had its own more complex cancellation logic where the user can choose to edit the previous word instead. This was accidentally buried behind the 'Do you really want to cancel' dialog.

This commit fixes it by only asking the user if they want to cancel when entering a password only, not for all trinary inputs.


Beerosagos (approved):
tACK!

PR #1020, opened 2022-11-18
releases: assertion v9.13.1 from Stadicus

Stadicus:
Add signatures of assertion after successful reproducible build process


benma (approved):
Thanks!

PR #1039, opened 2023-01-30
scripts: make dockerenv.sh and docker_exec.sh podman-compatible

benma:
podman is a (almost) drop-in replacment to docker that runs rootless and without a daemon. Since it runs rootless, it maps the host user to the root user inside the container. This clashes with the tricks used for docker to mount the volume with the same host user id and group id. This script removes these docker tricks when using podman.


scripts/docker_exec.sh
```diff
@@ -29,7 +36,14 @@ function docker_exec {
     PIDFILE=/tmp/docker-exec-$$
     shift
     trap 'kill $PID; docker_cleanup $IMAGE $PIDFILE' TERM INT
-    docker exec --user=dockeruser --workdir="$DIR/.." -i $IMAGE bash -c "echo \"\$\$\" > $PIDFILE; eval $*" &
+
+    USERFLAG=""
+    if [ "$RUNTIME" = "docker" ] ; then
+        # Only needed for docker - see the comment below.
```
benma, line 42: note to self - remove/rephrase

Beerosagos (approved):
tACK!

PR #1042, opened 2023-02-22
Subtract min sensor value per slider

douglasbakkum:
This improves touch position accuracy in noisy setups.

Touch position is calculated with a weighted average of the sensor readings. If properly calibrated, sensors on the opposite end of a finger touch would be zero and thus make no contribution to the weighted average. If the baseline sensor readings are elevated, the sensors on the opposite edge do contribute to the weighted average, making a positional artifact (i.e. the position is more central than it should be in reality). This artifact is higher when the finger is a bit distant while approaching and lower/negligible when the finger is fully touching the device. This can cause the position to move enough to enter "slide" mode and disable "tap" events being emitted.

### DIFF
--- a/src/qtouch/qtouch.c
+++ b/src/qtouch/qtouch.c
@@ -1,4 +1,3 @@
-
 /*============================================================================
 Filename : touch.c
 Project : QTouch Modular Library
@@ -570,7 +569,7 @@ uint16_t qtouch_get_sensor_node_signal_filtered(uint16_t sensor_node)
 {
     // Filter the sensor signal.
     //
-    // Smooth it out and saturate it so that values never go beyond 50.
+    // Smooth it out and saturate it so that values never go beyond DEF_SENSOR_CEILING.
     // This helps to mitigate 'jumpy' channels that exist at higher sensor readings when
     // in noisy environments.
     //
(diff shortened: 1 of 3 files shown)


src/qtouch/qtouch.c
```diff
@@ -630,21 +629,40 @@ void qtouch_process_scroller_positions(void)
...
         uint16_t weighted_sum = 0;
         uint16_t sensor_location[DEF_SCROLLER_NUM_CHANNELS] = {
             1, // Offset by `1` because a `0` location cannot be weight-averaged
             DEF_SCROLLER_RESOLUTION / 3,
             DEF_SCROLLER_RESOLUTION / 3 * 2,
             DEF_SCROLLER_RESOLUTION};
 
+        for (i = 0; i < DEF_SCROLLER_NUM_CHANNELS; i++) {
+            uint16_t value;
+            value = qtouch_get_sensor_node_signal_filtered(
```
benma, line 642: store these in a `uint16_t filtered_values[DEF_SCROLLER_NUM_CHANNELS]` array, then use the array values in the two loops below. this improves code clarity a lot, as one does not have to reason about what happens if the two loops read different values.

benma (changes requested):
tested, seems to work great. well done!

benma:
@douglasbakkum to format the code (fix the CI issue), run `./scripts/docker_exec.sh ./scripts/format` (outside dockerdev) or `./scripts/format` inside dockerdev.

src/qtouch/qtouch.c
douglasbakkum: ack. now done and CI passes

(used "readings" instead of "values" in order to match the naming in other variables)

douglasbakkum:
@benma 
👍  added

benma (changes requested):
Sorry about the nit but please squash the commits - commits should be atomic - think of it in terms of being able to revert a change: you wouldn't want to revert the touch fix but keep the changelog entry, etc.

douglasbakkum:
ack. happy for the nit and explanation. now squashed and pushed

PR #1046, opened 2023-03-09
scripts/container: add container selection for dockerinit

Beerosagos:
In cb16b81 and b9a05b1 we added podman support for containers, but `make dockerinit` was still fixed on docker, causing ad error when building and executing a new image with podman installed. This updates the Makefile to use podman if available also when building the image.


benma (approved):
thanks!

PR #1048, opened 2023-03-21
eth: fix protobuf generated file to remove Goerli

benma:
In 8b454550f4961b2d7380885fbd45526d8c5d42bd, Goerli was added, but
accidentally in the generated file, not in the source file. The
generated file is only refreshed if the source files change, which is
why the CI did not catch it.

Since we changed to using chainID instead of ETHCoin in
519a58b17e4692f5d2e6a5793f4c8fc09a312e2c, we remove it again from the
protobuf file.

Similarly, the generated files were not regenerated when updating Prost to
0.11. in 412e3400d627ab0b994df5a92f34c95478535ae5, which adds more
changes to the generated file, like formatting and the
`as_str_name`/`from_str_name` methods.

### DIFF
--- a/messages/eth.proto
+++ b/messages/eth.proto
@@ -18,9 +18,12 @@ package shiftcrypto.bitbox02;
 import "common.proto";
 import "antiklepto.proto";
 
+// Kept for backwards compatibility. Use chain_id instead, introduced in v9.10.0.
 enum ETHCoin {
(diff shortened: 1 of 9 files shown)


benma:
fyi @Beerosagos  - noticed this after the protobuf generated file was regenerated, removing GoerliEth again.

PR #1063, opened 2023-04-21
Update deps

benma:
Some deps are duplicated by this as different deps use different versions. Opening this as a draft for now.


benma:
Various other PRs have updated or removed these deps in the meantime.

PR #1071, opened 2023-05-25
Add Assertion and script check sign edit proposal

Inusual8:


releases/describe_signed_firmware.py
```diff
@@ -45,38 +45,47 @@ def main() -> int:
...
 
-    print(
-        "The following information assumes the provided binary was signed correctly; "
-        "the signatures are not being verified."
-    )
-    if magic == MAGIC_MULTI:
-        print("This is a Multi-edition firmware.")
-    elif magic == MAGIC_BTCONLY:
-        print("This is a Bitcoin-only edition firmware.")
+    if firmware_is_signed:
```
benma, line 52: the file is named `describe_signed_firmware.py` and its purpose is to check signed firmwares not unsigned ones.

To keep its focus narrow and this change simple, it would be better to exit with an error if the magic is not recognized, i.e. instead of printing

> Unrecognized firmware edition; magic = c8f10120

and then continuing, it could just stop after that error. Maybe expand this message with something like "This tool does not support unsigned firmware. You can use `sha256sum` directly on unsigned firmware binaries to get its hash" or similar.

benma (changes requested):
Thanks, left a comment. Please also keep the assertions and the code changes in separate commits (or even separate PRs), as they are unrelated to each other.

benma:
@Inusual8 reminder about this PR. Do you want to continue with it?

benma:
Closing due to inactivity.

PR #1079, opened 2023-06-12
rust/mnemonic: return CancelError when user aborts mnemonic setup

Beerosagos:


src/rust/bitbox02-rust/src/workflow/mnemonic.rs
```diff
@@ -125,7 +125,10 @@ pub async fn get() -> Result<zeroize::Zeroizing<String>, ()> {
     status(&format!("Enter {} words", num_words), true).await;
 
     // Provide all bip39 words to restrict the keyboard entry.
-    let bip39_wordlist = bitbox02::keystore::get_bip39_wordlist()?;
+    let bip39_wordlist = match bitbox02::keystore::get_bip39_wordlist() {
+        Ok(wordlist) => wordlist,
+        Err(_) => panic!("Could not retrieve bip39 wordlist"),
+    };
```
benma, line 131: This is correct but some tips:

- It is best to avoid wildcards `_`. By using explicit pattern matches, the compiler will tell you if `get_bip39_wordlist()` ever changes its error type, which might be relevant for propagation.

Also this whole block can be abbreviated with `get_bip39_wordlist().expect("Could not retrieve bip39 wordlist")`.

There is also `.unwrap()` which panics with a generic message. Maybe check which variant results in a smaller binary size.

benma (approved):
nice!

src/rust/bitbox02-rust/src/workflow/mnemonic.rs
Beerosagos: Thanks, `.unwrap()` resulted in the smaller binary :+1:

PR #1084, opened 2023-06-18
Add assertion for v9.14.0 btc-only from cstenglein; update cstenglein pubkey

cstenglein:
Haven't been active the last few months, sorry for that, so missed a few versions.

Updated my pubkey which extended the expiration date +  verified the current release (9.14.0)


benma:
Thank you very much as always!

PR #1087, opened 2023-06-20
api/bitcoin: allow exporting xpub at any keypath with confirmation

benma:
So far we prohibited unusual keypaths (e.g. non bip-44/49/84 keypaths).

When using descriptors/miniscript, there are no specific keypaths to whitelist - the user confirms the keypath at policy registration time. We need to allow exporting the xpub at any keypath so the a descriptor wallet app can construct a policy using BitBox02 xpubs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ## Firmware
 
 ### [Unreleased]
+- Allow exporting the xpub at any keypath after user confirmation
 
 ### 9.14.1
 - Improved security: keep seed encrypted in RAM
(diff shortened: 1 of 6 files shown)


Beerosagos (approved):
tACK!

Beerosagos:
@benma One nit I just noticed: Since the warning message is longer than the available display size, it doesn't look like a warning at first sight, as it starts with `unusual keypath..`. Would be nice to have it start with `warning: `, (maybe also in capital letters?) to allow a distracted user immediately understand that he need to pay attention to what is going on.

benma:
> @benma One nit I just noticed: Since the warning message is longer than the available display size, it doesn't look like a warning at first sight, as it starts with `unusual keypath..`. Would be nice to have it start with `warning: `, (maybe also in capital letters?) to allow a distracted user immediately understand that he need to pay attention to what is going on.

good point! fixed, also for a similar message in ethereum.

PR #1092, opened 2023-06-30
Miniscript: registration & signing

benma:
There is a example CLI that works with this PR which:

- creates a miniscript policy
- registers it on the BB02
- generates a receive address
- signs a tx
- verifies that all steps are correct

https://github.com/digitalbitbox/bitbox02-api-go/pull/82

Run with `go run cmd/miniscript/main.go`.

The first commit here is the same as https://github.com/digitalbitbox/bitbox02-firmware/pull/1087


Beerosagos (approved):
tACK!

PR #1100, opened 2023-07-17
eth: allow signing transactions with empty data and 0 value

Tomasvrba:
This is especially useful for signing mock transactions on all supported mainnets and testnets on different EVM networks, such as testing WalletConnect integration


benma (approved):
utACK. please add a CHANGELOG entry

PR #1102, opened 2023-07-29
rust: remove bs58 dep in favor of bitcoin::base58

benma:
Duplicate implementation of the same thing. Saves ~1.3kB in binary space.


Beerosagos (approved):
utACK

PR #1108, opened 2023-08-15
btc/miniscript: activate mainnet, add address unit tests

benma:
The unit test addresses were verified with an independent miniscript implementation at https://github.com/benma/miniscript-go.


src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
```diff
@@ -38,7 +38,7 @@ const MAX_KEYS: usize = 20;
 
 // We only support Bitcoin testnet for now.
```
Beerosagos, line 39: Pls update the comment :)

Beerosagos (approved):
utACK with minor nit :)

PR #1109, opened 2023-08-16
rust/miniscript: make deps optional, enable by features

benma:
The bitcoin/miniscript deps don't need to be compiled for all targets.

xpubcache was marked as unused in a warning when compiling
factory-setup, so we guard this also by the features only enabled in
the firmware editions.


Beerosagos (approved):
LGTM

PR #1110, opened 2023-08-16
update from C99 to C11

benma:
This change does not modify the resulting images at all, the hashes stay the same. It allows however to use C11 features in future commits, such as static
asserts (https://en.cppreference.com/w/c/language/_Static_assert), which can help improve some compile time checks.


Beerosagos (approved):
utACK

PR #1116, opened 2023-08-23
bitcoin/signtx: stricter index checking

benma:
The send_message.py example had an invalid index which should have resulted in an InvalidInput error. This fixes the example and the check.


Beerosagos (approved):
LGTM

PR #1119, opened 2023-08-28
py/bitbox02: parse eth eip-712 uints as hex

benma:
Some apps use EIP-712 typed messages but provide uints in hex format instead of decimal format.

### DIFF
--- a/py/send_message.py
+++ b/py/send_message.py
@@ -1036,7 +1036,8 @@ class SendMessage:
         ],
         "Person": [
             { "name": "name", "type": "string" },
-            { "name": "wallet", "type": "address" }
+            { "name": "wallet", "type": "address" },
+            { "name": "age", "type": "uint8" }
         ],
         "Mail": [
             { "name": "from", "type": "Person" },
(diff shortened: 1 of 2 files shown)


py/send_message.py
```diff
@@ -1055,11 +1056,13 @@ def _sign_eth_typed_message(self) -> None:
     "message": {
         "from": {
             "name": "Cow",
-            "wallet": "0xCD2a3d9F938E13CD947Ec05AbC7FE734Df8DD826"
+            "wallet": "0xCD2a3d9F938E13CD947Ec05AbC7FE734Df8DD826",
+            "age": 20,
```
Beerosagos, line 1060: This ',' should not be here. It makes Python rise a `json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 36 column 9 (char 1153)` error

benma: Strange I thought I had tested it. Anyway, removed the comma, thx.

Beerosagos (approved):
tACK!

PR #1121, opened 2023-09-01
releases/build.sh: apply patch to fix reproducible build for v9.15.0

benma:
See the diff for the exact issue that is solved.

The patch is in build.sh so that the documented way (see ./releases/README.md) of reproducing the firmware continues to work. The change is small and well-documented and only applies to v9.15.0, so build.sh continues to be easy to audit.

### DIFF
--- a/releases/build.sh
+++ b/releases/build.sh
@@ -18,8 +18,27 @@ cd temp;
 # work.
 git fetch --tags;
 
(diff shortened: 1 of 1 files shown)


sutterseba (approved):
Worked for me with both versions!

PR #1133, opened 2023-09-27
verifying firmware for BTC only and adding assertation and my GPG pubkey

guggerf:


benma:
Thank you very much!

benma:
Could you please also sign your commit using the same gpg key? Thanks!

guggerf:
> Could you please also sign your commit using the same gpg key? Thanks!

sorry I forgott that. It is signed now ;-)

PR #1138, opened 2023-10-16
Eip1559

Tomasvrba:


benma (commented):
Would be great to add some unit tests in sign.rs for the eip1559 transactions where one can see that the displayed fee is correct. Check the existing unit tests for how to do this.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -231,22 +357,16 @@ pub async fn process(request: &pb::EthSignRequest) -> Result<Response, Error> {
...
-        chain_id: params.chain_id,
-    })
-    .or(Err(Error::InvalidInput))?;
+    let hash: [u8; 32] = match request {
+        Transaction::Legacy(legacy) => hash_legacy(params.chain_id, recipient, legacy)?,
+        Transaction::Eip1559(eip1559) => hash_eip1559(params.chain_id, recipient, eip1559)?,
+    };
 
-    let host_nonce = match request.host_nonce_commitment {
+    let host_nonce = match request.host_nonce_commitment() {
```
benma, line 365: for Eip1559, please force the use of anti-klepto. No need to make it optional because there are backwards compatibility concerns as with legacy, where anti-klepto was introduced after wallets had already been using the non-anti-klepto signing.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -70,16 +133,59 @@ fn parse_erc20(request: &pb::EthSignRequest) -> Option<([u8; 20], BigUint)> {
...
 
 // fee: gas limit * gas price:
-fn parse_fee<'a>(request: &pb::EthSignRequest, params: &'a Params) -> Amount<'a> {
-    let gas_price = BigUint::from_bytes_be(&request.gas_price);
-    let gas_limit = BigUint::from_bytes_be(&request.gas_limit);
-    Amount {
-        unit: params.unit,
-        decimals: WEI_DECIMALS,
-        value: gas_price.mul(gas_limit),
+fn parse_fee<'a>(request: &Transaction<'a>, params: &'a Params) -> Amount<'a> {
```
benma, line 136: Use `Transaction<'_>` not `'a` - the  resulting Amount lifetime only depends on the lifetime of `params`.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -70,16 +133,59 @@ fn parse_erc20(request: &pb::EthSignRequest) -> Option<([u8; 20], BigUint)> {
 }
 
 // fee: gas limit * gas price:
```
benma, line 135: please extend the docstring, the more detail the better, as this fee stuff in both legacy and eip1559 is not easy to grasp. would be good to mention both legacy and eip1559 and what is displayed for the fee.

example

>             // We show the user the max possible fee, but the actual fee might be lower

but in legacy this is also true (actual gas used could be lower than gas limit), so understanding the difference would be helpful.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -70,16 +133,59 @@ fn parse_erc20(request: &pb::EthSignRequest) -> Option<([u8; 20], BigUint)> {
...
+            Amount {
+                unit: params.unit,
+                decimals: WEI_DECIMALS,
+                value: max_fee_per_gas.mul(gas_limit),
+            }
+         }
     }
 }
 
+fn hash_legacy(chain_id: u64, recipient: [u8; 20], request: &pb::EthSignRequest) -> Result<[u8; 32], Error> {
```
benma, line 159: you can drop the recipient arg, it is already in `request.recipient`.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -70,16 +133,59 @@ fn parse_erc20(request: &pb::EthSignRequest) -> Option<([u8; 20], BigUint)> {
...
+        recipient: &recipient,
+        value: &request.value,
+        data: &request.data,
+        chain_id,
+    })
+    .map_err(|_| Error::InvalidInput)?;
+    Ok(hash)
+}
+
+fn hash_eip1559(chain_id: u64, recipient: [u8; 20], request: &pb::EthSignEip1559Request) -> Result<[u8; 32], Error> {
```
benma, line 173: same here, drop the recipient arg in favor of request.recipient.

you can also drop the `chain_id` arg here as it is always present in `request.chain_id` (the same is not true in hash_legacy, as there the coin could also be specified with `request.coin` with a zero `request.chain_id`).

benma (changes requested):
Reviewed the first commits, will review `eip1559: add eip1559 sighash` soon.

src/rust/bitbox02-rust/src/hww/api/ethereum/sighash.rs
```diff
@@ -90,18 +104,52 @@ fn hash_params<W: Write>(writer: &mut W, params: &Params) {
...
+fn hash_params_eip1559<W: Write>(writer: &mut W, params: &ParamsEIP1559) {
+    hash_u64(writer, params.chain_id);
+    hash_element(writer, params.nonce);
+    hash_element(writer, params.max_priority_fee_per_gas);
+    hash_element(writer, params.max_fee_per_gas);
+    hash_element(writer, params.gas_limit);
+    hash_element(writer, params.recipient);
+    hash_element(writer, params.value);
+    hash_element(writer, params.data);
+    hash_header(writer, RLP_SMALL_TAG, RLP_LARGE_TAG, 0_u16); // access list not currently supported and hashed as empty list
```
benma, line 116: `_u16` is not needed here, the `0` literal should work as `hash_header()` defines the type.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
benma: nevermind - the client libraries can/should enforce antiklepto for eip1559, the firmware does not have to. you can also enforce it in the firmware if it is eay, but that is not so important.

benma (changes requested):
also checked the sighash.rs changes - very nice!

Tomasvrba:
should be all fixed, check fixup commits and https://github.com/digitalbitbox/bitbox02-firmware/pull/1138/commits/9e8eb638970117d48fe0b4fa36ffcbf0924e1a63

I can squash them all once ACKed

py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -796,6 +785,59 @@ def eth_sign(self, transaction: bytes, keypath: Sequence[int], chain_id: int = 1
 
             return signature
 
+        if is_eip1559:
+            self._require_atleast(semver.VersionInfo(9, 16, 0))
+            (
+                decoded_chain_id,
```
benma, line 791: I added a check that this is the same as the passed chain_id param (seemed easier than forcing the arg to be `None` in case of eip1559).

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -177,40 +287,60 @@ async fn verify_standard_transaction(
 }
 
 /// Verify and sign an Ethereum transaction.
-pub async fn process(request: &pb::EthSignRequest) -> Result<Response, Error> {
-    let coin = pb::EthCoin::from_i32(request.coin).ok_or(Error::InvalidInput)?;
-    let params = super::params::get_and_warn_unknown(coin, request.chain_id).await?;
+pub async fn process(request: &Transaction<'_>) -> Result<Response, Error> {
+    let coin = pb::EthCoin::from_i32(request.coin()).ok_or(Error::InvalidInput)?;
+    let params = super::params::get_and_warn_unknown(coin, request.chain_id()).await?;
```
benma, line 292: if the chain_id is 0, there is no error and ETH will be used, because `request.coin()` above defaults to ETH.

This is a bug, a chain ID of 0 should not be signed for as ETH.

I noticed this because the unit test you added had a chain_id of 0 and it didn't cause an error :P

benma (changes requested):

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -612,12 +878,30 @@ mod tests {
...
+                keypath: KEYPATH.to_vec(),
+                nonce: b"\xb9".to_vec(),
+                max_priority_fee_per_gas: b"".to_vec(),
+                max_fee_per_gas: b"\x3b\x9a\xca\x00".to_vec(),
+                gas_limit: b"\x01\x09\x85".to_vec(),
+                recipient: b"\x9c\x23\xd6\x7a\xea\x7b\x95\xd8\x09\x42\xe3\x83\x6b\xcd\xf7\xe7\x08\xa7\x47\xc1".to_vec(),
+                value: b"".to_vec(),
+                data: b"\xa9\x05\x9c\xbb\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x85\x7b\x3d\x96\x9e\xac\xb7\x75\xa9\xf7\x9c\xab\xc6\x2e\xc4\xbb\x1d\x1c\xd6\x0e\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x98\xa6\x3c\xbe\xb8\x59\xd0\x27\xb0".to_vec(),
+                host_nonce_commitment: None,
+                chain_id: 0,
```
benma, line 898: 1

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -531,14 +724,69 @@ mod tests {
...
+                keypath: KEYPATH.to_vec(),
+                nonce: b"\x1f\xdc".to_vec(),
+                max_priority_fee_per_gas: b"\x3b\x9a\xca\x00".to_vec(),
+                max_fee_per_gas: b"\x01\x65\xa0\xbc\x00".to_vec(),
+                gas_limit: b"\x52\x08".to_vec(),
+                recipient: b"\x04\xf2\x64\xcf\x34\x44\x03\x13\xb4\xa0\x19\x2a\x35\x28\x14\xfb\xe9\x27\xb8\x85".to_vec(),
+                value: b"\x07\x5c\xf1\x25\x9e\x9c\x40\x00".to_vec(),
+                data: b"foo bar".to_vec(),
+                host_nonce_commitment: None,
+                chain_id: 0,
```
benma, line 781: 1

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -426,7 +568,58 @@ mod tests {
...
+            max_fee_per_gas: b"\x01\x65\xa0\xbc\x00\x00".to_vec(),
+            gas_limit: b"\xa2\x08".to_vec(),
+            recipient:
+                b"\x04\xf2\x64\xcf\x34\x44\x03\x13\xb4\xa0\x19\x2a\x35\x28\x14\xfb\xe9\x27\xb8\x85"
+                    .to_vec(),
+            // 530564000000000000
+            value: b"\x07\x5c\xf1\x25\x9e\x9c\x40\x00".to_vec(),
+            data: b"".to_vec(),
+            host_nonce_commitment: None,
+            chain_id: 0,
```
benma, line 621: 1

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -372,12 +496,30 @@ mod tests {
...
+                keypath: KEYPATH.to_vec(),
+                nonce: b"\x1f\xdc".to_vec(),
+                max_priority_fee_per_gas: b"".to_vec(),
+                max_fee_per_gas: b"\x01\x65\xa0\xbc\x00".to_vec(),
+                gas_limit: b"\x52\x08".to_vec(),
+                recipient: b"\x04\xf2\x64\xcf\x34\x44\x03\x13\xb4\xa0\x19\x2a\x35\x28\x14\xfb\xe9\x27\xb8\x85".to_vec(),
+                value: b"\x07\x5c\xf1\x25\x9e\x9c\x40\x00".to_vec(),
+                data: b"".to_vec(),
+                host_nonce_commitment: None,
+                chain_id: 0,
```
benma, line 516: 1

benma (changes requested):

benma:
@Tomasvrba I fixed the bug with chain_id=0 (and added a eip1559_unhappy test for it), bumped the fw version to 9.16.0, and rebased (fixed conflicts).

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
benma: fixed

py/bitbox02/bitbox02/bitbox02/bitbox02.py
benma: fixed

Tomasvrba:
> @Tomasvrba I fixed the bug with chain_id=0 (and added a eip1559_unhappy test for it), bumped the fw version to 9.16.0, and rebased (fixed conflicts). I also squashed all commits.

Nice, thanks! Anything else that needs to be done here?

benma:
> > @Tomasvrba I fixed the bug with chain_id=0 (and added a eip1559_unhappy test for it), bumped the fw version to 9.16.0, and rebased (fixed conflicts). I also squashed all commits.
> 
> Nice, thanks! Anything else that needs to be done here?

Don't think so, I just wanted to test it in the BitBoxApp to make sure it works before merging it.

benma (approved):
tACK!

PR #1143, opened 2023-10-31
Verify firmware for btc and btc-only - v9.15.0

JuliusWiedemann:


benma (approved):
Thanks a lot, appreciate it!

PR #1144, opened 2023-11-02
rust: update toolchain to 1.73.0

benma:
Saves around 10kB in binary size.


asi345 (approved):
After updating the Docker image and container, it works without a problem.
tACK!

PR #1153, opened 2023-11-16
eth tx verification: show same address as in pc

asi345:
ETH addresses can be encoded all lowercase, all uppercase, or mixed case (https://eips.ethereum.org/EIPS/eip-55). We always show the address in mixed case when verifying a tx, but some exchanges and other wallets don't. If a user copies an address in lowercase or uppercase, the user is confused why the bb02 display doesn't match the wallet display.

The bb02 should show it in the same way the user entered the address.

That is why ETH signing requests are extended with a new string field: "address". This field hold the raw recipient address that the user enters. After the recipient hex address is also derived from the request and it is approved, this derived address is compared with the raw provided address(case-insensitive). If they match, transaction continues and bb02 shows the raw provided address(same as in pc) in the transaction confirmation scree.


messages/eth.proto
```diff
@@ -56,6 +56,7 @@ message ETHSignRequest {
   AntiKleptoHostNonceCommitment host_nonce_commitment = 9;
   // If non-zero, `coin` is ignored and `chain_id` is used to identify the network.
   uint64 chain_id = 10;
+  string address = 11;
```
benma, line 59: i am now thinking that it might be better to provide a `enum ETHAddressCase { MIXED = 0; LOWER = 1; UPPER = 2; }` instead and format the address to be displayed/return according to this. 

The benefit is that it is explicit. An address string can contain many invalid values while the enum can only have valid values.

What do you think? the downside i could see is that figuring out which value to provide would then be offloaded to the client libraries (or their users), but it's generally better to have libraries deal with such complexities than the firmware.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -266,7 +278,11 @@ async fn verify_standard_transaction(
         .await?;
     }
 
-    let address = super::address::from_pubkey_hash(&recipient);
+    let address_verify = super::address::from_pubkey_hash(&recipient);
+    let address = format!("0x{}", request.address());
```
benma, line 282: why does the request.address() not contain `0x`? that prefix is part of the address.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -266,7 +278,11 @@ async fn verify_standard_transaction(
...
+    if address.to_lowercase() != address_verify.to_lowercase() {
+        return Err(Error::InvalidInput);
```
benma, line 284: should also fail with `InvalidInput` if `address` is not all lowercase, not all uppercase, or if mixed case, is not exactly `address_verify` (mixed case must follow the checksum algorithm).

that's why in the other comment i suggest using an enum instead which is easier to handle, as it can only contain valid values.

py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -818,6 +818,7 @@ def handle_antiklepto(request: eth.ETHRequest) -> bytes:
                     recipient=recipient,
                     value=value,
                     data=data,
+                    address=recipient.hex(),
```
benma, line 821: the method itself should take an `address` parameter so that users of the library can provide the address as the user typed it in. by adding `recipient.hex()` here, you force the display of the address on the bitbox02 to be lowercase, which is not the goal.

benma (changes requested):

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -206,7 +213,12 @@ async fn verify_erc20_transaction(
     let erc20_params =
         bitbox02::app_eth::erc20_params_get(params.chain_id, parse_recipient(request.recipient())?);
     let formatted_fee = parse_fee(request, params).format();
-    let recipient_address = super::address::from_pubkey_hash(&erc20_recipient);
+    let recipient_address_verify = super::address::from_pubkey_hash(&erc20_recipient);
+    let recipient_address = format!("0x{}", hex::encode(&erc20_recipient));
+    if recipient_address.to_lowercase() != recipient_address_verify.to_lowercase() {
+        return Err(Error::InvalidInput);
+    }
+
```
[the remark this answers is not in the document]
asi345: The problem here is that `recipient_address` is constructed from `erc20_recipient`, not from `recipient` coming from `request`. That's why using directly `request.address()` is not a solution here. Do you have a suggestion about this?

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
asi345: Okay, I will add `0x` when the request is first created

py/bitbox02/bitbox02/bitbox02/bitbox02.py
asi345: Then how can I take the `address` as the user types in? I couldn't see a direct way to get that. Only thing I saw was all fields of `request`, including `recipient`, was decoded from bytes of `transaction` that was given as input.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
asi345: Oh, I see more clearly now. I thought a string with random capitalisation was okay. Will fix that.

messages/eth.proto
asi345: That's an okay idea. But figuring out which value to provice you mention confuses me. Then how are we supposed to interpret user input in the code? Mixed, lower or upper? Do we check all 3 cases and then construct `ETHAddressCase` from whichever that suits? Do we just throw an `InvalidInputError` when it's not a valid string?

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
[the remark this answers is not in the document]
benma: Yes, if use the enum solution instead as suggested in another comment, you don't need to compare any addresses, but just use the casing as defined by the enum value.

messages/eth.proto
benma: the enum would be instead of the address (the new address field would be dropped again). the client library then simply takes the enum value as a method argument in `eth_sign()`.

py/bitbox02/bitbox02/bitbox02/bitbox02.py
benma: the address as the user types it (or the enum value if we go for the enum solution, which i think we should) would be provided by the user of the client library. the user of the client library, e.g. the BitBoxApp or Rabby (or any wallet) knows what the user typed and can infer the right value.

PR #1154, opened 2023-11-18
rust: update toolchain to 1.74.0

benma:
firmware.bin size before and after:

before: 761096
after: 759736

-1360 bytes.


asi345 (approved):
tACK, it really reduces the binary size. Even I could not measure the exact size by linux commands, the difference is noticable and around 1.5 KB for me.

benma:
> Even I could not measure the exact size by linux commands

@asi345 why not? 

    du -b build/bin/firmware.bin

asi345:
I was using `ls -lah` and `du -h`, and they were not precise in the byte dimension. Seems that `du -b` is a more preferable option.

PR #1156, opened 2023-11-23
Update btc deps

benma:
Builds on https://github.com/digitalbitbox/bitbox02-firmware/pull/1154

The two commits (without Rust toolchain update) yield a binary size reduction of 17524 bytes.


asi345 (approved):
tACK, I also experienced around 17 KBs of size reduction in the firmware binary.

PR #1158, opened 2023-12-29
rust: add workaround for potential compiler bug

benma:
Apply workaround for a potential miscompilation bug: See
- https://github.com/rust-embedded/cortex-m/discussions/503
- https://github.com/rust-lang/rust/issues/118867 We have not observed any abnormal behavior even though we fulfil all the criteria to be affected (opt-level='z', thumbv7em-none-eabi target, Rust toolchain 1.74.0 being >= 1.73.0), but we apply the workaround just in case.
This increases the binary size of the `make firmware` (Multi) build by 11568 bytes at the time of adding this workaround. This can be removed again once the issue above is fixed and we have updated to a Rust toolchain that contains the fix.

This workaround is one of three suggested alternatives. The other two are:
- Downgrade the Rust toolchain - dismissed as it is harder to do, as it involves re-building the Docker image
- Switch from opt-level='z' to opt-level='s' - this increases the binary size by 74416 bytes for the Multi, which is much more than the workaround in this commit.


Beerosagos (approved):
utACK

PR #1159, opened 2024-01-04
rust/src: add stdout print

asi345:
Due to the cyclic dependency issues, rust part of the firmware can not import std create and therefore it does not have io capabilities, including 'println!'. As a result, C 'printf' function is exposed to rust as an external function.

It is a useful feature to have it because as
developers we may need to debug many parts of the
rust code when we are not using debuggers. Also,
in projects such as simulator, firmware is mainly
debugged with gdb. Hence there is limited flexibility of debugging when execution jumps to rust. In addition, simulator could print some of the ui elements of
firmware to stdout in this way.


src/rust/bitbox02/src/lib.rs
```diff
@@ -176,6 +176,12 @@ pub fn reboot() -> ! {
     panic!("reboot called")
 }
 
+pub fn print_stdout(msg: &str) {
```
benma, line 179: please add `#[cfg(feature = "testing")]` to only activate this in tests, as this should not be used in production firmware.

benma (changes requested):

src/rust/bitbox02/src/lib.rs
asi345: It must have slipped from my eye. Fixing.

benma (approved):
utACK

PR #1162, opened 2024-01-15
Verified firmware-btc v9.15.0 binary hash and added my gpg pubkey

junghoon2lee:
I just verified firmware-btc-v9.15.0 binary hash, and added my gpg pubkey in the pubkeys directory.
Thank you!
Junghoon


benma (approved):

benma:
Thank you! Could you please sign your commit as well with the same key?

junghoon2lee:
I just signed my commit with the same key. 
Thank you!

PR #1166, opened 2024-01-24
rust: add c-unit-testing feature

benma:
The `testing` feature is enabled for *Rust* unit tests, but not for the Rust code when compiled to be linked to the *C* unit tests.

We want to add a simulator in the unit-test directory (as C unit tests are already standalone host binaries that link the full C/Rust firmware code), but need a way to change behavior for the simulator only. The `testing` flag is not enabled there as mentioned above.

We add a `c-unit-testing` feature that is active only in Rust code when linked to the C unit tests (including the simulator) for this purpose.

We wanted to simply enable `testing` there as well, but ran into compilation problems when compiling the Rust unit tests:

```
/usr/bin/ld:
.../build-build-rust-unit-tests/src/../lib/libbitbox_merged.a(bitbox02_rust_c-06ac99f37ae6bb45.bitbox02_rust_c.ff44fdfedc2996e0-cgu.0.rcgu.o):(.init_array.00099+0x0):
multiple definition of `std::sys::unix::args::imp::ARGV_INIT_ARRAY';
.../build-build-rust-unit-tests/src/rust/target/all-features/release/deps/bitbox02-5ad04c78e548693d.bitbox02.5d97720ccc25f2c4-cgu.0.rcgu.o:(.init_array.00099+0x0):
first defined here
```

### DIFF
--- a/src/rust/bitbox02-rust-c/Cargo.toml
+++ b/src/rust/bitbox02-rust-c/Cargo.toml
@@ -59,7 +59,8 @@ target-c-unit-tests = [
   # enable these features
   "app-bitcoin",
   "app-ethereum",
-  "firmware"
+  "firmware",
+  "c-unit-testing",
 ]
 
 platform-bitbox02 = []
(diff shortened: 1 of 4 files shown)


asi345 (approved):
utACK

PR #1167, opened 2024-01-25
simulator: implement a simulator for bitbox02 device

asi345:
HWI is thinking of updating its support policy such that supported wallets must implement a simulator/emulator. See https://github.com/bitcoin-core/HWI/issues/685. That's why a simulator is implemented for bitbox02, supporting functionalities of its API.

This first version of the simulator is capable of nearly every functionality of a normal Bitbox02 device, without promising any security or production use. Its main aim is to be able to run unit tests for features and test the API.

In addition, it will be configured to run automated tests in CI, which helps both us and HWI integration.

Right now, the simulator has 3 different ways to communicate with a client: giving inputs/getting output from CLI, using pipes or opening sockets. Socket is the most convenient and reliable choice in this version. It expects the clients to open a socket on port 15432, which is selected intentionally to avoid possible conflicts.

The simulator resides with C unit-tests since it uses same mocks, therefore it can be built by `make unit-test`.

Lastly, Python client implemented in `py/send_message.py` is updated to support communicating with simulator with the socket configuration mentioned above. Client can be started up with `./py/send_message.py --simulator` command. To run the simulator, `build-build/bin/test_simulator` command is sufficient.


src/rust/bitbox02/src/ui/ui_stub_c.rs
```diff
@@ -0,0 +1,163 @@
+// Copyright 2020 Shift Crypto AG
```
benma, line 1: wrong year

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,368 @@
...
+    usb_processing_set_send(usb_processing_hww(), send_message);
+    printf("USB setup success\n");
+    hww_setup();
+    printf("HWW setup success\n");
+    // Memory initialization
+    sd_success = sd_format();
+    printf("Sd card setup %s\n", sd_success ? "success" : "failed");
+    mock_memory_factoryreset();
+    memory_interface_functions_t ifs = {
+        .random_32_bytes = _memory_setup_rand_mock,
```
benma, line 339: `= random_32_bytes_mcu` and `#include<random.h>` at the top, which uses C's `rand()` to get pseudorandom numbers, no need for the fixtures.

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,368 @@
...
+    };
+    memory_success = memory_setup(&ifs);
+    printf("Memory setup %s\n", memory_success ? "success" : "failed");
```
benma, line 342: properly abort/fail on errors

src/rust/bitbox02-rust/src/workflow/mnemonic.rs
```diff
@@ -13,26 +13,36 @@
...
-use super::trinary_choice::{choose, TrinaryChoice};
-use super::trinary_input_string;
 
-use alloc::boxed::Box;
 use alloc::string::String;
-use alloc::vec::Vec;
-use core::cell::RefCell;
 
-use sha2::{Digest, Sha256};
+cfg_if::cfg_if! {
```
benma, line 20: i think with that many changes, it's better to add a new file mnemonic_stub_c_unit_tests.rs, so the real firmware code does not become obfuscated by all this

src/rust/bitbox02-rust/src/workflow/mnemonic_c_unit_tests.rs
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+pub use super::cancel::Error as CancelError;
+use super::confirm;
+
+use alloc::string::String;
+use alloc::string::ToString;
+
+pub async fn show_and_confirm_mnemonic(words: &[&str]) -> Result<(), CancelError> {
+    let _ = confirm::confirm(&confirm::Params {
```
benma, line 22: not sure what the purpose is of the contents of this function - it could just be empty right?

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,155 @@
...
+#include <sd.h>
+#include <stdio.h>
+#include <unistd.h>
+
+#include <netdb.h>
+#include <netinet/in.h>
+#include <sys/socket.h>
+
+#define BUFFER_SIZE 1024
+#define SOCKET
```
benma, line 36: unused

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,155 @@
...
+    rust_async_usb_spin();
+    usb_processing_process(usb_processing_hww());
+}
+
+int main(void)
+{
+    char buffer[BUFFER_SIZE];
+    uint8_t input[BUFFER_SIZE];
+    int (*get_message)(char*, uint8_t*);
+    void (*send_message)(void);
```
benma, line 82: not needed? can just use `get_usb_message_socket` and `send_usb_message_socket` directly

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,155 @@
...
+}
+
+void send_usb_message_socket(void)
+{
+    struct queue* q = queue_hww_queue();
+    const uint8_t* data = queue_pull(q);
+    if (data != NULL) {
+        data_len = 256 * (int)data[5] + (int)data[6];
+        for (int i = 0; i < USB_HID_REPORT_OUT_SIZE; i++) {
+            out_buffer[i] = (char)data[i];
```
benma, line 59: what's the point of this var? why nt just `data` in the next line in `write()`?

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,155 @@
...
+        }
+        if (!write(sockfd, out_buffer, USB_HID_REPORT_OUT_SIZE)) {
+            perror("ERROR, could not write to socket");
+            exit(1);
+        }
+    }
+}
+
+void simulate_firmware_execution(uint8_t* input)
```
benma, line 68: `const uint8_t*`

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,155 @@
...
+{
+    memcpy(_out_report, input, sizeof(_out_report));
```
benma, line 70: no need for _out_report and this copy? can just use `input` on the next line

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,155 @@
...
+}
+
+int main(void)
+{
+    char buffer[BUFFER_SIZE];
+    uint8_t input[BUFFER_SIZE];
+    int (*get_message)(char*, uint8_t*);
+    void (*send_message)(void);
+    bool memory_success, sd_success;
+    int temp_len;
```
benma, line 84: no need to declare the vars at the top, code is imo nicer when the vars are just declared on first use, since we don't care much about stack use.  `bool memory_success = memory_setup(...)` etc.

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,155 @@
...
+
+    // Establish socket connection with client
+    int portno;
+    struct sockaddr_in serv_addr;
+    struct hostent* server;
+    portno = 15423;
```
benma, line 90: same, one line is better than two lines ;) `int portno = 15423`. could also prefix with `const` since it's not changing.

benma (changes requested):
looking nice, just a few nits!

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,135 @@
...
+#include <unistd.h>
+
+#include <netdb.h>
+#include <netinet/in.h>
+#include <sys/socket.h>
+
+#define BUFFER_SIZE 1024
+
+int data_len;
+int sockfd;
```
benma, line 38: make `sockfd*` an argument to the get_usb_message_socket function, then you don't need a global sockfd variable and it can be local. 

looks like we can't do much about data_len being global for now, that seems it would require some refactoring of the usb processing functions, which we don't need to go into now :smile:

test/unit-test/test_simulator.c
```diff
@@ -0,0 +1,135 @@
...
+    smarteeprom_bb02_config();
+    bitbox02_smarteeprom_init();
+    idle_workflow_blocking();
+
+    // BitBox02 firmware loop
+    uint8_t input[BUFFER_SIZE];
+    int temp_len;
+    while (1) {
+        if (!get_usb_message_socket(input)) break;
+        simulate_firmware_execution(input);
```
benma, line 125: some comments here explaining how long packets are read and send would be helpful for the future, it's not very obvious because the usb stack is so convoluted.

benma (approved):
tACK - very nice work!

PR #1169, opened 2024-02-05
Makefile had some bugs from previous way we compiled semihosting bina…

NickeZ:
Makefile had some bugs from previous way we compiled semihosting binaries


benma (approved):
utACK

PR #1170, opened 2024-02-08
Simulator setup improvements

asi345:
- Simulator is moved to its own directory under `test`
- Added shortcuts in Makefile to build only the simulator and run it
- Socket communication sides are inverted such that simulator acts as a server now and clients connect to them


test/simulator/simulator.c
```diff
@@ -67,27 +67,31 @@ int main(void)
...
-    memcpy((char*)&serv_addr.sin_addr.s_addr, (char*)server->h_addr_list[0], server->h_length);
+    serv_addr.sin_addr.s_addr = INADDR_ANY;
     serv_addr.sin_port = htons(portno);
-    if (connect(sockfd, (struct sockaddr*)&serv_addr, sizeof(serv_addr)) < 0) {
-        perror("ERROR, could not connect to client");
-        return 1;
+    int serv_addr_len = sizeof(serv_addr);
+    if (bind(sockfd, (struct sockaddr*)&serv_addr, serv_addr_len) < 0) {
+        perror("ERROR binding socket");
+        exit(EXIT_FAILURE);
```
benma, line 82: why not `return 1`?

test/simulator/CMakeLists.txt
```diff
@@ -0,0 +1,191 @@
+# Copyright 2018 Shift Cryptosecurity AG
```
benma, line 1: 2024 Shift Crypto AG

test/simulator/CMakeLists.txt
```diff
@@ -0,0 +1,191 @@
...
+  ${DBB-FIRMWARE-USB-SOURCES}
+  )
+
+set(IGNORE_SOURCES
+    "src/screen.c"
+    "src/memory/nvmctrl.c"
+    "src/memory/smarteeprom.c"
+    "src/memory.mpu.c"
+    )
+#f9b3ceaa91e
```
benma, line 34: remove this comment here and in the unit-test folder, seems like some local note leftover

test/simulator/CMakeLists.txt
```diff
@@ -0,0 +1,191 @@
...
+endif()
+if(SANTIZE_UNDEFINED)
+  target_compile_options(bitbox_objects-simulator PUBLIC "-fsanitize=undefined")
+  target_compile_options(bitbox-simulator PUBLIC "-fsanitize=undefined")
+endif()
+if(COVERAGE)
+  target_link_libraries(bitbox-simulator PUBLIC "--coverage")
+  target_compile_options(bitbox_objects-simulator PUBLIC "--coverage")
+  target_compile_options(bitbox-simulator PUBLIC "--coverage")
+endif()
```
benma, line 176: don't need coverage stuff here i think, as the simulator is not unit testing and not about seeing which lines are covered during execution

test/simulator/framework/eh_personality.c
```diff
@@ -0,0 +1,28 @@
...
+//
+// See
+// https://doc.rust-lang.org/unstable-book/language-features/lang-items.html#writing-an-executable-without-stdlib.
+//
+// One could get rid of this and also considerably shrink the binary size by compiling core instead
+// of using pre-built binaries. See a proof of concept implementation here:
+// https://github.com/digitalbitbox/bitbox02-firmware/tree/build-std-PoC.  We decided against doing
+// this for now as the feature seems immature and because of the warnings against using it in
+// production:
+// https://github.com/rust-lang/wg-cargo-std-aware/tree/81765f0eb744b9c47840c16f43a32c9f61fd7f0c#mvp-implementation
```
benma, line 26: not about unit tests here but about the simulator.  could also drop the comment (except for the first paragraph)  as we don't need to repeat all this detail here.

test/simulator/framework/includes/mock_qtouch.h
```diff
@@ -0,0 +1,34 @@
...
+#define _QTOUCH_H_
+
+#include <stdint.h>
+
+void __wrap_qtouch_process(void);
+
+uint8_t __wrap_qtouch_get_scroller_is_active(uint16_t sensor_node);
+
+uint16_t __wrap_qtouch_get_scroller_position(uint16_t sensor_node);
+
```
benma, line 25: wrap stuff not needed, it's only used in unit tests to replace functions with the `-Wl,wrap=...` linker args

test/simulator/framework/includes/test_random.h
```diff
@@ -0,0 +1,25 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef _TEST_RANDOM_H
+#define _TEST_RANDOM_H
+
+int __wrap_rand(void);
+int __wrap_wally_sha256(
+    const unsigned char* bytes,
```
benma, line 20: same

test/simulator/framework/mock_memory.c
```diff
@@ -0,0 +1,107 @@
...
+}
+
+bool __wrap_memory_get_encrypted_seed_and_hmac(
+    uint8_t* encrypted_seed_and_hmac_out,
+    uint8_t* len_out)
+{
+    *len_out = _encrypted_seed_and_hmac_len;
+    memcpy(encrypted_seed_and_hmac_out, _encrypted_seed_and_hmac, *len_out);
+    return true;
+}
```
benma, line 93: same

test/simulator/framework/mock_memory.c
```diff
@@ -0,0 +1,107 @@
...
+
+static uint8_t _salt_root[32] = {
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11,
+    0x22, 0x22, 0x22, 0x22, 0x22, 0x22, 0x22, 0x22, 0x33, 0x33, 0x33, 0x33, 0x33, 0x33, 0x33, 0x33,
+};
+void mock_memory_set_salt_root(const uint8_t* salt_root)
+{
+    memcpy(_salt_root, salt_root, 32);
+}
+bool __wrap_memory_get_salt_root(uint8_t* salt_root_out)
```
benma, line 103: same

benma (changes requested):

test/simulator/simulator.c
asi345: True, it is a mistake from copy paste

test/simulator/CMakeLists.txt
```diff
@@ -0,0 +1,185 @@
...
+  ${CMAKE_SOURCE_DIR}/external/fatfs/source
+  $<TARGET_PROPERTY:wallycore,INTERFACE_INCLUDE_DIRECTORIES>
+  $<TARGET_PROPERTY:secp256k1,INTERFACE_INCLUDE_DIRECTORIES>
+  $<TARGET_PROPERTY:fatfs,INTERFACE_INCLUDE_DIRECTORIES>
+)
+target_include_directories(
+  bitbox_objects-simulator
+  PUBLIC
+  ${INCLUDES}
+  ${CMAKE_CURRENT_SOURCE_DIR}/framework/includes
```
benma, line 108: we have a bit of a mess with the includes:

the libc-unit-tetsts library built in src/CMakeLists adds the  `unit-test/framework/includes`  directory (see https://github.com/digitalbitbox/bitbox02-firmware/blob/1bcb6dd934635a454d5ddb13524da569ec5ea1e3/src/CMakeLists.txt#L266), and the resulting lib is used in both the unit tests *and* the simulator.

the simulator then also has a copy of most/all of the same includes and adds it here.

untangling this could be annoying, let's postpone it? i think you can just use this instead

    ${CMAKE_CURRENT_SOURCE_DIR}/../unit-test/framework/includes

and delete the simulator/framework/includes folder.

Then all have the same includes, which might not be ideal, but at least consistent.

benma (changes requested):
tested the simulator, seems to work fine also with the connection reversed:)

test/simulator/framework/includes/mock_memory.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include <flags.h>
+
+void mock_memory_factoryreset(void);
+bool memory_write_to_address_mock(uint32_t base, uint32_t addr, const uint8_t* chunk);
+bool memory_write_chunk_mock(uint32_t chunk_num, const uint8_t* chunk);
+void memory_read_chunk_mock(uint32_t chunk_num, uint8_t* chunk_out);
+// Size: `FLASH_SHARED_DATA_LEN`.
+void memory_read_shared_bootdata_mock(uint8_t* chunk_out);
+void mock_memory_set_salt_root(const uint8_t* salt_root);
```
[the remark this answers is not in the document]
asi345: I could leave out only the last line. Build fails if I delete the others(lines 26 and 28)

test/simulator/CMakeLists.txt
asi345: simulator only uses 3 files from `framework/includes` folder, while unit-test uses 8. so actually simulator uses less stuff from there, but still we can postpone this for now since the solution would be untrivial. i'll change the `includes` path as `${CMAKE_CURRENT_SOURCE_DIR}/../unit-test/framework/includes` in the build of simulator.

PR #1172, opened 2024-02-08
releases: add sutterseba signatures for v9.16.0

sutterseba:
Signatures for both v9.16.0 releases


benma:
That was fast :grin: thanks for confirming the builds again!

benma (approved):

PR #1174, opened 2024-02-12
Minor tweaks

NickeZ:


scripts/set-new-screen.jlink
```diff
@@ -0,0 +1,4 @@
+mem 0xe002 1
+w1 0xe002 1
```
benma, line 2: can you also add scripts to set it back to the old one? on raw pcb's it's easy to switch screens.

probably not necessary for the chip as you can't switch that :)

benma (changes requested):
pls run `./scripts/docker_exec.sh ./scripts/format` to fix up the CI error complaining about the formatting.

scripts/set-new-screen.jlink
NickeZ: done!

benma (approved):
utACK

PR #1176, opened 2024-02-14
Correct flash memory layouts

NickeZ:
Document the actual values of the flash partition offsets and update linker scripts to match.

### DIFF
--- a/bootloader.ld
+++ b/bootloader.ld
@@ -35,7 +35,7 @@ SEARCH_DIR(.)
 /* Memory Spaces Definitions */
 MEMORY
 {
-  /* Reserve (at least) 32 bytes of space for factory install random bytes */
+  /* Reserve 32 bytes of space for factory install random bytes */
   rom      (rx)  : ORIGIN = 0x00000000, LENGTH = 0x0000DFE0
   ram      (rwx) : ORIGIN = 0x20000000, LENGTH = 0x00040000
   bkupram  (rwx) : ORIGIN = 0x47000000, LENGTH = 0x00002000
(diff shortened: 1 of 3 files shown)


NickeZ:
Please double check my math :)

bootloader.ld
```diff
@@ -35,8 +35,7 @@ SEARCH_DIR(.)
 /* Memory Spaces Definitions */
 MEMORY
 {
-  /* Reserve (at least) 32 bytes of space for factory install random bytes */
-  rom      (rx)  : ORIGIN = 0x00000000, LENGTH = 0x0000DFE0
+  rom      (rx)  : ORIGIN = 0x00000000, LENGTH = 0x0000E000
```
benma, line 38: the previous value was correct - the last 32 bytes are reserved for the factory randomness:

https://github.com/digitalbitbox/bitbox02-firmware/blob/1bcb6dd934635a454d5ddb13524da569ec5ea1e3/src/random.c#L96

That should maybe be put into its own definition in flags.h.

Please update the diagram too (reduce bootloader area by 32 bytes, add the 32 bytes factory randomness item).

firmware.ld
```diff
@@ -35,7 +35,7 @@ SEARCH_DIR(.)
 /* Memory Spaces Definitions */
 MEMORY
 {
-  rom      (rx)  : ORIGIN = 0x00010000, LENGTH = 0x000F0000
+  rom      (rx)  : ORIGIN = 0x00010000, LENGTH = 0x000D8000
```
benma, line 38: i re-computed and came to the same conclusion.

benma (changes requested):

bootloader.ld
NickeZ: Interesting! I'll update it

benma (approved):
thanks, that was a very good and important catch!

PR #1177, opened 2024-02-16
Add support for optional sccache in build system

NickeZ:
sccache is a binary cache that wraps all calls to gcc/rustc and caches build output in `~/.cache`. This speeds up a build after `make clean` by about 3x. I choose not to put it in the dev dockerfile by default since caching might have unexpected side effects? But I put documentation in the build docs of how to install it.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -2,7 +2,10 @@ include(ExternalProject)
 
 if(CMAKE_CROSSCOMPILING)
   set(CONFIGURE_FLAGS
-    --host=${CMAKE_SYSTEM_PROCESSOR}-none-eabi --build=${CMAKE_HOST_SYSTEM_PROCESSOR}-linux-gnu)
+    --host=${CMAKE_SYSTEM_PROCESSOR}-none-eabi
+    --build=${CMAKE_HOST_SYSTEM_PROCESSOR}-linux-gnu
+    "$<$<BOOL:${CMAKE_C_COMPILER_LAUNCHER}>:CC=${CMAKE_C_COMPILER_LAUNCHER} ${CMAKE_C_COMPILER}>"
+  )
 endif()
 
 # Remove parameters to make build identical to older build
--- a/BUILD.md
+++ b/BUILD.md
@@ -179,3 +179,12 @@ make coverage # or make -C build-build coverage
 ```sh
 make -C build-build coverage-lcovr
 ```
+
+### SCCache / CCache
+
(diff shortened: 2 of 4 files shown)


external/CMakeLists.txt
```diff
@@ -2,7 +2,10 @@ include(ExternalProject)
 
 if(CMAKE_CROSSCOMPILING)
   set(CONFIGURE_FLAGS
-    --host=${CMAKE_SYSTEM_PROCESSOR}-none-eabi --build=${CMAKE_HOST_SYSTEM_PROCESSOR}-linux-gnu)
+    --host=${CMAKE_SYSTEM_PROCESSOR}-none-eabi
+    --build=${CMAKE_HOST_SYSTEM_PROCESSOR}-linux-gnu
+    "$<$<BOOL:${CMAKE_C_COMPILER_LAUNCHER}>:CC=${CMAKE_C_COMPILER_LAUNCHER} ${CMAKE_C_COMPILER}>"
```
benma, line 7: doesn't cmake automatically prepend/use the CMAKE_C_COMPILER_LAUNCHER? why is this line necessary?

benma (commented):
I think it's a good call to not put it into the container so there are for sure no unexpected side effects.

Can you explain what the point is of this vs simply not calling `make clean`? cmake/rust already have incremental builds.

NickeZ:
> I think it's a good call to not put it into the container so there are for sure no unexpected side effects.
> 
> Can you explain what the point is of this vs simply not calling `make clean`? cmake/rust already have incremental builds.

Sometimes when I jump back and forth between branches it rebuilts almost everything. And before pushing I would like to be able to call `.ci/ci` to speed up the feedback cycle. Some outputs are probably also the same in `build` and `build-semihosting` and then sccache can reuse those compiled units. Right now I have to wait 6-7 minutes on my machine for a clean build, but with scache that takes 1.5 minute. And those minutes add up when I'm trying to get the PR to pass CI.

Especially when I'm dealing with cmake dependencies. I don't trust cmake so I do make clean between biulds.

external/CMakeLists.txt
benma: good to know - pls add a comment :smile:

benma (approved):
utACK

external/CMakeLists.txt
benma: nevermind, i thought we were in a different CMakeLists.txt. It's in external/, so that's clear enough.

PR #1180, opened 2024-02-20
releases/describe_signed_firmware: add helpful error msg

benma:
When this script is called on an unsigned binary, there is a warning

    Unrecognized firmware edition

but the script then still parses and dumps information that is completely wrong.

This commit outputs a helpful error message instead and exists.


asi345 (approved):
utACK! good thinking

PR #1187, opened 2024-02-29
simulator: print simulator msgs only in simulator, not in unit tests

benma:
The `c-unit-testing` is also active in Rust unit tests, so the simulator msgs were printed there. This commit moves the stdout prints to `ui_stub_c_unit_tests.rs`, which is only used in the simulator.


asi345 (approved):
tACK, works as intended. thanks for help!

PR #1189, opened 2024-03-08
add firmware-btc-v9.17.0/assertion-JunghoonLee.sig

junghoon2lee:
I just verified firmware-btc-v9.17.0 binary hash, and so added my assertion.  
Note that my gpg pubkey (JunghoonLee.asc) is already in the releases/pubkeys directory, which was used when I verified firmware-btc-v9.15.0, so I didn't add it this time. 
Thank you!
Junghoon


benma (approved):
Thank you so much!

PR #1197, opened 2024-04-17
simulator : build configuration update and multiple sequential client…

asi345:
… support

Since simulator and c unit tests use the same build script(build-build), sanitization options defined for c unit tests were also passed to simulator and this was causing more dependencies for simulator binary. Of course this contributes to more size. Simulator now has this sanitization options set to false to automatically avoid including them.

Another issue that was faced while implementing HWI integration was multiple client support. In HWI, one client could access to simulator for initialization and then disconnect, while newer clients could connect and directly test the functionalities of wallet API, assuming that the device is already initialized. Parallel handling of clients are still not supported, also not expected from HWI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@
 
 .DEFAULT_GOAL := firmware
 SANITIZE ?= ON
+simulator: SANITIZE = OFF
 
 bootstrap:
 	git submodule update --init --recursive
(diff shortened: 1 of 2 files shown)


benma (approved):
tACK, good improvement and makes the simulator more useful as a debugging tool too.

PR #1202, opened 2024-04-24
py/bitbox02: allow infering version/edition from OP_INFO

benma:
For when no USB descriptor is available, where we normally get the info from. We need this now to connect to the simulator with the right params. OP_INFO is available since v5.0.0, and the created after that, so for the simulator this will always work.


asi345 (approved):
Seems to be working as expected, tACK!

PR #1211, opened 2024-05-08
securechip_random : multiple tries

asi345:
secure_chip random function uses atca library for randomness, but this call to library may not be successful and error codes might be returned by atca. In this case, firmware was also calling `Abort()`. This stops the usb connection and terminates the program. Should it happen during a factory reset, then it means the device is left in a half-reset state.

That's why it is a good practice to try this atca library call for randomness multiple times, making it possible to avoid rare unexpected errors from atca random implementation. Furthermore, this commit also specifies a `ATCA_STATUS` code in the abort message, hence allowing debugging a potential error coming from atca.

### DIFF
--- a/src/reset.c
+++ b/src/reset.c
@@ -48,11 +48,25 @@ void reset_reset(bool status)
 {
     keystore_lock();
 #if !defined(TESTING)
-    if (!securechip_update_keys()) {
+    bool sc_result_update_keys = false;
+    for (int retries = 0; retries < 5; retries++) {
+        sc_result_update_keys = securechip_update_keys();
+        if (sc_result_update_keys) {
+            break;
+        }
+    }
+    if (!sc_result_update_keys) {
         Abort("Could not reset secure chip.");
     }
 #if APP_U2F == 1
-    if (!securechip_u2f_counter_set(0)) {
+    bool sc_result_u2f_counter_set = false;
+    for (int retries = 0; retries < 5; retries++) {
+        sc_result_u2f_counter_set = securechip_u2f_counter_set(0);
+        if (sc_result_u2f_counter_set) {
+            break;
+        }
+    }
+    if (!sc_result_u2f_counter_set) {
         Abort("Could not initialize U2F counter.");
     }
 #endif
--- a/src/securechip/securechip.c
+++ b/src/securechip/securechip.c
@@ -605,7 +605,12 @@ bool securechip_monotonic_increments_remaining(uint32_t* remaining_out)
 
 bool securechip_random(uint8_t* rand_out)
 {
-    return atcab_random(rand_out) == ATCA_SUCCESS;
+    for (int retries = 0; retries < 5; retries++) {
+        if (atcab_random(rand_out) == ATCA_SUCCESS) {
+            return true;
+        }
+    }
+    return false;
 }
 
 // Length of priv_key must be 32 bytes


src/securechip/securechip.c
```diff
@@ -605,7 +605,18 @@ bool securechip_monotonic_increments_remaining(uint32_t* remaining_out)
...
-    return atcab_random(rand_out) == ATCA_SUCCESS;
+    ATCA_STATUS result;
+    uint8_t retries = 0, max_retries = 5;
+    do {
+        result = atcab_random(rand_out);
+    } while (result != ATCA_SUCCESS && retries++ < max_retries);
+    if (result != ATCA_SUCCESS) {
+        char msg[70];
+        sprintf(msg, "securechip_random is not successful with ATCA STATUS %d", result);
+        Abort(msg);
```
benma, line 616: I wouldn't abort here, as this stops execution of the firmware. If one has to Abort, better do it at the callsite, which is what is happening now already anyway.

benma (changes requested):
https://github.com/digitalbitbox/bitbox02-firmware/blob/master/src/reset.c#L51-L57

The original problem does not seem to be completely solved with this, as `securechip_update_keys()` could fail or `securechip_u2f_counter_set()` could fail, due to `random_32_bytes_sec()` (as it `Abort()`s directly) or due to many other securechip functions that are called, resulting in an `Abort()` in `reset_reset()`.

So the device could end up half-reset in many more cases than just securechip_random failing.

Having `random_32_bytes_sec()` return a true/false and not aborting would be quite invasive and add error handling in many places, so keeping the re-trying code in securechip_random is probably good, as you have done. 

In addition, to solve the above issue, it might be good to retry `securechip_update_keys()` and `securechip_u2f_counter_set()`  inside reset_reset() a few times before aborting there.

src/securechip/securechip.c
asi345: You're right, but that denies the opportunity of printing the atca status code from the error. Therefore we again can not have an idea about which error we get.

asi345:
I understand the concern about `securechip_update_keys()` and `securechip_u2f_counter_set()`, on the other hand, is it really necessary to retry both of them and also `securechip_random()` in extra? Wasn't just retrying `securechip_update_keys()` and `securechip_u2f_counter_set()` could work?

src/securechip/securechip.c
benma: If you want to display the error code, then it is best to return it in the function signature and do it at the call site, like how it is done in `securechip_setup()`.

benma:
> I understand the concern about `securechip_update_keys()` and `securechip_u2f_counter_set()`, on the other hand, is it really necessary to retry both of them and also `securechip_random()` in extra? Wasn't just retrying `securechip_update_keys()` and `securechip_u2f_counter_set()` could work?

Just repeating the latter would work, but only if `random_32_bytes_sec()` did not Abort but returned an error instead, otherwise the reset can still just abort at this function. As mentioned previously, having it return an error and propagate and handle this everywhere could be quite annoying, so repeating it in addition might be an okay solution. If you want you can try returning an error instead and seeing how it goes.

src/securechip/securechip.c
asi345: I did not want to change function signature for this, therefore I discarded the error message for now. It is a random generator function anyway, I don't think the type of error we get there would help debugging something. Just retrying seems enough for me in that sense, I don't it's worth to change function signature and its calls.

asi345:
`void (*const random_32_bytes)(uint8_t* buf)`  is an intermediate function since it calls `random_32_bytes_sec` and `random_32_bytes_mcu` while being called by many places including `_update_kdf_key` and therefore `securechip_update_keys`. What I'm saying is this function signature is used in many critical places, also in the simulator. Error propagation from `securechip_update_keys` requires also propagating it through `random_32_bytes`. Assigning a return type for this function will generate so many warnings in all the callers because of `return value not used` and then we need to insert return value checks everywhere.
I have been trying to achieve this but as you said it is annoying(although doable) and seems messy, in my opinion. So my suggestion would also be leaving as it is and inserting retry loops in all of `securechip_random`, `securechip_update_keys` and `securechip_u2f_counter_set`.

benma:
Agree, but imo it would be clearer to not repeat inside securechip_update_keys and securechip_u2f_counter_set, but inside reset_reset(), as this is where it's critical. Another reason is that if we replace the chip (which is planned), we don't have to repeat this task with the new secure chip code.

asi345:
I'm doing the retries in `reset_reset`, not the functions itself. I agree that the securechip functions should stay as it is as much as possible.
In addition, I updated pr title, commit title and body accordingly.

src/reset.c
```diff
@@ -47,12 +47,20 @@ static void _show_reset_label(bool status)
 void reset_reset(bool status)
 {
     keystore_lock();
+    uint8_t retries = 0, max_retries = 5;
```
benma, line 50: better to simply use `int` (more common and probably more efficient than uint8_t), and `const` for max_retries

this function also does 6 attempts not 5, but it's hard to read because the do..while with in-line increment is kind of hard to read. how about doing a simpler but easier to read loop like this:

```
   ATCA_STATUS result;
    int retries = 0;
    const int max_retries = 5;

    while (retries < max_retries) {
        if (atcab_random(rand_out) == ATCA_SUCCESS) {
            return true;
        }
        retries++;
    }
    return false;
```

src/reset.c
```diff
@@ -47,12 +47,20 @@ static void _show_reset_label(bool status)
...
-    if (!securechip_update_keys()) {
+    while (!sc_result && retries++ < max_retries) {
+        sc_result = securechip_update_keys();
+    }
+    if (!sc_result) {
         Abort("Could not reset secure chip.");
     }
 #if APP_U2F == 1
-    if (!securechip_u2f_counter_set(0)) {
+    while (!sc_result && retries++ < max_retries) {
```
benma, line 60: I think this never executes because `sc_result` is `true` after `securechip_update_keys()` succeeds?

benma (changes requested):

src/reset.c
asi345: okay, i was trying to make it faster and shorter. but of course i can be more verbose

src/reset.c
asi345: i thought the `#if` cases were mutual exclusive, i missed this 😄

src/reset.c
benma: clarity & correctness first :smile: shorter is not better if it's harder to read, and in this case it had a "bug" (too many retries) as a result

src/reset.c
```diff
@@ -48,11 +48,31 @@ void reset_reset(bool status)
...
+    bool sc_result_update_keys = false;
+    while (retries_update_keys < max_retries_update_keys) {
+        sc_result_update_keys = securechip_update_keys();
+        if (sc_result_update_keys) {
+            break;
+        }
+        retries_update_keys++;
+    }
+    if (!sc_result_update_keys) {
         Abort("Could not reset secure chip.");
```
benma, line 62: sorry for one more nit but i now realized that all these loops are simply for loops:

```
bool sc_result_update_keys = false;
    for (int retries = 0; retries < 5; retries++) {
        sc_result_update_keys = securechip_update_keys();
        if (sc_result_update_keys) {
            break;
        }
    }
    if (!sc_result_update_keys) {
        Abort("Could not reset secure chip.");
    }
```

(in this case no need for a different retry var name as it is locally scoped to the for loop).

same in securechip.c

benma (changes requested):

benma (approved):
LGTM

PR #1214, opened 2024-05-13
Rust 1.78

benma:
The toolchain bump and removing the async_recursion dep reduces the firmware binary by 5832 bytes compared to the v9.18.0 release.


benma:
I tested U2F (due to the small `bitbox02-rust-c/src/workflow.rs ` which is used for U2F confirmation screens) and signing ETH typed msgs manually on a device to confirm everything works as expected.

PR #1215, opened 2024-05-14
erc20 params binary efficiency improvements

benma:
Multi firmware before: 755504 bytes
After: ~~747472~~ 745872 bytes
Diff: ~~-8032~~  9632 bytes


src/rust/Cargo.toml
```diff
@@ -21,8 +21,14 @@ members = [
...
     "bitbox02",
     "bitbox02-sys",
+    "erc20_params",
 ]
 
+# Needed if edition 2018 and 2021 are mixed in the workspace crates, to fix
+# a cargo warning.
+# When all are switched to 2021, this can be removed.
+resolver = "2"
+
```
benma, line 31: Now part of https://github.com/digitalbitbox/bitbox02-firmware/pull/1216, which should be merged first, then this can be rebased.

asi345: I tested and reviewed that pr first, it's ready to merge.

asi345 (approved):
built and tested the firmware, all works fine. binary size is exactly 747472 in my system too, perfect!
tACK 👍

benma:
@asi345 I removed another ~1 byte per token by removing the null terminator byte for the unit strings and grouping the tokens by unit string length too.

PR #1217, opened 2024-05-15
rust: inherit deps from the workspace to keep versions in sync

benma:
Specify the dep and its version only once.

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table


asi345 (approved):
good idea for efficiency, tACK!

benma:
Merged as part of another PR, closing this.

PR #1222, opened 2024-06-03
.ci: also build simulator

benma:
To catch regressions.


asi345 (approved):
Good idea to test the state of the simulator too, utACK

PR #1223, opened 2024-06-04
bitcoin: allow multisig at arbitrary keypaths

benma:
Before, we would restrict the account-level keypaths for multisig to be:

- m/48'/coin'/account'/1' for P2WSH_P2SH
- m/48'/coin'/account'/2' for P2WSH

Since the keypath is verified by the user and the coin network is confirmed for every receive/send, ransom and isolation bypass attacks are mitigated sufficiently, and we can lift this restriction. Note that for wallet policies (of which multisig is a subset of), arbitrary keypaths are already allowed.

When exporting an xpub, we furthermore warned about "unusual" keypaths. In addition to the above keypaths, we also allow exporting the xpub at m/45' without warning, as this path is used by Unchained for their vaults.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 
 ### [Unreleased]
 - Bitcoin: add support for payment requests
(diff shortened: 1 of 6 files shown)


asi345 (approved):
seems good, tACK 👍

PR #1224, opened 2024-06-05
bitcoin: allow spending UTXOs at very high address indices

benma:
In a BIP-44 keypath `m/purpose/coin/account/{0,1}/addressIdx`, we
disallow addressIdx >= 10000 when verifying receive addresses to
mitigate ransom attacks (attacker as victim verify an address at a
very high index, so the victim cannot find the funds).

The same check is applied to change outputs in transactions for the
same reason.

We also had the same check for inputs in the transaction, but there,
the check is not necessary. We want to be able to spend UTXOs that
were received on high indices.


benma:
Rebased now that the other PR is merged.

asi345 (approved):
i don't see any issues, approved.

PR #1227, opened 2024-06-12
sdcard: allow exit from 'Insert SD Card' screen

asi345:
Currently the user has no option to go back from the 'Insert SD Card' screen prompt except by unplugging the device. Maybe the user clicks the button to manage backups in the settings by mistake and they don't have the SD card available and need to cancel the request.

This commit introduces a cancel option on the 'Insert SD Card to continue' screen, allowing going back to the original menu. Thus, unplugging is not needed anymore and the device can continue without an interruption.


benma (changes requested):
very nice, tACK!

Please add a CHANGELOG entry, sth like "Allow exiting the screen asking to insert the microSD card".

benma (approved):
LGTM

PR #1228, opened 2024-06-12
Policy refactor

benma:


benma:
> tACK, i can't see a problem. However I could not understand why this change was needed. Can you summarize shortly?

Thanks. Have you checked the commit messages? The script configs preprocess commit has a long explanation. The others are to prepare adding a `Tr(...)` Taproot enum entry next to `Wsh()` in `ParsedPolicy`, as Taproot policies will need re-use this.

asi345:
Oh you're right. Sorry I just missed the commit message paragraphs. Thanks!

PR #1229, opened 2024-06-12
hww: _api_info write byte for initialized status

NicolaLS:
Write another byte for the initialized status in the _api_info function. This field can be used to check whether a device can be unlocked or not. The unlocked byte is not sufficient because unlocked = false could mean the device is locked or the device is uninitialized.

### DIFF
--- a/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
+++ b/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
@@ -546,7 +546,7 @@ class BitBoxCommonAPI:
             elif device_info["product_string"] == BITBOX02BTC:
                 edition = BitBox02Edition.BTCONLY
         else:
-            version, _, edition, _ = self.get_info(transport)
+            version, _, edition, _, _ = self.get_info(transport)
 
         self.edition = edition
         try:
@@ -680,11 +680,14 @@ class BitBoxCommonAPI:
         return True
 
     @staticmethod
-    def get_info(transport: TransportLayer) -> Tuple[str, Platform, Union[BitBox02Edition], bool]:
+    def get_info(
+        transport: TransportLayer,
+    ) -> Tuple[str, Platform, Union[BitBox02Edition], bool, Optional[bool]]:
         """
-        Returns (version, platform, edition, unlocked).
-        This is useful to get the version of the firmware when a usb descriptor is not available
-        (via BitBoxBridge, etc.).
+        Returns (version, platform, edition, unlocked, initialized).
+        This is useful to get the version of the firmware or the device unlocked/initialized status
+        when a usb descriptor is not available (via BitBoxBridge, etc.). The initialized status is
+        supported from firmware v9.20.0 (it is None if not supported).
         This call does not use a versioned BitBoxProtocol for communication, as the version is not
         available (this call is used to get the version), so it must work for all firmware versions.
         """
@@ -704,9 +707,14 @@ class BitBoxCommonAPI:
         else:
             raise Exception("Unknown platform: {}".format(platform))
 
-        unlocked_byte = response[0]
+        unlocked_byte, response = response[0], response[1:]
         unlocked = {0x00: False, 0x01: True}[unlocked_byte]
-        return (version_str, platform, edition, unlocked)
+
+        initialized = None
+        if parse_device_version(version_str) >= semver.VersionInfo(9, 20, 0):
+            initialized_byte = response[0]
+            initialized = {0x00: False, 0x01: True}[initialized_byte]
+        return (version_str, platform, edition, unlocked, initialized)
 
     def check_min_version(self) -> None:
         """
--- a/src/hww.c
+++ b/src/hww.c
@@ -16,6 +16,7 @@
 
 #include <hardfault.h>
 #include <keystore.h>
+#include <memory/memory.h>
 
 #include <platform_config.h>
 #include <rust/rust.h>
@@ -65,6 +66,7 @@ typedef struct {
  * - For the BitBoxBase platform (deprecated):
  " - 0x00 - Standard
  * 1 byte: 0x00 if the device is locked, 0x01 if the device is unlocked.
+ * 1 byte: 0x00 if the device is uninitialized, 0x01 if the device is initialized.
  * @param[out] buf serialize info to this buffer.
  * @return number of bytes written
  */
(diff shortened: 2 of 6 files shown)


src/hww.c
```diff
@@ -97,6 +99,10 @@ static size_t _api_info(uint8_t* buf)
     *current = keystore_is_locked() ? 0x00 : 0x01;
     current++;
 
+    // 1 byte initialized status
+    *current = memory_is_initialized() ? 0x00 : 0x01;
```
benma, line 103: wouldn't `0x01` for "is_initialized" be more intuitive? If one interprets 0x00 as false and 0x01 as true.

benma (changes requested):
Awesome, thanks. Please add a CHANGELOG entry

src/hww.c
NicolaLS: True, I just copy and pasted the above without thinking much. I'll change it, and also change the `switch` in the go api then.

benma (changes requested):
Thanks! Please also, in a new commit, adjust the Py lib in the same repo, add a CHANGELOG entry to the Py lib and bump the pylib version (major version bump as it's an incompatible change)

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py#L683

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -680,11 +680,13 @@ def reboot(
...
-    def get_info(transport: TransportLayer) -> Tuple[str, Platform, Union[BitBox02Edition], bool]:
+    def get_info(
+            transport: TransportLayer
+    ) -> Tuple[str, Platform, Union[BitBox02Edition], bool, bool]:
         """
-        Returns (version, platform, edition, unlocked).
-        This is useful to get the version of the firmware when a usb descriptor is not available
-        (via BitBoxBridge, etc.).
+        Returns (version, platform, edition, unlocked, initialized).
+        This is useful to get the version of the firmware or the device unlocked/initialized status
```
NicolaLS, line 688: I think the _This is useful (...)_ part could be improved but I don't understand it well enough and also I'm bad at constructing concise sentences.

-> this is useful if there was no handshake yet too, otherwise we could just call the other info request that already has the initialized attribute in the response...how could this be phrased ? (when I tried to use the other request in `api-go` it would throw an error like this: `handshake must come first`

but I guess it is fine to leave it as is?

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -704,9 +706,12 @@ def get_info(transport: TransportLayer) -> Tuple[str, Platform, Union[BitBox02Ed
         else:
             raise Exception("Unknown platform: {}".format(platform))
 
-        unlocked_byte = response[0]
+        unlocked_byte, response = response[0], response[1:]
         unlocked = {0x00: False, 0x01: True}[unlocked_byte]
-        return (version_str, platform, edition, unlocked)
+
+        initialized_byte = response[0]
+        initialized = {0x00: False, 0x01: True}[initialized_byte]
```
benma, line 713: This will crash on firmware versions that don't support this yet. Maybe one way of dealing with this is to return `Optional[bool]` and document that it is `None` if not supported.

benma (changes requested):

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
NicolaLS: ah yes thanks, it would throw index out of range for `initialized_byte = response[0]` should I just `try except` it and set `initialized` to `None` in `except` or is there a better way of handling this (checking somehow if it is supported ?)

benma: You can check the firmware version, there are other examples. For now let's expect this to go into v9.20.0, but we can adjust the number if it lands quicker.

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -680,11 +680,14 @@ def reboot(
...
-        This is useful to get the version of the firmware when a usb descriptor is not available
-        (via BitBoxBridge, etc.).
+    def get_info(
+            transport: TransportLayer
+    ) -> Tuple[str, Platform, Union[BitBox02Edition], bool, Optional[bool]]:
+        """
+        Returns (version, platform, edition, unlocked, initialized).
+        This is useful to get the version of the firmware or the device unlocked/initialized status
+        when a usb descriptor is not available (via BitBoxBridge, etc.). In case the firmware does not
+        respond with the initialized byte yet None is returned.
```
benma, line 690: The docstring could be more explicit: "The initialized status is supported from firmware v9.20.0 (it is `None` if not supported)"

benma (changes requested):
CI complains about code formatting. Apply `black py/bitbox02/bitbox02/communication/bitbox_api_protocol.py` to format the file.

NicolaLS:
@benma thanks! rebased and re-formatted/changed docstring as suggested. did this in one push

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -704,9 +707,14 @@ def get_info(transport: TransportLayer) -> Tuple[str, Platform, Union[BitBox02Ed
         else:
             raise Exception("Unknown platform: {}".format(platform))
 
-        unlocked_byte = response[0]
+        unlocked_byte, response = response[0], response[1:]
         unlocked = {0x00: False, 0x01: True}[unlocked_byte]
-        return (version_str, platform, edition, unlocked)
+
+        initialized = None
+        if version_str >= semver.VersionInfo(9, 20, 0):
```
benma, line 714: Please test your changes - this check does not work, it crashes because `version_str` is not a valid version string.

Use parse_device_version(version_str) instead.

benma (changes requested):
please also bump version strings to 9.20.0, which is now possible after we released 9.19.0

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/CMakeLists.txt#L99-L100

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
NicolaLS: right, thanks.

benma (approved):
tACK, thanks!

PR #1232, opened 2024-06-16
Policies improvements

benma:
See commit messages.


asi345 (approved):
Seems good and clear commits, tACK

PR #1234, opened 2024-06-20
bitcoin: add Pocket pubkey for payment requests

benma:
pocketbitcoin.com will be able to make payment requests, which will be useful for the Pocket sell feature in the BitBoxApp.


benma:
@davidknezic please confirm that this pubkey is correct. Thanks!

davidknezic (approved):
Yes, I can confirm the compressed pub key in db9e536535fa0701cf5cc7e33091c937ffbbb3b8 corresponds to Pocket's SLIP-0024 signing key. Recipient name is correct, too.

Beerosagos (approved):
LGTM!

PR #1239, opened 2024-06-23
rust: add `dep:` prefix when enabling deps

benma:
For clarity.

The bitcoin dep is now always enabledas it's needed in bip32.rs, which is used not only by the bitcoin app.


NickeZ (approved):
utACK

PR #1242, opened 2024-06-26
simulator: add `--port` flag

benma:
We might want to use different ports or run multiple simulators in parallel for testing, in which case one must be able to specify the port.


asi345 (approved):
Looks perfect, tACK

PR #1243, opened 2024-06-26
simulator: add --version argument

benma:
One purpose is to be able to track the current simulator version in git history, so when we release it, we know which version it should be.

bitbox02 prefix as we will have simulators for future BitBox versions too.

The firmware version is there so we know which firmware is simulated.

The "simulator<VERSION>" part is to identify the simulator and its version.


asi345 (approved):
tACK, works as intended!

PR #1248, opened 2024-07-10
Firmware v9.19.0 verification

yaziza:


benma (approved):
Thanks a ton, highly appreciated.

PR #1249, opened 2024-07-11
pairing: remove impossible panic

benma:
Good to avoid panics, plus it save some bytes in the binary for code that never executes.


asi345 (approved):
tACK, nice finding!

PR #1253, opened 2024-07-13
rust: remove usage of sha2::Sha512

benma:
Relies on https://github.com/BitBoxSwiss/rust-bip32-ed25519/pull/4 to be merged first.

Reduces binary size by ~12.7kB. We use sha2::Sha512 only in
Cardano (sighash when signing and bip32-ed25519 bip32 derivations).

We can use the other sha512 we already have: wally_sha512. This one is
also used in wally's bip32, so we can use it "for free".

If we remove wally in the future, we'd use rust-bitcoin's bip32 impl,
which uses rust-bitcoin's own sha512 impl, in which case we would use
that one for Cardano as well.


asi345 (approved):
huge improvement!

PR #1256, opened 2024-07-13
transaction/fee: don't use fmt with "{.1}" to reduce binary size

benma:
Using this pulls in
`core::fmt::float::float_to_decimal_common_shortest` into the binary,
which is a huge function.

By changing the impl to something simpler, we save 20112 bytes in the binary :O


asi345 (approved):
this is huge, wow, very nice!

PR #1258, opened 2024-07-14
format_datetime improvements

benma:


asi345 (approved):
i don't see any problem, tACK!

PR #1260, opened 2024-07-22
external/libwally-core: update to v1.3.0

benma:
Calling autogen.sh instead of autoreconf as autogen.sh also does that
but also calls autogen.sh on secp256k1. Without this change secp256k1
would not build anymore b/c of a build system change in libwally.

USE_BASIC_CONFIG dropped, it's not used anymore in secp256k1.

Darin hack TODO dropped, as libwally's autogen.sh contains it.

Submodule now points to:

https://github.com/BitBoxSwiss/libwally-core/tree/bitbox02-firmware-20240722

which has a custom patch to point its secp256k1 submodule to
https://github.com/BitBoxSwiss/secp256k1-zkp/tree/bitbox02-20240722.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -25,12 +25,10 @@ endif()
 if(SANITIZE_UNDEFINED)
   set(LIBWALLY_CFLAGS "${LIBWALLY_CFLAGS} -fsanitize=undefined")
 endif()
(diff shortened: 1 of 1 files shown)


asi345 (approved):
all works fine, tACK!

PR #1262, opened 2024-07-24
Address case in ETH tx verification on BitBox02 and tx verification on simulator

asi345:
See commit messages for detailed explanation.


src/rust/bitbox02-rust/src/hww/api/ethereum/address.rs
```diff
@@ -58,6 +74,28 @@ mod tests {
...
+                b"\xf4\xc2\x17\x10\xef\x8b\x5a\x5e\xc4\xbd\x37\x80\xa6\x87\xfe\x08\x34\x46\xe6\x7b",
+                pb::AddressCase::Lower as i32,
+            ),
+            "0xf4c21710ef8b5a5ec4bd3780a687fe083446e67b"
+        );
+        assert_eq!(
+            from_pubkey_hash(
+                b"\xf4\xc2\x17\x10\xef\x8b\x5a\x5e\xc4\xbd\x37\x80\xa6\x87\xfe\x08\x34\x46\xe6\x7b",
+                3,
             ),
```
benma, line 99: remove, same as first test

src/rust/bitbox02-rust/src/hww/api/ethereum/address.rs
```diff
@@ -17,23 +17,39 @@ use sha3::digest::Digest;
...
+                *e -= 32; // convert to uppercase
+            }
+        }
+    } else if case == pb::AddressCase::Lower as i32 { // lower case
+        for e in hex.iter_mut() {
+            if *e > b'9' && *e < b'a' {
+                *e += 32; // convert to lowercase
+            }
+        }
+    } else { // take mixed case as default, case = pb::AddressCase::Mixed as i32
```
benma, line 39: I guess hex::encode() returns lowercase, so you don't need to do anything for to lowercase it?

Just use hex::encode_upper() instead if uppercase is needed for simplicity? Then we don't need be careful about if all the conditions here are correct. https://docs.rs/hex/latest/hex/fn.encode_upper.html


Also when you use `case: pb::EthAddressCase`, you can make a nice pattern match:

```
match case {
   EthAddressCase::EthAddressCaseLower => { ... }
    ... 
}
```

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -90,6 +90,12 @@ impl<'a> Transaction<'a> {
             Transaction::Eip1559(_) => Ok(None),
         }
     }
+    fn case(&self) -> i32 {
```
benma, line 93: return `Result<pb::EthAddressCase, Error>`, and convert like for example above in `fn coin()` using try_from.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -490,6 +496,7 @@ mod tests {
                 data: b"".to_vec(),
                 host_nonce_commitment: None,
                 chain_id: 0,
+                case: 2,
```
benma, line 499: please use explicit names for clarity like you did above, i.e. ` pb::AddressCase::Mixed as _`. I consider it a big shortcoming of prost (the Rust protobuf lib we use) that enum fields are not decoded into the actual type but into i32 :( See

https://github.com/tokio-rs/prost/issues/276
https://github.com/tokio-rs/prost/issues/303

messages/eth.proto
```diff
@@ -27,6 +27,12 @@ enum ETHCoin {
   RinkebyETH = 2;
 }
 
+enum AddressCase {
+  UPPER = 0;
+  LOWER = 1;
+  MIXED = 2;
```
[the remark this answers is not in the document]
asi345: ah, i see. then i think `MIXED = 0` is a better solution since we are already okay with mixed case being the default.

src/rust/bitbox02-rust/src/hww/api/ethereum/address.rs
asi345: but it was the test case when the case variable is not valid(not 0,1,2)

src/rust/bitbox02-rust/src/hww/api/ethereum/address.rs
```diff
@@ -17,23 +17,39 @@ use sha3::digest::Digest;
 use alloc::string::String;
 use core::convert::TryInto;
 
+use super::pb;
+
 /// Generates a checksummed ethereum hex address from a 20 byte recipient.
 /// `recipient` - 20 byte tail (last 20 bytes of the pubkeyhash).
-pub fn from_pubkey_hash(recipient: &[u8; 20]) -> String {
+pub fn from_pubkey_hash(recipient: &[u8; 20], case: i32) -> String {
```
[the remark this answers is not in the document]
asi345: nice! that was why i set it to i32 in the firsthand

asi345:
Okay, I implemented the changes you have pointed out.

py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -788,7 +788,13 @@ def eth_pub(
         )
         return self._eth_msg_query(request, expected_response="pub").pub.pub
 
-    def eth_sign(self, transaction: bytes, keypath: Sequence[int], chain_id: int = 1) -> bytes:
+    def eth_sign(
+        self,
+        transaction: bytes,
+        keypath: Sequence[int],
+        address_case: eth.ETHAddressCase.ValueType,
```
benma, line 795: ` = ETH_ADDRESS_CASE_MIXED` to set a default.

src/rust/bitbox02-rust/src/shiftcrypto.bitbox02.rs
```diff
@@ -1543,6 +1547,35 @@ impl EthCoin {
         }
     }
 }
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
+#[repr(i32)]
+pub enum EthAddressCase {
+    Mixed = 0,
```
benma, line 1553: nice, i don't know how prost decided to drop the long prefix, but it's very welcome :D

benma (changes requested):
good job! left just a small nit

benma:
Oh also please add a CHANGELOG entry

src/rust/bitbox02-rust/src/shiftcrypto.bitbox02.rs
asi345: yes, i also couldn't believe when I was looking at new enum field names 😄

asi345:
Did my last push now, according to your requests

py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -788,7 +788,13 @@ def eth_pub(
         )
         return self._eth_msg_query(request, expected_response="pub").pub.pub
 
-    def eth_sign(self, transaction: bytes, keypath: Sequence[int], chain_id: int = 1) -> bytes:
+    def eth_sign(
+        self,
+        transaction: bytes,
+        keypath: Sequence[int],
+        address_case: eth.ETHAddressCase.ValueType = eth.ETH_ADDRESS_CASE_MIXED,
```
benma, line 795: sorry I forgot: also add an entry to /py/bitbox02/CHANGELOG.md, as this changes the Python API.

benma (changes requested):

asi345:
Okay, I also added an entry to CHANGELOG of Python API

PR #1264, opened 2024-07-25
rust: update toolchain to 1.80.0

benma:
And update the prost dep version.

This reduces the binary size by 1692 bytes. 1332 bytes of it due to the Rust 1.80 update, the rest due to the prost update.


Beerosagos (approved):
utACK!

PR #1265, opened 2024-08-06
Build rust stdlib

NickeZ:
Building the rust core and alloc libraries ourselves with abort as panic handler reduced the memory footprint from

```
Generating binary firmware.bin
   text	   data	    bss	    dec	    hex	filename
 680876	  19268	 206520	 906664	  dd5a8	firmware.elf
```

to

```
Generating binary firmware.bin
   text	   data	    bss	    dec	    hex	filename
 630452	  19268	 206520	 856240	  d10b0	firmware.elf
```

To test this you need to rebuild the docker build container.


src/rust/vendor-rust-std.sh
```diff
@@ -0,0 +1,12 @@
+RUST_DIR="$(rustc --print=sysroot)"
+
+# The following two commands must be run as root in the docker container.
+# Copy the Cargo.lock for Rust to place the `cargo vendor` will see
+# cp "$RUST_DIR/lib/rustlib/src/rust/Cargo.lock" "$RUST_DIR/lib/rustlib/src/rust/library/test/"
+# chmod 777 $RUST_DIR/lib/rustlib/src/rust/library/test/Cargo.lock
```
benma, line 6: I am not sure what this means. I just ran this script inside docker and it seemed to work.

src/rust/vendor-rust-std.sh
```diff
@@ -0,0 +1,12 @@
+RUST_DIR="$(rustc --print=sysroot)"
```
benma, line 1: The name of the script sounds like it vendors only the stdlib, but it vendors all our normal deps too and should be used instead of `cargo vendor`, right?

Maybe rename the script and add a section to src/rust/README.md about how to vendor.

src/rust/vendor-rust-std.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# The following two commands must be run as root in the docker container.
+# Copy the Cargo.lock for Rust to place the `cargo vendor` will see
+# cp "$RUST_DIR/lib/rustlib/src/rust/Cargo.lock" "$RUST_DIR/lib/rustlib/src/rust/library/test/"
+# chmod 777 $RUST_DIR/lib/rustlib/src/rust/library/test/Cargo.lock
+
+
+RUSTC_BOOTSTRAP=1 cargo vendor \
```
benma, line 9: If I run this script in docker, I already get a diff - did the stdlibs change? Shouldn't they stay fixed per Rust toolchain version?

benma (changes requested):
Tested and works nicely, saving 50424 bytes in the Multi firmware binary :tada:

benma:
> ```
> Generating binary firmware.bin
>    text	   data	    bss	    dec	    hex	filename
>  630452	  19268	 206520	 856240	  d10b0	firmware.elf
> ```

:+1: 

`du -sb firmware.bin` results in 649720, which is the number of bytes actually taken up on the device.

src/rust/vendor-rust-std.sh
benma: Ok I see now that these are needed, but:

why not uncomment and run them automatically?

And could you add some more comments about why this is needed, and why the library/test/ directory, etc.? Maybe some reference? A bit hard to understand this.

src/rust/vendor-rust-std.sh
```diff
@@ -0,0 +1,12 @@
...
+    --manifest-path Cargo.toml \
+    --sync "$RUST_DIR/lib/rustlib/src/rust/library/test/Cargo.toml" \
```
benma, line 11: Could you add a comment why it's this Cargo.toml? `/test/` sounds wrong, so it could help to add some sort of reference.

src/rust/vendor-rust-std.sh
NickeZ: I moved the commands to the dockerfile

NickeZ:
I found yet another flag to reduce the size of `core`: `optimize_for_size`.

benma:
> I found yet another flag to reduce the size of `core`: `optimize_for_size`.
> 
> ```
>    text	   data	    bss	    dec	    hex	filename
>  629980	  19268	 206520	 855768	  d0ed8	firmware.elf
> ```

Compared to before this, firmware.bin is now 472 bytes smaller. One would expect more for such a promising-sounding flag :joy:

NickeZ:
> 
> Compared to before this, firmware.bin is now 472 bytes smaller. One would expect more for such a promising-sounding flag 😂

Perhaps it has a bigger impact on `std` :).

src/rust/vendor.sh
```diff
@@ -0,0 +1,25 @@
...
+# working directory.
+#
+# For some reason Cargo needs to find the dependencies of all rust std libs.
+# Since "test" depends on all the other ones, we take the toml-file from it.
+# This means that we vendor libs that we don't use in the end (like hashbrown
+# and getopts).
+#
+# The invocation below depends on the fact that rust std libs "Cargo.lock" has
+# been manually copied to be next to the Cargo.toml file in the test
+# directory.
```
benma, line 16: Maybe mention here that the Dockerfile is doing that.

benma (approved):
tACK, very nice :heart_eyes:

benma (approved):
Thanks! Please amend the commits in a way where each commit can build. Could just squash all into one.

NickeZ:
OK, now I'm pretty sure every commit passes `.ci/ci`. Do you also want to test?

benma:
> OK, now I'm pretty sure every commit passes .ci/ci. Do you also want to test?

No need I think, the commits look good! Thanks.

PR #1267, opened 2024-08-13
Nickez/test speeding up docker pull

NickeZ:
This is a test. no need to review yet


NickeZ:
Using a single layer only saved about 10 seconds on the build. so not good enough

PR #1268, opened 2024-08-13
Nickez/slim down docker image

NickeZ:
this is a ci test


Dockerfile
```diff
@@ -64,101 +67,105 @@ RUN apt-get update && apt-get install -y \
...
+    && rm -rf /var/lib/apt/lists/*
 
 # Python modules
-COPY py/bitbox02 /tmp/bitbox02
-RUN python3 -m pip install /tmp/bitbox02
-RUN rm -r /tmp/bitbox02
-COPY py/requirements.txt /tmp
-RUN python3 -m pip install --upgrade --requirement /tmp/requirements.txt
-RUN rm /tmp/requirements.txt
+RUN --mount=source=py,target=/mnt,rw \
```
benma, line 94: TIL

Dockerfile
```diff
@@ -64,101 +67,105 @@ RUN apt-get update && apt-get install -y \
...
+    rustup component add rustfmt && \
+    rustup component add clippy && \
+    rustup component add rust-src && \
+    if [ "${TARGETPLATFORM}" = "linux/arm64" ]; then \
+      CARGO_HOME=/opt/cargo cargo install cbindgen --version ${CBINDGEN_VERSION} --locked && \
+      CARGO_HOME=/opt/cargo cargo install bindgen-cli --version ${BINDGEN_VERSION} --locked --target-dir=/tmp/bindgen-target && rm -r /tmp/bindgen-target; \
+    else \
+      curl -sSL https://github.com/rust-lang/rust-bindgen/releases/download/v${BINDGEN_VERSION}/bindgen-cli-x86_64-unknown-linux-gnu.tar.xz | tar -xJ --strip-components=1 -C /opt/cargo/bin bindgen-cli-x86_64-unknown-linux-gnu/bindgen && \
+      curl -sSL https://github.com/mozilla/cbindgen/releases/download/${CBINDGEN_VERSION}/cbindgen -o /opt/cargo/bin/cbindgen && chmod +x /opt/cargo/bin/cbindgen; \
+    fi && \
```
benma, line 161: imho this adds more long term maintenance effort than is worth it to have two different ways of how to get cbindgen/bindgen-cli depending on the platform. 

it's not often that we need to build the image anyway, so I'd prefer to keep it simple

not a very strong opinion though, feel free to overrule :)

benma (approved):
utACK, but prefer to keep `cargo install` for the two rust tools.

If your goal is to reduce layers, you could also move all of it to a docker_install.sh script and call that in a single RUN. @cstenglein has done that in this open PR https://github.com/BitBoxSwiss/bitbox02-firmware/pull/992, but somehow that PR was never merged.

NickeZ:
> utACK, but prefer to keep `cargo install` for the two rust tools.
> 
> If your goal is to reduce layers, you could also move all of it to a docker_install.sh script and call that in a single RUN. @cstenglein has done that in this open PR #992, but somehow that PR was never merged.

I have a few goals I guess:
* Reducing number of layers (this PR cuts layers in half, which I think is a decent improvement)
* Reducing size of individual layers (Remove all temporary files in the same layer they are created so that they don't end up in the final container, mount files that are only needed temporarily)
* Working towards a functionally reproducible container.
   - What is left here is to pin the ubuntu base image and remove `apt upgrade`. And then probably move all `apt installs` to a new "base image" that we almost never rebuild.
* Increasing speed of building image. This is good when you are building many images, like if you are testing all different GCC toolchains. 2 minutes saved per build, might make you save 20 minutes if you test 10 different versions.

I do think you can take the released versions of bindgen and cbindgen where available. We aren't building any other tools from source. Perhaps we could build our own linux/arm64 versions and publish somewhere in case you don't want to have two approaches in the dockerfile.

The reason a reproducible container is important is so that we can build older versions of the firmware at all. Today, since the image isn't reproducible we depend on docker hub having a copy of all the versions. You can imagine some future version of a debian or pip package after some "security update" isn't compatible anymore.

I think calling out to a single `install.sh` script that installs everything is a bit extreme. If you then for example want to try out different toolchains or something you have to rebuild the whole image and cannot use any caching at all. Defeats one of the purposes of docker. I guess it depends a bit on what you do in the install script.

benma:
> The reason a reproducible container is important is so that we can build older versions of the firmware at all. 

That would be nice, but I would rank this as low priority for two reasons:

1. we address this problem directly [here](https://github.com/BitBoxSwiss/bitbox02-firmware/tree/master/releases#what-is-the-purpose-of-this) and gather signatures for a historic record

> What is the purpose of this?

> The purpose is twofold:
> - [...]
> The firmware binaries are only practically reproducible for a limited amount of time after the release. In the future, the dependencies of our Dockerfile might have changed, or Docker itself might even become incompatible or obsolete. This historic record allows you to be reasonably sure that the released binary was created correctly, even if it becomes infeasible to re-build the binary.

> 2. Failing to reproduce old firmwares has never come up as an issue so far. Users I guess are interested the reproducibility of the firmware they want to run, which is usually the latest version.

benma (approved):
utACK

Thompson1985 (approved):

PR #1270, opened 2024-08-14
Sd auto detect for insert and remove

asi345:
Check commit messages for detailed explanation.

Let me know if you have a better suggestion for polling interval. You can test and let me know if it takes too long.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 - Bitcoin: UX improvements for payment request confirmations
 - Ethereum: display the addresses on device in the same case as the user input
 - Allow exiting the screen asking to insert the microSD card
+- SD card: Remove API to prompt removal of the microSD card
+- SD card: add auto-detection of insertion
 
 ### 9.19.0
 - Display device name on screen before unlock
--- a/py/bitbox02/CHANGELOG.md
+++ b/py/bitbox02/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+# 8.0.0
+- SD card: Remove API to prompt removal of the microSD card from the device
+
 # 7.0.0
 - get_info: add optional device initialized boolean to returned tuple
 - eth_sign: add address_case field, which should be initialized by the client
--- a/py/bitbox02/bitbox02/bitbox02/bitbox02.py
+++ b/py/bitbox02/bitbox02/bitbox02/bitbox02.py
@@ -648,16 +648,6 @@ class BitBox02(BitBoxCommonAPI):
         )
         self._msg_query(request, expected_response="success")
 
-    def remove_sdcard(self) -> None:
-        # pylint: disable=no-member
-        request = hww.Request()
-        request.insert_remove_sdcard.CopyFrom(
-            bitbox02_system.InsertRemoveSDCardRequest(
-                action=bitbox02_system.InsertRemoveSDCardRequest.REMOVE_CARD
-            )
-        )
-        self._msg_query(request, expected_response="success")
-
     def root_fingerprint(self) -> bytes:
         """
         Get the root fingerprint from the bitbox02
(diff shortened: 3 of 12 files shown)


src/ui/components/sdcard.c
```diff
@@ -19,10 +19,15 @@
 #include <hardfault.h>
 #include <screen.h>
 #include <sd.h>
-#include <string.h>
 #include <touch/gestures.h>
 #include <ui/screen_stack.h>
 
+// TODO: use a TIMER interrupt to get a more accurate timer.
+// 250 is ~0.5 sec. Unit: rendering rate.
+#define ACTIVE_AFTER 250
```
benma, line 27: ACTIVE_AFTER is a bit misnamed., it's more like CHECK_INTERVAL?

src/ui/components/sdcard.c
```diff
@@ -19,10 +19,15 @@
...
+
+static int _count = 0;
```
benma, line 29: move this to the data_t struct to avoid globals and so it is initialized to 0 when the component is created.

src/ui/components/sdcard.c
```diff
@@ -47,17 +71,6 @@ static const component_functions_t _component_functions = {
 
 /********************************** Create Instance **********************************/
 
-static void _continue_callback(component_t* component)
-{
-    data_t* data = (data_t*)component->parent->data;
-    if (!data->insert || sd_card_inserted()) {
```
benma, line 53: For removing the card, the continue button served to exit the workflow without an error. You changed it to show a cancel button only, which errors with UserAbort when pressed, which is a breaking change.

Please leave the semantics as it was - leave the continue button/callback and hide the cancel button/callback for removal. Otherwise we have to worry about how it will behave in the app :sweat_smile:

benma (changes requested):

py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -648,16 +648,6 @@ def insert_sdcard(self) -> None:
         )
         self._msg_query(request, expected_response="success")
 
-    def remove_sdcard(self) -> None:
```
benma, line 651: This implies a major version bump of the bitbox02 Python library (7.0.0 => 8.0.0) and a CHANGELOG entry to py/bitbox02/CHANGELOG.md

src/ui/components/sdcard.c
```diff
@@ -19,22 +19,41 @@
...
+    // 250 is ~0.5 sec. Unit: rendering rate.
+    int check_interval;
+    int count;
 } data_t;
 
+static void _insert_poll_callback(component_t* component)
+{
+    data_t* data = (data_t*)component->data;
+    if (sd_card_inserted()) {
+        if (data->callback) {
```
benma, line 38: `if(data->callback && sd_card_inserted())` ? not sure if `sd_card_inserted()` is cheap or expensive, but if expensive, it's better to call it only when needed.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 - Ethereum: allow signing EIP-712 messages containing multi-line strings
 - Allow exiting the screen asking to insert the microSD card
 - HWW: add initialized status byte to _api_info response
+- SD card: delete remove sd card api
```
benma, line 19: confusing wording, maybe better like "Remove API to prompt removal of the microSD card"

benma (changes requested):
I tested the main change, works very nicely! 

Some small nits left.

py/bitbox02/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+# 8.0.0
```
benma, line 5: Also bump the actual package version please: https://github.com/BitBoxSwiss/bitbox02-firmware/blob/878ea00dc91d2675de378bba24062f188cf104b8/py/bitbox02/bitbox02/bitbox02/__init__.py#L19

NickeZ: We can take a look at best practices. I think there also are some changelog tooling, maybe that helps.

benma (approved):

PR #1271, opened 2024-08-14
chore: Readme clarity improvements

RostarMarek:
Added a few parts to readme files to incentivize using Docker setup instead of local environment. Also made sure these aprts link to dockerize setup in Build.md.
Once I have a better understanding of simulator and scripts I will also probably try to write up some at least small guide for those.


NickeZ (approved):
lgtm

py/bitbox02/README.md
```diff
@@ -15,3 +15,16 @@ bitbox02.
...
+
+
+### Development environment
+
+It is highly recommended that you use the dockerized setup while developing, guide for setting it up
+can be found in [BUILD.md](../../BUILD.md).
+
+If you intended to develop instead on your local environment be vary that some development procedures
+such as regenerating files have their libraries NOT in requirements.txt as they are not always necessary to run python
+scripts.
```
benma, line 27: suggestionn:

not always necessary to run the python scripts => not necessary for the bitbox02 Python package.

They are never necessary for the bitbox02 package, as these are dev-dependencies not relevant for the package users.

benma (changes requested):
Thanks a lot. I left a a nit about the wording for Python.

benma:
Please rebase, there appears to be a conflict already :innocent:

benma (approved):

benma (changes requested):
Thanks. Please squash so there is one commit instead of two.

BUILD.md
```diff
@@ -36,12 +36,28 @@ make dockerinit
 make dockerdev
 ```
 
+If you do not want to build the docker image locally, or are not working on it, it may be more straightforward to
+pull the image from docker instead of building it. This should be a bit faster and there should not be any issues with
+`make dockerdev` expecting specific version of the image.
+
+```sh
+docker pull shiftcrypto/firmware_v2:latest
```
benma, line 44: So sorry for missing this :innocent:, but since last week we don't use `:latest` but the version in `.containerversion`. See https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1282/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R134

Please change this line to `make dockerpull`.

benma (changes requested):

BUILD.md
RostarMarek: Will amend the commit so it stays as 1 in a minute

benma (approved):

PR #1272, opened 2024-08-14
Allow cmake to find gcc with env variable

RostarMarek:
Allows cmake to find gcc even on macOS ARM where homebrew install folder is different. By setting an enviromental variable and then fetching it arm.cmake file. If the value is not set it defaults to previous value.

I know there are probably more changes needed for local environment development on macOS ARM, but I found this fix fairly easy to do and not intrusive to previous builds so I wanted to share it this way. 

I do not expect this to be merged, but based on our previous discussion I decided that this would be fairly striaghtforward way to share this.


arm.cmake
```diff
@@ -1,10 +1,14 @@
 set(CMAKE_SYSTEM_NAME "Generic")
 set(CMAKE_SYSTEM_PROCESSOR "arm")
 
+set(LOCATION_PATH $ENV{HOMEBREW_PATH})
```
NickeZ, line 4: I'm not a big fan of `HOMEBREW_PATH` in the `arm.cmake` file. What do you think about calling it `ARMROOT` or `GCCROOT` or something similar?

NickeZ (changes requested):
Looks good in general, I wrote a comment about the variable name. I like that it defaults to `/usr/local`.

arm.cmake
RostarMarek: Sure as mentioned in the opening comment, I do not expect this to get merged, just wanted to suggest a fix to a issue i found.

Thompson1985 (approved):

PR #1273, opened 2024-08-15
Switch to GCC 14

NickeZ:
To reduce the size of the binary a bit we could add -mno-unaligned-access because that made struct initialization generate quite a lot more code.


CMakeLists.txt
```diff
@@ -224,7 +224,7 @@ configure_file(src/bootloader/bootloader_version.h.in src/bootloader/bootloader_
 string(APPEND CMAKE_C_FLAGS " -std=c11 -pipe")
 if(CMAKE_CROSSCOMPILING)
   set(CMAKE_C_FLAGS "\
-    ${CMAKE_C_FLAGS} -mcpu=cortex-m4 -mthumb -mlong-calls \
+    ${CMAKE_C_FLAGS} -mcpu=cortex-m4 -mthumb -mlong-calls -mno-unaligned-access \
```
benma, line 227: Is this safe, or does this introduce risk? If it's safer to not use this flag we could postpone using this trick until we need to save bytes more urgently.

benma (approved):
utACK

NickeZ:
I will wait this this PR until I have some other CI related things merged and I've tested the device.

CMakeLists.txt
NickeZ: Yeah, not sure what the impact has. I think unaligned accesses can be problematic on some platforms, so I would assume that **no** unaligned access would be safer than allowing it.

NickeZ:
Firmware doesn't work. It doesn't pop up the "unlock" screen.

Thompson1985 (approved):

CMakeLists.txt
NickeZ: removed

NickeZ:
> Firmware doesn't work. It doesn't pop up the "unlock" screen.

Since usb stack refactoring modern GCC works.

Dockerfile
```diff
@@ -34,18 +34,16 @@ RUN echo "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-18 main" >> /etc/a
...
-      GNU_TOOLCHAIN_HASH=fb31fbdfe08406ece43eef5df623c0b2deb8b53e405e2c878300f7a1f303ee52 \
-      GNU_TOOLCHAIN_FORMAT=bz2; \
+      GNU_TOOLCHAIN=https://developer.arm.com/-/media/Files/downloads/gnu/14.2.rel1/binrel/arm-gnu-toolchain-14.2.rel1-x86_64-arm-none-eabi.tar.xz \
+      GNU_TOOLCHAIN_HASH=62a63b981fe391a9cbad7ef51b17e49aeaa3e7b0d029b36ca1e9c3b2a9b78823; \
     fi; \
-    wget -O gcc.tar.${GNU_TOOLCHAIN_FORMAT} ${GNU_TOOLCHAIN} &&\
-    echo "$GNU_TOOLCHAIN_HASH gcc.tar.${GNU_TOOLCHAIN_FORMAT}" | sha256sum -c &&\
-    tar -xvf gcc.tar.${GNU_TOOLCHAIN_FORMAT} -C /usr/local --strip-components=1 &&\
-    rm -f gcc.tar.${GNU_TOOLCHAIN_FORMAT}
+    curl -sSL -o gcc.tar.xz ${GNU_TOOLCHAIN} && \
```
benma, line 43: out of curiosity, why wget->curl?

benma (approved):
tACK

Dockerfile
NickeZ: I think we should pick one and not install the other and I think curl is better than wget.

PR #1277, opened 2024-08-18
ci: Fix per commit CI

NickeZ:
A new version that also works in my fork and hopefully works in the common repo as well.

### Benefits

You see small icons next to the appropriate commit because we use the github api to set statuses.

image

### Drawbacks

It is easier to by mistake introduce code that leaks github secrets. Currently we don't use github secrets in any job. But the idea is that we use github secrets in the `push` event so that we can authenticate with docker hub and do `docker push`.

It also requires the repository to give write access to github actions.

### Alternatives

It is possible to implement without pull_request_target by simply not setting the statuses on the commits. This means that you will have to look at the commit hash in the build job name to figure out where it failed.

image


benma (commented):
> Drawbacks

> It is easier to by mistake introduce code that leaks github secrets.

Can you post an example how that would happen for clarity?

Does the alternative you prevent that not have this problem? If so, imho we should go with that. I think the commit hash in the title to identify the commit is not too bad :shrug:

NickeZ:
> Can you post an example how that would happen for clarity?

https://nathandavison.com/blog/github-actions-and-the-threat-of-malicious-pull-requests

> 
> Does the alternative you prevent that not have this problem? If so, imho we should go with that. I think the commit hash in the title to identify the commit is not too bad 🤷

I was afraid you'd say that :P. I'll make a quick fix.

Btw, you should disable write access to the repo for actions.

NickeZ:
ready for review again

benma (approved):
lgtm

PR #1278, opened 2024-08-19
Nickez/build container in ci

NickeZ:
Build container in CI and other cleanup regarding ci


Makefile
```diff
@@ -129,7 +129,7 @@ jlink-flash-factory-setup: | build
 jlink-flash-firmware-semihosting: | build-semihosting
 	JLinkExe -if SWD -device ATSAMD51J20 -speed 4000 -autoconnect 1 -CommanderScript ./build-semihosting/scripts/firmware.jlink
 dockerinit:
-	./scripts/container.sh build --pull --platform linux/amd64 --force-rm --no-cache -t shiftcrypto/firmware_v2 .
+	./scripts/container.sh pull shiftcrypto/firmware_v2:$(shell cat .containerversion)
```
benma, line 132: Please leave the previous one as it was as I use it to build the docker image. If you a shortcut for it please add another command like `dockerpull:`.

Fyi, `make dockerdev` (./scripts/dockerenv.sh) pulls `:latest` if it does not exist. I guess that script could instead also use `.containerversion,, and we could stop using `:latest` altogether.

benma (changes requested):
Reviewed the commits except the main commits to build the image in CI, I hope I'll get to them tomorrow.

Generally, please add a quick comment explaining the "why" to every commit message body so it's easier to review and understand.

E.g. 

> ci: Rename TARGET_BRANCH to COMPARE_REV

Or

> ci: Use github checkout action to pull from git repo 

(in this one I am not sure how the commit title relates to the changes in the commit)

NickeZ:
this is ready for review again

.ci/README.md
```diff
@@ -0,0 +1,31 @@
+CI Design guidelines
+
+* It is more maintainable to create scripts in `.ci` and then call them from the workflows than to
+  have scripts inline in the workflows. However, it is also good to split up scripts in multiple
+  steps and jobs depending on what is being done.
+
+* The docker image is rebuilt if the `Dockerfile` or `.containerversion` file is modified. (In case
+  of a push event it is also automatically published to docker hub).
```
benma, line 8: What happens in the PR CI that modifies the Dockerfile? Technically the CI should run all commits starting from the one that modifies the Dockerfile against the new image, and the ones before against the old image.

Often Dockerfile modifications require other changes too, like fixing new clippy linters when updating the Rust toolchain, etc, and one wants to see if the PR passes CI with the new image.

NickeZ: Yes, this is exactly how it works. See my test here: https://github.com/NickeZ/bitbox02-firmware/pull/3

benma: Amazing!

benma (approved):
LGTM, nice!

PR #1282, opened 2024-08-20
ci: Move the current container version to .containerversion

NickeZ:
* Add helper `make dockerpull` to pull the current version of the container image.
* `make dockerinit` now tags with the current version. The logic is that if you are working on the container image, you should bump the version in `.containerversion`.


NickeZ:
I have no strong feelings about the name `.containerversion`. So we can change if there are any better ideas.

benma (approved):
Change scripts/dockerenv.sh too, like I mentioned here?

https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1278#discussion_r1722293261

NickeZ:
> Change scripts/dockerenv.sh too, like I mentioned here?
> 
> [#1278 (comment)](https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1278#discussion_r1722293261)

done!

PR #1283, opened 2024-08-20
btc/paymentrequest: split memo into multiple screens

benma:
One screen to introduce the memo, and one screen per line in the
memo. This will allow a nice UX for Pocket to display:

"Memo from POCKET"

"IBAN correct?"

"CH 93 ...."

title_short vs title_long to not change the UX of signmsg while being
able to re-use the same workflow.

Having Pocket add the wording about "IBAN correct?" (or similar) is
a hack/workaround. It would be a better fit to modify SLIP-24 to add a
IBAN memo type, so the BitBox can add specific wording about this. We
don't do this for now to not deviate from the standard, but we might
in the future.

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/bitcoin/payment_request.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/bitcoin/payment_request.rs
@@ -40,7 +40,7 @@ struct Identity {
 
 const IDENTITIES: &[Identity] = &[
     Identity {
-        name: "Pocket Bitcoin",
+        name: "POCKET",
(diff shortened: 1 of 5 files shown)


benma:
@jadzeidan will think of a better text for Pocket to use than "IBAN correct?", but that does not require a fw change.

benma:
@NickeZ I couldn't easily add an example to send_message.py as the Python lib is not adapted yet to support the payment requests.

The example script in bitbox02-api-go works. You just need to enable the Test Merchant in your firmware first.

See https://github.com/BitBoxSwiss/bitbox02-api-go/pull/106

NickeZ (approved):
tACK!

PR #1286, opened 2024-08-28
DX: Move clang-tidy config to config file

NickeZ:
Clangd (C LSP) will run some of the clang-tidy checks if it finds the config in the right place. This gives the developer immediate feedback.


benma (approved):
utACK

PR #1290, opened 2024-08-28
Update to CMSIS 6.1

NickeZ:
Upgrade CMSIS to 6.1.

This PR also solves the `clang-tidy`, `vfpcc`, issue. I guess the headers are changed to work better with clang.

bootloader.bin and firmware.bin do not change after this upgrade.


NickeZ:
Replaces https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1289

benma (approved):
I tested that the firmware binary does not change with this.

Would be nice if you could add info about how exactly you upgraded (where from, any modifications that were needed, ...) in the commit msg.

NickeZ:
> I tested that the firmware binary does not change with this.
> 
> Would be nice if you could add info about how exactly you upgraded (where from, any modifications that were needed, ...) in the commit msg.

I wrote a more elaborate commit msg. Will wait for CI before merging.

benma:
> > I tested that the firmware binary does not change with this.
> > Would be nice if you could add info about how exactly you upgraded (where from, any modifications that were needed, ...) in the commit msg.
> 
> I wrote a more elaborate commit msg. Will wait for CI before merging.

Perfect, thanks

PR #1293, opened 2024-08-31
Generate rust bindings in rust build scripts

NickeZ:
Various improvements, see commit msgs


benma (approved):
utACK

benma (changes requested):
CI fails

```
error: environment variable `CMAKE_CURRENT_BINARY_DIR` not defined at compile time
  --> bitbox02-sys/src/lib.rs:24:5
   |
24 |     env!("CMAKE_CURRENT_BINARY_DIR"),
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

NickeZ:
Ready for review again. It was a mistake to remove that environment variable in the cmake file.

NickeZ:
I realized that I forgot to forward an env variable (`CMAKE_SYSROOT`) in a couple of places.

NickeZ:
When looking into if the binary output was the same before and after this PR I realized that adding a build-script has an impact on symbol names and eventually impact on where functions are placed in the final binary.

So the minimal commit (95529d543e33136c8e2de2fab03a007b95a5b1d3) changes the firmware hash, but the second commit that actually moves the binding generation does not.

benma (approved):
utACK

PR #1294, opened 2024-09-02
rust: Running `make generate-protobufs` in build dir

NickeZ:
It seems we have updated the protobuf generator without rerunning the generation of protobufs.


benma (commented):
I thought `make clean; make firmware` re-generates them. Do you know why that does not happen?

benma (approved):

NickeZ:
> I thought `make clean; make firmware` re-generates them. Do you know why that does not happen?

Hmm, many thoughts. I can think of at least two things. The output (`*.rs` files) is in the source folder and therefore not removed by `make clean`. (`cd build; make clean`, does delete them, and that was how I ran into this issue). I guess the output is considered "up to date" since those files could be created later than the pb-files. Perhaps if you clone the repo many times it will sometimes consider them out-of-date.

benma:
> Hmm, many thoughts. I can think of at least two things. The output (`*.rs` files) is in the source folder and therefore not removed by `make clean`. (`cd build; make clean`, does delete them, and that was how I ran into this issue). I guess the output is considered "up to date" since those files could be created later than the pb-files. Perhaps if you clone the repo many times it will sometimes consider them out-of-date.

I see, it looks at if the .proto files are newer than the output files and regenerates in that case, which does not happen if we update prost. Thanks!

PR #1295, opened 2024-09-02
Nickez/rust version optional

NickeZ:
I broke this out of https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1293 since it has an impact on the final binary.

Unfortunately adding a `build.rs` script, even if it doesn't do anything, ends up changing symbol names. This in turn means that the linker outputs a slightly different binary where functions are placed differently.

I put the build script in its own commit so that you can verify that this is the reason for the binary output difference.

depends on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1294.


benma (approved):
utACK

PR #1297, opened 2024-09-02
rust: don't make bitcoin crate optional

benma:
It is used in bip32.rs, which is compiled unconditionally, so it is not only used by app-bitcoin.


NickeZ (approved):
lgtm

PR #1299, opened 2024-09-04
cardano: Allow vote delegation

RostarMarek:
Allow cardano based on [conway.cddl](https://github.com/IntersectMBO/cardano-ledger/blob/cardano-ledger-conway-1.12.0.0/eras/conway/impl/cddl-files/conway.cddl#L273) to sign messages containing Vote Delegation certificate. This is needed for withdrawing rewards.

I will add tests for firmware soon, just wanted to begin the review process as soon as possible. It has been tested through simulator and works as expected.

### DIFF
--- a/messages/cardano.proto
+++ b/messages/cardano.proto
@@ -85,16 +85,30 @@ message CardanoSignTransactionRequest {
     repeated AssetGroup asset_groups = 4;
   }
 
-  // See https://github.com/input-output-hk/cardano-ledger-specs/blob/d0aa86ded0b973b09b629e5aa62aa1e71364d088/eras/alonzo/test-suite/cddl-files/alonzo.cddl#L150
+  // See https://github.com/IntersectMBO/cardano-ledger/blob/cardano-ledger-conway-1.12.0.0/eras/conway/impl/cddl-files/conway.cddl#L273
   message Certificate {
     message StakeDelegation {
       repeated uint32 keypath = 1;
       bytes pool_keyhash = 2;
     }
+    message VoteDelegation {
+      enum CardanoDRepType {
+        KEY_HASH = 0;
+        SCRIPT_HASH = 1;
+        ALWAYS_ABSTAIN = 2;
+        ALWAYS_NO_CONFIDENCE = 3;
+      }
+
+      // keypath in this instance refers to stake credential
+      repeated uint32 keypath = 1;
+      CardanoDRepType type = 2;
+      optional bytes drep_credhash = 3;
+    }
     oneof cert {
       Keypath stake_registration = 1;
       Keypath stake_deregistration = 2;
       StakeDelegation stake_delegation = 3;
+      VoteDelegation vote_delegation = 10;
     }
   }
 
--- a/src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs
@@ -165,6 +165,45 @@ pub fn encode_transaction_body<W: Write>(
                     encode_stake_credential(&mut encoder, keypath)?;
                     encoder.bytes(pool_keyhash)?;
                 }
+                certificate::Cert::VoteDelegation(certificate::VoteDelegation {
+                    keypath,
+                    r#type,
+                    drep_credhash,
+                }) => {
+                    encoder.array(3)?.u8(9)?;
+                    encode_stake_credential(&mut encoder, keypath)?;
+                    let drep_type = certificate::vote_delegation::CardanoDRepType::try_from(*r#type)?;
+                    match drep_type {
+                        certificate::vote_delegation::CardanoDRepType::KeyHash => {
+                            encoder.array(2)?.u8(0)?;
+                            match drep_credhash {
+                                Some(hash) if hash.len() == 28 => { encoder.bytes(hash)?;}
+                                Some(_hash) => return Err(Error::InvalidInput),
+                                None => return Err(Error::InvalidInput),
+                            }
+                        }
+                        certificate::vote_delegation::CardanoDRepType::ScriptHash => {
+                            encoder.array(2)?.u8(1)?;
+                            match drep_credhash {
+                                Some(hash) if hash.len() == 28 => { encoder.bytes(hash)?;}
+                                Some(_hash) => return Err(Error::InvalidInput),
+                                None => return Err(Error::InvalidInput),
+                            }
+                        }
+                        certificate::vote_delegation::CardanoDRepType::AlwaysAbstain => {
+                            if !drep_credhash.is_none() {
+                                return Err(Error::InvalidInput);
+                            }
+                            encoder.array(1)?.u8(2)?;
+                        }
+                        certificate::vote_delegation::CardanoDRepType::AlwaysNoConfidence => {
+                            if !drep_credhash.is_none() {
+                                return Err(Error::InvalidInput);
+                            }
+                            encoder.array(1)?.u8(3)?;
+                        }
+                    }
+                }
             }
         }
     }
--- a/src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/certificates.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/certificates.rs
@@ -21,7 +21,7 @@ use alloc::vec::Vec;
 
 use pb::cardano_sign_transaction_request::{
     certificate,
-    certificate::Cert::{StakeDelegation, StakeDeregistration, StakeRegistration},
+    certificate::Cert::{StakeDelegation, StakeDeregistration, StakeRegistration, VoteDelegation},
     Certificate,
 };
 
@@ -87,6 +87,48 @@ pub async fn verify<'a>(
                     ..Default::default()
                 })
                 .await?;
+            }
+            VoteDelegation(certificate::VoteDelegation {
+                keypath,
+                r#type,
+                drep_credhash,
+            }) => {
+                validate_address_shelley_stake(keypath, Some(bip44_account))?;
+                signing_keypaths.push(keypath);
+                let drep_type_name =  match certificate::vote_delegation::CardanoDRepType::try_from(*r#type)? {
+                   certificate::vote_delegation::CardanoDRepType::KeyHash => "Key Hash",
+                   certificate::vote_delegation::CardanoDRepType::ScriptHash => "Script Hash",
+                   certificate::vote_delegation::CardanoDRepType::AlwaysAbstain => "Always Abstain",
+                   certificate::vote_delegation::CardanoDRepType::AlwaysNoConfidence => "Always No Confidence",
+                };
+                match drep_credhash {
+                    Some(hash) => { confirm::confirm(&confirm::Params {
+                        title: params.name,
+                        body: &format!(
+                            "Delegate voting for account #{} to type {} and drep {}?",
+                            keypath[2] + 1 - HARDENED,
+                            drep_type_name,
+                            hex::encode(hash),
+                        ),
+                        scrollable: true,
+                        accept_is_nextarrow: true,
+                        ..Default::default()
+                    })
+                    .await?; }
+                    None => { confirm::confirm(&confirm::Params {
+                        title: params.name,
+                        body: &format!(
+                            "Delegate voting for account #{} to type {}?",
+                            keypath[2] + 1 - HARDENED,
+                            drep_type_name,
+                        ),
+                        scrollable: true,
+                        accept_is_nextarrow: true,
+                        ..Default::default()
+                    })
+                    .await?; }
+                }
+
             }
         };
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ### [Unreleased]
 - Bitcoin: add support for sending to silent payment (BIP-352) addresses
 - Bitcoin: add support for regtest
+- Cardano: add support for vote delegation
 
 ### 9.20.0
 - Bitcoin: UX improvements for payment request confirmations
--- a/py/bitbox02/bitbox02/communication/generated/cardano_pb2.py
+++ b/py/bitbox02/bitbox02/communication/generated/cardano_pb2.py
@@ -14,15 +14,15 @@ _sym_db = _symbol_database.Default()
 from . import common_pb2 as common__pb2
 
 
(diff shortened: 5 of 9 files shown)


RostarMarek:
Tried testing it through devdevice, but even in docker environment when running `make firmware` and `make bootloader` there was an error that was strange 
```
dockeruser@aea6becef83c:/Users/marekrostar/git/forks/bitbox02-firmware$ make firmware
mkdir -p build
cd build && cmake -DCMAKE_TOOLCHAIN_FILE=arm.cmake ..
-- size: /opt/homebrew/bin/arm-none-eabi-size arm-none-eabi-size
CMake Error at CMakeLists.txt:116 (message):
  get_version firmware failed


-- Configuring incomplete, errors occurred!
See also "/Users/marekrostar/git/forks/bitbox02-firmware/build/CMakeFiles/CMakeOutput.log".
See also "/Users/marekrostar/git/forks/bitbox02-firmware/build/CMakeFiles/CMakeError.log".
make: *** [Makefile:26: build/Makefile] Error 1
dockeruser@aea6becef83c:/Users/marekrostar/git/forks/bitbox02-firmware$ make bootloader
mkdir -p build
cd build && cmake -DCMAKE_TOOLCHAIN_FILE=arm.cmake ..
-- size: /opt/homebrew/bin/arm-none-eabi-size arm-none-eabi-size
CMake Error at CMakeLists.txt:116 (message):
  get_version firmware failed


-- Configuring incomplete, errors occurred!
See also "/Users/marekrostar/git/forks/bitbox02-firmware/build/CMakeFiles/CMakeOutput.log".
See also "/Users/marekrostar/git/forks/bitbox02-firmware/build/CMakeFiles/CMakeError.log".
make: *** [Makefile:26: build/Makefile] Error 1
dockeruser@aea6becef83c:/Users/marekrostar/git/forks/bitbox02-firmware$ make bootloader-devdevice
make: *** No rule to make target 'bootloader-devdevice'.  Stop.
dockeruser@aea6becef83c:/Users/marekrostar/git/forks/bitbox02-firmware$ make firmware
mkdir -p build
cd build && cmake -DCMAKE_TOOLCHAIN_FILE=arm.cmake ..
-- size: /opt/homebrew/bin/arm-none-eabi-size arm-none-eabi-size
CMake Error at CMakeLists.txt:116 (message):
  get_version firmware failed


-- Configuring incomplete, errors occurred!
See also "/Users/marekrostar/git/forks/bitbox02-firmware/build/CMakeFiles/CMakeOutput.log".
See also "/Users/marekrostar/git/forks/bitbox02-firmware/build/CMakeFiles/CMakeError.log".
make: *** [Makefile:26: build/Makefile] Error 1
dockeruser@aea6becef83c:/Users/marekrostar/git/forks/bitbox02-firmware$ ./scripts/get_version 
py-bitbox02-6.3.0+200.dirty
dockeruser@aea6becef83c:/Users/marekrostar/git/forks/bitbox02-firmware$ ./scripts/get_version firmware
v9.19.0+131.dirty
dockeruser@aea6becef83c:/Users/marekrostar/git/forks/bitbox02-firmware$ ./scripts/get_version firmware --check-semver --check-gpg
```

looking into the error logs this error was reported
```
/usr/local/bin/../lib/gcc/arm-none-eabi/8.2.1/../../../../arm-none-eabi/bin/ld: /usr/local/bin/../lib/gcc/arm-none-eabi/8.2.1/../../../../arm-none-eabi/lib/libc.a(lib_a-exit.o): in function `exit':
exit.c:(.text.exit+0x2c): undefined reference to `_exit'
collect2: error: ld returned 1 exit status
```
which seems like the cmake have and the gcc specifically have some sort of internal error.

NickeZ:
Did you run this command? `git config --global --add safe.directory <the path to project>`? I seem to have to run it every time I run a new container.

RostarMarek:
> Did you run this command? `git config --global --add safe.directory <the path to project>`? I seem to have to run it every time I run a new container.

Not initially, but it was needed to build the simulator so eventually yes. The command `make simulator -j` works.

benma:
> Did you run this command? `git config --global --add safe.directory <the path to project>`? I seem to have to run it every time I run a new container.

Interesting, I guess I am not running into this as I use `podman` :thinking: 

We apply this fix here btw: https://github.com/BitBoxSwiss/bitbox02-firmware/blob/b399d44a64578c893fd3ff969e4c0222562e30d1/releases/build.sh#L51

Maybe it could be added in dockerenv.sh too so it automatically runs there too.

messages/cardano.proto
```diff
@@ -30,6 +30,13 @@ enum CardanoNetwork {
   CardanoTestnet = 1;
 }
 
+enum CardanoDRepType {
+  KeyHash = 0;
+  ScriptHash = 1;
+  AlwaysAbstain = 2;
+  AlwaysNoConfidence = 3;
```
benma, line 37: Unfortunately protobuf enum fields live in the global namespace, so all fields need a namespace prefix :( Please add `CARDANO_DREP_TYPE_`. In Rust, the protobuf generator will strip the prefix again I think, so it's not too bad.

If this message will never be used outside of the VoteDelegation message, you can also define it within VoteDelegation. In that case, you don't need the prefix, as the enum will be scoped.

Afaik the naming convention for enum fields is all uppercase too, so `KEY_HASH`, etc.

messages/cardano.proto
```diff
@@ -85,16 +92,22 @@ message CardanoSignTransactionRequest {
     repeated AssetGroup asset_groups = 4;
   }
 
-  // See https://github.com/input-output-hk/cardano-ledger-specs/blob/d0aa86ded0b973b09b629e5aa62aa1e71364d088/eras/alonzo/test-suite/cddl-files/alonzo.cddl#L150
+  // See https://github.com/IntersectMBO/cardano-ledger/blob/cardano-ledger-conway-1.12.0.0/eras/conway/impl/cddl-files/conway.cddl#L273
   message Certificate {
     message StakeDelegation {
-      repeated uint32 keypath = 1;
+      repeated uint32 stake_keypath = 1;
```
benma, line 98: Seems like an unrelated change that breaks downstream libraries. Is this necessary?

benma (commented):
Thanks!

I left some early comments, but the direction seems good!

messages/cardano.proto
RostarMarek: It was suggested during internal review to make naming consistent and in vote delegation it makes sense. I am not completely partial to it so I fine with keeping the naming `keypath` in `StakeDelegation` certificate

benma: This is also what seemed to break CI:

```
    --> bitbox02-rust/src/hww/api/cardano/sign_transaction.rs:526:29
    |
526 | ...                   keypath: vec![1852 + HARDENED, 1815 + HARDENED, HARDENED, 2, 0],
    |                       ^^^^^^^ `StakeDelegation` does not have this field
    |
```

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs
```diff
@@ -155,16 +155,52 @@ pub fn encode_transaction_body<W: Write>(
...
+                            Some(hash) => {
+                                encoder.bytes(hash)?;
+                            }
+                            None => return Err(Error::InvalidInput),
+                        }
+                    }
+                    else if matches!(r#type,2){
+                        encoder.array(1)?.u8(2)?;
+                    }
+                    else if matches!(r#type,3){
```
benma, line 197: please apply rustfmt on save to format the Rust code.

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs
```diff
@@ -155,16 +155,52 @@ pub fn encode_transaction_body<W: Write>(
...
                     encoder.bytes(pool_keyhash)?;
                 }
+                certificate::Cert::VoteDelegation(certificate::VoteDelegation {
+                    stake_keypath,
+                    r#type,
+                    drep_credhash,
+                }) => {
+                    encoder.array(3)?.u8(9)?;
+                    encode_stake_credential(&mut encoder, stake_keypath)?;
+                    if matches!(r#type,0){
```
benma, line 175: you can decode this to have a nicer pattern matching:

```rust
let drep_type = CardanoDRepType::try_from(r#type)?;
match drep_type {
    CardanoDRepType::KeyHash => { ... }
    // etc
}

File: src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs [hunk 5]
```diff
@@ -155,16 +155,52 @@ pub fn encode_transaction_body<W: Write>(
...
+                        encoder.array(2)?.u8(0)?;
+                        match drep_credhash {
+                            Some(hash) => {
+                                encoder.bytes(hash)?;
```
benma, line 179: should also check the expected length of the hash.

instead of making it `optional ...`, you could remove that keyword and check the length only. clients that omit it default to the empty bytestring.

messages/cardano.proto
RostarMarek: I do not think that in forseeable future with the current business case I was told of this will be used elsewhere, so I moved it into `VoteDelegation`. I put it on top with the casing based on other protobuf files and the naming in `CardanoNetwork` to keep consistency, but this should be nicer.

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs
RostarMarek: Added checking of length and I think using `optional` makes more sense here and makes it a bit more readable, maybe just because I dont like to rely on defaults from languages/libraries.

I'm thinking of adding a check in the cases where there should not be a `drep_credhash` sent to check it is none or throw an error. Functionality wise it does not change anything(even if it was sent it will not be encoded), but it might be nice to keep the requests clean. What do you think?

messages/cardano.proto
benma: Thanks. Indeed, CardanoNetwork uses a different style unfortunately. I checked, most other enums in the project are uppercase.

messages/cardano.proto
benma: I'd leave it as it was, as otherwise:

- renaming it here means that downstream libraries need to change and get a major version bump, and the Adalite integration with the bitbox-api NPM package then also needs to update the name.
- this update would seem to be inconsistent with the `stake_registration` and `stake_deregistration` fields, where the name is also still `keypath`.

Could it make sense to simply name the VoteDelegation keypath field `keypath` instead for consistency?

The TypeScript types right now are derived from the .proto field names like this:

```TypeScript
type CardanoCertificate =

  | {

      stakeRegistration: {

        keypath: Keypath

      }

    }

  | {

      stakeDeregistration: {

        keypath: Keypath

      }

    }

  | {

      stakeDelegation: {

        keypath: Keypath

        poolKeyhash: Uint8Array

      }

    };
```

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs
benma: Sounds good to me :+1:

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs
```diff
@@ -165,6 +165,47 @@ pub fn encode_transaction_body<W: Write>(
...
+                    encode_stake_credential(&mut encoder, stake_keypath)?;
+                    let drep_type = certificate::vote_delegation::CardanoDRepType::try_from(*r#type)?;
+                    match drep_type {
+                        certificate::vote_delegation::CardanoDRepType::KeyHash => {
+                            encoder.array(2)?.u8(0)?;
+                            match drep_credhash {
+                                Some(hash) => {
+                                    if hash.len() != 28 {
+                                        return Err(Error::InvalidInput);
+                                    }
```
benma, line 183: can also shorten this as `Some(hash) if hash.len() == 28 => { ... }`

messages/cardano.proto
RostarMarek: I think it can make sense, but I will add a comment above it so at least in code there is a comment that it is supposed to be `stake_keypath`

RostarMarek:
Also just to let it be known here this PR adds support for the whole certificate, Ada lite currently supports only `AlwaysAbstain`  but I think it is more valid to add support for all possible options instead of allowing just this option in firmware.

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/certificates.rs
```diff
@@ -88,6 +88,25 @@ pub async fn verify<'a>(
...
                 .await?;
             }
+            VoteDelegation(certificate::VoteDelegation {
+                keypath,
+                r#type: _,
+                drep_credhash: _,
+            }) => {
+                validate_address_shelley_stake(keypath, Some(bip44_account))?;
+                signing_keypaths.push(keypath);
+                confirm::confirm(&confirm::Params {
```
benma, line 98: Don't we also have to confirm the drep type and drep credhash?

benma (commented):
Please rebase and also add a line to CHANGELOG.md. Thanks!

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/certificates.rs
RostarMarek: Added a second confirm screen to do this, I didn't want to merge into one because it made it a bit unreadable in my opinion.

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/certificates.rs
```diff
@@ -88,6 +88,57 @@ pub async fn verify<'a>(
...
+                        scrollable: true,
+                        accept_is_nextarrow: true,
+                        ..Default::default()
+                    })
+                    .await?;
+                match drep_credhash {
+                    Some(hash) => { confirm::confirm(&confirm::Params {
+                        title: params.name,
+                        body: &format!(
+                            "to type #{} and drep #{}?",
```
benma, line 119: why the hashtag `#`?

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/certificates.rs
```diff
@@ -88,6 +88,57 @@ pub async fn verify<'a>(
...
+                let drep_type_name =  match certificate::vote_delegation::CardanoDRepType::try_from(*r#type)? {
+                   certificate::vote_delegation::CardanoDRepType::KeyHash => "Key Hash",
+                   certificate::vote_delegation::CardanoDRepType::ScriptHash => "Script Hash",
+                   certificate::vote_delegation::CardanoDRepType::AlwaysAbstain => "Always Abstain",
+                   certificate::vote_delegation::CardanoDRepType::AlwaysNoConfidence => "Always No Confidence",
+                };
+                confirm::confirm(&confirm::Params {
+                        title: params.name,
+                        body: &format!(
+                            "Delegate voting for account #{}?",
```
benma, line 107: I think UX would be a bit better if it was just a single line - it is already scrollable. Then the user does not have to worry about which type/drep it is, they can't see ahead.

If you think two screens are better anyway, it's better to not start it with "to ...", but maybe simply "Type: .... DRep: ...".

benma (changes requested):
Please also rebase again, a change was merged that modifies some protobuf files too, it will make it easire to copy them to bitbox-api-rs.

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/certificates.rs
RostarMarek: I just now realized it was there to signify number in the message so I removed them.

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/certificates.rs
RostarMarek: I am split as I think 2 screens are a little bit more readable, but also a bit more confusing as I am not sure how many users will actually know what which type of delegation means so it might just unnecessarily confuse them. I made it 1 screen in the end.

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/certificates.rs
```diff
@@ -88,6 +88,48 @@ pub async fn verify<'a>(
...
+                   certificate::vote_delegation::CardanoDRepType::KeyHash => "Key Hash",
+                   certificate::vote_delegation::CardanoDRepType::ScriptHash => "Script Hash",
+                   certificate::vote_delegation::CardanoDRepType::AlwaysAbstain => "Always Abstain",
+                   certificate::vote_delegation::CardanoDRepType::AlwaysNoConfidence => "Always No Confidence",
+                };
+                match drep_credhash {
+                    Some(hash) => { confirm::confirm(&confirm::Params {
+                        title: params.name,
+                        body: &format!(
+                            "Delegate voting for account #{} to type {} and drep {}??",
```
benma, line 108: typo: double `??` :smile:

benma (changes requested):
final nit before merge :innocent:

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/certificates.rs
RostarMarek: Yes sorry, fix pushed

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs
```diff
@@ -165,6 +165,45 @@ pub fn encode_transaction_body<W: Write>(
...
+                                Some(_hash) => return Err(Error::InvalidInput),
+                                None => return Err(Error::InvalidInput),
+                            }
+                        }
+                        certificate::vote_delegation::CardanoDRepType::ScriptHash => {
+                            encoder.array(2)?.u8(1)?;
+                            match drep_credhash {
+                                Some(hash) if hash.len() == 28 => { encoder.bytes(hash)?;}
+                                Some(_hash) => return Err(Error::InvalidInput),
+                                None => return Err(Error::InvalidInput),
```
benma, line 190: could collapse these into one `_ => return Err ...` arm. Same above

src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs
```diff
@@ -165,6 +165,45 @@ pub fn encode_transaction_body<W: Write>(
...
+                    keypath,
+                    r#type,
+                    drep_credhash,
+                }) => {
+                    encoder.array(3)?.u8(9)?;
+                    encode_stake_credential(&mut encoder, keypath)?;
+                    let drep_type = certificate::vote_delegation::CardanoDRepType::try_from(*r#type)?;
+                    match drep_type {
+                        certificate::vote_delegation::CardanoDRepType::KeyHash => {
+                            encoder.array(2)?.u8(0)?;
```
benma, line 178: I have no good of way of verifying this is correct, so we'll see in an end-to-end test, but

https://github.com/IntersectMBO/cardano-ledger/blob/cardano-ledger-conway-1.12.0.0/eras/conway/impl/cddl-files/conway.cddl#L273

> vote_deleg_cert = (9, stake_credential, drep)

I guess the `encoder.array(3?).u8(9)?` above starts the array and adds the first element. Shouldn't the `stake_credential` and `drep` now be added directly as two elements? Here it seems like you are adding them as an array of 2 elements.

Have you verified somehow that the cbor output here is correct? Could very well be that I am just mis-reading this, it's a bit hairy :sweat_smile:

benma: Nevermind, I just realized the stake_credential part is itself an array of 2!

benma (approved):
LGTM, thanks! Nice job!

PR #1300, opened 2024-09-05
releases: add v9.20.0 assertions

benma:


NickeZ (approved):
I got the same hashes

PR #1301, opened 2024-09-05
Assert 9.20 binaries

NickeZ:


benma (changes requested):
Please rename the file to assertion-nickez2.sig so the name matches the pubkey name.

NickeZ:
> Please rename the file to assertion-nickez2.sig so the name matches the pubkey name.

It's new subkeys so I replaced the old one. I guess that is correct since CI passes?

benma:
> > Please rename the file to assertion-nickez2.sig so the name matches the pubkey name.
> 
> It's new subkeys so I replaced the old one. I guess that is correct since CI passes?

Yes new subkeys should be fine, and your old assertions still pass :+1:

benma (approved):

PR #1302, opened 2024-09-07
bitcoin: add regtest support

benma:
It has a different bech32 address HRP (prefix).

Third party integrations sometimes are missing this and use testnet as a workaround, but proper support is nicer.


NickeZ (approved):
utAck

benma:
I also re-generated the Python protobuf files.

PR #1304, opened 2024-09-18
keystore: remove bip86 tweak flag from C, do it in Rust

benma:
A conflicting PR (minitapscript) also does most Taproot tweaking in Rust, deleting the _schnorr_bip86_keypair util function. We can do this here too to avoid relying on that util function in `keystore_secp256k1_get_private_key()`, doing in in Rust instead.

### DIFF
--- a/src/rust/bitbox02/src/keystore.rs
+++ b/src/rust/bitbox02/src/keystore.rs
@@ -224,16 +224,12 @@ pub fn encode_xpub_at_keypath(keypath: &[u32]) -> Result<Vec<u8>, ()> {
     }
 }
 
-pub fn secp256k1_get_private_key(
-    keypath: &[u32],
-    tweak_bip86: bool,
-) -> Result<zeroize::Zeroizing<Vec<u8>>, ()> {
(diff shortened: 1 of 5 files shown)


src/rust/bitbox02/src/keystore.rs
```diff
@@ -558,27 +554,16 @@ mod tests {
...
             "",
         );
 
         assert_eq!(
-            hex::encode(secp256k1_get_private_key(keypath, false).unwrap()),
+            hex::encode(secp256k1_get_private_key(keypath).unwrap()),
             "4604b4b710fe91f584fff084e1a9159fe4f8408fff380596a604948474ce4fa3"
         );
-
-        // See first test vector in
```
NickeZ, line 573: Why is this test not needed anymore?

benma: It tests the output of the function with the `tweak_bip86` param set to true, but this PR removes this param altogether and no bip86 tweaking is taking place here anymore.

NickeZ (approved):
utACK

PR #1315, opened 2024-10-03
rust: simplify bech32 hrp check

benma:


Thompson1985 (approved):

src/rust/bitbox02-rust/src/hww/api/cardano/address.rs
```diff
@@ -43,9 +43,8 @@ pub const ADDRESS_HASH_SIZE: usize = 28;
 fn decode_shelley_payment_address(params: &params::Params, address: &str) -> Result<Vec<u8>, ()> {
     let result =
         bech32::primitives::decode::CheckedHrpstring::new::<bech32::Bech32>(address).or(Err(()))?;
-    // TODO: use `result.hrp().as_str()` once bech32 has a new release.
-    let hrp: String = result.hrp().char_iter().collect();
-    if hrp != params.bech32_hrp_payment {
+    let hrp = result.hrp();
+    if hrp.as_str() != params.bech32_hrp_payment {
```
Beerosagos, line 47:
```suggestion
    if result.hrp().as_str() != params.bech32_hrp_payment {
```

src/rust/streaming-silent-payments/src/lib.rs
```diff
@@ -77,8 +77,7 @@ fn decode_address(address: &str, expected_hrp: &str) -> Result<SilentPaymentAddr
             .map_err(|_| ())?;
 
     let hrp = decoded_addr.hrp();
-    let hrp: &str = hrp.as_str();
-    if hrp != expected_hrp {
+    if hrp.as_str() != expected_hrp {
```
Beerosagos, line 80: Same here, I think you could get rid of the hrp variable.

Beerosagos (approved):
utACK, suggested to get rid of hrp local vars in the comments, but that's just a nit anyway

src/rust/bitbox02-rust/src/hww/api/cardano/address.rs
benma: Can't do this as as_str() returns a reference to something inside the hrp instance, so the hrp() value needs to stay alive :(.

Beerosagos: I see :+1: thanks

PR #1319, opened 2024-10-16
sdcard: solve backup bug when sd is re-inserted

asi345:
This issue was existent for a long time. While using the BitBox02, when the sdcard is plugged into the device and the user unplugs and replugs it, backup operations (e.g., list backups, restore from backup) failed and throwed error from the firmware side.

Sdcard backup operations first check if the sdcard is inserted before calling sdcard interface funtions. This commit fixes the re-insertion problem by reinitializing the sdcard whenever it is checked if it is inserted.

The problem earlier most probably stems from sdcard being in an unexpected state when it is re-inserted. The forced initialization step fixes the broken states.

### DIFF
--- a/src/sd.c
+++ b/src/sd.c
@@ -20,6 +20,7 @@
 #ifndef TESTING
 #include "driver_init.h"
 #include "sd_mmc.h"
+#include "sd_mmc/sd_mmc_start.h"
 #endif
 
 #include "flags.h"
@@ -107,7 +108,14 @@ static bool _mount(void)
     sd_mmc_resume_clock();
 #endif
     memset(&fs, 0, sizeof(FATFS));
-    if (f_mount(&fs, "SD", 1) == FR_INVALID_DRIVE) {
+    int res = f_mount(&fs, "", 1);
+    if (res == FR_DISK_ERR) {
+#ifndef TESTING
+        sd_mmc_start();
+#endif
+        res = f_mount(&fs, "", 1);
+    }
+    if (res == FR_INVALID_DRIVE) {
 #ifndef TESTING
         sd_mmc_pause_clock();
 #endif
@@ -117,11 +125,11 @@ static bool _mount(void)
 }
 
 /**
- * Unmunts an SD card and pauses the bus clock.
+ * Unmounts an SD card and pauses the bus clock.
  */
 static void _unmount(void)
 {
-    f_mount(NULL, "SD", 1);
+    f_unmount("");
 #ifndef TESTING
     sd_mmc_pause_clock();
 #endif
@@ -268,7 +276,8 @@ bool sd_list_subdir(sd_list_t* list_out, const char* subdir)
         if (list_out->num_files == allocated_files) {
             char** new_list_out_files;
             allocated_files *= 2;
-            new_list_out_files = (char**)realloc(list_out->files, sizeof(char*) * allocated_files);
+            new_list_out_files =
+                (char**)realloc((void*)list_out->files, sizeof(char*) * allocated_files);
             if (new_list_out_files == NULL) {
                 sd_free_list(list_out);
                 Abort("Error: realloc sd_list_subdir");
@@ -295,7 +304,7 @@ void sd_free_list(sd_list_t* list)
         util_zero(list->files[i], strlen(list->files[i]));
         free(list->files[i]);
     }
-    free(list->files);
+    free((void*)list->files);
     list->files = NULL;
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ### [Unreleased]
 - Update manufacturer HID descriptor to bitbox.swiss
 - Ethereum: remove deprecated Goerli network
+- SD card: solve backup bug when sd card is re-inserted
 
 ### 9.21.0
 - Bitcoin: add support for sending to silent payment (BIP-352) addresses


src/rust/bitbox02-rust/src/hww/api/sdcard.rs
```diff
@@ -23,6 +23,7 @@ use crate::workflow::sdcard;
 pub async fn process(
     &pb::InsertRemoveSdCardRequest { action }: &pb::InsertRemoveSdCardRequest,
 ) -> Result<Response, Error> {
+    bitbox02::sd::sdcard_init();
```
benma, line 26: Is there maybe an interrupt or sth to call this when an sdcard is inserted?

If not - would it make sense to call this in sd.c inside the mount function, so it is called whenever an sdcard operation is about to begin? It would be more robust than calling it in the API layer for when we check sd card insertion - a client might never even call this API call.

Is there a cost to calling this function, or is it a no-op if it has already been called?

Also cc @NickeZ, you have probably better ideas than me about how this should work.

benma (commented):
Amazing find!

src/rust/bitbox02-rust/src/hww/api/sdcard.rs
asi345: I was also thinking of some kind of an interrupt when the card is inserted, but how do you implement a sdcard insertion interruption? That is why I'm only calling this function only when it is needed.

I can move it to `_mount` function too, but I just wanted to fix the broken state of the sdcard as early as possible. Maybe other operations can also cause some kind of wrong state, this can fix it for everywhere, if the operation first calls `sd_card_inserted()`.

From what I learned from tracing, there should be a cost of calling this function, but it is less than doing it for the first time. But it is not a big cost I believe.

@NickeZ can have clearer suggestions on these too I think.

asi345:
Also, I don't understand why CI fails (I haven't touched anything related to the error).
And can you please check if the commit successfully addresses problem by trying out these steps:
```
- Have microSD inserted
- Plug in BitBox02
- Click on choose "recover from microSD"
- Go back to main menu (just hit "Enter")
- Take out MicroSD
- Plug MicroSD back in
- Click on "recover from microSD"
```

src/rust/bitbox02-rust/src/hww/api/sdcard.rs
NickeZ: The SDHC controller has an interrupt line, but I think that is more related to DMA / data transfers than card detect.

benma:
> Also, I don't understand why CI fails

It's because our clang-tidy check runs only on the files that are in the diff, which in this case is sdcard.c. Please try to figure out the error/fix. Could just be casting to `(void*)` on first glance.

NickeZ:
I took an even closer look at this today. We are a bit sloppy with the return value of `f_mount`. So we missed that it returns an error message after reinserting the sd card. Since @asi345 pointed out that re-initializing the driver fixes the issue, I found another way to also solve the issue:

```diff
diff --git a/src/sd.c b/src/sd.c
index 9ad53f47..836962df 100644
--- a/src/sd.c
+++ b/src/sd.c
@@ -18,6 +18,7 @@
 #include <string.h>
 
 #ifndef TESTING
+#include "sd_mmc/sd_mmc_start.h"
 #include "driver_init.h"
 #include "sd_mmc.h"
 #endif
@@ -107,7 +108,12 @@ static bool _mount(void)
     sd_mmc_resume_clock();
 #endif
     memset(&fs, 0, sizeof(FATFS));
-    if (f_mount(&fs, "SD", 1) == FR_INVALID_DRIVE) {
+    int res = f_mount(&fs, "", 1);
+    if (res == FR_DISK_ERR) {
+        sd_mmc_start();
+        res = f_mount(&fs, "", 1);
+    }
+    if (res == FR_INVALID_DRIVE) {
 #ifndef TESTING
         sd_mmc_pause_clock();
 #endif
@@ -121,7 +127,8 @@ static bool _mount(void)
  */
 static void _unmount(void)
 {
-    f_mount(NULL, "SD", 1);
+    f_unmount("");
+
 #ifndef TESTING
     sd_mmc_pause_clock();
 #endif

```

* Our unmount function is also incorrectly implemented. There is a helper macro that gets it right. I suggest we change to that.
* I see now that I also removed `SD` from the argument to mount/umount. Empty string should mount the "default" volume. We anyway set `FF_STR_VOLUME_ID` to 0, so we don't have support for string volume names. I guess the correct "logical drive number" in our case is `"0"`.

asi345:
okay, I'm implementing this version that you suggested, and removing the extra rust-c function calls that I have added.

NickeZ (approved):
lgtm, utack

src/sd.c
```diff
@@ -107,7 +108,14 @@ static bool _mount(void)
...
     memset(&fs, 0, sizeof(FATFS));
-    if (f_mount(&fs, "SD", 1) == FR_INVALID_DRIVE) {
+    int res = f_mount(&fs, "", 1);
+    if (res == FR_DISK_ERR) {
+#ifndef TESTING
+        sd_mmc_start();
+#endif
+        res = f_mount(&fs, "", 1);
+    }
+    if (res == FR_INVALID_DRIVE) {
```
benma, line 118: what if res is some other error?

NickeZ: Yeah, that check could be `res != FR_OK`. It would change the behavior of mount to return false in more cases. But that is probably for the good. Right now we are ignoring errors.

src/sd.c
```diff
@@ -107,7 +108,14 @@ static bool _mount(void)
     sd_mmc_resume_clock();
 #endif
     memset(&fs, 0, sizeof(FATFS));
-    if (f_mount(&fs, "SD", 1) == FR_INVALID_DRIVE) {
+    int res = f_mount(&fs, "", 1);
```
NickeZ, line 111: `int` here should probably be `FRESULT`.

asi345:
okay, I'm updating the code according to these reviews, but I am not sure whether I can reopen this pr.
at the worst case I'll open a new pr.

NickeZ:
> okay, I'm updating the code according to these reviews, but I am not sure whether I can reopen this pr. at the worst case I'll open a new pr.

Just make a new branch and a new PR with the fixes!

asi345:
Please check : https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1320 for these fixes

PR #1320, opened 2024-10-31
sdcard: improvement on reinsertion fix

asi345:


benma:
changelog

benma (commented):
utACK - @NickeZ pls double check, is pause_clock ok in all cases?

asi345:
Changelog is already updated in the previous commit, should I also update it here?

benma:
> Changelog is already updated in the previous commit, should I also update it here?

I had already deleted it :P I looked at the wrong branch and didn't see there was already an entry. No need for a second one of course.

NickeZ (approved):
LGTM

Yes, I think it is appropriate to turn of the clock always when you return `false`.

PR #1323, opened 2024-11-11
Nickez/add debug target

NickeZ:
This improves the developer experience. You can now run the following shortcuts in separate terminals.

Let the gdb-server and rtt-client run forever:
```sh
make gdb-server  # launches JLinkGDBServer
```

```sh
make rtt-client  # connects to gdb server with telnet to get the RTT output
```

Whenever you built the firmware, rerun this:
```sh
make run-debug  # runs gdb and flashes the firmware over the gdb-server before executing it
```

To compile the debug firmware, run:
```sh
make firmware-debug
```


benma (changes requested):
Please add docs for how to use this in BUILD.md.


869864 build-debug/bin/firmware.bin
715956	Build/bin/firmware.bin

This new debug firmware uses 849.48kB of the available 864kB, and adds 150kB on top of the release builds. This is super tight - when we add more code to the firmware, the debug build will be too large to fit. Any ideas what to do about it? 

150kB seems generally like way too much for RTT support, any idea what uses so much space? The panic strings, or something else too?

src/rust/Cargo.toml
```diff
@@ -48,3 +48,5 @@ lto = true
 
 [profile.dev]
 opt-level = 'z'
+# Right now lto="thin" takes to much space, we have to have "fat"/true
+lto = true
```
benma, line 52: Maybe `codegen-units = 1` also here would help with the size?

`debug = true` in the release build but not the dev build seems counter-intuitive.

NickeZ:
> Please add docs for how to use this in BUILD.md.
> 
> 869864 build-debug/bin/firmware.bin 715956 Build/bin/firmware.bin
> 
> This new debug firmware uses 849.48kB of the available 864kB, and adds 150kB on top of the release builds. This is super tight - when we add more code to the firmware, the debug build will be too large to fit. Any ideas what to do about it?
> 
> 150kB seems generally like way too much for RTT support, any idea what uses so much space? The panic strings, or something else too?

Debug builds (by default) check for example overflow in arithmetics, so they introduce a lot of new panic places, with that comes of course lots of static strings and probably some formatting.

Setting `debug-assertions = false` in `[profile.dev]` gives a binary size of 774264.

Setting `codege-units = 1` further reduces it to 750568.

We could add both, but eventually the release build will be equivalent to the debug build. So then we can just go back to only using release build and ignore the debug build.

src/rust/Cargo.toml
NickeZ: `debug = true` is the default in debug builds. This is simply to add elf debug info to the elf binary. That debug info is not copied when creating the binary file we eventually flash.

Makefile
```diff
@@ -112,6 +121,14 @@ jlink-flash-firmware-btc: | build
 	JLinkExe -if SWD -device ATSAMD51J20 -speed 4000 -autoconnect 1 -CommanderScript ./build/scripts/firmware-btc.jlink
 jlink-flash-factory-setup: | build
 	JLinkExe -if SWD -device ATSAMD51J20 -speed 4000 -autoconnect 1 -CommanderScript ./build/scripts/factory-setup.jlink
+jlink-flash-firmware-debug: | build
+	JLinkExe -if SWD -device ATSAMD51J20 -speed 4000 -autoconnect 1 -CommanderScript ./build-debug/scripts/firmware.jlink
+jlink-gdb-server:
+	JLinkGDBServer -nogui -if SWD -device ATSAMD51J20 -speed 4000
+rtt-client:
+	telnet localhost 19021
+run:
```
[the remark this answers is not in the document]
NickeZ: do you prefer `debug-run` or `run-debug`? In the future we could also add a shortcut for running the release build with gdb, so `gdb-run` doesn't seem good to me.

benma:
Is there any way to make rust_log in C have a sprintf signature like the screen debug equivalent?

    void screen_sprintf_debug(int duration, const char* fmt, ...) __attribute__((format(printf, 2, 0)));

then one much more conveniently log data, like `rust_log("result: %d", result);".

Makefile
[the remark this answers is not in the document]
benma: `run-debug` would fit better with all our other `run-` commands

NickeZ:
> Is there any way to make rust_log in C have a sprintf signature like the screen debug equivalent?
> 
> ```
> void screen_sprintf_debug(int duration, const char* fmt, ...) __attribute__((format(printf, 2, 0)));
> ```
> 
> then one can much more conveniently log data, like `rust_log("result: %d", result);`
> 
> I guess you could just add a wrapper around rust_log in C that builds the string that is passed to rust_log.

Good idea, I called it `util_log` because it does some stuff in C-land before calling `rust_log`.

benma:
To run the `make run` command inside the Docker container, the libncurses5 package must be installed. Would be useful to put it into the Dockerfile.

NickeZ:
> To run the `make run` command inside the Docker container, the libncurses5 package must be installed. Would be useful to put it into the Dockerfile.

Yeah, I noticed that as well.

benma:
Can you also add a util_log_hex() util function that logs a buffer as a hex string (simliar to screen_print_debug_hex)? :innocent: I find myself needing this a *lot* and calling util_uint8_to_hex() is inconvenient.

src/rust/bitbox02-rust-c/src/lib.rs
```diff
@@ -36,14 +36,60 @@ mod sha2;
...
+/// Don't use this in production, the returned pointer is only valid until this function is called
+/// again. This function is not thread safe.
+///
+/// Usage:
+///    uint8_t arr[2] = {1,2};
+///    util_log("%s", rust_hex_encode(rust_util_bytes(arr, sizeof(arr))));
+///
+#[no_mangle]
+#[cfg(feature = "rtt")]
+pub extern "C" fn rust_hex_encode(bytes: util::Bytes) -> *const ::util::c_types::c_char {
```
benma, line 83: we already have this in util.c

```
void util_uint8_to_hex(const uint8_t* in_bin, const size_t in_len, char* out)
{
    rust_util_uint8_to_hex(
        rust_util_bytes(in_bin, in_len), rust_util_bytes_mut((uint8_t*)out, in_len * 2 + 1));
}
```

with rust_util_uint8_to_hex in util.rs.

NickeZ: named it `rust_dbg_hex` and made it even easier to use from C. More in-line with your first comment.

NickeZ:
i have some bootloader specific changes that I will create separate PRs of.

src/rust/bitbox02-rust-c/src/lib.rs
```diff
@@ -36,14 +36,70 @@ mod sha2;
...
+///
+/// Max `len` is 64. This function panics if `len` is to large.
+///
+/// Usage:
+///    uint8_t arr[2] = {1,2};
+///    util_log("%s", rust_dbg_hex(arr, sizeof(arr)));
+///
+#[no_mangle]
+#[cfg(feature = "rtt")]
+pub extern "C" fn rust_dbg_hex(
```
benma, line 87: Out of curiosity, why did you do this in Rust? Seems a bit over-complicated, the C-equivalent would probably look much simpler right?

NickeZ: Because we have the `hex` crate in rust, and I lean towards rust. So if I can pick I choose rust over C.

I guess if I reuse `util_uint8_to_hex` it could be pretty short in C yes.

benma: util_uint8_to_hex is a wrapper around rust hex :joy: so you would in fact be using rust, just the wrapper would be simpler.

src/rust/bitbox02-rust-c/src/lib.rs
```diff
@@ -36,14 +36,70 @@ mod sha2;
...
+
+#[no_mangle]
+pub extern "C" fn rust_rtt_init() {
+    ::util::log::rtt_init();
+}
+
+// SAFETY: The pointer must point to a null terminated string
+#[no_mangle]
+#[cfg_attr(not(feature = "rtt"), allow(unused))]
+pub extern "C" fn rust_log(ptr: *const ::util::c_types::c_char) {
```
benma, line 60: If you use crate::util::Bytes here as the input argument this function can be safe.

benma:
> > To run the `make run` command inside the Docker container, the libncurses5 package must be installed. Would be useful to put it into the Dockerfile.
> 
> Yeah, I noticed that as well.
> 
> In the end the networking didn't work out for me. So I have to investigate why `--network=host` doesn't work with docker.

Even if you don't find a solution to the networking thing, it would still help me to just install libncurses5 in the docker image so I can use it. Currently I `apt-get update && apt-get install libncurses5` every time.

src/rust/bitbox02-rust-c/src/lib.rs
NickeZ: yeah, but I would say that would be less idiomatic C. When you pass strings in C they are typically null terminated and not ptr+len.

NickeZ:
> > > To run the `make run` command inside the Docker container, the libncurses5 package must be installed. Would be useful to put it into the Dockerfile.
> > 
> > 
> > Yeah, I noticed that as well.
> > In the end the networking didn't work out for me. So I have to investigate why `--network=host` doesn't work with docker.
> 
> Even if you don't find a solution to the networking thing, it would still help me to just install libncurses5 in the docker image so I can use it. Currently I `apt-get update && apt-get install libncurses5` every time.

I'll add that in a separate PR.

src/rust/bitbox02-rust-c/src/lib.rs
benma: True :+1:  You should mark the function `unsafe`.

BUILD.md
```diff
@@ -25,178 +10,304 @@ For security related issues please see [SECURITY.md](SECURITY.md).
...
 
-By default, the library will be installed into /usr/local/lib64 directory under Linux x86\_64.
-If the library is not on the library path by default, you might need to export the following environment variable:
+Run the following command to run GDB. GDB will connect to the J-Link GDB
+server, flash the debug firmware and then start execution from the bootloader
+(as if the device was just plugged in).
 
 ```sh
-export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/lib64/
+make run
```
benma, line 284: `make run-debug`

BUILD.md
```diff
@@ -25,178 +10,304 @@ For security related issues please see [SECURITY.md](SECURITY.md).
...
 ```
 
+> [!TIP]
+> After rebuilding the firmware, exit GDB and rerun `make run` to flash and reset the device.
```
benma, line 288: `make run-debug`

BUILD.md
```diff
@@ -25,178 +10,304 @@ For security related issues please see [SECURITY.md](SECURITY.md).
...
+
+    ./py/send_message.py --simulator
+
+If you choose to create a wallet by restoring a mnemonic, the simulator will automatically use this
+mnemonic:
+
+    boring mistake dish oyster truth pigeon viable emerge sort crash wire portion cannon couple enact box walk height pull today solid off enable tide
+
+
+#### Debugging using the J-Link probe and GDB
```
benma, line 252: would be useful to have some copy/paste code snippets ready for how to log stuff from C and from Rust:

`util_log()` and `::util::log::log!(...)`. 

An example for logging a hex string in C using the new helper function would also be helpful.

benma (approved):
Super mega useful :clap:

BUILD.md
```diff
@@ -25,178 +10,323 @@ For security related issues please see [SECURITY.md](SECURITY.md).
...
+> ```
+> in C you can also format with hex using `util_dbg_hex`:
+> ```c
+> uint8_t arr[] = {1,2};
+> util_log("%s", util_dbg_hex(arr, sizeof(arr)));
+> ```
+> in rust you can format with hex using the built in hex formatter:
+> ```rust
+> let arr = [1, 2];
+> log!("{:02x?}", arr)
```
benma, line 311: This prints `[01, 02]` which is not nice, especially for long hex strings (e.g. 32 bytes). `log!("{}", hex::encode(arr))` is better.

NickeZ: Added that as well. Sometimes I think it is easier if you can count the bytes and then the whitespace is helpful.

PR #1325, opened 2024-11-18
securechip: make securechip_slot_t an internal type

benma:
A step towards making the securechip interface generic so other chips can be used instead of ATECC608.


NickeZ (approved):
utACK

PR #1326, opened 2024-11-20
Secchip rename

benma:
Splitting the two rename commits out of https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1175/ and added a missing rename.

Will squash all commits before merge.


NickeZ (approved):
utACK

PR #1327, opened 2024-11-25
Make securechip an interface

benma:


src/securechip/securechip.c
```diff
@@ -0,0 +1,158 @@
...
+#if APP_U2F == 1
+        _fns.u2f_counter_inc = atecc_u2f_counter_inc;
+#endif
+        _fns.model = atecc_model;
+        break;
+    }
+    return true;
+}
+
+#define ABORT_IF_NULL(fn)                 \
```
NickeZ, line 73: This should probably compile to nothing `#ifdef NDEBUG`. NDEBUG is typically set in release builds. I think we can do it later when we add support for the optiga.

benma: This could also be a function instead of a macro. I guess it does not matter much in this case, but macros always feel a bit uncomfortable to me over normal functions.

src/memory/memory_shared.h
```diff
@@ -102,4 +102,8 @@ void memory_read_shared_bootdata(chunk_shared_t* chunk_out);
  */
 USE_RESULT uint8_t memory_get_screen_type(void);
 
+#define MEMORY_SECURECHIP_TYPE_ATECC 0xFF
+#define MEMORY_SECURECHIP_TYPE_OPTIGA 0x01
+USE_RESULT uint8_t memory_get_securechip_type(void);
```
benma, line 107: @NickeZ why is this in memory_shared and not in memory? shared means shared with the bootloader, but the bootloader does not interact with the chip.

NickeZ: If I may guess I wasn't aware at the time of the difference between the `memory.h` and the `memory_shared.h`. In my defence, `memory_shared.h` doesn't have a file-level comment specifying that it is only for things accessed from both firmware and bootloader.

benma: The struct you modified has that comment :stuck_out_tongue: 

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/8cf1e44c0d1c849d1f617f9d7cc8e5f7ab510df6/src/memory/memory_shared.h#L61

But anyway, I realized that it has to be here anyway, as the bootloader will have to be able to report the securechip model to the host, so we can keep the option open to split the firmware release by securechip model if we run out of binary space.

NickeZ: That isn't a file level comment. See this for example: https://github.com/torvalds/linux/blob/master/include/linux/list.h

src/securechip/securechip.c
NickeZ: Only because it Aborts. If you want the implementation to return it has to be a macro.

PR #1328, opened 2024-11-26
rustfmt

NickeZ:
It seems we missed rustfmt in the CI on these files?


benma (approved):
Indeed, these changes were provided by an external contributor and it seems we don't have a rustfmt CI check.

NickeZ:
We have added it, perhaps that PR sneaked in right before or maybe the PR Action was never rerun on the latest master branch before being merged.

PR #1329, opened 2024-11-27
external: add the optiga-trust-m dependency

benma:
Needed to interact with the Optiga secure chip.

The optiga-trust-m module was forked to the BitBoxSwiss org, along with its own submodule mbedtls, so we are safe against deps disappearing.

### DIFF
--- a/external/optiga_config.h
+++ b/external/optiga_config.h
@@ -1,0 +1,117 @@
+#ifndef _OPTIGA_LIB_CONFIG_BITBOX02_H_
+#define _OPTIGA_LIB_CONFIG_BITBOX02_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    /** @brief OPTIGA CRYPT random number generation feature enable/disable macro */
+    #define OPTIGA_CRYPT_RANDOM_ENABLED
+    /** @brief OPTIGA CRYPT hash feature enable/disable macro */
+    #define OPTIGA_CRYPT_HASH_ENABLED
+    /** @brief OPTIGA CRYPT ECC generate keypair feature enable/disable macro */
+    #define OPTIGA_CRYPT_ECC_GENERATE_KEYPAIR_ENABLED
+    /** @brief OPTIGA CRYPT ECDSA signature feature enable/disable macro */
+    #define OPTIGA_CRYPT_ECDSA_SIGN_ENABLED
+    /** @brief OPTIGA CRYPT verify ECDSA signature feature enable/disable macro */
+    #define OPTIGA_CRYPT_ECDSA_VERIFY_ENABLED
+    /** @brief OPTIGA CRYPT ECDH feature enable/disable macro */
+    #define OPTIGA_CRYPT_ECDH_ENABLED
+    /** @brief OPTIGA CRYPT ECC 521 feature enable/disable macro */
+    #define OPTIGA_CRYPT_ECC_NIST_P_521_ENABLED
+    /** @brief OPTIGA CRYPT ECC Brainpool feature enable/disable macro */
+    #define OPTIGA_CRYPT_ECC_BRAINPOOL_P_R1_ENABLED
+    /** @brief OPTIGA CRYPT TLS PRF sha256 feature enable/disable macro */
+    #define OPTIGA_CRYPT_TLS_PRF_SHA256_ENABLED
+    /** @brief OPTIGA CRYPT TLS PRF sha384 feature enable/disable macro */
+    #define OPTIGA_CRYPT_TLS_PRF_SHA384_ENABLED
+    /** @brief OPTIGA CRYPT TLS PRF sha512 feature enable/disable macro */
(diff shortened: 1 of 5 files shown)


src/CMakeLists.txt
```diff
@@ -473,6 +474,12 @@ if(CMAKE_CROSSCOMPILING)
     target_link_libraries(${elf} PRIVATE --specs=nosys.specs)
 
     target_link_libraries(${elf} PRIVATE ${firmware}_rust_c)
+
+    # Optiga config must be defined both when compiling the optiga lib, and also when compiling our
```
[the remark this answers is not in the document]
NickeZ: Or actually, shielded connection will be available but turned off in the default config:

https://github.com/BitBoxSwiss/optiga-trust-m/blob/856605f06ea824170cf310777c63306bd4d1a921/optiga/include/optiga/optiga_lib_config_m_v3.h#L105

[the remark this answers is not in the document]
benma: I checked that this is needed by pointing it to a non-existing file and compilation failed. I adjusted the comment.

benma:
I've been developing locally on top of this PR for a while now, communicating with the chip, and had no issues. @NickeZ ready for review & merge.

external/optiga_config.h
```diff
@@ -0,0 +1,125 @@
...
+     *         Any other value will lead to error
+     */
+    #define OPTIGA_COMMS_DEFAULT_RESET_TYPE     (1U)
+    
+    /** @brief NULL parameter check.
+     *         To disable the check, undefine the macro
+     */
+    #define OPTIGA_LIB_DEBUG_NULL_CHECK
+    /** @brief Maximum number of instance registration */
+    #define OPTIGA_CMD_MAX_REGISTRATIONS                (0x06)
```
NickeZ, line 83: Could this be 1, does that have any impact on size?

external/optiga_config.h
```diff
@@ -0,0 +1,125 @@
...
+    /** @brief Maximum buffer size required to communicate with OPTIGA */
+    #define OPTIGA_MAX_COMMS_BUFFER_SIZE                (0x615) //1557 in decimal
+
+    /** @brief Macro to enable logger \n
+    * Enable macro OPTIGA_LIB_ENABLE_UTIL_LOGGING for Util Service layer logging     \n
+    * Enable macro OPTIGA_LIB_ENABLE_CRYPT_LOGGING for Crypt Service layer logging     \n
+    * Enable macro OPTIGA_LIB_ENABLE_CMD_LOGGING for Command layer logging     \n
+    * Enable macro OPTIGA_LIB_ENABLE_COMMS_LOGGING for Communication layer logging     */
+    #define OPTIGA_LIB_ENABLE_LOGGING
```
NickeZ, line 92: Should logging be enabled, or should you enable that when you develop? Or perhaps make it conditional on `NDEBUG`?

external/optiga_config.h
```diff
@@ -0,0 +1,125 @@
...
+    /** @brief Macro to enable logger for Crypt service */
+    #define OPTIGA_LIB_ENABLE_CRYPT_LOGGING
+    /** @brief Macro to enable logger for Command layer */
+    #define OPTIGA_LIB_ENABLE_CMD_LOGGING
+    /** @brief Macro to enable logger for Communication layer */
+    #define OPTIGA_LIB_ENABLE_COMMS_LOGGING
+#endif
+/// @endcond
+
+    /* Below are the example macros for protected update not for any feature */
```
NickeZ, line 108: Maybe these examples are not necessary?

NickeZ (commented):
Overall looks good to me, left some comments

external/optiga_config.h
benma: Saves 288 bytes :smile:

external/optiga_config.h
benma: Removing this saves 4440 bytes! Good catch. I haven't seen any logs and have not been missing them, ok to disable for me.

external/optiga_config.h
benma: Yeah, I could remove them without any impact.

NickeZ (approved):
utACK

external/optiga_config.h
benma: @NickeZ actually 0x01 fails, 0x02 works. With 0x01, the second one fails:

```
    util = optiga_util_create(OPTIGA_INSTANCE_ID_0, optiga_lib_callback, NULL);
    if (NULL == util) {
        util_log("couldn't create optiga util");
        return false;
    }

    crypt = optiga_crypt_create(OPTIGA_INSTANCE_ID_0, optiga_lib_callback, NULL);
    if (NULL == crypt) {
        util_log("couldn't create optiga crypt");
        return false;
    }

```

NickeZ:
In the future we could probably reduce the size of the optiga lib by removing support for features as well.

PR #1330, opened 2024-11-27
sd: prompt user for formatting if format is not VFAT compatible

asi345:
This commit inserts a check to validate whether the sdcard where the backup will be put is VFAT compatible or not. VFAT compatilibity is crucial for using the card and lack of it could result in errors in the firmware. Therefore, when setting up the device, this commit validates it and if it is not compatible, it asks user whether he/she is okay with formatting or not. When the user confirms, SD card is formatted accordingly and the fresh SD card is ready to be used by BitBox02.


asi345:
@NickeZ can you check if the format check is sufficient? I saw that fat32 is the common format from my search, but I also want to confirm it. with you.
Btw @benma I don't think the commit adds too much to the firmware size as you worried (if the code is correct ofc 😅)

asi345:
And this is how the prompting screen looks on the device. Let me know what you think!

tempImagenw0awt

NickeZ:
Doesn't our code work with all fat file systems? Doesn't it make more sense to check if `f_mount() == FR_NO_FILESYSTEM`

src/sd.c
```diff
@@ -384,4 +393,3 @@ bool sd_format(void)
     uint8_t work[FF_MAX_SS] = {0};
     return f_mkfs("SD", &params, work, sizeof(work)) == FR_OK;
```
benma, line 394: What is FF_MAX_SS, and what does this param do in general?

Is this suitable for production or was it possibly only a good value for the tests/simulator? Asking b/c I am not familiar, but the sd_format function was originally added for unit tests only.

src/rust/bitbox02/src/sd.rs
```diff
@@ -30,6 +30,20 @@ pub fn sdcard_inserted() -> bool {
...
+pub fn sdcard_vfat_formatted() -> bool {
+    unsafe { bitbox02_sys::sd_card_vfat_formatted() }
+}
+
+#[cfg(feature = "testing")]
+pub fn sdcard_vfat_formatted() -> bool {
+    true
+}
+
+pub fn sd_format() -> bool {
```
benma, line 43: no need for the `sd` prefix, this is in the `sd` module and we don't need prefixes for namespacing like in C.

benma, line 43: `-> Result<(), ()>` is more idiotmatic Rust for error handling. You can see other examples for converting C bools to Results in the bitbox02/src files.

src/rust/bitbox02-rust/src/hww/api/backup.rs
```diff
@@ -89,6 +89,17 @@ pub async fn create(
...
 
+    let is_vfat_formatted = bitbox02::sd::sdcard_vfat_formatted();
+    if !is_vfat_formatted {
+        confirm::confirm(&confirm::Params {
+            title: "SD card\nformatting needed",
+            body: "This will erase all\ndata on the SD card.",
+            ..Default::default()
+        })
+        .await?;
+        bitbox02::sd::sd_format();
```
benma, line 100: one reason to make sd_format() return a Result<> is so that the compiler can warn you if you don't handle the result :grin:

asi345:
> Doesn't our code work with all fat file systems? Doesn't it make more sense to check if `f_mount() == FR_NO_FILESYSTEM`

You meant `f_mount() != FR_NO_FILESYSTEM` ? But does this check mitigate other kinds of error that could be returned?

NickeZ:
> > Doesn't our code work with all fat file systems? Doesn't it make more sense to check if `f_mount() == FR_NO_FILESYSTEM`
> 
> You meant `f_mount() != FR_NO_FILESYSTEM` ? But does this check mitigate other kinds of error that could be returned?

Not sure, what error messages do you get for a drive that is erased or a drive that has an incompatible file system. You can try with some apple fs or ntfs?

src/rust/bitbox02/src/sd.rs
asi345: Okay, converting that to Result object

src/rust/bitbox02-rust/src/hww/api/backup.rs
asi345: True, I intended to ask this in the pr message but forgot. Thanks, now it also handles when format fails 👍

src/sd.c
asi345: > This set of options defines the extent of sector size used for the low level disk I/O interface, disk_read and disk_write function. Valid values are 512, 1024, 2048 and 4096. FF_MIN_SS defines minimum sector size and FF_MAX_SS defines the maximum sector size. Always set both 512 for memory card and harddisk.
 
Taken from FatFs configurations, http://elm-chan.org/fsw/ff/doc/config.html#max_ss.  Only thing that boggled me was "SD" parameter there, but I did not encounter any issue while testing out. The card resetted and continued to work without an error.

asi345:
Okay, I'll try that out with my old windows computer which I can plug the sdcard. Then I'll write the result here.

asi345:
> > > Doesn't our code work with all fat file systems? Doesn't it make more sense to check if `f_mount() == FR_NO_FILESYSTEM`
> > 
> > 
> > You meant `f_mount() != FR_NO_FILESYSTEM` ? But does this check mitigate other kinds of error that could be returned?
> 
> Not sure, what error messages do you get for a drive that is erased or a drive that has an incompatible file system? You can try with some apple fs or ntfs?

I did a bit of research/trial and it seems the possible results for f_mount in that case would be:
- `FR_NO_FILESYSTEM`
- `FR_DISK_ERR`
- `FR_NOT_READY`
- `FR_INVALID_DRIVE`

In this case, I'm feeling like checking `f_mount() == FR_OK` could make sense.

NickeZ:
> > > > Doesn't our code work with all fat file systems? Doesn't it make more sense to check if `f_mount() == FR_NO_FILESYSTEM`
> > > 
> > > 
> > > You meant `f_mount() != FR_NO_FILESYSTEM` ? But does this check mitigate other kinds of error that could be returned?
> > 
> > 
> > Not sure, what error messages do you get for a drive that is erased or a drive that has an incompatible file system? You can try with some apple fs or ntfs?
> 
> I did a bit of research/trial and it seems the possible results for f_mount in that case would be:
> 
>     * `FR_NO_FILESYSTEM`
> 
>     * `FR_DISK_ERR`
> 
>     * `FR_NOT_READY`
> 
>     * `FR_INVALID_DRIVE`
> 
> 
> In this case, I'm feeling like checking `f_mount() == FR_OK` could make sense.

`FR_DISK_ERR` could also mean that the sd-card was re-inserted right?

I think with the message "SD-card needs to be formatted", we can only check `FR_NO_FILESYSTEM`. Otherwise the message must be something like "Failed to read SD-card".

We really, really, don't want people to erase their backups.

There should probably also be two screens, first one confirm/cancel, and then one "pinch gesture" confirm with something like "PERMANENTELY ERASES SD-CARD".

jadzeidan:
@asi345 @NickeZ 

Have you thought of how to avoid false positives? For example, if the the user plugs in a working microSD card with a backup, but it is not detected correctly (we have had many "error detecting mSD card" issues in the past that is usually solved by replugging the device).

Does this prompt only appear explicitly if the BitBox02 detects that the microSD card is in a known incompatible format? Or does the pompt appear anytime there is a mSD card detection error?

jadzeidan:
Regarding the UI. Would be good to use the "hold" gesture instead of the "check mark" since it's a permutant change.

asi345:
I pushed a new commit:

@NickeZ I changed the condition to be `f_mount() != FR_NO_FILESYSTEM`, which only extracts the condition where file system is corrupted/erased.


> And this is how the prompting screen looks on the device. Let me know what you think!
> 
> tempImagenw0awt

The first confirmation still looks like the above. Additionally I added this second confirmation screen:

tempImage5n99zo

If you can, I also suggest you trying out the code and check that you're not getting this prompt 😅

PR #1332, opened 2024-12-02
Nickez/dont boot to ff

NickeZ:
You can test with "Erase firmware" option in send message helper.

```sh
py/send_message.py
3
```

### DIFF
--- a/src/bootloader/bootloader.c
+++ b/src/bootloader/bootloader.c
@@ -900,7 +900,12 @@ static bool _devdevice_enter(secbool_u32 firmware_verified)
     UG_PutString(0, 0, "    <Enter bootloader>", false);
     UG_PutString(0, SCREEN_HEIGHT / 2 - 11, "DEV DEVICE", false);
     UG_PutString(0, SCREEN_HEIGHT / 2 + 2, "NOT FOR VALUE", false);
-    UG_PutString(0, SCREEN_HEIGHT - 9, "        <Continue>", false);
+    // Check that the firmware's reset handler isn't invalid.
+    if (((uint32_t*)FLASH_APP_START)[1] != 0xffffffff) {
+        UG_PutString(0, SCREEN_HEIGHT - 9, "        <Continue>", false);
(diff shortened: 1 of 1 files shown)


src/bootloader/bootloader.c
```diff
@@ -900,7 +900,11 @@ static bool _devdevice_enter(secbool_u32 firmware_verified)
     UG_PutString(0, 0, "    <Enter bootloader>", false);
     UG_PutString(0, SCREEN_HEIGHT / 2 - 11, "DEV DEVICE", false);
     UG_PutString(0, SCREEN_HEIGHT / 2 + 2, "NOT FOR VALUE", false);
-    UG_PutString(0, SCREEN_HEIGHT - 9, "        <Continue>", false);
+    if (*(uint32_t*)FLASH_APP_START != 0xffffffff) {
```
benma, line 903: is the first byte guaranteed not 0xffffffff for normal firmware?

Fyi there is also _compute_is_app_flash_empty if you want to check if the whole firmware area is wiped, though then that should first be refactored to not use a global :scream:

benma (approved):
utACK

src/bootloader/bootloader.c
NickeZ: yes, the first byte is the initial stack pointer, so it should point to the end of the RAM.

Thinking a bit more about it, it is actually the second word I care about, the second word is the Reset_Handler, and that can by definition not be `0xffffffff`. Because jumping there puts the CPU in LOCKUP state, which is not useful. Do you want me to switch to looking at the second word?

NickeZ: I really don't care about the whole firmware. If the initial stack pointer and reset handler are invalid (`0xffffffff`) the bootloader cannot and should not even try to boot it.

NickeZ: > Do you want me to switch to looking at the second word?

I'm looking at the reset handler pointer now

PR #1334, opened 2024-12-03
securechip: replace sign interface function with native Rust impl

benma:
The securechip unsafe sign key slot in the ATECC608 was only used as an alternative to adding firmware code for signing using hte NIST P-256 curve.

The Optiga Trust M does not support the same functionality - one can't write a private key to this chip and extract the corresponding pubkey.

Since we want one interface for both chips, we remove the unsafe sign functionality and replace it with the Rust p256 crate.

It is only used in U2F.

As an alternative, one could use the native MCU PUKCC/PUKCL feature to derive a pubkey and sign using this curve (see pukcc.c), but I could not get it to work. Using the Rust crate is much easier to implement / use.

This adds 10592 bytes of binary space. If we need it back, we should try again to use PUKCC instead and offload these operations to the MCU.

Useful site to test U2f registration & authentication: https://demo.yubico.com/webauthn-technical


NickeZ:
Did you run `vendor.sh` in the container?

benma:
> Did you run `vendor.sh` in the container?

Yes, why?

NickeZ:
> > Did you run `vendor.sh` in the container?
> 
> Yes, why?

you bumped core/alloc

src/atecc/atecc.c
```diff
@@ -34,7 +34,6 @@ typedef enum {
     ATECC_SLOT_ROLLKEY = 3,
     ATECC_SLOT_KDF = 4,
     ATECC_SLOT_ATTESTATION = 5,
-    ATECC_SLOT_ECC_UNSAFE_SIGN = 6,
```
NickeZ, line 37: Maybe we should keep this enumeration around, could rename it to DEPRECATED/RESERVED or something similar. So that we don't forget that it has been used.

NickeZ (approved):
utACK

PR #1335, opened 2024-12-03
Add libncurses5 to container

NickeZ:
This is a dependency for runnning arm-none-eabi-gdb.


Dockerfile
```diff
@@ -70,6 +70,9 @@ RUN apt-get update && apt-get install -y \
     doxygen \
     graphviz
 
+RUN apt-get update && apt-get install -y \
```
benma, line 73: I'd add a comment above this so it's clear what this is for (even though git blame would reveal it)

benma (approved):

Dockerfile
NickeZ: oh, I forgot that

PR #1337, opened 2024-12-09
platform: start timer independently of qtouch

benma:
If factorysetup, qtouch_init() is not called. We still need a timer there to talk to the Optiga Trust M, which requires a timer task for communication.

We start the timer at init time, tasks can be added to the timer afterwards


NickeZ (approved):
utACK!

PR #1344, opened 2024-12-18
securechip/optiga: limit unlock to 10 failed attempts

benma:
See commit messages.


src/optiga/optiga.c
```diff
@@ -0,0 +1,1817 @@
...
+    0x40,
+    // Allow reads, auth referencing 0xF1D3 (`OID_PASSWORD`), enforce shielded connection
+    0xD1,
+    0x07,
+    0x23,
+    0xF1,
+    0xD3,
+    // &&
+    0xFD,
+    0x20,
```
benma, line 355: I need to check if this should be 0x21 Int instead of 0x20 Conf

https://github.com/Infineon/optiga-trust-m-overview/blob/98b2b9c178f0391b1ab26b52082899704dab688a/docs/OPTIGA%E2%84%A2%20Trust%20M%20Solution%20Reference%20Manual.md#linkc15dfea4_2cc2_46ae_a53b_1e6ea9487f34

(test if with the current config, one could actually read without auth, and if then 0x21 Int enforces auth)

src/optiga/optiga.c
```diff
@@ -0,0 +1,1817 @@
...
+    FINAL_LCSO_STATE,
+    // Allow writes, auth referencing 0xF1D2 (`OID_PASSWORD_SECRET`), enforce shielded connection.
+    0xD0,
+    0x07,
+    0x23,
+    0xF1,
+    0xD2,
+    // &&
+    0xFD,
+    0x20,
```
benma, line 443: same, check if 0x21 Int would be right

benma:
Rebased

src/optiga/optiga.c
```diff
@@ -963,6 +1259,164 @@ static int _verify_metadata(
...
+    optiga_lib_status_t res =
+        _authorize(OID_PASSWORD, password_salted_hashed, sizeof(password_salted_hashed));
+    if (res != OPTIGA_LIB_SUCCESS) {
+        goto cleanup;
+    }
+
+    uint16_t password_secret_size = 32;
+    res = _optiga_util_read_data_sync(
+        _util, OID_PASSWORD_SECRET, 0, password_secret_out, &password_secret_size);
+    if (res != OPTIGA_LIB_SUCCESS || password_secret_size != 32) {
```
benma, line 1386: bug: the password size error does not set `res` to an error and so the function could succeed while it was supposed to fail.

src/optiga/optiga.c
benma: I tried to read the object without auth first and it failed with the expected error 0x07 (Auth error), so 0x21 (Int) does not seem necessary. If ChaGPT is to be believed, it would even be worse:

ChatGPT o3-mini-high thinks:

```
0x21 (Int) means the chip only applies a MAC for integrity—no encryption. For both encryption and integrity, use 0x20 (Conf) which establishes a full shielded connection.
```

This could be wrong, though, not sure how to verify it.

benma: https://github.com/Infineon/optiga-trust-m/discussions/133 clarified that ChatGPT was right and Int only applies to shielded connections, but only checks integrity, does not encrypt.

Our use of 0x20 Conf is correct then.

src/optiga/optiga.c
benma: fixed

NickeZ (approved):
tACK with optiga

PR #1345, opened 2024-12-19
simulator: clean shutdown of socket

NickeZ:
If we don't close the socket we have to wait a couple of minutes until the OS cleans up after us. This is quite annoying because then you can't launch the simulator with the same port immediately after you kill it.

### DIFF
--- a/test/simulator/simulator.c
+++ b/test/simulator/simulator.c
@@ -32,6 +32,7 @@
 #include <getopt.h>
 #include <netdb.h>
 #include <netinet/in.h>
+#include <signal.h>
 #include <sys/socket.h>
 
 static const char* _simulator_version = "1.0.0";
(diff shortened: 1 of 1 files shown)


benma (approved):
utACK

NickeZ:
I made some attempts at a clean shutdown when a client still is connected. I think the client has to do a proper shutdown when it detects that the server is down for the OS to release the port, so we might have to do some changes to the client as well. Anyway, it seems complicate enough that I can do that in a follow up PR. This PR still improves the situtation when no clients are connected.

Noxet:
You should avoid using non-reentrant functions, like printf, in signal handlers.
The only functions that POSIX guarantee to be async-signal-safe, e.g. write(), are [listed here](https://docs.oracle.com/cd/E19455-01/806-5257/gen-26/index.html)

NickeZ:
> You should avoid using non-reentrant functions, like printf, in signal handlers. The only functions that POSIX guarantee to be async-signal-safe, e.g. write(), are [listed here](https://docs.oracle.com/cd/E19455-01/806-5257/gen-26/index.html)

addressed in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1353, thanks for the feedback!

PR #1346, opened 2024-12-21
cardano: use custom cbor decoder to save binary space

benma:
Coded by ChatGPT o1. Reduces binary size by 2760 bytes.

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/cardano/address.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/cardano/address.rs
@@ -62,6 +62,166 @@ fn decode_shelley_payment_address(params: &params::Params, address: &str) -> Res
     Ok(data)
 }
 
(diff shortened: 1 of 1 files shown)


Egge21M:
My rust knowledge is still very limited, but I think I do understand this conceptually. 

Usually you you have some logic that reads byte strings dynamically and handles types accordingly. However if this is the only place where we need CBOR and the structure of each expected byte string is strictly defined, this LGTM.

benma:
> My rust knowledge is still very limited, but I think I do understand this conceptually.
> 
> Usually you you have some logic that reads byte strings dynamically and handles types accordingly. However if this is the only place where we need CBOR and the structure of each expected byte string is strictly defined, this LGTM.

Exactly, we only cbor-decode these legacy addresses, only to find out if they are valid and mainnet/testnet. Thanks!

PR #1347, opened 2024-12-23
bootloader: add OP_HARDWARE api endpoint

Beerosagos:
This will allow the bbapp to flash different FW versions, depending on the PCB variants. The initial implementation will respond with a single byte that identifies the Secure Chip: 0 for ATECC, 1 for Optiga.

Note: Based on the changes from #1341

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,9 +151,10 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 
 ## Bootloader
 
-### v1.0.7
+### v1.1.0
 - Update manufacturer HID descriptor to bitbox.swiss
 - Remove qtouch code from production bootloader
+- Implement OP_HARDWARE endpoint, to identify the secure chip model
 
 ### v1.0.6
 - Replace root pubkeys
--- a/py/bitbox02/bitbox02/bitbox02/bootloader.py
+++ b/py/bitbox02/bitbox02/bitbox02/bootloader.py
@@ -18,9 +18,12 @@ import typing
 import io
 import math
 import hashlib
+import enum
+from typing import TypedDict
 
 from bitbox02.communication import TransportLayer
-from bitbox02.communication.devices import DeviceInfo
+
+from bitbox02.communication.devices import DeviceInfo, parse_device_version
 
 BOOTLOADER_CMD = 0x80 + 0x40 + 0x03
 NUM_ROOT_KEYS = 3
@@ -42,6 +45,19 @@ VERSION_LEN = 4
 SIGNING_PUBKEYS_DATA_LEN = VERSION_LEN + NUM_SIGNING_KEYS * 64 + NUM_ROOT_KEYS * 64
 FIRMWARE_DATA_LEN = VERSION_LEN + NUM_SIGNING_KEYS * 64
 SIGDATA_LEN = SIGNING_PUBKEYS_DATA_LEN + FIRMWARE_DATA_LEN
+
+
+class SecureChipModel(enum.Enum):
+    """Secure chip model variants for the BitBox02 platform."""
+
+    ATECC = "ATECC"
+    OPTIGA = "Optiga"
+
+
+class Hardware(TypedDict):
+    """Hardware configuration containing secure chip model information."""
+
+    secure_chip_model: SecureChipModel
 
 
 def parse_signed_firmware(firmware: bytes) -> typing.Tuple[bytes, bytes, bytes]:
@@ -75,6 +91,10 @@ class Bootloader:
             "bb02btc-bootloader": SIGDATA_MAGIC_BTCONLY,
             "bitboxbase-bootloader": SIGDATA_MAGIC_BITBOXBASE_STANDARD,
         }.get(device_info["product_string"])
+        self.version = parse_device_version(device_info["serial_number"])
+        # Delete the prelease part, as it messes with the comparison (e.g. 3.0.0-pre < 3.0.0 is
+        # True, but the 3.0.0-pre has already the same API breaking changes like 3.0.0...).
+        self.version = self.version.replace(prerelease=None)
         assert self.expected_magic
 
     def _query(self, msg: bytes) -> bytes:
@@ -93,6 +113,25 @@ class Bootloader:
         response = self._query(b"v")
         firmware_v, signing_pubkeys_v = struct.unpack("<II", response[:8])
         return firmware_v, signing_pubkeys_v
+
+    def hardware(self) -> Hardware:
+        """
+        Returns (hardware variant).
+        """
+        secure_chip: SecureChipModel = SecureChipModel.ATECC
+
+        # Previous bootloader versions do not support the call and have ATECC SC.
+        if self.version >= "1.1.0":
+            response = self._query(b"W")
+            response_code = response[:1]
+
+            if response_code == b"\x00":
+                secure_chip = SecureChipModel.ATECC
+            elif response_code == b"\x01":
+                secure_chip = SecureChipModel.OPTIGA
+            else:
+                raise ValueError(f"Unrecognized securechip model: {response_code!r}")
+        return {"secure_chip_model": secure_chip}
 
     def get_hashes(
         self, display_firmware_hash: bool = False, display_signing_keydata_hash: bool = False
--- a/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
+++ b/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
@@ -557,9 +557,7 @@ class BitBoxCommonAPI:
 
         # Delete the prelease part, as it messes with the comparison (e.g. 3.0.0-pre < 3.0.0 is
         # True, but the 3.0.0-pre has already the same API breaking changes like 3.0.0...).
-        self.version = semver.VersionInfo(
-            self.version.major, self.version.minor, self.version.patch, build=self.version.build
-        )
+        self.version = self.version.replace(prerelease=None)
 
         # raises exceptions if the library is out of date
         self._check_max_version()
--- a/src/bootloader/bootloader.c
+++ b/src/bootloader/bootloader.c
@@ -67,6 +67,8 @@
 #define OP_SCREEN_ROTATE ((uint8_t)'f') /* 0x66 */
 // OP_SET_SHOW_FIRMWARE_HASH - Enable or disable the flag to automatically show the firmware hash.
 #define OP_SET_SHOW_FIRMWARE_HASH ((uint8_t)'H') /* 0x4A */
+// OP_HARDWARE - Return the secure chip variant.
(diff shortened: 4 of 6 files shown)


py/bitbox02/bitbox02/bitbox02/bootloader.py
```diff
@@ -94,6 +94,15 @@ def versions(self) -> typing.Tuple[int, int]:
         firmware_v, signing_pubkeys_v = struct.unpack("<II", response[:8])
         return firmware_v, signing_pubkeys_v
 
+    def hardware(self) -> str:
```
benma, line 97: Please no stringly types. An enum like this would be better imho:

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py#L187

Also this should return the ATECC variant if the bootloader version does not support this endpoint. Right now this call fails for old bootloaders.

src/bootloader/bootloader.c
```diff
@@ -766,6 +768,16 @@ static size_t _api_screen_rotate(uint8_t* output)
...
 }
 
+static size_t _api_hardware(uint8_t* output)
+{
+    uint8_t type = 0;
+    if (memory_get_securechip_type() == MEMORY_SECURECHIP_TYPE_OPTIGA) {
+        type = 1;
+    }
+    memcpy(output + BOOT_OP_LEN, &type, 1);
+    return BOOT_OP_LEN + 1;
```
benma, line 778: The second byte of the output for all api commands is the success/error byte, but it's missing here. The bootloader `._query()` call checks the second byte for error. It seems that you put the hardware type into this byte, so the Python client does not actually work as expected.

You should add it and set it by using `return _report_status(result, output) + ...;`

The memcpy of a single byte is a bit awkward. You can ust do `output[2] = type`

benma (changes requested):
Obligatory :smile: 
changelog

src/bootloader/bootloader.c
benma: _api_versions is also a bit borked: returns the size of three versions, but only returns two versions :facepalm:

Beerosagos:
> Obligatory 😄 

Let's make a PR template that includes a changelog checkbox, pls pls pls :pray: :joy:

src/bootloader/bootloader.c
Beerosagos: > Looking at some other api calls, `_api_set_show_firmware_hash` is borked - it says it returns a disabled/enabled byte and sets it like this
> 
> ```
> output[BOOT_OP_LEN] = data.fields.show_firmware_hash;
> return _report_status(result, output) + 1;
> ```
> 
> But _report_status call overwrites the second byte with the success byte. The client libraries also don't return any result here. Could you also add a commit or PR that cleans this up?

I don't think this is right. `BOOT_OP_LEN` is 2, so the `show_firmware_hash` result is in the 3rd byte of the array. Am I missing something?

Beerosagos: > _api_versions is also a bit borked: returns the size of three versions, but only returns two versions 🤦

Right :sweat_smile: The Py client is expecting only two versions. For backward compatibility, I guess it's better to change the return size instead of adding the third version. Wdyt?

py/bitbox02/bitbox02/bitbox02/bootloader.py
Beerosagos: Done :+1:

src/bootloader/bootloader.c
benma: The bootloader version is available through the HID descriptor serial field.

benma (changes requested):

CHANGELOG.md
```diff
@@ -150,6 +150,9 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 
 ## Bootloader
 
+### [Unreleased]
+- Implement OP_HARDWARE endpoint, to identify the SC model
```
benma, line 154: SC=secure chip. Imho abbreviations just make it harder to read, not everyone knows what it means.

Beerosagos:
rebased

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -191,6 +191,13 @@ class BitBox02Edition(enum.Enum):
     BTCONLY = "btconly"
 
 
+class Hardware(enum.Enum):
```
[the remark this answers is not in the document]
Beerosagos: Moved

src/bootloader/bootloader.c
Beerosagos: Fixed in a second commit

Beerosagos:
@benma this should be ready for a new review, PTAL :pray:

py/bitbox02/bitbox02/bitbox02/bootloader.py
```diff
@@ -94,6 +104,17 @@ def versions(self) -> typing.Tuple[int, int]:
         firmware_v, signing_pubkeys_v = struct.unpack("<II", response[:8])
         return firmware_v, signing_pubkeys_v
 
+    def hardware(self) -> Hardware:
+        """
+        Returns (hardware variant).
+        """
+        # Previous bootloader versions does not support the call and have ATECC SC.
+        if self.version > "1.0.6":
```
benma, line 112: ">= 1.1.0".

py/bitbox02/bitbox02/bitbox02/bootloader.py
```diff
@@ -94,6 +104,17 @@ def versions(self) -> typing.Tuple[int, int]:
...
+            response = self._query(b"W")
+            if response[:2] == b"\x01":
```
benma, line 114: how can two bytes equal one byte? :grin:

py/bitbox02/bitbox02/bitbox02/bootloader.py
Beerosagos: I did this initially, but it wasn't working well with dev version, e.g. `v1.1.0-pre+48561ef480.dev >= 1.1.0`, seemed to return false in my tests.

py/bitbox02/bitbox02/bitbox02/bootloader.py
Beerosagos: Shoot, how did I miss this? :see_no_evil:

Beerosagos: Lol, it worked by coincidence because the `query` call already strips the first 2 bytes, so `response[:2]` was actually 1 byte long

py/bitbox02/bitbox02/bitbox02/bootloader.py
Beerosagos: I checked again, it doesn't work for versions 1.1.0 with prerelease. I'd keep > 1.0.6, if you agree

py/bitbox02/bitbox02/bitbox02/bootloader.py
Beerosagos: Fixed :+1:

py/bitbox02/bitbox02/bitbox02/bootloader.py
benma: Check this out:

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/c016621b126c61c9c94e16ec0cf799d25c042bbd/py/bitbox02/bitbox02/communication/bitbox_api_protocol.py#L557-L562

The `>1.0.6` way does not work consistently, so imho better to apply the above here as well.

Beerosagos: Ah, that's nice! But why `>1.0.6` doesn't work consistently? Anyway, your suggestion sounds better :+1:

Beerosagos: Fixed. I found a more readble/compact way to remove the pre-release `self.version.replace(prerelease=None)`. Should I update also the code you linked while I'm here?

benma: >  Ah, that's nice! But why >1.0.6 doesn't work consistently? Anyway, your suggestion sounds better 👍

1.0.7 could be released that patches something in 1.0.6 before 1.1.0 goes out, for example. Less likely in the bootloader, but did happen in the firmware a few times.

Beerosagos: @benma Good point, I didn't think about it. Anyway, I updated the code. Hopefully this could be the last round of review :pray:

py/bitbox02/bitbox02/bitbox02/bootloader.py
```diff
@@ -75,6 +84,9 @@ def __init__(self, transport: TransportLayer, device_info: DeviceInfo):
             "bb02btc-bootloader": SIGDATA_MAGIC_BTCONLY,
             "bitboxbase-bootloader": SIGDATA_MAGIC_BITBOXBASE_STANDARD,
         }.get(device_info["product_string"])
+        self.version = parse_device_version(device_info["serial_number"])
+        # we remove the prerelease as it messes up same-version comparisons.
```
benma, line 88: I thought the other comment was clearer :innocent: 

```
        # Delete the prelease part, as it messes with the comparison (e.g. 3.0.0-pre < 3.0.0 is
        # True, but the 3.0.0-pre has already the same API breaking changes like 3.0.0...).
```

benma (commented):
Looking good, but I still want to test it. Left a tiny nit.

py/bitbox02/bitbox02/bitbox02/bootloader.py
Beerosagos: Copied the same comment here :+1:

benma (approved):
tACK

py/bitbox02/bitbox02/bitbox02/bootloader.py
```diff
@@ -94,6 +107,17 @@ def versions(self) -> typing.Tuple[int, int]:
         firmware_v, signing_pubkeys_v = struct.unpack("<II", response[:8])
         return firmware_v, signing_pubkeys_v
 
+    def hardware(self) -> Hardware:
```
benma, line 110: When adding the same API to the Go package I realized that this endpoint could return more HW info in the future, so having Hardware==Securechip like here will be a breaking change when that happens. It's not that important as I think there is no production use of the bootloader Python API, but technically Hardware should be an object (TypedDict?) with a SecureChipModel enum inside.

https://github.com/BitBoxSwiss/bitbox02-api-go/pull/116/

Beerosagos: Done, I also updated the check of the response to print unexpected values, so it behaves the same as your PR

PR #1350, opened 2024-12-28
Cardano: allow serialization using 258-tagged sets

benma:
Since the Conway era, the CBOR serialization of inputs and certificates
can optionally be tagged with the 258-tag. This will become mandatory
in a future hardfork.

- https://github.com/IntersectMBO/cardano-ledger/blob/6e2d37cc0f47bd02e89b4ce9f78b59c35c958e96/eras/conway/impl/cddl-files/extra.cddl#L5
- https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml

The protobuf field is named `tag_cbor_sets` to be consistent with
other tooling that names it the same way: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/4a59d8818402a1d1d4c9939a6a8f606940f51d55/src/types/public.ts#L1874

This change is needed now as some wallets/tools already use this tag,
e.g. delegating to a drep using the https://gov.tools/ DApp.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 - Update manufacturer HID descriptor to bitbox.swiss
 - Ethereum: remove deprecated Goerli network
 - SD card: solve backup bug when sd card is re-inserted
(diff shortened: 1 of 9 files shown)


NickeZ (approved):
utack

PR #1353, opened 2024-12-30
simulator: async-signal safety

NickeZ:
printf is not async-signal safe, see
https://docs.oracle.com/cd/E19455-01/806-5257/gen-26/index.html


asi345 (approved):
utACK

PR #1357, opened 2025-01-09
DX: Publish the simulators port

NickeZ:
This commit publishes the simulators default port 15423 on the development container so that processes on the host can access it

### DIFF
--- a/scripts/dockerenv.sh
+++ b/scripts/dockerenv.sh
@@ -72,13 +72,15 @@ dockerdev () {
         $RUNTIME rm "$CONTAINER_NAME"
     fi
 
-    # SYS_PTRACE is needed to run address sanitizer
+    # * `SYS_PTRACE` is needed to run address sanitizer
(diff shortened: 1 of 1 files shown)


scripts/dockerenv.sh
```diff
@@ -79,6 +79,7 @@ dockerdev () {
            --name="$CONTAINER_NAME" \
            -v "$repo_path":"$MOUNT_DIR" \
            --cap-add SYS_PTRACE \
+           -p 15423:15423 \
```
benma, line 82: please add a comment, it's not obvious what this is for when one does not know

benma (approved):

scripts/dockerenv.sh
NickeZ: Git blame ;), jk. I'll add a comment

NickeZ: done, force pushed

PR #1358, opened 2025-01-13
rust: update toolchain to 1.84.0 and update bindgen/cbindgen

benma:
This reduces the binary size by about 3kB.

New now are warnings (which we interpret as errors) about shared mut
refs:

```
 = note: shared references to mutable statics are dangerous; it's undefined behavior if the static is mutated or if a mutable reference is created for it while the shared reference lives
   = note: `-D static-mut-refs` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(static_mut_refs)]`
error: creating a shared reference to mutable static is discouraged
```

The commit includes `UnsafeSyncRefCell` and uses it in some unit tests
to remove the warning. In `src/rust/bitbox02-rust-c/src/workflow.rs`
we disable the warning altogether as it's hard to fix it there, so we
delay it there for a future PR.


src/rust/util/src/lib.rs
```diff
@@ -50,6 +50,25 @@ impl<'a, T> Survive<'a, T> {
     }
 }
 
+/// A wrapper that adds the Sync trait to RefCell. We can use this to avoid mutexes as our target is
+/// single-threaded and our unit tests run single-threaded.
```
NickeZ, line 54: > our target is single-threaded

Our target isn't single-threaded, it has interrupts. So the comment should say more something in line with the fact that we don't want to fix this right now, that we are breaking the rust soundness, and that we intend to fix it later. Maybe put a "TODO" next to the `unsafe impl Sync`.

src/rust/bitbox02-rust/src/xpubcache.rs
```diff
@@ -127,24 +127,20 @@ mod tests {
...
 
         impl Xpub for MockXpub {
             fn derive(&self, keypath: &[u32]) -> Result<Self, ()> {
                 let mut kp = Vec::new();
                 kp.extend_from_slice(&self.0);
                 kp.extend_from_slice(keypath);
-                unsafe {
-                    CHILD_DERIVATIONS += keypath.len() as u32;
-                }
+                *CHILD_DERIVATIONS.borrow_mut() += keypath.len() as u32;
```
NickeZ, line 138: It would be better if we used raw pointers and kept the `unsafe` blocks. Using this `UnsafeSyncRefCell` it is not obvious anymore where the unsafe code is.

src/rust/util/src/lib.rs
benma: Hmm good point. For now I moved it to testing.rs as it's only used in unit tests. Is the assumption there okay as we run single-threaded and there are no interrupts there?

Btw, we use the same trick also in production code since a long time (probably incorrect):
https://github.com/BitBoxSwiss/bitbox02-firmware/blob/160e154d8cae9211feb26e72f85bb84f4c5be70c/src/rust/bitbox02-rust/src/async_usb.rs#L44-L49
 
Though I have never seen a panic in the use of this particular RefCell myself or from a customer. 

Possibly useful reference: https://docs.rust-embedded.org/book/concurrency/

benma:
I wonder why the compiler is not warning about all the other static muts, there are quite many, e.g.:

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/160e154d8cae9211feb26e72f85bb84f4c5be70c/src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs#L1769

src/rust/bitbox02-rust/src/xpubcache.rs
NickeZ: technically it is still unsafe / unsound, because when you did `unsafe impl Sync` you promised that you have appropriate synchronization primitives internally. RefCell only manages the borrowchecker at runtime instead of compile time.

src/rust/util/src/lib.rs
NickeZ: you also have this: https://doc.rust-lang.org/nomicon/send-and-sync.html

NickeZ (approved):

PR #1360, opened 2025-01-30
bitcoin: display account name when sending to self

benma:
Before, we did this only when sending to the same account. This commit extends this to work when sending to an address of a *different* account of the same keystore.

The bitcoin transaction script_configs field is used for inputs and change outputs. We add a similar list, `output_script_configs`, which allows to attach script configs of different accounts to outputs, with `output_script_config_index` to reference them in the output.

To ensure backwards compatibility, `output_script_config_index` is added as a new field, which takes precedence over
`script_config_index` when set.

The output script configs are validated, but since one can send to any address of the same keystore, the output script config validation allows any combination of script configs, unlike the input script configs. To not duplicate lots of code, a few refactors are done:

- validate_script_configs is now used for generic validation and is used to validate the output script configs as well as the input script configs
- validate_input_script_configs performs additional validation as before (can't mix multisig and single sig inputs, can't mix inputs from different accounts, etc).
- the multisig/policy name is stored in ValidatedScriptConfig, so we don't have to fetch it twice

The Python lib version was downgraded from 8.0.0 to 7.0.0, it seems it
was bumped before in error - the previous release tag and release is 6.3.0.

### DIFF
--- a/messages/btc.proto
+++ b/messages/btc.proto
@@ -115,6 +115,9 @@ message BTCSignInitRequest {
   }
   FormatUnit format_unit = 8;
   bool contains_silent_payment_outputs = 9;
+  // used script configs for outputs that send to an address of the same keystore, but not
+  // necessarily the same account (as defined by `script_configs` above).
+  repeated BTCScriptConfigWithKeypath output_script_configs = 10;
 }
 
 message BTCSignNextResponse {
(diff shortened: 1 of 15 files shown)


messages/btc.proto
```diff
@@ -174,12 +177,19 @@ message BTCSignOutputRequest {
...
   uint32 script_config_index = 6;
   optional uint32 payment_request_index = 7;
   // If provided, `type` and `payload` is ignored. The generated output pkScript is returned in
   // BTCSignNextResponse. `contains_silent_payment_outputs` in the init request must be true.
   SilentPayment silent_payment = 8;
+  // If ours is true. If set, `script_config_index` is ignored. References an output script config
+  // from BTCSignInitRequest. This enables verification that an output belongs to the same keystore,
+  // even if it is from a different account than we spend from, allowing us to display this info to
+  // the user.
+  optional uint32 output_script_config_index = 9;
```
benma, line 192: Wondering it this should just be `optional BTCScriptConfigWithKeypath output_script_config` instead of a reference into a vector, to simplify the code and API of client libraries.

The `script_configs` vector for inputs/changes is there so 
- we can validate that multisig/singlesig inputs are not mixed and that different accounts are not mixed, and 
- that we can confirm multisig details at the beginning of a transaction 
- and to have a speedup to not repeat validation for each input
 
All that is not necessary for the output script infos, and the validation performance is not so important there as the user is anyway manually confirming the output :thinking:

Beerosagos: Makes sense to me :+1:

py/bitbox02/CHANGELOG.md
```diff
@@ -2,13 +2,12 @@
 
 ## [Unreleased]
 
-# 8.0.0
-- SD card: Remove API to prompt removal of the microSD card from the device
-- Add support for regtest (supported from firmware version v9.21.0)
-
 # 7.0.0
```
Beerosagos, line 5: If the last released version is 6.3.0, shouldn't all these changes go under `Unreleased` for now?

Beerosagos (approved):
tACK with a couple of comments :pray: 

Sorry for the waiting, but has been quite a challenging review :D

PR #1365, opened 2025-02-05
Debug bootloader without menu

NickeZ:
Based on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1362


NickeZ:
I don't need this anymore as I can jump straight to fw in gdb.

PR #1371, opened 2025-02-12
make product string in HID descriptor and the platform value in the REQ_INFO dynamic, not hardcoded

benma:
Builds on #1370

Allows us to have one firmware (per edition) for both bb02 and bb02-plus while still allowing the client to differentiate between them.


NickeZ (approved):
utACK

PR #1378, opened 2025-03-11
bitcoin: enable message signing on testnet and regtest

### LINKED ISSUE #1376: Better error message for testnet sign_message
As discussed in https://github.com/bitcoin-core/HWI/issues/777  Bitbox does not support sign_message for testnet.  
It would be great if either, it would support it, or the error message would be clear (like "sign_message is only available on Mainnet"), to not confuse developers and users.

benma:
Fixes #1376

### DIFF
--- a/py/send_message.py
+++ b/py/send_message.py
@@ -896,33 +896,59 @@ class SendMessage:
     def _sign_btc_message(self) -> None:
         # pylint: disable=no-member
 
-        keypath = [49 + HARDENED, 0 + HARDENED, 0 + HARDENED, 0, 0]
-        script_config = bitbox02.btc.BTCScriptConfig(
-            simple_type=bitbox02.btc.BTCScriptConfig.P2WPKH_P2SH
-        )
-        address = self._device.btc_address(
-            keypath=keypath, script_config=script_config, display=False
-        )
-
-        print("Address:", address)
-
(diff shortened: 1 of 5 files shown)


benma:
@Beerosagos I added his to send_message.py for easier manual testing as discussed.

py/send_message.py
```diff
@@ -896,33 +896,59 @@ def _sign_btc_tx(self) -> None:
...
+                simple_type=bitbox02.btc.BTCScriptConfig.P2WPKH_P2SH
             )
-            print("Signature:", base64.b64encode(sig65).decode("ascii"))
-        except UserAbortException:
-            print("Aborted by user")
+            sign(bitbox02.btc.BTC, keypath, script_config)
+
+        def sign_testnet() -> None:
+            keypath = [49 + HARDENED, 1 + HARDENED, 0 + HARDENED, 0, 0]
+            script_config = bitbox02.btc.BTCScriptConfig(
```
Beerosagos, line 937: nit: since you use the same script_config both times maybe you can directly include it in the sign function?

Beerosagos (approved):
tACK, with a small nit

py/send_message.py
benma: I think I'll leave it, as the script config is tightly coupled to the keypath, so being able to modify one but not the other locally seems off.

PR #1379, opened 2025-03-11
Pylint

benma:


benma:
Damn, the fixed-up lint script now correctly exits with non-zero exist code while before it failed to report the error:

```
py/bitbox02/setup.py:17:0: W0402: Uses of a deprecated module 'distutils.core' (deprecated-module)
```

Need to fix this too now I guess :sweat_smile:

PR #1380, opened 2025-03-11
Evm sidechain screen

Tomasvrba:
add a new screen confirmation to the user verification workflow for known L2s and sidechains

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ## Firmware
 
 ### [Unreleased]
+- Ethereum: add confirmation screen for known networks, change base unit to ETH for Arbitrum and Optimism
 
 ### 9.22.0
 - Update manufacturer HID descriptor to bitbox.swiss
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 
 ### [Unreleased]
 - Ethereum: add confirmation screen for known networks, change base unit to ETH for Arbitrum and Optimism
+- Ethereum: add Base and Gnosis Chain to known networks
 
 ### 9.22.0
 - Update manufacturer HID descriptor to bitbox.swiss
--- a/src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
@@ -299,6 +299,16 @@ pub async fn process(request: &Transaction<'_>) -> Result<Response, Error> {
     }
     super::keypath::warn_unusual_keypath(&params, params.name, request.keypath()).await?;
 
+    // Show chain confirmation only for known networks
+    if super::params::is_known_network(request.coin()?, request.chain_id()) {
+        confirm::confirm(&confirm::Params {
+            body: &format!("Sign transaction on\n\n{}", params.name),
+            accept_is_nextarrow: true,
+            ..Default::default()
+        })
+        .await?;
+    }
+
     // Size limits.
     if request.nonce().len() > 16
         || request.gas_limit().len() > 16
@@ -472,6 +482,11 @@ mod tests {
         const KEYPATH: &[u32] = &[44 + HARDENED, 60 + HARDENED, 0 + HARDENED, 0, 0];
 
         mock(Data {
+            ui_confirm_create: Some(Box::new(|params| {
+                assert_eq!(params.body, "Sign transaction on\n\nEthereum");
+                assert!(params.accept_is_nextarrow);
+                true
+            })),
             ui_transaction_address_create: Some(Box::new(|amount, address| {
                 assert_eq!(amount, "0.530564 ETH");
                 assert_eq!(address, "0x04F264Cf34440313B4A0192A352814FBe927b885");
@@ -546,6 +561,11 @@ mod tests {
                     UI_COUNTER
                 } {
                     1 => {
+                        assert_eq!(params.body, "Sign transaction on\n\nEthereum");
+                        assert!(params.accept_is_nextarrow);
+                        true
+                    }
+                    2 => {
                         assert_eq!(params.title, "High fee");
                         assert_eq!(params.body, "The fee is 12.0%\nthe send amount.\nProceed?");
                         assert!(params.longtouch);
@@ -575,7 +595,7 @@ mod tests {
             address_case: pb::EthAddressCase::Mixed as _,
         })))
         .is_ok());
-        assert_eq!(unsafe { UI_COUNTER }, 1);
+        assert_eq!(unsafe { UI_COUNTER }, 2);
     }
 
     /// Test an EIP-1559 transaction with an unusually high fee.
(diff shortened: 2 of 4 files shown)


src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -299,6 +299,17 @@ pub async fn process(request: &Transaction<'_>) -> Result<Response, Error> {
...
 
+    // Show chain confirmation only for known non-mainnet networks
+    if super::params::is_known_non_mainnet(params.chain_id) {
+        confirm::confirm(&confirm::Params {
+            title: "Confirm Chain",
+            body: &format!("Sign transaction on\n{}", params.name),
+            accept_is_nextarrow: true,
+            ..Default::default()
+        })
+        .await?;
```
benma, line 310: In BTC, we have this wording currently:


Title: `Spend from`
Body: `<account name>`

For consistency you could do the same, but simply use `body: params.name.into()`?

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -1175,4 +1191,171 @@ mod tests {
         );
         assert_eq!(unsafe { CONFIRM_COUNTER }, 4);
     }
+
+    /// Test that the chain confirmation screen appears for known non-mainnet networks.
+    #[test]
+    pub fn test_chain_confirmation_for_l2_networks() {
```
benma, line 1197: is it fair to call all of them l2 (or sidechains)? Aren't some normal base-layer chains in their own right?

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -299,6 +299,17 @@ pub async fn process(request: &Transaction<'_>) -> Result<Response, Error> {
     }
     super::keypath::warn_unusual_keypath(&params, params.name, request.keypath()).await?;
 
+    // Show chain confirmation only for known non-mainnet networks
+    if super::params::is_known_non_mainnet(params.chain_id) {
```
benma, line 303: why not confirm this unconditionally? Imho it would be good UX to first confirm what we want to do regardless.

benma (commented):
nice improvement!

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
Tomasvrba: sure, can change the name

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
Tomasvrba: hm, `Spend from\n Optimism` doesn't sound quite right, imo you spend `from` your account, but `on` a network
Also since it's not BTC, you might not actually be spending anything, you could just be signing a tx, like `Approve` to approve a token spend, or bidding on something and so on

benma: :+1: 

I'd drop the "Confirm chain" as it does not add much info and sounds a bit too technical for users that only use mainnet in the app.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -299,6 +299,16 @@ pub async fn process(request: &Transaction<'_>) -> Result<Response, Error> {
     }
     super::keypath::warn_unusual_keypath(&params, params.name, request.keypath()).await?;
 
+    // Show chain confirmation only for known networks
+    if super::params::is_known_network(params.chain_id) {
```
benma, line 303: `let is_known_network = super::params::get(request.coin()?, request.chain_id()).is_some()` and delete the new helper function you had, as `get()` is backwards compatible with clients that use the `coin` parameter instead of the `chain_id` parameter.

Maybe it even makes sense to delete `get_and_warn_unknown()` and move the confirmation screen there to here, so they are all close together, or re-factor it into `fn warn_unknown()` and call it here. The warning also appears when displaying a receive address, but I am not sure anymore if this is needed there at all, as the address is the same everywhere anyway.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -1175,4 +1251,131 @@ mod tests {
...
+    pub fn test_no_chain_confirmation_for_unknown_networks() {
+        const KEYPATH: &[u32] = &[44 + HARDENED, 60 + HARDENED, 0 + HARDENED, 0, 0];
+        static mut CONFIRM_COUNTER: u32 = 0;
+
+        // Test with an unknown network (chain_id 999999)
+        mock(Data {
+            ui_confirm_create: Some(Box::new(|params| {
+                unsafe {
+                    // Only the warning confirmations should appear, not chain confirmation
+                    if params.body.starts_with("Sign transaction on") {
```
benma, line 1350: this test is a bit brittle, as if the wording changes it wouldn't catch a regression. would be safer to check all expected screens.

benma (changes requested):

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ## Firmware
 
 ### [Unreleased]
+- Ethereum: add confirmation screen for known L2 networks and sidechains and change their unit to ETH
```
benma, line 10: pls update this, it is now out of date

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
Tomasvrba: actually, looking at existing tests, this one is not needed since "test_process_unknown_network" will already catch this, removed

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
Tomasvrba: ok, refactored as you suggested (I think, lmk if I misunderstood what you meant)
see the fixup commit for the diff

benma: Sorry for the trouble, but let's revert it :sweat_smile:  The reasons are:

- using `::get()` instead of `::get_and_warn_unknown()` in the pubrequest means the call will fail for unknown networks, which is not desirable. We should think about if we want to remove the coin/chain_id param from that call completely, but that should be a separate discussion/PR.
- warn_unknown does not only warn, it also produces params, and the prompts are still not closer together than they were before :shrug: 

I did not foresee these before, sorry.

Let's revert these parts, but make `is_known_network(coin, chain_id)` simply call `params::get(coin, chain_id).is_some()`.

Tomasvrba: > warn_unknown does not only warn, it also produces params, and the prompts are still not closer together than they were before 

hmm, yeah, I see that now.

> Let's revert these parts, but make is_known_network(coin, chain_id) simply call params::get(coin, chain_id).is_some()

Then keep the change making the getter public right, to call it from sign.rs?

benma (approved):
utACK, thank you!

PR #1382, opened 2025-03-26
external/asf4: Upgrade to latest version

NickeZ:
Upgrade to latest generated drivers


NickeZ:
Added license text and wrote some more in the commit message.

external/asf4-drivers/hal/documentation/flash.rst
```diff
@@ -0,0 +1,52 @@
+The Flash Driver
```
benma, line 1: Kind of nice to have some docs included here now.

src/platform/driver_init.c
```diff
@@ -198,6 +198,7 @@ static void _mci_init(void)
         GCLK, SDHC0_GCLK_ID, CONF_GCLK_SDHC0_SRC | (1 << GCLK_PCHCTRL_CHEN_Pos));
     hri_gclk_write_PCHCTRL_reg(
         GCLK, SDHC0_GCLK_ID_SLOW, CONF_GCLK_SDHC0_SLOW_SRC | (1 << GCLK_PCHCTRL_CHEN_Pos));
+    mci_sync_init(&MCI_0, SDHC0);
```
benma, line 201: Why is this newly needed?

benma (commented):
I tested sdcard insertion check and backup listing and re-insertion, works fine.

I also tested touch, and obviously the screen, and the reboot API command.

There are a lot of driver compile warnings, e.g.

```
/bitbox02-firmware/external/asf4-drivers/hal/src/hal_usb_device.c: In function '_usb_d_trans':
/bitbox02-firmware/external/asf4-drivers/hal/src/hal_usb_device.c:144:9: warning: cast discards 'const' qualifier from pointer target type [-Wcast-qual]
      = {(uint8_t *)buf, size, dir ? (uint8_t)(ep | USB_EP_DIR) : (uint8_t)(ep & USB_EP_N_MASK), zlp};
         ^
```

Could/should we disable all warnings in the `asf4-drivers` folder?

benma:
Note to self: test bootloader

src/platform/driver_init.c
NickeZ: We put that line in sd_mmc.c:930, but this is a line that belongs to user code and not driver code. I noticed that I missed removing it from sd_mmc.c when I reverted that file. And then my autoformatter also made it harder to see the diff. Should be better now.

NickeZ:
> I tested sdcard insertion check and backup listing and re-insertion, works fine.
> 
> I also tested touch, and obviously the screen, and the reboot API command.
> 
> There are a lot of driver compile warnings, e.g.
> 
> ```
> /bitbox02-firmware/external/asf4-drivers/hal/src/hal_usb_device.c: In function '_usb_d_trans':
> /bitbox02-firmware/external/asf4-drivers/hal/src/hal_usb_device.c:144:9: warning: cast discards 'const' qualifier from pointer target type [-Wcast-qual]
>       = {(uint8_t *)buf, size, dir ? (uint8_t)(ep | USB_EP_DIR) : (uint8_t)(ep & USB_EP_N_MASK), zlp};
>          ^
> ```
> 
> Could/should we disable all warnings in the `asf4-drivers` folder?

I removed the warnings for the asf4-drivers.

Can you please retest sd-card. I did some changes there that shouldn't have any impact.

PR #1384, opened 2025-03-29
debugging: Run firmware without bootloader

NickeZ:
When debugging firmware it is easier to skip the bootloader altogether and go straight to firmware code.

Since VTOR is reset on core reset we need to manually set the PC and SP registers. Issuing a reset will simply try to boot from address 0 again.

If we merge this PR `make run-debug` will always go straight to application code.


benma (approved):
utACK

PR #1385, opened 2025-03-29
rust: remove SafeInputString

benma:
This was a custom C string wrapper with auto-zeroing, but it is much easier to simply use zeroize and CStr (which were both introduced to the codebase after SafeInputString). This also allows us to call bip39 unlock more easily with a Rust string in the future (see testing.rs for example).


NickeZ (approved):
utack

PR #1387, opened 2025-03-30
api: centralize state check

benma:
We had two separate checks to decide if we will allow a protobuf message API call to proceed: in hww.rs (forbid if initialized and locked) and in api.rs in the can_call() function that defines which states the API call can run in.

This makes it a bit hard to follow as split into two locations, and it also prevents us from introducing a protobuf call that is only allowed to run if the device is initialized but still locked. `Unlock` would be such a useful new API call if we want to allow entering the passphrase on the host.

The hww.rs check passed if `!initialized || !locked`, which means the same as `uninitialized || seeded || unlocked`.

We change the `can_call` state checks to include this condition. For this we split the local `Initialized` state into
`InitializedAndLocked` and `InitializedAndUnlocked`. For each line in can_call, the two conditions are combined. The equivalent state check in the end is simply to convert all `Initialized` checks to `InitializedAndUnlocked`.

The `true` entries become explicit as they can't include `InitializedAndLocked` (e.g. can't change the name using `DeviceName` if initialized but not yet unlocked).

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api.rs
+++ b/src/rust/bitbox02-rust/src/hww/api.rs
@@ -85,20 +85,27 @@ async fn process_api_btc(_request: &Request) -> Result<Response, Error> {
 
 /// Checks if the device is ready to accept/handle an api endpoint.
 fn can_call(request: &Request) -> bool {
-    // We have three main states:
-    // Creating a wallet on an uninitialized device goes through those states in order.
-    // Restoring a backup skips the seeded state and goes straight to `initialized`.
+    // We have four main states:
+    // Creating a wallet on an uninitialized device goes from Uninitialized to Seeded, and when the
+    // backup is created to `Initialized*`.
+    // Restoring a backup skips the seeded state and goes straight to `Initialized*`.
     // Each state has a set of valid api calls associated.
     enum State {
         // Uninitialized (reset).
         Uninitialized,
         // Seeded (password defined, seed created/loaded).
         Seeded,
-        // Initialized (seed backuped up on SD card).
-        Initialized,
+        // InitializedAndLocked (seed backuped up on SD card, keystore locked).
+        InitializedAndLocked,
+        // InitializedAndUnlocked (seed backuped up on SD card, keystore unlocked).
+        InitializedAndUnlocked,
     }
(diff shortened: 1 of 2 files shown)


src/rust/bitbox02-rust/src/hww/api.rs
```diff
@@ -94,11 +94,17 @@ fn can_call(request: &Request) -> bool {
         Uninitialized,
         // Seeded (password defined, seed created/loaded).
         Seeded,
-        // Initialized (seed backuped up on SD card).
-        Initialized,
+        // InitializedAndLocked (seed backuped up on SD card, keystore locked).
+        InitializedAndLocked,
+        // InitializedAndUnlocked (seed backuped up on SD card, keystore unlocked).
+        InitializedAndUnlocked,
```
Beerosagos, line 100: State enum comment could be updated to reflect the changes

src/rust/bitbox02-rust/src/hww/api.rs
```diff
@@ -108,33 +114,54 @@ fn can_call(request: &Request) -> bool {
...
         // Deprecated call, last used in v1.0.0.
         Request::PerformAttestation(_) => false,
-        Request::DeviceInfo(_) => true,
-        Request::Reboot(_) => true,
-        Request::DeviceName(_) => true,
-        Request::DeviceLanguage(_) => true,
-        Request::CheckSdcard(_) => true,
-        Request::InsertRemoveSdcard(_) => true,
-        Request::ListBackups(_) => true,
+        Request::DeviceInfo(_) => matches!(
```
Beerosagos, line 117: Using some helper function to check for the state condition could help readability and reduce code redundancy, wdyt?

src/rust/bitbox02-rust/src/hww/api.rs
benma: Done

src/rust/bitbox02-rust/src/hww/api.rs
benma: Adding helper vars didn't work to help readability, but I grouped the messages that have the same checks, which seems a bit better. Please check.

Beerosagos (approved):
tACK!

PR #1388, opened 2025-03-31
Spi memory integration

Beerosagos:
add configuration for SPI_MEM

CLOCKS:
- SPI_MEM Core CLK: generic 0 (120MHz)
- SPI_MEM Slow CLK: generic 3 (48MHz)
- Sercom baud rate: 8MHz

The MX25 module needs 80MHz for high performance mode and 8MHz in low
power. Since the minimum divider for the baud rate generator is 4 (with
BAUD reg set to 1, see 33.6.2.3 of SAM D5 datasheet), we need an input
clock of 320MHz in high performance, which is not achievable with our
mcu. We'll go with the low-power mode.

I am not sure we actually need the low power clock.

PORTS:
Enabled SPI_MEM interface on SERCOM4 with the following pinout:
- CLK -> PA12
- MOSI -> PA15
- MISO -> PA13
- CS -> PA14
- WP -> PB07
- HOLD -> PA20


NickeZ (commented):
You need to add checks if `memory_get_platform` is bb02plus when calling init

src/platform/driver_init.c
```diff
@@ -115,6 +121,57 @@ static void _spi_init(void)
     SPI_OLED_enable();
 }
 
+static void _spi_mem_CLOCK_init(void)
```
benma, line 124: why uppercase clock and not lowercase? (in our own code we usually have lowercase snake case, and in this file there is also a `_ptc_clock_init()` function named similarly).

benma (commented):
Can you add to the commit msg where you got the hpl/spi folder from, for future reference? Is it from atmel start or something like that?

src/platform/driver_init.c
Beerosagos: Right, it was a leftover of the original Atmel start function. Fixed :+1:

Beerosagos:
@benma yeah it is based on Atmel start code. Added a line in the commit msg to specify it :pray: PTAL

NickeZ (approved):
utACK

benma (approved):
tACK, also on devices without the chip.

PR #1389, opened 2025-04-01
DX: Mount home directory in container

NickeZ:
This commit mounts the users whole `$HOME` directory instead of only the project path. This means that relative paths between projects work.

Git needs to be able to see the `.git` folder of the original clone when using multiple work trees.

### DIFF
--- a/scripts/dockerenv.sh
+++ b/scripts/dockerenv.sh
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This script only works if the work tree is in $HOME.
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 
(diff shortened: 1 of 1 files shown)


benma:
What other projects? Not very comfortable mounting the whole home folder, seems like unnecessary risk. Maybe there's a different solution, i.e. a flag of other folders to mount, or flags that are just passed along to docker, or something like that.

NickeZ:
> What other projects? Not very comfortable mounting the whole home folder, seems like unnecessary risk. Maybe there's a different solution, i.e. a flag of other folders to mount, or flags that are just passed along to docker, or something like that.

How about mounting the directory above the project folder?

Another benefit is that you don't have to have one docker container running per work tree. I currnetly have 4 work trees checked out. one PR i'm reviewing, one for simulator, one for main and one for a branch i'm working on.

A third benefit of using work trees is that I don't have to rebuild the world when I switch between branches.

NickeZ:
I started maintaining my own dockerdev executable instead.

PR #1390, opened 2025-04-01
DX: Add script for quick erase

NickeZ:


NickeZ:
Not really needed anymore when the default gdb script jumps directly to the firmware.

PR #1391, opened 2025-04-01
factory-setup: Remove C implementation of RTT

NickeZ:
The debug builds use the rust implementation of RTT. The two implementations cannot coexist. To be able to build debug builds of the factory-setup the C imeplementation is therefore removed.

The factory-setup image is changed to use the second channel for API calls, since the first is used for debug console.


src/factorysetup.c
```diff
@@ -130,44 +128,43 @@ typedef enum {
     ERR_UNKNOWN_COMMAND,
 } error_code_t;
 
+static error_code_t ble_configuration_status = ERR_INVALID_INPUT;
```
benma, line 131: Remove? Not used in this PR I think.

benma:
`target_link_libraries(factory-setup.elf PRIVATE rtt)` in CMakeLists needs to be removed I think.

NickeZ:
> `target_link_libraries(factory-setup.elf PRIVATE rtt)` in CMakeLists needs to be removed I think.

Yeah, fixed

NickeZ:
rebased

NickeZ:
rebased and fixed an issue with non-debug builds

NickeZ:
@benma found the bug in non-debug builds, fixed, rebased and pushed, ready for testing again

benma (approved):
tACK :tada:

benma:
I rebased this resolving two small conflicts.

PR #1393, opened 2025-04-01
Host passphrase

benma:
This is a quick prototype for optionally allowing entering the passphrase on the host by adding a new UnlockRequest API call that allows this. 

We might split this into two calls, one for unlocking the device and one for entering the passphrase, so the bbapp knows which state we are in. Alternatively, we could keep one UnlockRequest API, but send state updates to the app via `next_request()`. It is not clear yet which would be better.


benma:
Replaced by https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2115

PR #1395, opened 2025-04-01
newlib-compat: macro conflict between asf4 and std

NickeZ:
When upgradeing to newer GCC toolchain and newlib this conflict of macro definitions breaks compilation.

More specifically without this I cannot compile on darwin with arm GCC 13 toolchain.


benma (approved):
utACK

PR #1397, opened 2025-04-02
screen.c: do not call delay_ms if duration is 0

NickeZ:
The new ASF drivers changed the behavior of delay_ms. Calling it with 0 seems to mean delay forever and not 0 ms.

### DIFF
--- a/src/screen.c
+++ b/src/screen.c
@@ -43,7 +43,7 @@ void screen_print_debug(const char* message, int duration)
     UG_FontSelect(&font_font_a_9X9);
     UG_PutString(0, 0, print, false);
     UG_SendBuffer();
-    delay_ms(duration);
+    if (duration > 0) delay_ms(duration);
 }
 
 void screen_sprintf_debug(int duration, const char* fmt, ...)


benma (commented):
Is is a crazy change, who wants delay(0) to mean hang forever?! :exploding_head: 

Please also add the check to:

- https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/src/rust/bitbox02/src/lib.rs#L82
- https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/src/optiga/pal/pal_os_timer.c#L63

NickeZ:
> Is is a crazy change, who wants delay(0) to mean hang forever?! 🤯
> 
> Please also add the check to:
> 
>     * https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/src/rust/bitbox02/src/lib.rs#L82
> 
>     * https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/src/optiga/pal/pal_os_timer.c#L63

The first one already seem to be covered, the second one really never should be called with 0.

I would almost define calling a "sleep function" with 0 UB. It is not obvious to me what the correct implementation would be.

benma (approved):

PR #1400, opened 2025-04-02
optiga-pal: Fix assert

NickeZ:
ASF4 timers do not support to register same task twice, so we have to take care and remove it when it was cancelled before it was run.

I ran into an assert in hal with DEBUG enabled

This fixes the assert, but I haven't checked that the optiga still works 100%.

### DIFF
--- a/src/optiga/pal/pal_os_event.c
+++ b/src/optiga/pal/pal_os_event.c
@@ -37,7 +37,7 @@
 
 #include "pal_os_event.h"
 #include "hal_timer.h"
-#include "util.h"
+
 extern struct timer_descriptor TIMER_0;
 
 static pal_os_event_t pal_os_event_0 = {0};
@@ -70,7 +70,6 @@ static struct timer_task scheduler;
 
 void pal_os_event_trigger_registered_callback(void)
 {
-    // traceln("%s: called", __func__);
     register_callback callback;
 
     if (pal_os_event_0.callback_registered) {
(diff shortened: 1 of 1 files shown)


src/optiga/pal/pal_os_event.c
```diff
@@ -56,6 +59,9 @@ void pal_os_event_start(
 void pal_os_event_stop(pal_os_event_t* p_pal_os_event)
 {
     p_pal_os_event->is_event_triggered = 0;
+    if (!scheduler_was_triggered) {
+        timer_remove_task(&TIMER_0, &scheduler);
+    }
```
benma, line 64: The optiga lib's libusb PAL example removes the timer in `pal_os_event_destroy()` like this:

https://github.com/BitBoxSwiss/optiga-trust-m/blob/5b924c8615f563334136d5a84431f1682cfce3ea/extras/pal/libusb/pal_os_event.c#L287-L289

I am not sure what is going on under the hood yet, but maybe that would be the proper solution?

NickeZ: Hmm, yeah I guess what we missed was actually to remove the task in the `destroy` function. That also solves the issue. made the PR much smaller.

src/optiga/pal/pal_os_event.c
```diff
@@ -103,6 +102,10 @@ void pal_os_event_register_callback_oneshot(
 void pal_os_event_destroy(pal_os_event_t* pal_os_event)
 {
     (void)pal_os_event;
```
benma, line 104: can remove this line now

benma (approved):
tACK, but left a small nit

src/optiga/pal/pal_os_event.c
NickeZ: good catch!

PR #1402, opened 2025-04-07
make blocking workflows before the mainloop async

benma:
The last two remaining non-async workflows are now made async. This will help with BB02+ Bluetooth chip integration, whose firmware we want to stream in an async fashion at boot while the orientation screen and idle workflows are running, to not cause a noticable delay before or after these workflows.


NickeZ (approved):
tACK, had one question.

PR #1404, opened 2025-04-08
memory: return unique default device name for BB02+

benma:
To disambiguate multiple BitBoxes in the Bluetooth devices list, we use a unique name of the form "BitBox ABCD" (four random uppercase letters) until a device name is set by the user.

The random letters are not picked perfectly uniformely, but this is not security critical - it's merely to create a unique name to be able to disambiguate.

This change only applies to BB02+, and this random device name is displayed on the lockscreen visible before pairing.

### DIFF
--- a/src/memory/memory.c
+++ b/src/memory/memory.c
@@ -250,6 +250,10 @@ bool memory_set_device_name(const char* name)
     _read_chunk(CHUNK_1, chunk_bytes);
     util_zero(chunk.fields.device_name, sizeof(chunk.fields.device_name));
     snprintf((char*)&chunk.fields.device_name, MEMORY_DEVICE_NAME_MAX_LEN, "%s", name);
+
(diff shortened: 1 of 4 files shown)


src/memory/memory.c
```diff
@@ -250,16 +250,55 @@ bool memory_set_device_name(const char* name)
...
+static void _random_name(char* name_out)
+{
+    static char cached_name[MEMORY_DEVICE_NAME_MAX_LEN] = {0};
+
+    if (cached_name[0] == 0x00) {
+        // Generate 4 random uppercase letters
+        uint8_t random[32] = {0};
+        _interface_functions->random_32_bytes(random);
+        uint8_t letters[4];
+        for (int i = 0; i < 4; i++) {
```
NickeZ, line 269: Nit, but `sizeof(letters)` instead of 4

NickeZ (approved):
utACK, left a nit

NickeZ:
Tested now as well, works for me 🎉

PR #1405, opened 2025-04-08
da14531: Serial link protocol

NickeZ:


src/da14531/da14531_protocol.c
```diff
@@ -0,0 +1,428 @@
...
+    return true;
+}
+
+void da14531_protocol_init(void)
+{
+    _firmware_loader_init(&da14531_protocol.loader, da14531_firmware_start, da14531_firmware_size);
+    _serial_link_in_init(&da14531_protocol.serial_link);
+
+    // Reset the device if possible, if we cannot reset it over SWD, it must already be running
+    if (!_swd_reset_da14531()) {
```
benma, line 422: Guard this with FACTORYSETUP

src/da14531/da14531_protocol.c
```diff
@@ -0,0 +1,422 @@
...
+};
+
+struct da14531_protocol {
+    struct serial_link_in serial_link;
+    struct firmware_loader loader;
+};
+
+static struct da14531_protocol _protocol;
+
+#if !defined(NDEBUG)
```
benma, line 65: remove this, as this function is called below unconditionally

src/memory/memory_shared.h
```diff
@@ -82,6 +86,10 @@ typedef union {
         uint8_t encryption_key_split[32];
         uint8_t platform;
         uint8_t reserved[3]; // align to 4 bytes
+        uint8_t ble_bond_db[MEMORY_BLE_BOND_DB_LEN]; // 2 bytes len, then data
```
NickeZ, line 89: should split out ble_bond_db in `uint16_t ble_bond_db_len`.

Could also reduce `MEMORY_BLE_BOND_DB_LEN` with 2 to avoid padding later.

src/da14531/da14531_protocol.c
```diff
@@ -0,0 +1,421 @@
...
+    struct da14531_protocol_frame* frame = NULL;
+    if (_protocol.loader.state != FIRMWARE_LOADER_STATE_DONE) {
+        _firmware_loader_poll(&_protocol.loader, in_buf, in_buf_len, out_queue);
+    } else {
+        frame = _serial_link_in_poll(&_protocol.serial_link, in_buf, in_buf_len);
+    }
+    return frame;
+}
+
+static bool _swd_reset_da14531(void)
```
benma, line 373: guard with `#if FACTORYSETUP == 1 || !defined(NDEBUG)` like at the callsite, otherwise the compiler complains that this is unused (e.g. in `make firmware`)

src/memory/memory.c
```diff
@@ -335,6 +335,12 @@ bool memory_setup(const memory_interface_functions_t* ifs)
     chunk_0_t chunk = {0};
     CLEANUP_CHUNK(chunk);
     _read_chunk(CHUNK_0_PERMANENT, chunk_bytes);
+    // Factory setup image always does a factory reset of HWW
+#if FACTORYSETUP == 1
+    if (!memory_reset_hww()) {
```
benma, line 340: this is not a factory reset. `reset_reset()` is :thinking: not sure I like this change, it shouldn't be necessary to run it.

src/memory/memory.c
```diff
@@ -727,6 +759,20 @@ bool memory_bootloader_set_flags(auto_enter_t auto_enter, upside_down_t upside_d
...
+void memory_set_ble_bond_db(uint8_t* data, int16_t data_len)
+{
+    ASSERT(data_len <= 1022);
+    chunk_shared_t chunk = {0};
+    memory_read_shared_bootdata(&chunk);
+    ((int16_t*)chunk.fields.ble_bond_db)[0] = data_len;
+    memcpy(&chunk.fields.ble_bond_db[2], data, data_len);
+    if (memcmp((uint8_t*)(FLASH_SHARED_DATA_START), chunk.bytes, FLASH_SHARED_DATA_LEN) != 0) {
+        util_log("Updated bond db");
+        _write_to_address(FLASH_SHARED_DATA_START, 0, chunk.bytes);
```
benma, line 771: check result and return return false it it fails; convert function to `bool memory_set_ble_bond_db(...)`

src/memory/memory.c
NickeZ: Can I do a factory reset with the factory setup image?

NickeZ: Or should we have a separate factory reset image?

src/memory/memory.c
```diff
@@ -727,6 +759,20 @@ bool memory_bootloader_set_flags(auto_enter_t auto_enter, upside_down_t upside_d
     return false;
 }
 
+void memory_set_ble_bond_db(uint8_t* data, int16_t data_len)
+{
+    ASSERT(data_len <= 1022);
+    chunk_shared_t chunk = {0};
```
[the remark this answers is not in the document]
NickeZ: Ah, yeah, I forgot we use destructors.

src/memory/memory.c
NickeZ: I removed this code, but I really think a device should be factory reset after running the factory setup. Perhaps there is some more appropriate way to achieve that.

NickeZ: @benma would you like to resolve this?

src/firmware_main_loop.c
```diff
@@ -47,7 +73,19 @@ void firmware_main_loop(void)
 #endif
 
     while (1) {
-        // Do USB I/O
+        // Do UART I/O
+        if (!_usb_hww_request_seen) {
```
benma, line 77: also check that this is the bb02+ platform - currently this PR does not work on a regular bb02 because of it.

NickeZ: done

src/rust/bitbox02-rust/src/hww/api/system.rs
```diff
@@ -21,7 +21,7 @@ use pb::response::Response;
 use crate::hal::Ui;
 use crate::workflow::confirm;
 
-pub async fn reboot(
+pub async fn reboot_to_bootloader(
```
benma, line 24: also rename the tests and the calls to `reboot()` in the tests below

src/rust/bitbox02/src/lib.rs
```diff
@@ -212,8 +213,8 @@ pub fn format_datetime(
 }
 
 #[cfg(not(feature = "testing"))]
-pub fn reboot() -> ! {
-    unsafe { bitbox02_sys::reboot() }
+pub fn reboot_to_bootloader() -> ! {
```
benma, line 216: also rename the testing version below

PR #1409, opened 2025-04-14
usb: Move usb handling out into main loop

NickeZ:
It best practice to do as little work as possible in the interrupt handlers. In this commit the parsing is moved from the interrupts handlers to the main loop.

This is also a requirement for having multiple transports.

### DIFF
--- a/src/bootloader/startup.c
+++ b/src/bootloader/startup.c
@@ -16,6 +16,7 @@
 #include "mpu_regions.h"
 #include "platform_config.h"
 #include "platform_init.h"
+#include "usb/class/hid/hww/hid_hww.h"
 
 #include <driver_init.h>
 #include <hardfault.h>
@@ -56,8 +57,22 @@ int main(void)
 #endif
     bootloader_jump();
 
+    const uint8_t* hww_data = NULL;
+    uint8_t hww_frame[USB_REPORT_SIZE] = {0};
+
     // If did not jump to firmware code, begin USB processing
     while (1) {
+        if (hid_hww_read(&hww_frame[0])) {
+            usb_packet_process((const USB_FRAME*)hww_frame);
+        }
+        if (!hww_data) {
+            hww_data = queue_pull(queue_hww_queue());
+        }
+        if (hww_data) {
+            if (hid_hww_write_poll(hww_data)) {
+                hww_data = NULL;
+            }
+        }
         usb_processing_process(usb_processing_hww());
     }
     return 0;
--- a/src/firmware_main_loop.c
+++ b/src/firmware_main_loop.c
@@ -15,43 +15,94 @@
 #include "firmware_main_loop.h"
 
 #include "hardfault.h"
+#include "hid_hww.h"
 #include "hww.h"
 #include "touch/gestures.h"
-#include "u2f.h"
 #include "ui/screen_process.h"
 #include "ui/screen_stack.h"
+#include "usb/class/hid/hww/hid_hww.h"
 #include "usb/usb.h"
+#include "usb/usb_frame.h"
 #include "usb/usb_processing.h"
 #include "workflow/orientation_screen.h"
 #include <rust/rust.h>
+#if APP_U2F == 1
+#include "u2f.h"
+#include "u2f/u2f_packet.h"
+#include "usb/class/hid/u2f/hid_u2f.h"
+#endif
 
 void firmware_main_loop(void)
 {
     // This starts the async orientation screen workflow, which is processed by the loop below.
     orientation_screen();
 
+    const uint8_t* hww_data = NULL;
+    uint8_t hww_frame[USB_REPORT_SIZE] = {0};
+
+#if APP_U2F == 1
+    u2f_packet_init();
+    const uint8_t* u2f_data = NULL;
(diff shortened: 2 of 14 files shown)


src/firmware_main_loop.c
```diff
@@ -30,28 +35,59 @@ void firmware_main_loop(void)
     // This starts the async orientation screen workflow, which is processed by the loop below.
     orientation_screen();
 
+    const uint8_t* hww_data = NULL;
+    const uint8_t* u2f_data = NULL;
```
benma, line 39: guard u2f things with `#if APP_U2F == 1`, it's confusing if it's guarded below but not here.

src/usb/usb_processing.c
```diff
@@ -38,7 +38,7 @@ extern struct timer_descriptor TIMER_0;
  * Amount of time to wait before an outstanding operation times out
  * (if the client is closed).
  */
-#define USB_OUTSTANDING_OP_TIMEOUT_MS (500)
+#define USB_OUTSTANDING_OP_TIMEOUT_MS (30000)
```
benma, line 41: not the right PR for this :smile:

src/firmware_main_loop.c
```diff
@@ -30,28 +35,66 @@ void firmware_main_loop(void)
...
+            }
+        }
+#endif
+
         screen_process();
         /* And finally, run the high-level event processing. */
 
         rust_workflow_spin();
 
-        if (usb_is_enabled()) {
```
benma, line 39: Why is this check gone?

src/bootloader/startup.c
```diff
@@ -56,8 +58,24 @@ int main(void)
 #endif
     bootloader_jump();
 
+    const uint8_t* hww_data = NULL;
+    uint8_t hww_frame[USB_REPORT_SIZE] = {0};
+
     // If did not jump to firmware code, begin USB processing
     while (1) {
+        if (hid_hww_read(&hww_frame[0])) {
```
benma, line 66: I don't mind this, but I wondered why you use this syntax over just `hww_frame` (saw it consistently in your code and was curious)

NickeZ: `hww_frame` is a `uint8_t[64]` whereas `hid_hww_read` takes a `uint8_t*` I think modern compilers distinguish between these two types. But I could be wrong.

src/firmware_main_loop.c
NickeZ: `hid_usb_read` will return false if usb is disabled.

src/bootloader/startup.c
NickeZ: Is this resolved?

NickeZ:
Fixed CI, tested bootloader, u2f and signing transactions.

src/bootloader/startup.c
benma: I think you might be wrong, the two are equivalent and you can just pass `hww_frame`.

NickeZ: I just think it is clearer that I don't intend to pass the whole array as a value. But I'm open to change if you think it is less clear and you prefer the implicit decaying

src/firmware_main_loop.c
```diff
@@ -30,28 +35,72 @@ void firmware_main_loop(void)
...
+    while (1) {
+        // Do USB I/O
+        if (!hww_data) {
+            hww_data = queue_pull(queue_hww_queue());
+        }
 #if APP_U2F == 1
-            usb_processing_process(usb_processing_u2f());
+        // Generate timeout packets
+        uint32_t timeout_cid;
+        while (u2f_packet_timeout_get(&timeout_cid)) {
```
benma, line 55: We have to wait for a timeout? Why?

benma (approved):
I also tested the bootloaders and firmwares and the simulator, seems to work well.

Left a question about the u2f timeout while loop, I don't understand it, but it works :smile:

src/bootloader/startup.c
benma: No, as I said at the top, I don't mind it at all, I was just curious.

src/firmware_main_loop.c
NickeZ: This is a loop checking if any of the current u2f commands have timed out. It was moved from the usb_processing file because it called `send` and that function has been removed since all Io should be done from the main loop.

benma: In the original, `send()` was called inside the loop. I am not really following the logic, but wouldn't that mean that here we should turn the  `while` into a `if`, so if there are multiple timeouts, their packets all get processed properly? Or is the queue large enough so it doesn't matter?

NickeZ: Yeah, queue is very large

src/firmware_main_loop.c
```diff
@@ -15,12 +15,17 @@
 #include "firmware_main_loop.h"
 
 #include "hardfault.h"
+#include "hid_hww.h"
 #include "hww.h"
 #include "touch/gestures.h"
 #include "u2f.h"
+#include "u2f/u2f_packet.h"
```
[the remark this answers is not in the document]
NickeZ: Done

PR #1413, opened 2025-04-15
optiga: u2f_counter_inc return true for success

NickeZ:
u2f_counter_inc (return type bool) incorrectly returned 0 (false) for success.


benma (approved):
utACK

PR #1415, opened 2025-04-17
Spi memory lock

Beerosagos:
This adds api calls to lock/unlock and write a protected area of the memory.

This PR includes a test function to be removed before merging

### DIFF
--- a/src/memory/spi_mem.c
+++ b/src/memory/spi_mem.c
@@ -14,6 +14,8 @@
 
 #include "spi_mem.h"
 #include "bitbox02_pins.h"
+#include "random.h"
+#include "screen.h"
 #include "util.h"
 #include <hal_delay.h>
 #include <spi_lite.h>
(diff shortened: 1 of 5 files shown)


src/memory/spi_mem.c
```diff
@@ -258,6 +290,47 @@ int32_t spi_mem_smart_erase(void)
...
+    _spi_mem_read_cr(&reg[1]);
+
+    // clean and update status register with protection bits
+    reg[0] &= ~SR_PROTECT_BITS_MASK;
+    reg[0] |= protection & SR_PROTECT_BITS_MASK;
+
+    // set the top/bottom protection bit.
+    // This is an OTP bit,so the write will have an effect
+    // only the first time.
+    reg[1] = reg[1] | CR1_TB_BIT_BOTTOM;
```
Beerosagos, line 308: Not sure if it is ok to add this here, or if we want to do it inside the driver during the init function.

benma:
Assigned to @NickeZ who is more familiar with all this.

src/bootloader/startup.c
```diff
@@ -55,6 +56,7 @@ int main(void)
 #ifdef BOOTLOADER_DEVDEVICE
     qtouch_init();
 #endif
+    spi_mem_protected_area_lock();
```
NickeZ, line 59: perhaps do it in `platform_init`? It should be conditional on `memory_get_platform() == MEMORY_PLATFORM_BITBOX02PLUS`

NickeZ (commented):
In general very good! Only some minor comments.

Beerosagos:
@NickeZ updated, PTAL :pray:

Beerosagos:
> ACK, I assume you will remove all changes from bootloader/startup.c and firmware.c before merging.

Sure, I guess we'll call `spi_mem_protected_area_lock()` somewhere in the factory setup code in a future PR, right?

> I just realized, does it make sense to move the test to factorysetup.c?

I don't think so.. the test takes a lot of time to execute, I don't think we want that, wdyt?

NickeZ:
> > ACK, I assume you will remove all changes from bootloader/startup.c and firmware.c before merging.
> 
> Sure, I guess we'll call `spi_mem_protected_area_lock()` somewhere in the factory setup code in a future PR, right?

I'm pretty sure factory setup calls `platform_init()`. So as long as it is there we are good.

> 
> > I just realized, does it make sense to move the test to factorysetup.c?
> 
> I don't think so.. the test takes a lot of time to execute, I don't think we want that, wdyt?

True, yeah, we want factory setup to be as minimal as possible. Since we write the ble firmware to the chip we know if it works during factory setup.

PR #1419, opened 2025-04-21
use Hal trait to combine Ui, Sd and future traits

benma:
We want to inject deps into, for BB02, testing and in the future for
other BitBox products.

For this we create a Hal trait that can be used anywhere with
`&mut impl Hal`, without needing to repeat the trait bounds in
every function.

The alternative was to have a trait that contains all functions, which
did not seem like a good thing for modularity.

Another alternative was to colour every function with `<U: Ui,
S: Sd, ...>` trait bounds, which is ugly and does not scale to more
traits easily.

The `Workflows` trait is renamed to `Ui` (in the `hal` module only for
now, later we will untangle the workflows module and rename things
there too).


benma:
Discussed with @NickeZ to rename `Workflows` to `Ui` and `Context` to something better, TBD.

NickeZ (approved):
utACK! Very nice!

PR #1420, opened 2025-04-22
hal: add rest of the functions to the Sd trait

benma:
And use the hal layer to access sdcard. This allows local mocking in unit tests and other BitBox products to plug their own impl.

### DIFF
--- a/src/rust/bitbox02-rust/src/hal.rs
+++ b/src/rust/bitbox02-rust/src/hal.rs
@@ -15,8 +15,15 @@
 use crate::workflow::RealWorkflows;
 pub use crate::workflow::Workflows as Ui;
 
+use alloc::string::String;
+use alloc::vec::Vec;
+
 pub trait Sd {
     fn sdcard_inserted(&mut self) -> bool;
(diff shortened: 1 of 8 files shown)


src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -58,24 +86,66 @@ impl Hal for BitBox02Hal {
...
+        fn write_bin(&mut self, filename: &str, dir: &str, data: &[u8]) -> Result<(), ()> {
+            self.files
+                .entry(dir.into())
+                .or_default()
+                .insert(filename.into(), data.to_vec());
+            Ok(())
+        }
     }
 
     impl TestingHal<'_> {
```
NickeZ, line 151: Isn't it a bit weird to have the `impl` before the `struct`?

benma: Moved down one too much :see_no_evil: fixed

benma:
Merged but somehow GH is not detecting it, closing manually.

PR #1423, opened 2025-04-22
bootloader-fw-link: Use RAM to communicate

NickeZ:
Instead of writing to non-volatile storage between mcu resets, write to a well-known location.

RAM is random after hardware reset, but does not get cleared on MCU reset.


src/factorysetup.c
```diff
@@ -312,19 +313,8 @@ int main(void)
...
-        };
-        upside_down_t upside_down = {
-            .value = false,
-        };
-        if (!memory_bootloader_set_flags(auto_enter, upside_down)) {
-            // Not much we can do here.
-        }
-    }
+    // After reset we prefer to stay in bootloader
+    auto_enter = sectrue_u32;
```
NickeZ, line 317: The factory-setup could also just erase the first 8 bytes of itself when it is done, making it unbootable (and thus the bootloader would not try to boot it).

NickeZ:
this was mostly a proof of concept

PR #1426, opened 2025-04-25
[wip] convert cipher to Rust

benma:
This is using https://docs.rs/cbc/ and https://docs.rs/aes/latest/aes/.

For an alternative using rust-ctaes, see https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1427


benma:
Closing in favor of https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1427, which results in ~4kB smaller binaries.

PR #1427, opened 2025-04-26
convert cipher to Rust using rust-ctaes

benma:


src/rust/bitbox-aes/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+version = "0.1.0"
+authors = ["Shift Crypto AG <support@bitbox.swiss>"]
+edition = "2024"
+license = "Apache-2.0"
+
+[dependencies]
+bitbox02 = { path = "../bitbox02", optional = true }
+sha2 = { workspace = true }
+hmac = { version = "0.12", default-features = false, features = ["reset"] }
+ctaes = { git = "https://github.com/BitBoxSwiss/rust-ctaes", rev = "0df5295d9afe2da49e8bbe5eba8fd74e2d53f8ed" }
```
benma, line 26: Once https://github.com/BitBoxSwiss/rust-ctaes/pull/1 is merged, I could make a tag and reference that, or even push it to crates.io and get it from there instead.

NickeZ: yeah, push to crates.io sounds good to me.

benma: Done and updated this PR.

src/rust/bitbox-aes/src/lib.rs
```diff
@@ -0,0 +1,3065 @@
...
+    // The subtraction can't underflow due to the above conditions.
+    buf.truncate(buf.len() - pad_len);
+    Ok(())
+}
+
+fn encrypt(iv: &[u8; 16], key: &[u8; 32], plain: &[u8]) -> Vec<u8> {
+    // Allocate enough space so copying plain into it and padding it does not create copies that
+    // won't get zeroed.
+    let mut padded = zeroize::Zeroizing::new(Vec::with_capacity(pad_length(plain.len())));
+    padded.extend_from_slice(plain);
```
NickeZ, line 54: It would be nice to add `debug_assert(padded.as_ref().capacity() == padded_len)` after every place it could reallocate to ensure that it doesn't.

src/rust/bitbox-aes/src/lib.rs
```diff
@@ -0,0 +1,3065 @@
...
+fn pad_length(buf_len: usize) -> usize {
+    BLOCK_SIZE - buf_len % BLOCK_SIZE
+}
+
+fn pkcs7_pad(buf: &mut Vec<u8>) {
+    let pad_len = pad_length(buf.len());
+    buf.resize(buf.len() + pad_len, pad_len as u8);
+}
+
+fn pkcs7_unpad(buf: &mut Vec<u8>) -> Result<(), ()> {
```
NickeZ, line 37: If you are returning an error with 0 information you might as well panic.

src/rust/bitbox-aes/src/lib.rs
```diff
@@ -0,0 +1,3065 @@
...
+    if buf.len() % BLOCK_SIZE != 0 {
+        return Err(());
+    }
+    let pad_len = *buf.last().ok_or(())? as usize;
```
NickeZ, line 41: I don't understand `let pad_len = *buf.last().ok_or(())?`. Is pad_len the last in the cipher text?

src/rust/bitbox-aes/src/lib.rs
```diff
@@ -0,0 +1,3065 @@
...
+    if pad_len == 0 || pad_len > BLOCK_SIZE {
```
NickeZ, line 42: Isn't this a programmer error? i.e. shouldn't you panic? (can also use `debug_assert` or `assert` of course).

src/rust/bitbox-aes/src/lib.rs
benma: An `()` error is absolutely not the same thing as a panic. The caller will know there was an error and can handle it.

Not sure what the point is of adding more info if the caller does not care about which error path it was, can you explain?

src/rust/bitbox-aes/src/lib.rs
benma: pkcs7 padding you fill the padding bytes with `pad_len`. To unpad, you read the pad_len and remove that many bytes.

Actually I could/should also check that the last `pad_len` bytes are all `pad_len` as a sanity check.

src/rust/bitbox-aes/src/lib.rs
```diff
@@ -0,0 +1,3065 @@
...
+pub fn encrypt_with_hmac(iv: &[u8; 16], key: &[u8], plain: &[u8]) -> Vec<u8> {
+    let hash: [u8; 64] = sha512(key);
+    let (encryption_key, authentication_key) = hash.split_at(32);
+    let mut cipher = encrypt(iv, encryption_key.try_into().unwrap(), plain);
+    let mac: [u8; 32] = hmac_sha256(authentication_key, &cipher);
+    cipher.extend_from_slice(&mac);
+    cipher
+}
+
+pub fn decrypt_with_hmac(key: &[u8], cipher: &[u8]) -> Result<zeroize::Zeroizing<Vec<u8>>, ()> {
```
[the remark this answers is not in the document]
benma: I am not sure how you deifne "programmer issue" exactly. Doesn't that just mean the programmer has to check the condition instead of this function? In the bitbox02-rust case, the input is coming from flash memory.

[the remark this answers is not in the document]
benma: What's the point of a more concrete error than `()`  if the caller never inspects it or matches on it?

src/rust/bitbox-aes/src/lib.rs
benma: `debug_assert` is enabled in "non optimized builds". Do we have that? Our debug builds enable DEBUG, but I don't see where/how it translates to Rust build flags.

src/rust/bitbox-aes/src/lib.rs
benma: Also the caller can always `.unwrap()` if the caller prefers a panic over graceful error handling. The caller should be able to decide that. 

It's sth I dislike about the rust std: half the functions return Result/Option, the other half panics, it's quite inconsistent. But Result/Option is always better, the caller can always unwrap if they are sure their input is good.

NickeZ: Result/Option is always less performant and generates more code. I don't think the library should force the developer to use `unwrap`. I think code bases in general should be `unwrap`-free unless they are "work in progress".

stdlib will panic when it makes sense, like if allocation fails or if indexing fails.

NickeZ: So errors that are due to bugs are fine to panic on.

src/rust/bitbox-aes/src/lib.rs
NickeZ: Cool, I leave it up to you if you want to write that as a comment. I would have understood it without reading the pkcs7 padding spec if that was stated in a comment.

src/rust/bitbox-aes/src/lib.rs
NickeZ: debug_asserts are disabled in both builds in our case. to get a more similar code size of release and debug builds.

Both release and debug includes debug data (like names of symbols). This is not related to debug asserts.

NickeZ: unit tests have debug_asserts enabled for example.

src/rust/bitbox-aes/src/lib.rs
NickeZ: How will the user code handle the error if it doesn't know what path failed?

src/rust/bitbox-aes/src/lib.rs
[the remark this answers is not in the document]
NickeZ: If you return a Result I expect the code to be able to handle the issue. If the point is to propagate the issue all the way up to the end-user. then it should be as precise as possible so that the end-user can change the input and try again.

src/rust/bitbox-aes/src/lib.rs
NickeZ: In this specific case I guess it isn't a programmer bug. But if you don't return the cause for the error then the library user will have to write all the logic for figuring out what is wrong to inform the end-user?

benma: I don't disagree in principle, a concrete error enum with different failure cases makes sense if the caller needs to now. But in our case the caller does not care, and within our project I don't care much for adding types just for the sake of it if every call-site just discards it. 

At least in the original C implementation the callsites don't care about this error - if you feel like the errors should be handled specifically based on the failure case, it's a separate discussion.

benma: > So errors that are due to bugs are fine to panic on.

I could agree, but even with programmer bugs, you'd prefer an API call to fail gracefully in case of a bug instead of crashing everything.

But "programmer bug" is not clearly defined. Is it a programmer bug if the flash memory contains something unexpected? Is it the caller's job to verify the input it's valid or is it this module's job? I don't know if there is a clear line.

src/rust/bitbox-aes/src/lib.rs
benma: Added the comments and the check that all `pad_len` bytes must equal `pad_len`.

src/rust/bitbox-aes/src/lib.rs
benma: Amazing, by adding and running into the assert, I found that I had a bug - the capacity should not be `pad_length(plain.len())` but `plain.len() + pad_length(plain.len())`!

I used `assert!()` though not `debug_assert!()`, since `cargo test` actually did not trigger `debug_assert!()` :thinking:

NickeZ (approved):
utACK

PR #1428, opened 2025-04-27
optiga: move config sanity check to factorysetup only

benma:
It causes a long delay before the orientation animation. It's a sanity
check only, so doing it as part of the factory setup only suffices.

We also enable it in debug builds, so it catches a misconfigured
device (e.g. if you forgot to run factorysetup on a new PCB).


NickeZ (approved):
utACK!

PR #1429, opened 2025-04-28
Small keccak

benma:
https://github.com/RustCrypto/hashes/issues/561 mentioned the no_unroll feature which I wasn't aware of until then.


NickeZ (approved):
utACK

PR #1432, opened 2025-05-01
fonts: Make fonts closer to original

NickeZ:
Fonts are generated with `ugui.h` as the path to `ugui.h`. We add the full path to `ugui.h` as an include directory which let's us use fonts with fewer manual modifications.

Bump the tool so it supports homebrew and use our constant for `UG_FONT_DATA`.


benma (approved):
Does it even make sense to have the tool as a submodule of this repo?

NickeZ:
> Does it even make sense to have the tool as a submodule of this repo?

I guess it makes as much sense as dockerdev and docker_exec? No strong opinion here. I use my personal forks of those scripts for example.

PR #1433, opened 2025-05-04
rust: allow cancelling an async task from within

benma:
Calling `cancel()` within a running Rust task before could have lead to the panic `"spin: illegal executor state"` if there was a yield point after the `cancel()` call (in other words, if polling the task afterwards returned `Pending`, not `Ready`).

We don't techncially need this commit, as if one actually wanted to `cancel()`, you could also return right away, there is no point in `.await`ing anyhing else afterwards. Then again, there is no reason to panic in this scenario.


benma:
Closing as we don't really need to allow this :shrug: It should always be sufficient to simply return without another await point.

PR #1435, opened 2025-05-05
build: osx support in build builds

NickeZ:
Make it work to build on osx when building for the build machine. Unfortunately tests doesn't work as there is no linker that supports `--wrap`. At least the simulator works.

Remove our bespoke FindCMocka.cmake in favor of standardized "pkg-config", which works on osx and in the development container.


benma (approved):
utACK

PR #1440, opened 2025-05-07
Add latest ble firmware

NickeZ:
new firmware that works better with 320 byte ble packets.


benma (changes requested):
- Also update the hash in bluetooth.rs (must)
- Would be nice if the file comitted was the actual .bin file, not the .o file, and you add a CMake thing to turn it into the object file that will be linked to the factorysetup (in this or a separate PR)

NickeZ:
> * Also update the hash in bluetooth.rs
Done
 
>     * Would be nice if the file comitted was the actual .bin file, not the .o file, and you add a CMake thing to turn it into the object file that will be linked to the factorysetup (in this or a separate PR)

Will make a separate PR

benma (approved):
utACK

PR #1442, opened 2025-05-08
da14531: Use binary file as source instead of object file

NickeZ:
CI doesn't pass. It think it might be due to old binutils generating some code that the ancient version of gcc doesn't recognize. I'll rebase this on master after we upgrade to gcc-14


benma (approved):
utACK, thanks!

PR #1443, opened 2025-05-08
add communication_mode_is_ble() and expose to Rust

benma:
Bluetooth upgrade will need to know whether to reset the BLE chip based on this.

### DIFF
--- a/src/communication_mode.c
+++ b/src/communication_mode.c
@@ -1,0 +1,40 @@
+// Copyright 2025 Shift Cryptosecurity AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "communication_mode.h"
+
+#include "memory/memory_shared.h"
+
+static bool _usb_hww_request_seen = false;
+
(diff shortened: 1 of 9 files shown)


src/communication_mode.c
```diff
@@ -0,0 +1,29 @@
...
+static bool _usb_hww_request_seen = false;
+
+void communication_mode_enable_usb(void)
+{
+    _usb_hww_request_seen = true;
+}
+
+bool communication_mode_is_ble(void)
+{
+    return !_usb_hww_request_seen && memory_get_platform() == MEMORY_PLATFORM_BITBOX02_PLUS;
```
NickeZ, line 28: this is quite a lot less efficient since it always goes out and reads flash in case `_usb_hww_request_seen==false`.

You could add a 
```
static has_ble = memory_get_platform() == MEMORY_PLATFORM_BITBOX02_PLUS;
return !_usb_hww_request_seen && has_ble;
```

src/communication_mode.c
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "communication_mode.h"
+
+#include "memory/memory_shared.h"
+
+static bool _usb_hww_request_seen = false;
+
+void communication_mode_enable_usb(void)
```
NickeZ, line 21: I would prefer if the two functions match. So more something like

`disable_ble` / `ble_enabled` or `enable_usb` / `usb_enabled`.

I think I also prefer to suffix the verb, so `usb_disable` / `usb_enabled`

With the current naming you have to implicitly understand that enabling usb disables ble.

NickeZ (changes requested):
Would like some change to naming.

src/communication_mode.c
benma: Renamed :+1:

NickeZ (approved):
utACK

benma:
Merged but GH failed to detect it. Closing.

PR #1447, opened 2025-05-09
ble: Wait with setting the product string to after orientation

NickeZ:


src/da14531/da14531.h
```diff
@@ -36,9 +36,12 @@ void da14531_power_down(struct ringbuffer* uart_out);
...
 void da14531_reset(struct ringbuffer* uart_out);
 
-// product is an array of characters to be set as product characteristic
+// product is an array of characters to be set as product characteristic (not null terminated)
 // procuct_len is the number of characters in the product array
 // uart_out is the queue where to put the outgoing serially encoded bytes
-void da14531_set_product(const char* product, uint16_t product_len, struct ringbuffer* uart_out);
+void da14531_set_product(
+    volatile const uint8_t* product,
+    volatile uint16_t product_len,
```
benma, line 44: why volatile?

benma (approved):
tACK

src/da14531/da14531.h
NickeZ: Everything you touch from interrupt context must be volatile. And since we modify this in a timer callback it must be volatile. We have quite a lot of UB because we aren't really following this.

benma: Maybe add a comment for clarity.

PR #1448, opened 2025-05-12
ble: Make bootloader pairing confirmation better

NickeZ:


src/bootloader/bootloader.c
```diff
@@ -310,6 +310,30 @@ static void _render_default_screen(void)
...
+    bootloader_pairing_request = true;
+    uint32_t pairing_code_int = (*(uint32_t*)&bootloader_pairing_code_bytes[0]) % 1000000;
+    unsigned int part0 = pairing_code_int / 10000;
+    unsigned int part1 = (pairing_code_int / 100) % 100;
+    unsigned int part2 = pairing_code_int % 100;
+    char code_str[10] = {0};
+    snprintf(code_str, sizeof(code_str), "%02u %02u %02u", part0, part1, part2);
+    UG_ClearBuffer();
+    if (confirmed) {
+        UG_PutString(15, 0, "Confirm on app", false);
```
benma, line 327: This flashes for a split second even if the pairing was accepted in the app before the BitBox. Could this be fixed?

src/bootloader/bootloader.c
```diff
@@ -310,6 +310,30 @@ static void _render_default_screen(void)
...
+
+void bootloader_render_ble_confirm_screen(bool confirmed)
+{
+    bootloader_pairing_request = true;
+    uint32_t pairing_code_int = (*(uint32_t*)&bootloader_pairing_code_bytes[0]) % 1000000;
+    unsigned int part0 = pairing_code_int / 10000;
+    unsigned int part1 = (pairing_code_int / 100) % 100;
+    unsigned int part2 = pairing_code_int % 100;
+    char code_str[10] = {0};
+    snprintf(code_str, sizeof(code_str), "%02u %02u %02u", part0, part1, part2);
```
benma, line 324: Would be nice to show a title here, like "Pairing code" (same as in the noise pairing).

src/bootloader/bootloader.c
```diff
@@ -310,6 +310,30 @@ static void _render_default_screen(void)
...
+    UG_ClearBuffer();
+    if (confirmed) {
+        UG_PutString(15, 0, "Confirm on app", false);
+    } else {
+        image_cross(0, 0, IMAGE_DEFAULT_CROSS_HEIGHT);
+        image_checkmark(SCREEN_WIDTH - 20, 0, IMAGE_DEFAULT_CHECKMARK_HEIGHT);
```
benma, line 330: The positions seem off. Use the same offsets as in the firmware? See

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/src/ui/components/icon_button.c#L53-L59

src/bootloader/bootloader.c
benma, line 324: Why a different format to https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/src/da14531/da14531_handler.c#L142 (and different code to derive it)? Should be consistent I guess, and could be put into a helper function shared by both BL and FW.

benma (changes requested):

src/bootloader/bootloader.c
NickeZ: done

src/bootloader/bootloader.c
NickeZ: done

NickeZ: done

src/bootloader/bootloader.h
```diff
@@ -15,6 +15,10 @@
 #ifndef _BOOTLOADER_H_
 #define _BOOTLOADER_H_
 
+#include <stdbool.h>
+
 void bootloader_jump(void);
+void bootloader_render_default_screen(void);
+void bootloader_render_ble_confirm_screen(bool confirmed);
```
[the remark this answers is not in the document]
NickeZ: done

src/bootloader/bootloader.c
NickeZ: It will require a lot of work. Should we consider doing it later?

benma (approved):
tACK

PR #1459, opened 2025-05-14
da14531: invalid firmware

NickeZ:
If there is no valid firmware UART communication is turned off to avoid spending a lot of cpu cycles trying to boot the da14531.

The firmware version reading function must also fail "gracefully" so that the "device info" endpoint gets some data. If the metadata_version is wrong we return all 0s.


benma (approved):
utACK

PR #1461, opened 2025-05-15
da14531: Add concept of private name

NickeZ:
This name will only ever be shared by the BLE chip if there is an established encrypted connection

embed ble firmware v3


NickeZ:
Not needed

PR #1462, opened 2025-05-15
usb_processing: increase watchdog timeout once

NickeZ:
The firmware typically expects keep-alive requests every 300ms from the app. However keystore unlock takes about 5-6s so for that single call we reset the watchdog counter to a lower value allowing approximately a 7s gap between messages from the app.

In practice this has worked anyway for the usb stack due to the order of things in the main loop. I.e. check the next incoming message before checking the timeout counter.

### DIFF
--- a/src/usb/usb_processing.c
+++ b/src/usb/usb_processing.c
@@ -39,7 +39,7 @@ extern struct timer_descriptor TIMER_0;
  * Amount of time to wait before an outstanding operation times out
  * (if the client is closed).
  */
-#define USB_OUTSTANDING_OP_TIMEOUT_MS (7000)
+#define USB_OUTSTANDING_OP_TIMEOUT_MS (500)
 #define USB_TIMER_TICK_PERIOD_MS (100)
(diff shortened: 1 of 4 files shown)


src/usb/usb_processing.c
```diff
@@ -39,7 +39,7 @@ extern struct timer_descriptor TIMER_0;
  * Amount of time to wait before an outstanding operation times out
  * (if the client is closed).
  */
-#define USB_OUTSTANDING_OP_TIMEOUT_MS (7000)
+#define USB_OUTSTANDING_OP_TIMEOUT_MS (300)
```
benma, line 42: The previous value was 500. Why lower it? Seems a little risky, a sit leaves less time for the host to send the poll/retry msg.

NickeZ: ah, I must have misremembered.

benma: The commit msg also mentions the host polls every 300ms, but it's every 200ms.

NickeZ: It mentions that the firmware expects within 300ms, after updating to 500ms I woudl say it expects a keepalive within 500ms. Of course the app can send every 200ms if it wants to.

NickeZ: anyway, comment addressed

benma (approved):
tACK

PR #1464, opened 2025-05-19
Logo

benma:


src/ui/components/ui_images.c
```diff
@@ -122,3 +125,142 @@ void image_sdcard(bool mirror)
     UG_DrawLine(
         x - m * (w - c - 1), y - m * h / 2, x - m * (w - 1), y - m * h / 2 + m * c, C_WHITE);
 }
+
+// TODO: Make new logos for plus!
```
NickeZ, line 129: can't you remove this TODO?

benma: i can :smile: and did now

src/ui/components/ui_images.c
```diff
@@ -220,14 +231,54 @@ static const uint8_t IMAGE_BITBOX02_PLUS_LOGO[] = {
...
+            },
+    };
+    return result;
+#else
+#error "unknown platform"
+#endif
+}
+#endif // defined(BOOTLOADER)
+
+#if !defined(BOOTLOADER)
```
NickeZ, line 280: why not just `#else` here?

NickeZ (commented):
Why did you go with two function names when they are "mutually exclusive" so both could have the same name?

benma:
> Why did you go with two function names when they are "mutually exclusive" so both could have the same name?

Just for code clarity - I thought two functions with clear purposes are easier to read and understand than a monster function with a ton of `#if/#else` inside the function body.

NickeZ:
> > Why did you go with two function names when they are "mutually exclusive" so both could have the same name?
> 
> Just for code clarity - I thought two functions with clear purposes are easier to read and understand than a monster function with a ton of `#if/#else` inside the function body.

But only one of them are instantiated, so they could have the same name?

benma:
> But only one of them are instantiated, so they could have the same name?

Yeah, I still thought it's clearer. Changed it now to be the same name.

NickeZ (approved):
utack

PR #1466, opened 2025-05-19
ethereum: allow fetching xpub at m/44'/60'/0' and m/44'/1'/0'

benma:
NuFi wants to fetch the xpub at the hardened prefix only and derive children from there on the host for consitency with other HWWs.

The BitBox still only allows displaying addresses and signing txs at `m/44'/coin'/0'/0/*`.


Beerosagos (approved):
utACK

PR #1467, opened 2025-05-21
usb: reboot to bootloader on Abort

benma:
If there is a bug in our USB processing, the unlock flow could Abort there, which would brick a device. We set auto_enter to true so that upon reconnect, we load the bootloader so there is a chance to upgrade the firmware.


NickeZ (approved):
lgtm

PR #1468, opened 2025-05-21
make noise pairing confirmation optional over BLE

benma:
Connecting over Bluetooth requires Bluetooth pairing including pairing code confirmation, so a second noise pairing confirmation is redundant.

This does not prevent the host from requesting a pairing code confirmation on demand anyway at any time, it just does not require confirmation on the device to proceed with communicating over noise, when communicting through the BLE chip.


NickeZ (approved):
tACK

PR #1471, opened 2025-05-21
da14531: Send over name on boot

NickeZ:
If the bootloader has booted the da14531 chip, we need to send over the current name.

### DIFF
--- a/src/firmware_main_loop.c
+++ b/src/firmware_main_loop.c
@@ -56,6 +56,12 @@ void firmware_main_loop(void)
     uint8_t uart_write_buf[UART_OUT_BUF_LEN];
     ringbuffer_init(&uart_write_queue, &uart_write_buf, UART_OUT_BUF_LEN);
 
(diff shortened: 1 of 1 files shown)


src/firmware_main_loop.c
```diff
@@ -56,6 +56,13 @@ void firmware_main_loop(void)
     uint8_t uart_write_buf[UART_OUT_BUF_LEN];
     ringbuffer_init(&uart_write_queue, &uart_write_buf, UART_OUT_BUF_LEN);
 
+    /// If the bootloader has booted the BLE chip, it isn't aware of the private name
+    // Send over the private name
```
benma, line 60: Outdated comment, there is no "private name"

benma (changes requested):

src/firmware_main_loop.c
NickeZ: fixed

benma (approved):
utACK

PR #1474, opened 2025-05-24
Nickez/fix u2f safari

NickeZ:


src/u2f.c
```diff
@@ -865,35 +824,20 @@ static void _cmd_authenticate(const Packet* in_packet, Packet* out_packet)
...
 /**
  * Process a U2F message
  */
 static void _cmd_msg(const Packet* in_packet, Packet* out_packet, const size_t max_out_len)
 {
     (void)max_out_len;
+
+    // Give the user 5s to respond to u2f requests (the FIDO Client in u2f is not required to ping
+    // every 200ms like the BitBoxApp does even though most implementations do).
+    usb_processing_timeout_reset(-50);
```
benma, line 836: If it times out the confirm screen just disappears? 

Do you know of a browser that does not ping to keep it alive?

src/u2f/u2f_app.c
```diff
@@ -219,7 +219,9 @@ void u2f_app_confirm_start(enum u2f_app_confirm_t type, const uint8_t* app_id)
 async_op_result_t u2f_app_confirm_retry(enum u2f_app_confirm_t type, const uint8_t* app_id)
 {
     if (_state.outstanding_confirm != type || !MEMEQ(app_id, _state.app_id, 32)) {
-        Abort("Arbitration failed for U2F confirmation.");
+        // CID used a new app_id, clearly invalid
+        // TODO: Clear u2f state
+        return ASYNC_OP_NOT_READY;
```
benma, line 224: Did you run into this in the wild? The error return sounds like it will be ready later, but it won't be?

benma (approved):
utACK

benma (changes requested):
Oh, please add a changelog entry, so we announce it properly when released

src/u2f.c
benma: Isn't 5s seems a bit too short. I'd bump it to at least 10s, maybe 15s. Would be frustrating to miss it. Do you know what timeout Yubikey uses?

NickeZ: So we can probably remove this timeout, even if 5s is what I found in other open source implementations.

(btw it wouldn't actually mean that the user would only have 5s. because I guess all FIDO clients try again within that 5s.) I think FIDO clients typically retry up to at least 30s. maybe longer

benma (approved):
utACK

PR #1475, opened 2025-05-26
memory/spi_mem: add read check after write/erase

Beerosagos:
This adds a read back of the written page data to check that the writing completed successfully.

### DIFF
--- a/src/memory/spi_mem.c
+++ b/src/memory/spi_mem.c
@@ -174,6 +174,21 @@ bool spi_mem_sector_erase(uint32_t sector_addr)
     // --- Wait for write to end ---
     _spi_mem_wait();
 
+    // --- Check that sector has been actually erased ---
+    uint8_t page_data[SPI_MEM_PAGE_SIZE];
+    for (size_t i = 0; i < SPI_MEM_SECTOR_SIZE / SPI_MEM_PAGE_SIZE; i++) {
+        uint32_t page_addr = sector_addr + i * SPI_MEM_PAGE_SIZE;
+        if (!spi_mem_page_read(page_addr, page_data)) {
+            util_log("Read after sector erase at %p failed", (void*)(uintptr_t)page_addr);
+            return false;
+        }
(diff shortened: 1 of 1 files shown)


src/memory/spi_mem.c
```diff
@@ -242,6 +242,18 @@ static bool _spi_mem_page_write(uint32_t page_addr, const uint8_t* input)
     // --- Wait for write to end ---
     _spi_mem_wait();
 
+    // --- Check input data has been properly written ---
```
benma, line 245: I wonder if this slows down writing the BLE firmware in factorysetup. There, it's not needed as we anyway read it back and verify it.

Could you check if if this causes a slowdown? Could also just exclude it there using `#if FACTORYSETUP == 0`

benma: Thanks for checking :+1:

Beerosagos:
Updated and added a check for sector erase as well. @benma PTAL :pray:

src/memory/spi_mem.c
```diff
@@ -174,6 +174,21 @@ bool spi_mem_sector_erase(uint32_t sector_addr)
...
 
+    // --- Check that sector has been actually erased ---
+    uint8_t page_data[SPI_MEM_PAGE_SIZE];
+    for (int i = 0; i < SPI_MEM_SECTOR_SIZE / SPI_MEM_PAGE_SIZE; i++) {
+        uint32_t page_addr = sector_addr + i * SPI_MEM_PAGE_SIZE;
+        if (!spi_mem_page_read(page_addr, page_data)) {
+            util_log("Read after sector erase at %p failed", (void*)(uintptr_t)page_addr);
+            return false;
+        }
+        for (int j = 0; j < SPI_MEM_PAGE_SIZE; j++) {
```
benma, line 185: Afaik size_t is normally the type for array indices, though I am not sure if there is a real downside to using `int` :thinking: @NickeZ does it matter which type to use?

benma (approved):
utACK

PR #1477, opened 2025-05-28
keystore: port bip85_ln() to Rust

benma:


src/rust/bitbox02-rust/src/hash.rs
```diff
@@ -0,0 +1,54 @@
...
+/// Implements the digest traits for Sha512 backing it with the wally_sha512 C function. This is
+/// done to avoid using a second sha512 implementation like `sha2::Sha512`, which bloats the binary
+/// by an additional ~12.7kB (at the time of writing).
+///
+/// This implementation accumulates the data to be hashed in heap, it does **not** hash in a
+/// streaming fashion, even when using `update()`.
+#[derive(Default, Clone)]
+pub struct Sha512(Vec<u8>);
+
+impl digest::HashMarker for Sha512 {}
```
NickeZ, line 26: Maybe name it `Sha512Message`, `Sha512Data` or something? It took me a while to understand that it contains the message and not the digest while reading the code.

src/rust/bitbox02-rust/src/hash.rs
```diff
@@ -0,0 +1,54 @@
...
+use alloc::vec::Vec;
+
+/// Implements the digest traits for Sha512 backing it with the wally_sha512 C function. This is
+/// done to avoid using a second sha512 implementation like `sha2::Sha512`, which bloats the binary
+/// by an additional ~12.7kB (at the time of writing).
+///
+/// This implementation accumulates the data to be hashed in heap, it does **not** hash in a
+/// streaming fashion, even when using `update()`.
+#[derive(Default, Clone)]
+pub struct Sha512(Vec<u8>);
```
NickeZ: Maybe name it `Sha512Message`, `Sha512Data` or something? It took me a while to understand that it contains the message and not the digest while reading the code.

NickeZ: I changed my mind, I suppose `Sha512` is a good name. It is just strange that it contains the message. But it is described in the comment.

I guess
```rust
pub struct Sha512 {
    message: Vec<u8>
}
```

would make it slightly clearer.

benma: Makes sense, I named the struct field as you suggested now.

src/rust/bitbox02-rust/src/hash.rs
```diff
@@ -0,0 +1,56 @@
...
+
+impl digest::Update for Sha512 {
+    fn update(&mut self, data: &[u8]) {
+        self.message.extend(data);
+    }
+}
+
+impl digest::Reset for Sha512 {
+    fn reset(&mut self) {
+        self.message = vec![];
```
NickeZ: This works, but `Vec::clear` is probably faster because it would not allocate a new vec.

https://doc.rust-lang.org/std/vec/struct.Vec.html#method.clear

benma: Fixed :+1:

src/keystore.h
```diff
@@ -80,7 +80,7 @@ USE_RESULT keystore_error_t keystore_create_and_store_seed(
 /** Unlocks the keystore seed or checks the password:
  * If the keystore is locked, it decrypts and loads the seed, unlocking the keystore:
  * 1) loads the stored seed and tries to decrypt using password.
- * 2) if successful, the bip39 seed should be derived using keystore_unlock_bip39().
+ * 2) if successful, the bip39 seed should be derived using keystoreu_nlock_bip39().
```
Beerosagos: Typo :)

Beerosagos (approved):
utACK, with a small typo

PR #1481, opened 2025-06-03
default pw keyboard to digits

benma:


src/rust/bitbox02-rust/src/workflow/password.rs
```diff
@@ -62,14 +87,12 @@ pub async fn enter(
 
 pub enum EnterTwiceError {
     DoNotMatch,
-    Cancelled,
+    EnterError(EnterError),
```
benma, line 90: Not very happy about all this error wiring to pass the `Memory` error up in case `memory::get_securechip_type()` fails, but not sure how to improve it easily.

NickeZ (approved):
tACK

src/rust/bitbox02-rust/src/workflow/password.rs
NickeZ: There are many tradeoffs to the various ways of handling errors, so we would need a separate discussion on that.

If you just want to minimize the boilerplate there are crates like `thiserror` that provides macros to do that.

PR #1482, opened 2025-06-03
usb: update strings

benma:


NickeZ:
What about https://github.com/BitBoxSwiss/bitbox02-firmware/blob/f570425e5b4d07f4003102aa1e3bac09dad8018f/src/bootloader/startup.c#L43?

benma:
> What about
> 
> https://github.com/BitBoxSwiss/bitbox02-firmware/blob/f570425e5b4d07f4003102aa1e3bac09dad8018f/src/bootloader/startup.c#L43
> ?

That's not the USB descriptor string but the peripheral characteristic string, which we want to keep short. (Same in the orientation workflow where we set the corresponding FW string)

PR #1489, opened 2025-06-04
memory/smarteeprom: skip initialization reboot during factorysetup

benma:
During factorysetup, we don't do any smarteeprom read/writes, but the reboot is messing with RTT by initializing it twice.


NickeZ (approved):
utACK

PR #1490, opened 2025-06-04
api/bluetooth: use protected write

benma:
We introduced a write lock in the SPI mem, so we need to use the appropriate function to be able to write to the protected area.


NickeZ (approved):
utACK

PR #1495, opened 2025-06-05
scripts/get_version: Ignore untracked files

NickeZ:


NickeZ:
Just add paths to be ignored to `.git/info/exclude` instead.

PR #1497, opened 2025-06-06
CI: Rewrite to native github actions

NickeZ:
Reduces CI time to 5 minutes when docker container doesn't have to be rebuilt.


.github/workflows/ci-common.yml
```diff
@@ -0,0 +1,219 @@
...
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+          submodules: recursive
+
+      - name: Mark directory as safe
+        run: git config --global --add safe.directory $GITHUB_WORKSPACE
+
+      - name: Build unit-tests
+        run: make -j$(($(nproc) + 1)) unit-test
```
benma, line 98: Isn't simply `-j` enough? Why nproc + 1, did you measure that this is faster?

NickeZ: `-j` means parallelize as much as possible, so building all c files at the same time. Usually it is recommended to use 1 + number of CPUs to get max overall speed

NickeZ: There seems to have been some discussion recently to use fewer parallel jobs on systems with high cpu/memory ratio:

https://unix.stackexchange.com/questions/208568/how-to-determine-the-maximum-number-to-pass-to-make-j-option

But the second answer still mentions the nproc+1, which is what I historically have used.

PR #1500, opened 2025-06-10
factorysetup: don't skip BLE if already flashed

benma:
If the BLE boot check failed further below, repeating the factory setup made it report success, as _verify_ble() only checks if the BLE firmware was flashed properly already, and the boot check is skipped in that case.

For now we simply remove this check so every attempt tries the whole BLE setup & verification procedure. To be refactored later.


NickeZ:
replaced by https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1501

NickeZ (approved):
LGTM

PR #1506, opened 2025-06-12
evm: add HyperEVM and Sonic to known networks

Tomasvrba:


src/rust/bitbox02-rust/src/hww/api/ethereum/params.rs
```diff
@@ -99,6 +99,20 @@ const PARAMS: &[Params] = &[
         name: "Gnosis Chain",
         unit: "xDAI",
     },
+    Params {
+        coin: None,
+        bip44_coin: 60 + HARDENED,
+        chain_id: 999,
```
benma, line 105: https://evmchainlist.org/?search=999 says it's Wanchain Testnet

Tomasvrba: I usually use: https://chainlist.org/?search=hype

In any case, I tested it on the BB and FW says unknown id 999 when interacting with HypeEVM

PR #1508, opened 2025-06-17
build: Remove workaround for bug in 1.74 and 1.75.

NickeZ:
The fix for this bug was merged in 1.76 and we use a more recent compiler.

https://github.com/rust-lang/rust/pull/119802


NickeZ:
> Nice! Did we get back ~11.5kB of space?

18.6kB :)

PR #1516, opened 2025-06-18
Rust tests

benma:


src/rust/bitbox02/src/keystore.rs
```diff
@@ -509,6 +509,64 @@ mod tests {
         assert!(bip39_mnemonic_from_seed(b"foo").is_err());
     }
 
+    #[test]
+    fn test_unlock() {
+        mock_memory();
+        lock();
+
+        assert!(matches!(unlock("password"), Err(Error::Unseeded)));
```
NickeZ, line 558: would `assert_eq!(...)` work in all the cases you use `assert!(matches!(...))`

NickeZ: I guess the `Error` type would need to derive `PartialEq` for that to work. Maybe not worth it..

benma: Let me know if you have a preference, otherwise I'll leave it :P

NickeZ: leave it as is!

PR #1518, opened 2025-06-19
use Rust bitcoin::bip32 over libwally for bip32 derivations

benma:
In an effort to go full Rust for all the benefits of Rust over C that we love.


NickeZ (approved):
utACK

PR #1519, opened 2025-06-20
Bluetooth typos

benma:


messages/bitbox02_system.proto
```diff
@@ -29,7 +29,7 @@ message DeviceInfoResponse {
   message Bluetooth {
     // Hash of the currently active Bluetooth firmware on the device.
     bytes firmware_hash = 1;
-    // Firmware version, formated as "major.minor.patch".
+    // Firmware version, formated as an unsigned integer "1", "2", etc.
     string firmware_version = 2;
```
benma, line 33: we should have turned this into an integer when we changed from semver :see_no_evil:

PR #1521, opened 2025-06-23
ssd1312: enforce column addressing to prevent horizontal offset/wrap

benma:
Explicitly set column address to 0 during initialization and screen updates to resolve intermittent ~20px horizontal offset and wrapping.

Details:
- Initialize columns to 0 in `ssd1312_configure()` via SSD1312_CMD_SET_LOW_COL(0) and SSD1312_CMD_SET_HIGH_COL(0).
- Reset columns for every page in `ssd1312_update()` to override residual state.
- Fixes rare display corruption caused by uninitialized column addressing in Page Mode, where stale column values shifted pixels horizontally.

### DIFF
--- a/src/ui/oled/ssd1312.c
+++ b/src/ui/oled/ssd1312.c
@@ -83,6 +83,8 @@ static uint8_t* _frame_buffer;
 void ssd1312_configure(uint8_t* buf)
 {
     _frame_buffer = buf;
+    oled_writer_write_cmd(SSD1312_CMD_SET_LOW_COL(0));
+    oled_writer_write_cmd(SSD1312_CMD_SET_HIGH_COL(0));
     oled_writer_write_cmd(SSD1312_CMD_SET_DISPLAY_OFF);
     oled_writer_write_cmd_with_param(SSD1312_CMD_SET_CONTRAST_CONTROL, 0xff);
     oled_writer_write_cmd_with_param(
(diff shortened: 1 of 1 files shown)


src/ui/oled/ssd1312.c
```diff
@@ -118,6 +120,8 @@ void ssd1312_update(void)
     /* The SSD1312 has one page per 8 rows. One page is 128 bytes. Every byte is 8 rows */
     for (size_t i = 0; i < 64 / 8; i++) {
         oled_writer_write_cmd(SSD1312_CMD_SET_PAGE_START_ADDRESS(i));
```
NickeZ, line 122: Could you add a comment that this only has been experienced on a single NOVA device? Because this fix fixes the symptom and not the underlying issue.

The column address is autoincremented for every access so if we end up with an off-by-one or more it means not all bytes arrived at the screen.

NickeZ (approved):
utACK, I looked at the data sheet and this looks reasonable.

src/ui/oled/ssd1312.c
benma: Done, sounds good? Not sure if saying it was one Nova helps, it could probably be reproduced on other novas (or even bb02s), but reproducing is a pain so we haven't tried on more devices.

Any idea for the actual cause?

NickeZ: could be faulty hardware

PR #1524, opened 2025-07-15
ui: Fix null ptr derefs

NickeZ:
When these helpers were called with `parent==NULL`, they dereferenced null.

### DIFF
--- a/src/ui/ui_util.c
+++ b/src/ui/ui_util.c
@@ -128,11 +128,15 @@ static void _position_left(component_t* parent, component_t* child)
 
 static void _position_right(component_t* parent, component_t* child)
 {
-    child->position.left = parent->position.left + parent->dimension.width - child->dimension.width;
+    uint16_t left = 0;
+    if (parent) {
+        left = parent->position.left + parent->dimension.width;
+    }
+    child->position.left = left - child->dimension.width;
     // The frame of some SSD1312 screens slightly covers around 0.5px on each side, cutting off the
     // text a bit, especially when looking at the device at an angle. We move by one pixel to
     // improve visibility.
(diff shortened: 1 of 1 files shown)


src/ui/ui_util.c
```diff
@@ -128,11 +128,15 @@ static void _position_left(component_t* parent, component_t* child)
 
 static void _position_right(component_t* parent, component_t* child)
 {
-    child->position.left = parent->position.left + parent->dimension.width - child->dimension.width;
+    uint16_t left = 0;
```
benma, line 131: _position_right() without a parent does not make any sense, but if anything, it should start at SCREEN_WIDTH, otherwise it will not be right-positioned. How did this even work before to right-align if the parent was NULL?

Below `left - child->dimension.width` would also underflow if `left` is `0`.

Alternatively, Abort() if parent is NULL?

benma (changes requested):
I would expect this PR to rather provide the parent to `confirm_gesture_create()` instead.

NickeZ:
> I would expect this PR to rather provide the parent to `confirm_gesture_create()` instead.

Ok, I assumed parent==NULL meant that the parent was the entire screen.

benma:
> > I would expect this PR to rather provide the parent to `confirm_gesture_create()` instead.
> 
> Ok, I assumed parent==NULL meant that the parent was the entire screen.

But the function does not do that (before or now). Which made me take a second look, and `ui_util_position_right_top(parent, confirm_gesture);` in `confirm_gesture.c` may be pointless? The component has a _render() function that manually renders the longtouch arrows on the right side. Maybe just deleting that line is enough?

PR #1526, opened 2025-07-17
usb: Fix timeout bug

NickeZ:
The FW was waiting for the "IN" endpoint to be ready when it should've been waiting for the "OUT" endpoint.


benma (approved):
tACK

PR #1532, opened 2025-07-24
Add missing dependencies.

FilipLaurentiu:
Source:
https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/include/secp256k1_ecdsa_s2c.h https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/include/secp256k1_ecdsa_adaptor.h https://github.com/ElementsProject/libwally-core/blob/master/include/wally_crypto.h https://github.com/ElementsProject/libwally-core/blob/master/include/wally_bip39.h https://github.com/ElementsProject/libwally-core/blob/master/include/wally_bip32.h https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1.h

### DIFF
--- a/src/secp256k1.h
+++ b/src/secp256k1.h
@@ -1,0 +1,885 @@
+#ifndef SECP256K1_H
+#define SECP256K1_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stddef.h>
+
+/** Unless explicitly stated all pointer arguments must not be NULL.
+ *
+ * The following rules specify the order of arguments in API calls:
+ *
+ * 1. Context pointers go first, followed by output arguments, combined
+ *    output/input arguments, and finally input-only arguments.
+ * 2. Array lengths always immediately follow the argument whose length
+ *    they describe, even if this violates rule 1.
+ * 3. Within the OUT/OUTIN/IN groups, pointers to data that is typically generated
+ *    later go first. This means: signatures, public nonces, secret nonces,
+ *    messages, public keys, secret keys, tweaks.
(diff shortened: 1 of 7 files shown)


benma:
What problem are you trying to solve? There are no missing deps, similar/same files exist in the source tree already, check the `external/libwally-core` folder.

FilipLaurentiu:
> What problem are you trying to solve? There are no missing deps, similar/same files exist in the source tree already, check the `external/libwally-core` folder.

I was not aware that I should pull the git submodule and the repo didn't compile. My bad, sorry.

But I can take this opportunity and ask you something about [this article](https://benma.github.io/2020/10/16/rust-zeroize-move.html) . The reason I pull this repo is to take a look at the implementation, but I see there is no use of `Box`, as you suggested in the post, why ?
I wanted to PM you on X but I couldn't

benma:
@FilipLaurentiu since Rust has no strong guarantees about zeroing, we do it on a best effort basis only. 

That being said, in the firmware we mostly deal with secret bytestrings or strings, so you'll find `Zeroizing<Vec<u8>>` or `Zeroizing<String>`. Both `Vec<u8>` and `String` are heap objects like `Box<...>`.

FilipLaurentiu:
> @FilipLaurentiu since Rust has no strong guarantees about zeroing, we do it on a best effort basis only.
> 
> That being said, in the firmware we mostly deal with secret bytestrings or strings, so you'll find `Zeroizing<Vec<u8>>` or `Zeroizing<String>`. Both `Vec<u8>` and `String` are heap objects like `Box<...>`.

Thank you for the explanation :+1:

PR #1534, opened 2025-08-07
keystore: pass private key to nonce_commit() and sign()

benma:
End-goal: reduce the number of secure chip ops when signing a BTC transaction, to reduce the chance of going over the Optiga chip's "rate limit", which induces throttling.

With antiklepto, we derived the private key twice for each input that is signed: once to commit to the nonce, and after that to sign.

This commit decouples the nonce commit and sign functions from the underlying keystore, and allows reusing a private key instead of re-deriving it, which requires secure chip operations.

This halves the number of secure chip ops needed per input when signing a BTC transaction.

We do not reuse the privkey for the other instances of antiklepto (signing a msg, signing an ETH tx), as there it's one commit/sign pair only and unlikely to cause secure chip throttling.

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/bitcoin/signmsg.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/bitcoin/signmsg.rs
@@ -98,7 +98,10 @@ pub async fn process(
         // Engage in the anti-klepto protocol if the host sends a host nonce commitment.
         Some(pb::AntiKleptoHostNonceCommitment { ref commitment }) => {
             let signer_commitment = keystore::secp256k1_nonce_commit(
-                keypath,
+                crate::keystore::secp256k1_get_private_key(keypath)?
+                    .as_slice()
+                    .try_into()
+                    .unwrap(),
                 &sighash,
                 commitment
                     .as_slice()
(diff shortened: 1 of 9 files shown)


src/rust/bitbox02-rust/src/hww/api/bitcoin/signmsg.rs
```diff
@@ -98,7 +98,10 @@ pub async fn process(
         // Engage in the anti-klepto protocol if the host sends a host nonce commitment.
         Some(pb::AntiKleptoHostNonceCommitment { ref commitment }) => {
             let signer_commitment = keystore::secp256k1_nonce_commit(
-                keypath,
+                crate::keystore::secp256k1_get_private_key(keypath)?
```
Beerosagos, line 101: Would it make sense to zeroize the privkey here and in the other places?

benma: The return type of `secp256k1_get_private_key` is `zeroize::Zeroizing<Vec<u8>>`, so it should zeroize automatically at the end of the statement.

Beerosagos: Aha! I missed that, sorry :)

PR #1536, opened 2025-08-13
xpubcache: allow choosing if xpub computation should be repeated

benma:
End-goal: reduce the number of secure chip ops when signing a BTC transaction, to reduce the chance of going over the Optiga chip's "rate limit", which induces throttling.

By default keystore::get_xpub computed the xpub twice, to mitigate potential bitflips, which could be bad when delivering the wrong xpub (or derivatives) to the host.

When signing a transaction however, one does not need the extra protection - if there is a bit flip, the resulting signature will be invalid.

This commit reduces the number of secure chip ops needed when the bitflip mitigation is not required.

The existing method `get_xpub` was renamed so the compiler can tell us all the instances where we need to decide between one or the other.

### DIFF
--- a/src/rust/bitbox02-rust/src/xpubcache.rs
+++ b/src/rust/bitbox02-rust/src/xpubcache.rs
@@ -37,7 +37,7 @@ pub trait Xpub: Sized {
 
 /// Implements a cache for xpubs. Cached intermediate xpubs are used to derive child xpubs.
 ///
-/// The cache must be configured using `cache_keypath()`, otherwise no caching occurs. The reason
+/// The cache must be configured using `add_keypath()`, otherwise no caching occurs. The reason
 /// for this is that automatic caching is harder to get right and reason about, e.g. in a BTC tx, we
 /// shouldn't cache xpubs at the address level (e.g. m/84/0'/0'/0/0), as they don't repeat and there
 /// can be many of them.
--- a/src/rust/bitbox02-rust/src/xpubcache.rs
+++ b/src/rust/bitbox02-rust/src/xpubcache.rs
@@ -17,12 +17,22 @@ use super::keystore;
 use crate::bip32;
 use alloc::vec::Vec;
 
+#[derive(Copy, Clone)]
+pub enum Compute {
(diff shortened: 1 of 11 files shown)


PHAMVANPHUHP (commented):
vanphuhp2503@gmail.com

src/rust/bitbox02-rust/src/xpubcache.rs
```diff
@@ -17,12 +17,18 @@ use super::keystore;
...
     /// Derives a child xpub using the provided keypath.
-    fn derive(&self, keypath: &[u32]) -> Result<Self, ()>;
+    fn derive(&self, keypath: &[u32], compute: Compute) -> Result<Self, ()>;
 
     /// Derives an xpub from the root xpub using the provided keypath.
-    fn from_keypath(keypath: &[u32]) -> Result<Self, ()>;
+    fn from_keypath(keypath: &[u32], compute: Compute) -> Result<Self, ()>;
 }
 
 /// Implements a cache for xpubs. Cached intermediate xpubs are used to derive child xpubs.
```
Beerosagos, line 38: Unrelated to the change, but I think that the `cache_keypath()` in the comment should be changed to `add_keypath()`

src/rust/bitbox02-rust/src/xpubcache.rs
```diff
@@ -17,12 +17,18 @@ use super::keystore;
 use crate::bip32;
 use alloc::vec::Vec;
 
+#[derive(Copy, Clone)]
+pub enum Compute {
```
Beerosagos, line 21: nit: Maybe it could be useful to add a comment here (or somewhere else) explaining that this is used as an additional bitflips check for sensitive operations

src/rust/bitbox02-rust/src/xpubcache.rs
benma: Added separate commit.

src/rust/bitbox02-rust/src/xpubcache.rs
benma: Added comments.

PR #1540, opened 2025-08-20
Rust bip39

benma:


src/rust/bitbox02-rust/src/bip39.rs
```diff
@@ -0,0 +1,40 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use alloc::string::{String, ToString};
+
+/// `idx>` must be smaller than BIP39_WORDLIST_LEN.
```
NickeZ: there is a small typo here I think `idx>`

benma:
> I assume you checked that the u2f app id generated the same set of words in the previous impl.

Yeah, on the yubico demo website.

PR #1542, opened 2025-08-21
prost: improve error reporting

NickeZ:


benma (commented):
This needs a new docker image to take effect, so bump the container version?

NickeZ:
> This needs a new docker image to take effect, so bump the container version?

I will add more things to the dockerfile for the new simulator. And this fix is just for debugging / development. So I think we can go ahead without "deploying" it for now.

PR #1544, opened 2025-08-21
simulator: exclude from target `all`

NickeZ:
`make unit-test` builds `all` and shouldn't build simulator.


benma (approved):
I always wondered how that could be achieved :sweat_smile:

PR #1545, opened 2025-08-21
ugui.c: latin1 > utf8

NickeZ:


NickeZ:
> I wonder what made you do this change 😂

I honestly don't remember which tool complained. I think it was `sed`. Maybe it is specific to apples build of `sed`...

It crashed with very difficult to understand output :P

PR #1549, opened 2025-08-22
Maintenance/common hardware fakes

NickeZ:


src/CMakeLists.txt
```diff
@@ -275,25 +275,10 @@ add_custom_target(rust-cbindgen
...
-  list(APPEND RUST_INCLUDES -I${include})
-endforeach()
-if(NOT (CMAKE_CROSSCOMPILING AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin"))
-  list(APPEND RUST_INCLUDES -I${CMAKE_SOURCE_DIR}/test/unit-test/framework/includes)
-endif()
-
-
 # Test rust crates that contain business logic. Avoid testing crates that depend on hardware.
 if(NOT CMAKE_CROSSCOMPILING)
+  set(RUSTFLAGS_TESTS -L${CMAKE_ARCHIVE_OUTPUT_DIRECTORY} ${CMOCKA_LDFLAGS} -lbitbox_merged -lstatic=wallycore -lstatic=secp256k1 -lstatic=fatfs -lstatic=hardware-fakes --cfg=rust_secp_no_symbol_renaming)
```
benma: Shouldn't this also add the regular `$RUSTFLAGS`? `--cfg=rust_secp_no_symbol_renaming` seems relevant - how are the tests even working without it?

src/CMakeLists.txt
```diff
@@ -396,6 +379,11 @@ endif()
 #  *  firmware-btc_rust_c
 #  *  factory-setup_rust_c
 foreach(type ${RUST_LIBS})
+  # We build the rust standard libs so that everything from the rust runtime
+  # is included (like eh_personality).
+  if(type STREQUAL "c-unit-tests")
+      set(RUST_CARGO_FLAGS ${RUST_CARGO_FLAGS} -Zbuild-std)
```
benma, line 385: We already add `-Zbuild-std=core,alloc` if CROSS_COMPILING. How about adding it once for all targets outside of loops and conditions?

test/hardware-fakes/CMakeLists.txt
benma, line 1: Curious why you called this fakes and not mocks like all the files inside.

benma:
When I rebase https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1541 onto this PR, and compile Rust unit tests, I get:

```
bitbox02-firmware/test/hardware-fakes/src/mock_securechip.c:25:10: fatal error: rust/rust.h: No such file or directory
   25 | #include <rust/rust.h>
```

NickeZ:
> When I rebase #1541 onto this PR, and compile Rust unit tests, I get:
> 
> ```
> bitbox02-firmware/test/hardware-fakes/src/mock_securechip.c:25:10: fatal error: rust/rust.h: No such file or directory
>    25 | #include <rust/rust.h>
> ```

I'm probably missing a dependency on `rust-cbindgen` somewhere. I'm currently also struggling with the same on a branch that builds on this but with `rust-clippy`. It seems `rust.h` is being deleted because it is gone even if I generate it with `make -C build... rust-cbindgen`.

src/CMakeLists.txt
NickeZ: That flag is at the end.

src/CMakeLists.txt
NickeZ: Building `std` is only necessary for the C unit tests and the C simulator. I can look around and see if some other condition also encapsulates that.

The rust simulator can use the built-in std. No need to add all that compilation time.

test/hardware-fakes/CMakeLists.txt
NickeZ: Some parts are fakes and some parts are mocks. If we move away from cmocka maybe everything becomes fakes.

For example the sd-card, smarteeprom and memory is faked right? whereas the gestures are mocked?

NickeZ:
Maybe I should make two libs, `hardware-mocks` and `hardware-fakes`. Mocks should only be necessary in the c unit tests and non-gui simulator.

src/CMakeLists.txt
benma: Why would we want the c unit tests to use the compiled one and the simulator not? Wouldn't it be better if they all were as close as possible to the firmware?

What do the Rust unit tests use?

(Anyway, we can probably port all C unit tests to Rust and eliminate that folder at some point)

test/hardware-fakes/CMakeLists.txt
benma: Ah I don't think we follow that terminology. mock_securechip.c is not using cmocka for example, but just mocking it with some fixtures and re-implementing the expected SC behavior.

Anyway, fakes is fine if you prefer it, was just curious.

benma:
> Maybe I should make two libs, `hardware-mocks` and `hardware-fakes`. Mocks should only be necessary in the c unit tests and non-gui simulator.

Then you'd have to move/rename some of them, at least mock_securechip.c, mock_diskio.c, and probably more. But they are also used in the C unit tests, so I am not sure the distinction makes much sense.

NickeZ:
I asked chatgpt:

Screenshot 2025-08-25 at 21 58 25

I'll go with "precision" and create two libs. Also the mocks depends on cmocka and I think the distinction matters in our case.

edit: I gave up that pretty quickly..

src/CMakeLists.txt
NickeZ: When we use the rust std library and generate a static library to link with, like the case of c unit tests and c simulator, we get less extra work if we just compile the std library as well.

If we don't we have to provide an "exception handler personality".

I think it is easier to maintain this flag, than to maintain building a separate C-file in case we link in rust code.

NickeZ: The rust simulator will be a rust project, so it already links/contains everything necessary for building an executable.

NickeZ: And it will not even link to this rust-c-library.

test/hardware-fakes/CMakeLists.txt
NickeZ: If it doesn't provide a mocking interface it isn't a mock. Something that fakes something else is a fake.

I gave up because half of the memory.c file is mocks and the other half is fakes and I just don't want to detangle it right now. It is also made worse by the "TESTING" define that changes the implementation of things in the real `memory.c`.

NickeZ: mocking is when you can say "this function will return 'abc'" and then when you call that function it returns 'abc'. If it always just returns 'xyz', then it is a fake.

NickeZ:
> When I rebase #1541 onto this PR, and compile Rust unit tests, I get:
> 
> ```
> bitbox02-firmware/test/hardware-fakes/src/mock_securechip.c:25:10: fatal error: rust/rust.h: No such file or directory
>    25 | #include <rust/rust.h>
> ```

I tried adding a dependency on `rust-cbindgen` on the `hardware-mocks`. See if it works better now?

benma:
> > When I rebase #1541 onto this PR, and compile Rust unit tests, I get:
> > ```
> > bitbox02-firmware/test/hardware-fakes/src/mock_securechip.c:25:10: fatal error: rust/rust.h: No such file or directory
> >    25 | #include <rust/rust.h>
> > ```
> 
> I tried adding a dependency on `rust-cbindgen` on the `hardware-mocks`. See if it works better now?

Same:

```
/bitbox02-firmware/test/hardware-mocks/src/mock_securechip.c:25:10: fatal error: rust/rust.h: No such file or directory
   25 | #include <rust/rust.h>
      |          ^~~~~~~~~~~~~
```

src/CMakeLists.txt
benma: Oh :see_no_evil: Anyway, why not include $RUSTFLAGS instead? So tests are more like the real thing and future changes are automatically applied here too.

NickeZ:
> 
> Same:
> 
> ```
> /bitbox02-firmware/test/hardware-mocks/src/mock_securechip.c:25:10: fatal error: rust/rust.h: No such file or directory
>    25 | #include <rust/rust.h>
>       |          ^~~~~~~~~~~~~
> ```

I'm pretty sure I fixed it now.

src/CMakeLists.txt

test/hardware-fakes/CMakeLists.txt
benma: TIL :+1:

benma (approved):
tACK, works now :+1: :pray:

PR #1551, opened 2025-08-22
Maintenance/less cmake

NickeZ:
Merge after https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1579


src/rust/bitbox02-rust/src/u2f.rs
```diff
@@ -146,8 +146,8 @@ mod tests {
         let app_id = b"\xc5\x0f\x8a\x7b\x70\x8e\x92\xf8\x2e\x7a\x50\xe2\xbd\xc5\x5d\x8f\xd9\x1a\x22\xfe\x6b\x29\xc0\xcd\xf7\x80\x55\x30\x84\x2a\xf5\x81";
         let mut string = [0u8; 100];
         rust_u2f_app_string(
-            unsafe { crate::util::rust_util_bytes(app_id.as_ptr(), app_id.len()) },
-            unsafe { crate::util::rust_util_bytes_mut(string.as_mut_ptr(), string.len()) },
+            unsafe { ::util::bytes::rust_util_bytes(app_id.as_ptr(), app_id.len()) },
```
benma: prefixed with `::` here but not below. Probably use the one without `::` everywhere unless necessary?

src/rust/bitbox02-sys/build.rs
```diff
@@ -171,6 +171,92 @@ const RUSTIFIED_ENUMS: &[&str] = &[
     "trinary_choice_t",
 ];
 
+// BITBOX02_SOURCES are only used for native builds (simulator), comment out all sources that depends on
+// hardware
+//
+// todo(nd): Should we get this list through an environment variable from cmake instead?
```
benma: I prefer to not keep TODO's in the code, they tend to stay forever and become hard to deal with. If important, just open an issue for it.

src/rust/bitbox02-sys/build.rs
```diff
@@ -184,7 +270,7 @@ pub fn main() -> Result<(), &'static str> {
     }
 
     let target = env::var("TARGET").expect("TARGET not set");
-    let cross_compiling = target == "thumbv7em-none-eabi";
+    let cross_compiling = target.starts_with("thumb");
```
benma: why this?

src/rust/bitbox02-sys/build.rs
```diff
@@ -248,11 +344,23 @@ pub fn main() -> Result<(), &'static str> {
         ]);
     } else {
         // unit test framework includes
-        includes.push("../../../test/hardware-fakes/include")
+        includes.push("../../../test/hardware-fakes/include");
+        //includes.push("../../../test/unit-test/framework/include")
```
benma: remove

src/rust/bitbox02/src/util.rs
```diff
@@ -41,10 +43,20 @@ pub fn truncate_str(s: &str, len: usize) -> &str {
...
-        .into_bytes_with_nul())
+        .into_bytes_with_nul();
+    // Ideally we'd use Vec::into_raw_parts, but it's unstable.
+    // Before we disassemble `cstr` into its raw parts, make sure it
+    // does not get dropped!
+    let mut v = core::mem::ManuallyDrop::new(cstr);
+    let ptr = v.as_mut_ptr();
+    let len = v.len();
+    let cap = v.capacity();
+    Ok(unsafe { Vec::from_raw_parts(ptr.cast(), len, cap) })
```
benma: A bit scary and hard to review. 

First, if the stdlib is using ` Vec<u8>` for bytes, shouldn't we also? Can just cast the pointer in the end when passing through FFI if something expects `*const core::ffi::c_char`.

But if not, why not use something without `unsafe`, like `.into_iter().map(|b| b as core::ffi::c_char)).collect())`? Probably it is optimal with the stdlib funcs and compiler optimizations anyway.

src/rust/bitbox02/Cargo.toml
```diff
@@ -32,14 +32,20 @@ hex = { workspace = true }
...
 # Active when the Rust code is compiled to be linked into the C unit tests and simulator.
 c-unit-testing = []
 
 app-ethereum = []
 app-bitcoin = []
 app-litecoin = []
 app-u2f = []
+
+# Expose raw C bindings as bitbox02::bindings module
+bindings = []
```
benma: I can't see where this is used right now - is it used?

src/CMakeLists.txt
```diff
@@ -313,7 +313,7 @@ if(NOT CMAKE_CROSSCOMPILING)
     DEPENDS
       ${CMAKE_CURRENT_BINARY_DIR}/rust/rust.h
     )
-  add_dependencies(rust-test generate-protobufs bitbox_merged fatfs)
+    add_dependencies(rust-test secp256k1 generate-protobufs fatfs)
```
benma: indentation

src/CMakeLists.txt
```diff
@@ -436,10 +439,22 @@ foreach(type ${RUST_LIBS})
...
   add_library(${type}_rust_c STATIC IMPORTED GLOBAL)
-  # Any target that depends on the library also depends on the header file
-  # which is generated by rust-cbindgen. So we might as well depend on it here.
-  add_dependencies(${type}_rust_c ${type}-rust-target generate-protobufs rust-cbindgen)
+  add_dependencies(${type}_rust_c ${type}-rust-target generate-protobufs)
   set_property(TARGET ${type}_rust_c PROPERTY IMPORTED_LOCATION ${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}/lib${type}_rust_c.a)
+  # Add the include directory to find rust/rust.h
+  target_include_directories(${type}_rust_c INTERFACE ${CMAKE_CURRENT_BINARY_DIR})
+  if (NOT CMAKE_CROSSCOMPILING)
+    # Needs to match the definitions in `bitbox02-sys/build.rs`
```
benma, line 447: Also put a comment there so one can orient themselves from both sides.

src/rust/bitbox02-sys/build.rs
```diff
@@ -248,11 +344,23 @@ pub fn main() -> Result<(), &'static str> {
...
     }
 
     let out_path = PathBuf::from(env::var("OUT_DIR").unwrap()).join("bindings.rs");
     let out_path = out_path.into_os_string().into_string().unwrap();
+
+    let mut definitions = vec![
```
benma, line 352: Mention this needs to match CMakeLists

src/rust/bitbox02-sys/Cargo.toml
```diff
@@ -23,3 +23,6 @@ license = "Apache-2.0"
 
 [dependencies]
 util = {path = "../util"}
+
+[build-dependencies]
+cc = {version="1.2.31", features=["parallel"]}
```
benma: We already have cc=1.0.79 in the vendor/ folder, this adds another copy. Maybe we can avoid it and only use one version?

src/rust/bitbox02-sys/build.rs
NickeZ: future proofing, but not really necessary.

src/rust/bitbox02-sys/Cargo.toml
NickeZ: we also had `1.2.0` due to `core`. I downgraded to that version.

src/rust/bitbox02-rust/src/u2f.rs
NickeZ: fixed

src/rust/bitbox02-sys/build.rs
NickeZ: fixed

src/rust/bitbox02-sys/build.rs
NickeZ: fixed

src/rust/bitbox02-sys/build.rs
NickeZ: fixed

src/rust/bitbox02/src/util.rs
NickeZ: fixed

src/rust/bitbox02/Cargo.toml
NickeZ: fixed

src/CMakeLists.txt

src/CMakeLists.txt
NickeZ: fixed

src/rust/bitbox02-sys/build.rs
NickeZ: fixed

src/rust/bitbox02-noise/src/noise_xx.rs
```diff
@@ -20,7 +20,8 @@ use noise_rust_crypto::{ChaCha20Poly1305, Sha256, sensitive::Sensitive};
 
 /// Specialization of noise_protocol::HandshakeState, picking the implementations for Diffie
 /// Hellman, Cipher and Hash.
-pub type HandshakeState<R> = noise_protocol::HandshakeState<X25519<R>, ChaCha20Poly1305, Sha256>;
+/// cbindgen:ignore
+pub type HandshakeState<R> = ::noise_protocol::HandshakeState<X25519<R>, ChaCha20Poly1305, Sha256>;
```
benma: I guess also here the `::` are not needed? No big deal though.

NickeZ: fixed

PR #1553, opened 2025-08-26
remove the option of restoring from 18 recovery words

benma:


src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -329,12 +329,26 @@ pub fn trinary_choice_create<'a>(
...
     let component = unsafe {
         bitbox02_sys::trinary_choice_create(
             crate::util::str_to_cstr_vec(message).unwrap().as_ptr(), // copied in C
-            crate::util::str_to_cstr_vec(label_left).unwrap().as_ptr(), // copied in C
-            crate::util::str_to_cstr_vec(label_middle).unwrap().as_ptr(), // copied in C
-            crate::util::str_to_cstr_vec(label_right).unwrap().as_ptr(), // copied in C
+            // copied in C
+            label_left
+                .as_ref()
+                .map_or_else(|| core::ptr::null(), |label| label.as_ptr()),
```
NickeZ: I don't think you need a closure here, you can just provide the function `core::ptr::null`.

benma: Indeed :+1: fixed

NickeZ (approved):
utACK

PR #1555, opened 2025-08-26
ui: change title to include number of words

benma:
`1 of 24`, `2 of 24`, ... instead of `1st word`, `2nd word`, ...

This is to be more explicit about the number of words being restored.


sutterseba (approved):
tACK

PR #1556, opened 2025-08-26
rust: since rust 1.64, bindgen can use the c types from core

NickeZ:
There was a bug also, (u)int is 32bits on 64bit systems.

rust uses `u8` internally for strings but x86_64 uses `int8_t` for `c_char`. This means that we need to add a bunch of `.cast()` to handle both cases. (arm uses `uint8_t` for `c_char`.)


benma (approved):
TIL about `cast()`, I've been using `as _` or `as *const _` in before. If you want consistency, you could find them and change them into `cast()` too.

utACK

NickeZ:
> TIL about `cast()`, I've been using `as _` or `as *const _` in before. If you want consistency, you could find them and change them into `cast()` too.
> 
> utACK

I also learned about it today 🎉

PR #1557, opened 2025-08-26
rust: Make unit tests buildable on macos

NickeZ:
Mach-O has a different naming convention of linker segments and isn't compatible with `.segger_rtt_buf` that ELF uses. Let's only compile in RTT when we are building for embedded targets


NickeZ:
> utACK, but please be sure this does not break factorysetup 😁

I'll test some more before merging 😅

NickeZ:
The changes in this PR was already included in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1549

PR #1561, opened 2025-08-27
Move vendored rust deps

NickeZ:
`src` should only contain our sources


NickeZ:
> Reluctant ACK - a little weird that this Rust stuff is outside the rust/ folder, which contains all the other Rust stuff we have.

It is a bit strange to have a "rust" folder with all the rust stuff in it though. Do you think the rust simuator should be in the rust folder as well, and not a sibling folder to `simulator` in `test`?

Anyway, the other reason I want to move out the vendored deps is that I cannot easily grep / fuzzy search in our sources. I always have to exclude the src/rust/vendor dir.

PR #1563, opened 2025-09-01
Maintenance/simulator no cmocka dep

NickeZ:


NickeZ:
> ```
> $ ldd simulator 
>         linux-vdso.so.1 (0x00007fffdc9ca000)
>         libfatfs.so => bitbox02-firmware/build-build-noasan/external/libfatfs.so (0x000078693598e000)
>         libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000078692efd7000)
>         /lib64/ld-linux-x86-64.so.2 (0x000078693599a000)
> ```
> 
> libfats.so wasn't there before, it should be built statically

Fixed

benma (approved):
utACK

PR #1565, opened 2025-09-01
rust: bump to 2024

NickeZ:
We already had 2024 in one of the newer crates. This made rustfmt confused, as rustfmt.toml had 2021.


src/rust/bitbox02-rust/src/hww/api/bitcoin.rs
```diff
@@ -280,23 +280,23 @@ pub async fn process_pub(
...
         Some(Output::ScriptConfig(BtcScriptConfig {
-            config: Some(Config::Multisig(ref multisig)),
+            config: Some(Config::Multisig(multisig)),
         })) => address_multisig(hal, coin, multisig, &request.keypath, request.display).await,
         Some(Output::ScriptConfig(BtcScriptConfig {
-            config: Some(Config::Policy(ref policy)),
+            config: Some(Config::Policy(policy)),
         })) => address_policy(hal, coin, policy, &request.keypath, request.display).await,
         _ => Err(Error::InvalidInput),
     }
```
benma: Why this change? tests and clippy pass without it too.

benma (approved):
:+1:

src/rust/bitbox02-rust/src/hww/api/bitcoin.rs
NickeZ: I just removed all of the `ref` bindings in pattern matching. Maybe this case wasn't necessary.

NickeZ: @benma reverted this change and another similar change in cardanos signing module. Please rereview

PR #1567, opened 2025-09-02
rust: update bip32-ed25519

benma:


benma:
> utACK, but wouldn't it be easier to publish it to crates.io?

A toss-up, publishing needs more steps and is tied to personal accounts afaik.

PR #1571, opened 2025-09-03
keystore: reduce secure chip operations by precomputing fingerprint

benma:
The root fingerprint API call, calling
`bitbox02_rust::keystore::root_fingerprint()`, used two securechip
operations. Using too many operations too quickly in Optiga leads to
throttling, and the BitBoxApp fetches the root fingerprint every time
the BitBox is unlocked.

We can get away with not using hte securechip at all to get the root
fingerprint, by computing and storing it during unlock.

The global static mut could have lived in keystore.c with the other
static muts there, but adding more C code and Rust wrappers seemed
wrong. For now it lives in bitbox02::keystore, and would move over to
bitbox02_rust::keystore when the unlocking functions are migrated to Rust.

### DIFF
--- a/src/keystore.c
+++ b/src/keystore.c
@@ -448,7 +448,7 @@ keystore_error_t keystore_unlock(
     return result;
 }
 
-bool keystore_unlock_bip39(const char* mnemonic_passphrase)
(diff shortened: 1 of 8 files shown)


benma:
Rebased

benma:
Rebased

PR #1573, opened 2025-09-04
securechip: count optiga security events and integrate into tests

benma:
We count the number of security events the securechip functions take, and then assert in the keystore unit tests how many events the functions take there.

This is a basis for trying to reduce the count, so we can measure the changes properly. Too many events induce throttling.


NickeZ (approved):
😍 !

PR #1575, opened 2025-09-04
rust: make workdirs more coherent amongst various cargo invocations

NickeZ:
* Always execute cargo from the same place so that relative paths to CWD always are the same.
* Always use the "same depth" target dir, so that relative paths to the c-sources and files built by cmake stays the same.
* Give clippy its own target dir so that we don't have to wipe it before every run


benma (approved):
utACK

PR #1578, opened 2025-09-04
Reduce SC events during restore

benma:
19 --> 8 optiga security events during a restore.

### DIFF
--- a/src/keystore.c
+++ b/src/keystore.c
@@ -208,22 +208,35 @@ static keystore_error_t _get_and_decrypt_seed(
 }
 
 static bool _verify_seed(
-    const char* password,
+    const uint8_t* encryption_key,
     const uint8_t* expected_seed,
     size_t expected_seed_len)
 {
-    uint8_t decrypted_seed[KEYSTORE_MAX_SEED_LENGTH] = {0};
-    size_t seed_len;
-    UTIL_CLEANUP_32(decrypted_seed);
-    if (_get_and_decrypt_seed(password, decrypted_seed, &seed_len, NULL) != KEYSTORE_OK) {
+    uint8_t encrypted_seed_and_hmac[96];
+    UTIL_CLEANUP_32(encrypted_seed_and_hmac);
+    uint8_t encrypted_len;
+    if (!memory_get_encrypted_seed_and_hmac(encrypted_seed_and_hmac, &encrypted_len)) {
         return false;
     }
-    if (expected_seed_len != seed_len) {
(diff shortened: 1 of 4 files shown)


src/keystore.c
```diff
@@ -385,6 +305,93 @@ static void _delete_retained_seeds(void)
     _retained_bip39_seed_encrypted_len = 0;
 }
 
+keystore_error_t keystore_encrypt_and_store_seed(
```
NickeZ, line 338: moving these two functions makes it hard to review. Are there any differences, are they the same?

NickeZ (changes requested):
Can you make the second commit easier to review?

src/keystore.c
benma: Good point, my bad. Inserted a commit that moves them first, the diff is much smaller now.

src/keystore.c
```diff
@@ -208,22 +208,35 @@ static keystore_error_t _get_and_decrypt_seed(
...
+    uint8_t encrypted_len;
+    if (!memory_get_encrypted_seed_and_hmac(encrypted_seed_and_hmac, &encrypted_len)) {
         return false;
     }
-    if (expected_seed_len != seed_len) {
+    if (encrypted_len < 49) {
+        Abort("_verify_seed: underflow / zero size");
+    }
+    size_t decrypted_len = encrypted_len - 48;
+    uint8_t decrypted[decrypted_len];
```
NickeZ, line 225: it is better to have a fixed upper bound instead of dynamically allocating stack size.

benma: It's a copypaste of https://github.com/BitBoxSwiss/bitbox02-firmware/blob/dea7eff17b278e399822a2b2a2aa3ef45f48e11c/src/keystore.c#L191. I prefer to not change it in this PR, or not at all and port it to Rust (https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1425)

NickeZ: Can we at least have some upper bound so that nobody manages to smash the stack with a very large number?

benma: encrypted_len is uint8_t, so it is technically already upper bounded to a small value. Is that sufficient?

NickeZ: `decrypted_len` is a `size_t` AFAICT

NickeZ: ah, ok, I see that `encrypted_len` is indeed `uint8_t`.

PR #1582, opened 2025-09-08
simulator-ng init

NickeZ:
depends on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1585

@benma when you review this, can you please test that it builds/runs on your linux machine? (I only tested macos)


benma:
Regarding the sdl2 submodule: you don't need to make it a submodule, you can just add it as a git dependency in Cargo.toml, for example like we do here:

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/b99212dd3b1d6c4e91f6967689871c65fdba3e81/src/rust/bitbox02-rust/Cargo.toml#L45

Or was there a specific reason for not doing it that way?

(The reason I noticed is that the build now spits out warnings of the sdl crate, which it would not do I believe if it treated as an external/git dep)

benma:
> @benma when you review this, can you please test that it builds/runs on your linux machine? (I only tested macos)

It works! Though it crashes when trying to sign a BTC tx in send_message:

```
Exception: Did not read anything after 5000 seconds.
```

I did not wait 5000 seconds :joy:

NickeZ:
> > @benma when you review this, can you please test that it builds/runs on your linux machine? (I only tested macos)
> 
> It works! Though it crashes when trying to sign a BTC tx in send_message:
> 
> ```
> Exception: Did not read anything after 5000 seconds.
> ```
> 
> I did not wait 5000 seconds 😂

interesting, should probably run it in gdb and see where the segfault is :P

NickeZ:
> Regarding the sdl2 submodule: you don't need to make it a submodule, you can just add it as a git dependency in Cargo.toml, for example like we do here:
> 
> https://github.com/BitBoxSwiss/bitbox02-firmware/blob/b99212dd3b1d6c4e91f6967689871c65fdba3e81/src/rust/bitbox02-rust/Cargo.toml#L45
> 
> Or was there a specific reason for not doing it that way?
> 
> (The reason I noticed is that the build now spits out warnings of the sdl crate, which it would not do I believe if it treated as an external/git dep)

I'll look into it. I just wanted to avoid vendoring the crate because it includes the SDL sources (thousands of c-files...)

NickeZ:
Closing in favor of https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1628 which doesn't have C dependencies like SDL2

PR #1585, opened 2025-09-10
build: fix sanitizers

NickeZ:
The recommended way to enable sanitizers nowadays is to add `-fsanitize=*` as a compiler and linking flags.


NickeZ:
> utACK, but do we even use `make all` anywhere, or in what case does `EXCLUDE_FROM_ALL` apply?

yeah, when building the unit tests.

PR #1588, opened 2025-09-15
test: add option for the simulator to write to file.

bznein:
Disclaimer: This is a mixture of educated guesses, AI, and the little C I remember writing 15 years ago... so please keep that in mind while reviewing :D Anyway I tested it manually and it seems to work fine. 

If the env variable FAKE_MEMORY_FILEPATH is provided, the simulator will
write to file using three different files:

- ${FAKE_MEMORY_FILEPATH}_shared
- ${FAKE_MEMORY_FILEPATH}_app
- ${FAKE_MEMORY_FILEPATH}_smarteeprom

mimicking the behaviour used when reading/writing to memory.

### DIFF
--- a/test/hardware-fakes/src/fake_memory.c
+++ b/test/hardware-fakes/src/fake_memory.c
@@ -18,17 +18,60 @@
 #include <memory/memory_shared.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/stat.h>
+#include <stdlib.h>
+
+
+#define FAKE_MEMORY_ENV_VAR "FAKE_MEMORY_FILEPATH"
+#define SHARED_SUFFIX "shared"
+#define APP_SUFFIX "app"
+#define EEPROM_SUFFIX "eeprom"
 
 static uint8_t _memory_shared_data[FLASH_SHARED_DATA_LEN] = {0};
 static uint8_t _memory_app_data[FLASH_APPDATA_LEN] = {0};
 static uint8_t _memory_smarteeprom[SMARTEEPROM_RESERVED_FLASH_PAGES * FLASH_PAGE_SIZE] = {0};
+
+static void _init_file_if_needed(
+    const char* base_path,
+    const char* suffix,
+    const uint8_t* data,
+    size_t data_size)
+{
+    char path[512];
+    if (snprintf(path, sizeof(path), "%s_%s", base_path, suffix) >= sizeof(path)) {
+        fprintf(stderr, "file path %s_%s too long\n", base_path, suffix);
+        exit(EXIT_FAILURE);
+    }
+
+    struct stat st;
+    if (stat(path, &st) == 0) {
+        return; // The file already exists, do nothing.
+    }
+
+    FILE* f = fopen(path, "wb");
+    if (f) {
+        fwrite(data, 1, data_size, f);
+        fclose(f);
+    } else {
+        fprintf(stderr, "error: could not create fake memory file %s\n", path);
+        exit(EXIT_FAILURE);
+    }
+}
 
 void fake_memory_factoryreset(void)
 {
     memset(_memory_shared_data, 0xff, sizeof(_memory_shared_data));
     memset(_memory_app_data, 0xff, sizeof(_memory_app_data));
     memset(_memory_smarteeprom, 0xff, sizeof(_memory_smarteeprom));
+
+    const char* base_path = getenv(FAKE_MEMORY_ENV_VAR);
+    if (base_path) {
+        _init_file_if_needed(base_path, SHARED_SUFFIX, _memory_shared_data, sizeof(_memory_shared_data));
+        _init_file_if_needed(base_path, APP_SUFFIX, _memory_app_data, sizeof(_memory_app_data));
+        _init_file_if_needed(base_path, EEPROM_SUFFIX, _memory_smarteeprom, sizeof(_memory_smarteeprom));
+    }
 }
+
 #define ALLOWED_HASH                                                                               \
     "\x1e\x4a\xa8\x36\x4e\x93\x5c\x07\x85\xe4\xf8\x91\x20\x83\x07\xd8\x32\xf7\x88\x17\x2e\x4b\xf6" \
     "\x16\x21\xde\x6d\xf9\xec\x3c\x21\x5f"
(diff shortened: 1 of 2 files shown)


benma:
First of all, what is the reason for making the change? Please write it in the commit msg body every time :grin:

test/hardware-fakes/src/fake_memory.c
```diff
@@ -23,12 +23,41 @@ static uint8_t _memory_shared_data[FLASH_SHARED_DATA_LEN] = {0};
...
+static void _init_file_if_needed(
+    const char* base_path,
+    const char* suffix,
+    const uint8_t* data,
+    size_t data_size)
+{
+    char path[512];
+    snprintf(path, sizeof(path), "%s_%s", base_path, suffix);
+
+    FILE* f = fopen(path, "rb");
```
NickeZ: You can check if a file exists with `stat` instead. https://linux.die.net/man/2/stat

bznein:
> First of all, what is the reason for making the change? Please write it in the commit msg body every time 😁

Done. I was not sure on how to phrase it, left it to the end, and forgot about it 🤦‍♂️

(@benma feel free to review it, I assigned to Niklas only as he has context already as we discussed it together last week)

test/hardware-fakes/src/fake_memory.c
```diff
@@ -63,8 +92,73 @@ static uint8_t* _get_memory(uint32_t base)
...
+        const char* suffix = NULL;
+        switch (base) {
+        case FLASH_SHARED_DATA_START:
+            suffix = "shared";
+            break;
+        case FLASH_APPDATA_START:
+            suffix = "app";
+            break;
+        case FLASH_SMARTEEPROM_START:
+            suffix = "smartee";
```
NickeZ: `smartee` sounds a bit off, maybe just `eeprom`? The reason it is called "smart" is because it is an emulated eeprom in the MCU (using flash and a clever algorithm to reduce wear).

test/hardware-fakes/src/fake_memory.c
```diff
@@ -63,8 +92,73 @@ static uint8_t* _get_memory(uint32_t base)
...
+static void _write_file_chunk(const char* suffix, uint32_t offset, const uint8_t* chunk, size_t len)
+{
+    char path[512];
+    const char* base_path = getenv("FAKE_MEMORY_FILEPATH");
+    if (!base_path) {
+        return;
+    }
+    snprintf(path, sizeof(path), "%s_%s", base_path, suffix);
+    FILE* f = fopen(path, "r+b");
+    if (!f) {
```
NickeZ, line 121: Does this indicate a bug? The file should exist if `fake_memory_factoryreset` was called right?

Perhaps we should just abort here and say "Expected file to be present"?

If the file is truncated and opened (`wb`) it doesn't contain `0xff` as if it was real flash storage.

test/hardware-fakes/src/fake_memory.c
```diff
@@ -63,8 +92,73 @@ static uint8_t* _get_memory(uint32_t base)
...
+static void _read_file_chunk(const char* suffix, uint32_t offset, uint8_t* chunk, size_t len)
+{
+    char path[512];
+    const char* base_path = getenv("FAKE_MEMORY_FILEPATH");
+    if (!base_path) {
+        return;
+    }
+    snprintf(path, sizeof(path), "%s_%s", base_path, suffix);
+    FILE* f = fopen(path, "rb");
+    if (!f) {
```
NickeZ, line 142: maybe aborting/exit is better. This is in the end a developer tool and maybe it is better to be explicit that the file was expected?

test/hardware-fakes/src/fake_memory.c
```diff
@@ -23,12 +23,41 @@ static uint8_t _memory_shared_data[FLASH_SHARED_DATA_LEN] = {0};
...
+    }
+}
+
 void fake_memory_factoryreset(void)
 {
     memset(_memory_shared_data, 0xff, sizeof(_memory_shared_data));
     memset(_memory_app_data, 0xff, sizeof(_memory_app_data));
     memset(_memory_smarteeprom, 0xff, sizeof(_memory_smarteeprom));
+
+    const char* base_path = getenv("FAKE_MEMORY_FILEPATH");
```
NickeZ: the string "FAKE_MEMORY_FILEPATH", should be a define to avoid typos.

NickeZ: so should `"shared"`, `"app"` and `"eeprom"` also be.

test/hardware-fakes/src/fake_memory.c
```diff
@@ -23,12 +23,41 @@ static uint8_t _memory_shared_data[FLASH_SHARED_DATA_LEN] = {0};
...
 static uint8_t _memory_smarteeprom[SMARTEEPROM_RESERVED_FLASH_PAGES * FLASH_PAGE_SIZE] = {0};
 
+static void _init_file_if_needed(
+    const char* base_path,
+    const char* suffix,
+    const uint8_t* data,
+    size_t data_size)
+{
+    char path[512];
+    snprintf(path, sizeof(path), "%s_%s", base_path, suffix);
```
NickeZ: I think you should check the return value of `snprintf`. If it is 512 you should exit/abort and report an error to the user. Better than writing in the wrong place on the disk.

test/hardware-fakes/src/fake_memory.c
NickeZ: also this suffix isn't equal to the other suffix for the SMARTEEPROM. (but you will solve that when you use `#define`s instead.

test/hardware-fakes/src/fake_memory.c
NickeZ: I see you added a call to `stat` but didn't remove this `fopen`. Maybe I misunderstood the purpose of this `fopen`? Why cannot you open with `wb` immediately?

test/hardware-fakes/src/fake_memory.c
```diff
@@ -63,8 +108,78 @@ static uint8_t* _get_memory(uint32_t base)
...
 }
 
+static void _write_file_chunk(const char* suffix, uint32_t offset, const uint8_t* chunk, size_t len)
+{
+    char path[512];
+    const char* base_path = getenv(FAKE_MEMORY_ENV_VAR);
+    if (!base_path) {
+        return;
+    }
+    if (snprintf(path, sizeof(path), "%s_%s", base_path, suffix) >= 512) {
```
NickeZ: 512 should be `sizeof(path)`

test/hardware-fakes/src/fake_memory.c
```diff
@@ -63,8 +108,78 @@ static uint8_t* _get_memory(uint32_t base)
...
+        fprintf(stderr, "error writing fake memory file path");
```
NickeZ: the error should be "file path too long"

NickeZ:
It is probably a good idea to add a changelog entry as well

test/hardware-fakes/src/fake_memory.c
bznein: No you are right, I missed it, sorry!

PR #1593, opened 2025-09-16
Maint/cargo cc werror

NickeZ:


src/rust/bitbox02-sys/build.rs
```diff
@@ -285,6 +285,15 @@ pub fn main() -> Result<(), &'static str> {
         vec!["-DTESTING", "-D_UNIT_TEST_", "-DPRODUCT_BITBOX_MULTI=1"]
     };
 
+    // If user enables -Dwarnings for rust we also want to enable -Werror for C.
+    if let Ok(rustflags) = std::env::var("CARGO_ENCODED_RUSTFLAGS") {
+        for flag in rustflags.split('\x1f') {
```
benma, line 289: :exploding_head:

test/hardware-fakes/src/fake_memory.c
```diff
@@ -88,19 +88,14 @@ void memory_read_shared_bootdata_fake(uint8_t* chunk_out)
     memcpy(chunk_out, _memory_shared_data, (size_t)FLASH_SHARED_DATA_LEN);
 }
 
-static uint8_t _salt_root[32] = {
+uint8_t _salt_root[32] = {
```
benma: remove underscore, we normally use it to recognize statics.

NickeZ:
> Why not just always enable them instead of via env? Too annoying during development?

Yes, I definitely don't want to fix every single lint before I can test out code.

src/rust/bitbox02-sys/build.rs
NickeZ: They like doing things the right way over at rust land.

https://codepoints.net/U+001F?lang=en

PR #1595, opened 2025-09-18
ui: introduce double buffering

NickeZ:
Create a new module called "canvas" which is responsible for double buffering. Double buffering is required to enable asynchronous transfer of the frame buffer. While the "active" frame buffer is being transferred to the oled in the background, the ui will render to a "working" frame buffer. When the rendering is complete the buffers are flipped with "canvas_commit".

I tested with both ssd1312 and sh1107 screens.


src/ui/canvas.c
```diff
@@ -0,0 +1,47 @@
...
+    memset(canvas_working(), color, sizeof(_canvas[0]));
+}
+void canvas_clear(void)
+{
+    canvas_fill(0);
+}
+
+void canvas_commit(void)
+{
+    _canvas_active = (_canvas_active + 1) % 2;
```
benma: DeepSeek

--- 

**Potential Race Conditions**:
- The volatile `_canvas_active` switch isn't atomic. On 32-bit MCUs, this could cause torn buffer states if interrupted during update. Consider using atomic operations or disabling interrupts briefly during `canvas_commit()`.

src/ui/canvas.h
```diff
@@ -0,0 +1,58 @@
...
+
+// 8 pixels per byte in the canvas
+#define CANVAS_SIZE ((SCREEN_WIDTH * SCREEN_HEIGHT) / 8)
+
+#include <stdint.h>
+
+/*
+ * Initialize canvases
+ */
+void canvas_init(void);
```
benma, line 28: DeepSeek:

---

**Buffer Initialization**:
- `canvas_init()` is declared in canvas.h but not implemented.

src/ui/canvas.c
```diff
@@ -0,0 +1,47 @@
+// Copyright 2025 Shift Cryptosecurity AG
```
benma: wrong name :nerd_face:

src/ui/canvas.c
```diff
@@ -0,0 +1,47 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string.h>
+#include <ui/canvas.h>
+
+// This is sometimes updated from interrupts :( to be undone eventually
```
benma: If you know what "sometimes" means exactly, please specify for clarity. BIP39 unlock? More?


"to be undone" you mean the `volatile` qualifier right?

src/ui/canvas.h
```diff
@@ -0,0 +1,58 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#ifndef CANVAS_H
+#define CANVAS_H
+
+#include <screen.h>
+
+// 8 pixels per byte in the canvas
+#define CANVAS_SIZE ((SCREEN_WIDTH * SCREEN_HEIGHT) / 8)
```
benma, line 20: Maybe move this to canvas.c, as it's not used anywhere else.

src/ui/canvas.h
```diff
@@ -0,0 +1,58 @@
+// Copyright 2025 Shift Cryptosecurity AG
```
benma: same

test/hardware-fakes/src/fake_oled.c
```diff
@@ -0,0 +1,15 @@
+// Copyright 2025 Shift Cryptosecurity AG
```
benma: same

benma (commented):
Slightly uncomfortable with the removal of the screen clearing before rendering/flushing, as it's a bit hard to reason about if the working buffer is really clean (from a previous commit) in all these cases.

src/ui/canvas.c
NickeZ: deepseek can fight gemini
Screenshot 2025-09-22 at 11 44 48

NickeZ:
> Slightly uncomfortable with the removal of the screen clearing before rendering/flushing, as it's a bit hard to reason about if the working buffer is really clean (from a previous commit) in all these cases.

I guess I could return a dirty buffer and keep all the clears instead. I have no strong opinion, just thought it would be nicer like this.

(if you didn't commit/switch the buffers the current buffer will never be drawn, so I guess it would be hard to miss?)

src/ui/canvas.h
NickeZ: it will be used by the ssd1312 oled driver

src/ui/canvas.c
NickeZ: yup, volatile can be removed when we simplify all interrupts to just set flags.

src/ui/canvas.c
benma: Well, you are not just writing the value, you are reading it too.

PR #1598, opened 2025-09-20
rust: update to toolchain 1.90.0

benma:
This saves ~2-3kB in binary space in the multi firmware.

`clippy::collapsible-if` is disabled because I didn't like some of the collapse suggestions. Readability seems worse rather than better.


benma:
@NickeZ the vendor folder now has two copies of libc, but src/rust/Cargo.lock only contains one libc dep entry. Not sure what is going on...

benma:
Continuing in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1599, CI can't submit the docker image if the branch is not on upstream.

PR #1599, opened 2025-09-20
rust: update to toolchain 1.90.0

benma:
This saves ~2-3kB in binary space in the multi firmware.

`clippy::collapsible-if` is disabled because I didn't like some of the collapse suggestions. Readability seems worse rather than better.


benma:
@NickeZ the vendor folder now has two copies of libc, but src/rust/Cargo.lock only contains one libc dep entry. Not sure what is going on...

NickeZ:
> @NickeZ the vendor folder now has two copies of libc, but src/rust/Cargo.lock only contains one libc dep entry. Not sure what is going on...

rust std libs also depend on libc. you should enforce the same version in our lockfile with `cargo update --precise`

src/rust/bitbox02-rust/src/bb02_async.rs
```diff
@@ -67,4 +67,4 @@ pub fn block_on<O>(task: impl core::future::Future<Output = O>) -> O {
             return result;
         }
     }
-}
+}
```
NickeZ: please configure emacs to end files with a newline.

benma: Lol what an insult :joy: Of course it is. But I can't explain why this file didn't end in a newline though :thinking: Anyway, fixed

benma:
> > @NickeZ the vendor folder now has two copies of libc, but src/rust/Cargo.lock only contains one libc dep entry. Not sure what is going on...
> 
> rust std libs also depend on libc. you should enforce the same version in our lockfile with `cargo update --precise`

Done.

PR #1600, opened 2025-09-22
Async bip39

benma:


external/vendor/bip39/src/pbkdf2.rs
```diff
@@ -126,3 +126,41 @@ pub(crate) fn pbkdf2<M>(mnemonic: M, unprefixed_salt: &[u8], c: usize, res: &mut
...
+			salt
+		};
+
+		for i in 1..c {
+			let mut prfc = prf.clone();
+			prfc.input(&salt);
+			salt = hmac::Hmac::from_engine(prfc).to_byte_array();
+
+			xor(chunk, &salt);
+			yield_now(i).await;
```
NickeZ: I'm just a little bit skeptical about passing in `i`. won't `i` be a sequence like `1,2,3,2,3,3` for example if `c = 4`? Isn't that a very awkward interface? Wouldn't it be better with something linear so that you can make a progress bar or whatever you want?

src/rust/util/src/bb02_async.rs
```diff
@@ -74,6 +74,38 @@ pub fn yield_now() -> impl core::future::Future<Output = ()> {
     })
 }
 
+/// Executes both futures in tandem, returning the results of both when both are done.
```
NickeZ: This is fine, but I want to point out that there are libraries already implementing it:

* https://docs.rs/futures-lite/latest/futures_lite/future/fn.zip.html
* https://docs.rs/futures/latest/futures/macro.join.html
* [nightly only] https://doc.rust-lang.org/std/future/macro.join.html
* https://docs.rs/futures-concurrency/latest/futures_concurrency/future/trait.Join.html#tymethod.join

src/ui/components/unlock_animation.c
```diff
@@ -25,7 +26,11 @@
 // This many iterations times the slowdown factor to render the whole animation.
 #define LOCK_ANIMATION_N_FRAMES (38)
 
-#define SLOWDOWN_FACTOR (10)
+// Since BIP39 unlock takes 2048 iterations, and the screen frame rate is 30 (SCREEN_FRAME_RATE,
+// render is called only every 30th iteration), if we want both to finish at the same time, the
+// slowdown factor becomes the following. 10% is added so the animation takes a bt longer than the
+// actual unlock.
+#define SLOWDOWN_FACTOR (1.1f * (2048 / ((float)LOCK_ANIMATION_N_FRAMES * (float)SCREEN_FRAME_RATE)))
```
NickeZ: This is very close to 2. Could we set it to that instead, to avoid floats?

src/ui/components/unlock_animation.c
```diff
@@ -0,0 +1,208 @@
...
+    dimension_t dim = {.width = LOCK_ANIMATION_FRAME_WIDTH, .height = LOCK_ANIMATION_FRAME_HEIGHT};
+    in_buffer_t image = {.data = _get_frame(frame), .len = LOCK_ANIMATION_FRAME_SIZE};
+    graphics_draw_image(&pos, &dim, &image);
+    data->frame++;
+}
+
+static const component_functions_t _component_functions = {
+    .cleanup = ui_util_component_cleanup,
+    .render = _render,
+    .on_event = ui_util_on_event_noop,
```
NickeZ: `on_event` is allowed to be NULL, so no need to call a noop implementation.

src/rust/util/src/bb02_async.rs
benma: Yeah I was debating if I should pull in a dep or have a local implementation. Do you have a preference?

NickeZ: I would lean towards `futures_light` as long as we don't already have like `futures` in our dependency tree.

external/vendor/bip39/src/pbkdf2.rs
benma: Good question. `c` is 2048, the bip39 stretch count, and there is only one chunk in bip39 seeds, so i will go from `0..2048`. 

Maybe I'll simply omit that parameter altogether, seeing we don't actually make use of it. I guess hosts can emulate it by incrementing a counter inside the callback if they need to keep track for some reason.

src/ui/components/unlock_animation.c
benma: What's the issue with floats exactly?

I can do it and move the formula to the comments, but if one wanted to tune it differently, it will not work without going back to floats.

benma: Done

src/ui/components/unlock_animation.c
benma: Done

external/vendor/bip39/src/pbkdf2.rs
benma: Done

src/rust/util/src/bb02_async.rs
benma: Removed custom yield_now/join and added futures_light dep. Binary size is even 384 bytes smaller now :shrug:

PR #1602, opened 2025-09-23
simulator: fix task poll speed

benma:
The simulator didn't behave like the real firmware: it only polled/spun the async tasks when there was USB input available, which happens at the client poll interval during a long-running workflow.

Instead, usb_packet_process should be called if there is USB input, but the async tasks should be polled independently at max speed.

We noticed this issue when making bip39-unlock async, and it would only advance by one bip39 stretch iteration every few hundred milliseconds.

### DIFF
--- a/test/simulator/simulator.c
+++ b/test/simulator/simulator.c
@@ -33,6 +33,7 @@
 #include <netdb.h>
 #include <netinet/in.h>
 #include <signal.h>
+#include <sys/select.h>
 #include <stdbool.h>
 #include <sys/socket.h>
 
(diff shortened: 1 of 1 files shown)


NickeZ:
Does this make the simulator use 100% of one cpu thread?

Maybe a quickfix could be to sleep 10 ms in the main loop?

(the real fix is to change to an event loop that blocks if there is nothing more to do)

benma:
> Does this make the simulator use 100% of one cpu thread?
> 
> Maybe a quickfix could be to sleep 10 ms in the main loop?
> 
Yeah I have a version where it does a timeout like this (pushed the commit here), but there is no good balance.

- 10ms still means an unbearably long unlock
- 0.05ms is still slow with cpu at about 10% (for me)

> > (the real fix is to change to an event loop that blocks if there is nothing more to do)

That is not an easy thing to do at the moment right?

For now I am abandoning this PR and simply making bip39 unlock not yield in every iteration in the simulator.

PR #1604, opened 2025-09-23
Move bb02_async

benma:


src/rust/util/Cargo.toml
```diff
@@ -31,5 +31,6 @@ bitcoin = {workspace = true}
 
 [features]
 rtt = ["dep:rtt-target", "dep:cortex-m"]
+testing = []
 c-unit-testing = []
-firmware = []
+firmware = []
```
NickeZ: still something off with your editor :)

src/rust/util/src/bb02_async.rs
```diff
@@ -51,7 +51,6 @@ pub fn option<O>(option: &RefCell<Option<O>>) -> AsyncOption<'_, O> {
 }
 
 /// Polls a future until the result is available.
-#[cfg(feature = "testing")]
```
NickeZ: I don't see why it is needed in non-testing builds again.

src/rust/util/Cargo.toml
benma: Hmm it seems my cargo/toml mode did indeed not set `require-final-newline` :thinking: I now enabled it globally for all files :crossed_fingers: 

Fixed

src/rust/util/src/bb02_async.rs
benma: Rebase error, fixed.

NickeZ (approved):
utACK

PR #1605, opened 2025-09-24
bitcoin/common: use bitcoin crate to compute pkscript

benma:


Beerosagos:
Deferring to @NickeZ :innocent:

PR #1607, opened 2025-09-25
events: Remove events unrelated to sliders

NickeZ:


NickeZ:
Fixed bug where keyboard was activated by wrong events. There was a logical bug in the event handler of trinary_input_char.

src/ui/components/trinary_input_string.c
```diff
@@ -419,16 +434,14 @@ static void _letter_chosen(component_t* trinary_char, char chosen)
     }
 
     if (data->string_index + 1 >= INPUT_STRING_MAX_SIZE) {
-        event_t e;
-        e.id = EVENT_CONFIRM;
-        emit_event(&e);
+        _confirm_button_cb(trinary_char);
```
benma: DeepSeek (verified by me)

---

   - In `_letter_chosen()`, when `data->string_index + 1 >= INPUT_STRING_MAX_SIZE`, it calls `_confirm_button_cb(trinary_char)` but passes the wrong parameter. It should pass `trinary_input_string` instead of `trinary_char`.

src/ui/components/right_arrow.c
```diff
@@ -33,6 +33,8 @@
 typedef struct {
     uint8_t location;
     bool active; // Marker is 'active', i.e., touched
+    void (*callback)(void* user_data);
+    void* user_data;
```
benma, line 37: I'd prefer, for consistency, if you stayed with the naming pattern of callback params as we have it, which is `<callbackname>_param`. In this case `callback_param` instead of `user_data`.

Screenshot_2025-09-30_13-26-25

src/ui/components/left_arrow.c
```diff
@@ -58,33 +60,21 @@ static void _render(component_t* component)
 static void _on_event(const event_t* event, component_t* component)
 {
     left_arrow_data_t* data = (left_arrow_data_t*)component->data;
-    const gestures_slider_data_t* slider_data = (const gestures_slider_data_t*)event->data;
+    if (event->data.source != data->location) {
+        return;
```
benma: Previously, `data->active = false` was set in the switch default case, now missing. Somehow, by touching the bottom slider while the arrow is active in the top slicer, active still seems to be reset to false and behave the same way as before, but I am not sure why.

In the show-mnemonic workflow, the arrows are on the bottom slider. Interestingly there, the arrow stays active even if touching the top slider, so the behavior is not symmetric.

src/ui/components/right_arrow.c
```diff
@@ -58,36 +60,23 @@ static void _render(component_t* component)
 static void _on_event(const event_t* event, component_t* component)
 {
     right_arrow_data_t* data = (right_arrow_data_t*)component->data;
-    const gestures_slider_data_t* slider_data = (const gestures_slider_data_t*)event->data;
+    if (event->data.source != data->location) {
+        return;
```
benma: Same, `data->active = false` missing now.

src/ui/components/trinary_input_string.c
```diff
@@ -337,62 +337,77 @@ static void _set_alphabet(component_t* trinary_input_string, bool maybe_autocomp
...
-        }
-        _set_alphabet(component, false);
+    case UPPER_CASE:
+        trinary_input_char_set_alphabet(trinary_char, _alphabet_uppercase, 1);
+        break;
+    case DIGITS:
+        trinary_input_char_set_alphabet(trinary_char, _digits, 1);
+        break;
+    case SPECIAL_CHARS:
+        trinary_input_char_set_alphabet(trinary_char, _special_chars, 2);
```
benma: This is a copy/paste of the last branch of _set_alphabet(), and before this PR, _set_alphabet() was called on keyboard switching. Why not just call _set_alphabet() here instead of repeating the code?

Or at least move this to a helper function and call it from both places.

src/ui/components/trinary_input_string.c
```diff
@@ -337,62 +337,77 @@ static void _set_alphabet(component_t* trinary_input_string, bool maybe_autocomp
...
 
-static void _on_event(const event_t* event, component_t* component)
+static void _confirm_button_cb(void* user_data)
 {
-    data_t* data = (data_t*)component->data;
-
-    if (event->id == EVENT_CONFIRM && data->can_confirm) {
+    component_t* self = (component_t*)user_data;
+    data_t* data = self->data;
+    // is this check needed?
```
benma: I think so, as the confirm callback could still be called even if the confirm button is not actually rendered. The confirm button is not rendered while entering a BIP39 word before it is complete.

src/ui/components/right_arrow.c
NickeZ: Can I rename "callback_param" to "user_data" everywhere, because "user_data" is the conventional term? I will never remember to call it "callback param".

https://stackoverflow.com/questions/50874154/what-is-the-use-of-userdata-in-callback-register-function-in-c

src/ui/components/trinary_input_string.c
NickeZ: comment removed

src/ui/components/left_arrow.c
NickeZ: Not related to this PR anymore

src/ui/components/right_arrow.c
NickeZ: https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1613

src/ui/components/right_arrow.c
NickeZ: Not related to this PR anymore

src/ui/components/trinary_input_string.c
NickeZ: `_set_alphabet()` seems to be some kind of "god function" that behaves very differently based on input. I can see again if I can reduce the duplicity, but sometimes just writing it out is simpler than having an extra function.

NickeZ:
> I am midway through the 2nd commit, but want to look at trinary_input_string.c more closely.
> 
> For review feedback, please add commits (no amend), and squash later, as it's a big PR and revisiting all files is time-consuming.

Sorry, I started fixing the issues before reading this comment.

src/ui/components/trinary_input_string.c
NickeZ: called the callback in `_set_alphabet()` to keep it DRY.

benma: Imho calling `_on_keyboard_switch_cb()` from `_set_alphabet()` even though there was no keyboard switch is less intuitive than calling `_set_alphabet()` from `_on_keyboard_switch_cb()` :joy: as you say, it's the "god" function that sets the keyboard correctly whenever called, so it's appropriate to call it.

But not that important, if you prefer it much more this way, then ok.

NickeZ: i made a separate function to update the `input_char` from the `keyboard_mode`.

I really didn't like that the `else` clause could cause a nullptr-deref in case the keyboard switch component was NULL, so I added an extra check. I'm sure the other checks implicitly meant that that was impossible. But it just felt so unsafe.

src/ui/components/trinary_input_string.c
NickeZ: fixed

benma (approved):
utACK, feel free to squash everything

PR #1608, opened 2025-09-29
sd.c: always pause clock after sdcard check

benma:
When the sdcard is inserted, the sdcard check runs into the retry loop. In that case, the clock was not paused, which could lead to touch issues.


NickeZ (approved):
tACK

PR #1610, opened 2025-09-29
releases: add v9.23.2

benma:


NickeZ (approved):
reproduced ACK

benma:
This is merged, GH does not seem to understand it :shrug: https://github.com/BitBoxSwiss/bitbox02-firmware/commit/1597e16b

PR #1612, opened 2025-09-30
Maint/reduce events merge sliders

NickeZ:
based on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1607


src/ui/components/left_arrow.c
```diff
@@ -58,33 +60,21 @@ static void _render(component_t* component)
 static void _on_event(const event_t* event, component_t* component)
 {
     left_arrow_data_t* data = (left_arrow_data_t*)component->data;
-    const gestures_slider_data_t* slider_data = (const gestures_slider_data_t*)event->data;
+    if (event->data.source != data->location) {
+        return;
```
benma, line 64: Copied from https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1607#discussion_r2390960245

Previously, `data->active = false` was set in the switch default case, now missing. Somehow, by touching the bottom slider while the arrow is active in the top slicer, active still seems to be reset to false and behave the same way as before, but I am not sure why.

In the show-mnemonic workflow, the arrows are on the bottom slider. Interestingly there, the arrow stays active even if touching the top slider, so the behavior is not symmetric.

NickeZ: I think it still is there, see line 83.

NickeZ: all the cases that `event->data.source != data->location` covers lead to `break` before and not to the `default` case in the switch.

src/ui/components/keyboard_switch.c
```diff
@@ -90,18 +90,19 @@ static void _render(component_t* component)
 static void _on_event(const event_t* event, component_t* component)
 {
     keyboard_switch_data_t* ks_data = (keyboard_switch_data_t*)component->data;
-    const gestures_slider_data_t* slider_data = (const gestures_slider_data_t*)event->data;
     switch (event->id) {
-    case EVENT_TOP_CONTINUOUS_TAP:
-        if (ks_data->location == top_slider && slider_data->position > SLIDER_POSITION_ONE_THIRD &&
-            slider_data->position <= SLIDER_POSITION_TWO_THIRD) {
+    case EVENT_CONTINUOUS_TAP:
```
benma, line 96: just remove ks_data->location? it is always top_slider, and the event handling wouldn't even work if it was bottom_slider. (separate commit please if you do it)

NickeZ: done as separate commit

PR #1613, opened 2025-09-30
maintenance: Rename "callback param" to "user data"

NickeZ:
"user data" is the conventional term for the user provided pointer that is returned in a callback.


src/rust/bitbox02/src/ui/ui.rs
benma, line 1: This file has a bunch of comments and vars containing `param` that should also be renamed, e.g. `cancel_cb_param`, and `// passed to c_confirm_callback as `param`.`

NickeZ: fixed!

benma (approved):
utACK

PR #1616, opened 2025-10-01
Fix/bottom tap while gesture confirm

NickeZ:


src/ui/components/trinary_input_string.c
```diff
@@ -245,6 +240,12 @@ static void _render(component_t* component)
     if (data->title_on_top || show_title) {
         data->title_component->f->render(data->title_component);
     }
+
+    // Render confirm button or gesture (render confirm gesture last so that bottom triangle is
+    // above keyboard)
```
benma: > so that bottom triangle is above keyboard

The keyboard is not rendered while the confirm gesture is active. Is this comment about something else?

NickeZ: Not true, you can clearly see the bottom triangle "animating out" beneath the keyboard.

benma: So the keyboard (and everything else) is not rendered while the bottom triangle is animating in, but it is rendered when it is animated out. Not a big deal, but I wonder if that should be changed so it waits until the bottom triangle is gone before rendering the rest again.

Quick patch to achieve this (not sure if this is a good way, but it works):

Screenshot_2025-10-13_09-49-12
 
Feels pretty good in my opinion, please try it.

NickeZ: also tried it, I really like it. Switched to that implementation.

PR #1618, opened 2025-10-01
avoid flicker when changing from lockscreen to waiting screen

benma:
The BitBox shows the lockscreen ("See the BitBoxApp") first, and when the noise connection is initiazed, switches to the BitBox logo, which stays there as the waiting/default screen until reconnected.

The noise connection is initiated by the host after unlock finishes, so in between, "See the BitBoxApP" flashes for a split second.

This commit solves this by

1) not pushing the lockscreen as an extra component to be popped, but folding it into the top/default component

2) updating that screen when the app initiates the connection if the device is not initialized

3) if initialized, updating the screen on OP_UNLOCK (see comment in code)

This prevents flashing/flickering if the device is initialized. There is still a small flicker when the device is uninitialized, as the screen is updated before the pairing confirmation appears. This was the case before this commit too.


src/ui/components/waiting.c
```diff
@@ -57,14 +63,36 @@ component_t* waiting_create(void)
...
+
+void waiting_host_connected(component_t* component)
+{
+    data_t* data = (data_t*)component->data;
+    if (data->host_connected) {
+        return;
+    }
+    data->host_connected = true;
+
+    if (component->sub_components.amount != 1) {
```
NickeZ, line 81: this should be an ASSERT. The subcomponent is clearly added in the constructor.

src/ui/screen_process.c
```diff
@@ -32,16 +32,21 @@ void ui_screen_render_component(component_t* component)
     UG_SendBuffer();
 }
 
+static component_t* _waiting_screen = NULL;
```
NickeZ: This static could be inside `_get_waiting_screen`.

src/ui/components/waiting.c
```diff
@@ -57,14 +63,36 @@ component_t* waiting_create(void)
     waiting->dimension.height = SCREEN_HEIGHT;
     waiting->position.top = 0;
     waiting->position.left = 0;
+    waiting->data = data;
+
+    ui_util_add_sub_component(waiting, lockscreen_create());
```
NickeZ, line 68: Since the lockscreen isn't a screen anymore, but a subcomponent, should we rename it to something like `locked_device_background` instead?

src/ui/components/waiting.c
```diff
@@ -57,14 +63,36 @@ component_t* waiting_create(void)
...
         logo.buffer.len,
         logo.dimensions.width,
         logo.dimensions.height,
         CENTER,
-        waiting);
-    ui_util_add_sub_component(waiting, bb2_logo);
-    return waiting;
+        component);
+
+    component->sub_components.sub_components[0] = bb2_logo;
```
NickeZ, line 98: Wouldn't it be less hacky if the waiting screen had two sub-components, the `connected` background and the `not connected` and only rendered one of them based on `host_connected`?

src/ui/screen_process.h
```diff
@@ -27,6 +27,11 @@ void ui_screen_render_component(component_t* component);
  */
 component_t* screen_process_get_top_component(void);
 
+/**
+ * Wraps `waiting_host_connected()` for the waiting screen.
```
NickeZ: Shouldn't this be either in `communication_mode` module or in some `connection_status` module? `screen_process` doesn't feel like the right place to keep track of if the device is connected to a host.

NickeZ: I don't see a point in this function just wrapping `waiting_host_connected()`, why can't callers of `screen_process_host_connected()` just call `waiting_host_connected()`?

src/ui/components/waiting.c
benma: The problem with ASSERT is that they don't run in prod, and if the assumption is wrong for some reason, there will be a memory bug. I could do `Abort(...)`  instead?

src/ui/components/waiting.c
benma: Not sure why that would be less hacky. I chose it this way because it needs less memory, which I thought could be relevant as this whole waiting component is never deallocated.

src/ui/screen_process.h
benma: Because they don't have access the `component_t*` static instance.

benma: > Shouldn't this be either in `communication_mode` module or in some `connection_status` module? `screen_process` doesn't feel like the right place to keep track of if the device is connected to a host.

I don't think so, because the moment we call it is not generic, it's tailored to update this specific screen in a way that reduces flashing/flickering (e.g. delayed until OP_UNLOCK is called, even though the host has connected potentially much earlier already). If you prefer, I could still move it somewhere else and document it that way, but I don't see much benefit in that.

benma: Maybe better to just rename "waiting_host_connected" to "waiting_update_display" or similar?

src/ui/components/waiting.c
NickeZ: yeah, Abort is fine as well if you want it to be checked in prod

src/ui/components/waiting.c
NickeZ: You are essentially implementing `ui_util_replace_sub_component` in the waiting screen. It doesn't feel like the right place for that code to be. But I guess we can have it here until we have a second use case for it.

src/ui/screen_process.h
NickeZ: why do you think it needs the indirection via `screen_process`. How is it related to the screen processing?

src/rust/bitbox02-rust/src/hww.rs
```diff
@@ -110,6 +110,13 @@ fn api_attestation(usb_in: &[u8]) -> Vec<u8> {
 }
 
 async fn _process_packet(hal: &mut impl crate::hal::Hal, usb_in: Vec<u8>) -> Vec<u8> {
+    // Update the waiting screen from "See the BitBoxApp" to the logo, now that the host is
+    // connected. When the device is initialized, we delay this until the unlock call, otherwise
+    // there would be a flicker where the logo would be shown before the host invokes unlock.
+    if !bitbox02::memory::is_initialized() || usb_in.as_slice() == [OP_UNLOCK] {
+        bitbox02::ui::screen_process_host_connected();
```
NickeZ: i think the name `screen_process_host_connected()` here must be clearer, I have a very hard time understanding what it does. Is it supposed to be `waiting_screen_switch_to_logo()?`

NickeZ: `host_connected()` makes me think it is check, like `if host_connected() { do_something }`.

src/ui/components/waiting.c
```diff
@@ -21,13 +22,12 @@
...
+#include <stdbool.h>
 #include <string.h>
 
-static void _render(component_t* component)
-{
-    // TODO - add an interesting animation?
-    ui_util_component_render_subcomponents(component);
-}
+typedef struct {
+    bool host_connected;
```
NickeZ: based on other comments I think the name of this variable is wrong. Is this truly `true` if and only if the host is connected?

NickeZ:
Could be done in a later PR. But what do you think about renaming `waiting_screen` to `idle_screen` and have `ui_screen_stack_top` return the Idle screen in case the stack is empty?

benma:
> Could be done in a later PR. But what do you think about renaming `waiting_screen` to `idle_screen` and have `ui_screen_stack_top` return the Idle screen in case the stack is empty?

`screen_process_get_top_component()` basically does what you suggest, and it seems better to keep it in screen_process than to have ui_screen_stack take care of it, but not a very strong opinion.

Renaming waiting_screen to idle_screen, I also don't have a strong opinion, can do.

src/ui/screen_process.h
benma: screen_process.c is where  `static component_t* _waiting_screen = NULL;` is defined, because `screen_process_get_top_component(void)` calls `_get_waiting_screen(void)` which instantiates it. So if you want to call `waiting_host_connected(<waiting component>)` (or waiting_switch_to_logo or whatever the name will be), it's wrapped in screen_process.c because it needs to pass the component instance.

src/rust/bitbox02-rust/src/hww.rs
benma: Yeah I like your name (switch_to_logo), will use that :+1:

src/ui/components/waiting.c
benma: Well no, as explained in another comment :) will make an update and rename everything

src/ui/screen_process.h
NickeZ: aha ok

NickeZ:
> > Could be done in a later PR. But what do you think about renaming `waiting_screen` to `idle_screen` and have `ui_screen_stack_top` return the Idle screen in case the stack is empty?
> 
> `screen_process_get_top_component()` basically does what you suggest, and it seems better to keep it in screen_process than to have ui_screen_stack take care of it, but not a very strong opinion.
> 
> Renaming waiting_screen to idle_screen, I also don't have a strong opinion, can do.

Let's refactor later...

src/ui/screen_process.c
benma: No, as it is needed by `screen_process_host_connected()` (now renamed to `screen_process_waiting_switch_to_logo()`.

src/ui/components/waiting.c
benma: Done

src/ui/screen_process.h
benma: Renamed now

src/ui/components/waiting.c
benma: I don't see a need, in the end it's doing the same as before :shrug:

src/ui/components/waiting.c
benma: Renamed everything now

src/rust/bitbox02-rust/src/hww.rs
benma: Done

src/ui/screen_process.c
NickeZ: no, you are getting a pointer to the component using `_get_waiting_screen()` in `screen_process_waiting_switch_to_logo()` so you should not need to have it file-visible.

This works fine for example:
```c
#include <stdint.h>
#include <stdio.h>

int32_t *get_static(void) {
  static int32_t a = 5;
  return &a;
}

void inc_static(void) {
  int32_t *a = get_static();
  (*a)++;
}

int main(void) {
  printf("%d\n", *get_static());
  inc_static();
  printf("%d\n", *get_static());
}
```

outputs:
```
5
6
```

This is a typical singleton pattern in C and the static really doesn't need to be file visible.

benma: Ah of course, and this is how it was before I changed it in this PR for no reason :facepalm: reverted that part.

PR #1619, opened 2025-10-02
keystore: fewer securechip calls when checking password

benma:
The sanity check to see if the seed has changed does not need a securechip operation, it can use the retained seed hash instead, same as `unlock_bip39()`. This reduces the number of securechip operations needed to do a password check, which reduces the risk of running into the Optiga throttling security mechanism.

This is based on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1621 so the reduction can be seen in the show_mnemonic unit test too

### DIFF
--- a/src/keystore.c
+++ b/src/keystore.c
@@ -422,6 +422,23 @@ keystore_error_t keystore_create_and_store_seed(
     return keystore_encrypt_and_store_seed(seed, host_entropy_size, password);
 }
 
+// Checks if the retained seed matches the passed seed.
+static bool _check_retained_seed(const uint8_t* seed, size_t seed_length)
+{
+    if (!_is_unlocked_device) {
+        return false;
+    }
+    uint8_t seed_hashed[32] = {0};
+    UTIL_CLEANUP_32(seed_hashed);
+    if (_hash_seed(seed, seed_length, seed_hashed) != KEYSTORE_OK) {
+        return false;
+    }
+    if (!MEMEQ(seed_hashed, _retained_seed_hash, sizeof(_retained_seed_hash))) {
+        return false;
+    }
+    return true;
+}
+
 keystore_error_t keystore_unlock(
     const char* password,
     uint8_t* remaining_attempts_out,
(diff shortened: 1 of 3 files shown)


src/keystore.c
```diff
@@ -455,14 +472,9 @@ keystore_error_t keystore_unlock(
...
-            uint8_t current_seed[KEYSTORE_MAX_SEED_LENGTH] = {0};
-            size_t current_seed_len = 0;
-            if (!keystore_copy_seed(current_seed, &current_seed_len)) {
-                return KEYSTORE_ERR_DECRYPT;
-            }
-            if (seed_len != current_seed_len || !MEMEQ(current_seed, seed, current_seed_len)) {
+            if (!_check_retained_seed(seed, seed_len)) {
                 Abort("Seed has suddenly changed. This should never happen.");
-            }
+            };
```
NickeZ: seems like an extra `;` here

src/rust/bitbox02/src/keystore.rs
```diff
@@ -466,15 +466,18 @@ mod tests {
...
+        assert_eq!(crate::securechip::fake_event_counter(), 6);
+
         // Loop to check that unlocking works while unlocked.
-        for _ in 0..3 {
-            // First call: unlock Further calls perform a password check. The first onedoes a seed
-            // rentention (1 securechip event). The password check does not do the rentention but a
-            // copy_seed() instead to check the seed, so they end up hacing the same number of
-            // events.crate::securechip::fake_event_counter_reset();
+        for _ in 0..2 {
+            // Further calls perform a password check.The password check does not do the rentention
```
NickeZ: spelling 
> rentention

NickeZ (approved):
tACK, two nits

src/keystore.c
benma: Fixed :+1:

src/rust/bitbox02/src/keystore.rs
benma: Fixed

PR #1620, opened 2025-10-02
workflows: use HAL for showing mnemonic and mnemonic quiz

benma:
`show_and_confirm_mnemonic(hal: ...)` did not use the HAL for all UI: show_mnemonic and confirm_mnemonic directly used the BitBox02 menu workflow without going through HAL. Can't use `hal().ui().menu(...)`, becaus:

- these functions bolt a cancel prompt on top
- show_mnemonic is not a normal menu where one can pick an entry, it's just a scroll-through for displaying the mnemonic

This should make it possible to mock/fake/test the mnemonic workflow functions, and provide different implementations for them in future BitBox hardware.


NickeZ (approved):
utACK

PR #1621, opened 2025-10-02
Test show mnemonic

benma:
Builds on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1620

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
@@ -62,3 +62,149 @@ pub async fn process(hal: &mut impl crate::hal::Hal) -> Result<Response, Error> 
     hal.ui().status("Backup created", true).await;
     Ok(Response::Success(pb::Success {}))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use alloc::boxed::Box;
+
+    use crate::hal::testing::TestingHal;
+    use crate::workflow::testing::Screen;
+    use bitbox02::testing::mock_memory;
+    use util::bb02_async::block_on;
+
+    /// When not yet initialized, we show the mnemonic without a password check. This happens during
+    /// wallet setup.
+    #[test]
+    fn test_process_uninitialized() {
+        mock_memory();
+        bitbox02::keystore::encrypt_and_store_seed(
+            hex::decode("c7940c13479b8d9a6498f4e50d5a42e0d617bc8e8ac9f2b8cecf97e94c2b035c")
+                .unwrap()
+                .as_slice(),
+            "password",
+        )
+        .unwrap();
+
+        assert!(!bitbox02::memory::is_initialized());
+
+        let mut mock_hal = TestingHal::new();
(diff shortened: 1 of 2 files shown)


src/rust/bitbox02-rust/src/workflow.rs
```diff
@@ -79,6 +79,20 @@ pub trait Workflows {
...
+    /// The default implementation is implemented in terms of `self.show_mnemonic()`,
+    /// `self.quiz_mnemonic_word()`, etc.
+    ///
+    /// This function is defined in the HAL so unit tests can easily mock it. Real implementations
+    /// should leave the default implementation.
+    async fn show_and_confirm_mnemonic(&mut self, words: &[&str]) -> Result<(), cancel::Error>
+    where
+        Self: Sized,
+    {
+        mnemonic::show_and_confirm_mnemonic(self, words).await
```
NickeZ, line 94: This is the first time the trait implements the functionality, are we sure we want that?

benma: It was at least my intention, as per the docstring:

```
/// This function is defined in the HAL so unit tests can easily mock it. Real implementations
/// should leave the default implementation.
```

Assuming say a BB03 would also implement HAL show_mnemonic and quiz_mnemonic_word, the default impl of this should work fine.

I considered not adding this trait fn here but doing a local `trait Funcs` in api/show_mnemonic.rs for mocking locally there, but that was a ton of boilerplate, repeating much of testing `TestingWorkflows`. And I'd need to put it somewhere reusable as api/bip85.rs also would need it.

NickeZ: But it is calling a specific implementation. I would assume it would call other trait functions.

NickeZ: > Assuming say a BB03 would also implement HAL show_mnemonic and quiz_mnemonic_word, the default impl of this should work fine.

Isn't this true for all trait functions? Should all of them have a default impl and we just simplify RealWorkflows to not override any trait functions?

In case we want the workflow trait to have a default implementation I think that implementation should be either in the same module as the trait (`worfklow`) or in a **private** module inside `workflow` (`workflow::mnemonic` is currnetly taken by a public module) because then it is an imlpementation detail of the trait and not something that should be callable from everywhere.

benma: > But it is calling a specific implementation. I would assume it would call other trait functions.

That implementation is taking `self` and is implemented purely in terms of the other trait functions.
 I could inline it here, but thought that would be too much code here, and mnemonic.rs is the more appropriate home for it.

> But it is calling a specific implementation. I would assume it would call other trait functions.


What would be the point of that?

With this particular trait function, the difference is that a BB03 does not need to override it as the default implementation will work for BB03. For the other trait functions, that is generally not the case - there a BB03 would provide different implementations based on different UI and a different input method.

NickeZ: Ok, if it only calls other trait functions it makes more sense to me. I still think module visibility is a bit off, but we will refactor that once we have a bb03 I suppose.

PR #1624, opened 2025-10-06
sd-card: Increase usb processing timeout if sd card presence is checked.

NickeZ:
Some users experienced a bug where they got "backup creation failed" even though the backup was successfully created.

This issue could be reproduced by forcing `sd_card_inserted` to always take 1s. The logs showed that the usb processing incorrectly timed out. The processing timeout is therefore reset to about 1s more than the default.

For good measure the loop checking the sd card presence is also changed to iterate more times and wait shorter times. One common case is that it runs a single iteration of the loop, and that is now much quicker.

### DIFF
--- a/src/sd.c
+++ b/src/sd.c
@@ -29,6 +29,7 @@
 #include "screen.h"
 #include "sd.h"
 #include "util.h"
+#include <usb/usb_processing.h>
 
 #include <ff.h>
 
@@ -311,14 +312,17 @@ void sd_free_list(sd_list_t* list)
 
 bool sd_card_inserted(void)
 {
+    // The sd card is allowed up to 1s to initialize, therefore the usb processing timeout must also
+    // be increased by about 1s.
+    usb_processing_timeout_reset(-10);
 #ifdef TESTING
     return true;
 #else
     sd_mmc_err_t err = sd_mmc_check(0);
     /* If initialization is ongoing, wait up to 1 second for it to initialize */
     if (err == SD_MMC_INIT_ONGOING) {
-        for (int i = 0; i < 10; ++i) {
-            delay_ms(100);
+        for (int i = 0; i < 100; ++i) {
+            delay_ms(10);
             err = sd_mmc_check(0);
             if (err != SD_MMC_INIT_ONGOING) {
                 break;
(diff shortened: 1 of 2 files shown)


src/sd.c
```diff
@@ -311,14 +312,17 @@ void sd_free_list(sd_list_t* list)
 
 bool sd_card_inserted(void)
 {
+    // The sd card is allowed up to 1s to initialize, therefore the usb processing timeout must also
+    // be increased by about 1s.
+    usb_processing_timeout_reset(-10);
```
benma, line 317: Did you check if we actually sleep in case of backup create, assuming the app checked sd card presence beforehand once (which it does)? So we can more confidently release a hotfix knowing it's the right fix. Maybe one of the other operations in backup creation (copy_seed, writing backup) also causes delays and we should give more time there?

NickeZ: I checked now and during backup creation it only sleeps once, either if the sd-card was inserted before the device was started or after a replug of the sd-card.

NickeZ: When I create a backup from the app it calls `sd_card_inserted()` 88 times

NickeZ: during the device initialization it didn't sleep during the backup creation for me. But I don't think that is conclusive either way.

NickeZ: Maybe sleeping in `sd_card_inserted()` isn't the culprit, but the `timeout_reset` still solves the issue because `sd_card_inserted()` is called from many places.

benma: If the culprit is not sd_card_inserted (which it probably is not if there is no sleep and the check is fast), then the timeout reset should *not* be here right? 

How about measuring how long backup.rs:create() takes, roughly?

 > When I create a backup from the app it calls sd_card_inserted() 88 times

That is crazy and clearly a bug, but can you specify when it is called? The app can't make API calls *during* another API call, so if it's the app making all these calls, it should technically not interfere with backup creation.

NickeZ: I can dig a little, but we don't call it from so many places and in one place we call it in the render function of a component. Probably that is why?

benma: Ah yeah, if you run into the blocking insert prompt, that must be it. But that should not interfere with the backup creation api call :thinking:

benma:
This can probably be closed in favor of https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1629 -> sdcard was probably not the cause for the timeouts, and the other PR also decreases the sleep time.

PR #1626, opened 2025-10-07
maintenance: allow reuse of address and port for the simulator.

bznein:
This way, the simulator can be killed and immediately restarted without encountering the "address already in use" error.

I'm not sure if there is a better way to achieve this, I got some AI help to arrive to this :) But it does make my automatic testing more reliable, while without this option it would fail when restarting the simulator with the error mentioned above.

### DIFF
--- a/test/simulator/simulator.c
+++ b/test/simulator/simulator.c
@@ -143,6 +143,20 @@ int main(int argc, char* argv[])
         perror("ERROR opening socket");
         return 1;
     }
+
+    int yes = 1;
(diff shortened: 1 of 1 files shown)


bznein:
Moving it back to draft as I'm not sure it fixes the issue :(

bznein:
Reopened: I verified and it does indeed solve the issue in my tests. Using the simulator without this patch makes it fail when starting again after killing it, while with this version it works

NickeZ:
Please try this solution instead. I'm not sure what SO_REUSEADDR implicates.

https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1627

bznein:
Closing in favour of #1627

PR #1628, opened 2025-10-09
simulator-graphical: Initial commit

NickeZ:
todo
- [x] Implement denying connections until orientation is picked
- [x] Implement slide event
- [x] Fix bug with gesture confirm (not being able to move away and cancel)
- [x] Alpha blend away black pixels from the screen.
- [x] Fix resizing, resize everything and not excluding the margins like now


src/rust/bitbox02-rust-c/Cargo.toml
```diff
@@ -68,6 +68,16 @@ target-c-unit-tests = [
   "firmware",
   "c-unit-testing",
 ]
+target-simulator-graphical = [
+  "dep:bitcoin", # for rust_base58_encode_check()
+  "app-bitcoin",
+  "app-litecoin",
+  "app-ethereum",
+  "app-cardano",
+  "app-u2f",
```
benma: Probably don't need that in the simulator as it can't be used anyway?

test/simulator-graphical/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
+# Copyright 2024 Shift Cryptosecurity AG
```
benma: wrong year and name. 

```# SPDX-License-Identifier: Apache-2.0``` if you want to already start using that, or fix the line for now and we will run a script later to replace all.

**Review by benma — COMMENTED** [review]
in send_massage, I did `9 1` (sign normal btc transaction), and it panicked with

```
thread 'main' panicked at /home/user/coding/bb/bitbox02-firmware/src/rust/bitbox02-rust/src/hww.rs:72:19:
called `Option::unwrap()` on a `None` value
```

File: src/rust/bitbox02-rust-c/Cargo.toml (hunk 1 above)
Comment (NickeZ) [reply src/rust/bitbox02-rust-c/Cargo.toml]: fixed

File: test/simulator-graphical/CMakeLists.txt (hunk 2 above)
Comment (NickeZ) [reply test/simulator-graphical/CMakeLists.txt]: fixed

**Comment by NickeZ** [comment]
> in send_massage, I did `9 1` (sign normal btc transaction), and it panicked with
> 
> ```
> thread 'main' panicked at /home/user/coding/bb/bitbox02-firmware/src/rust/bitbox02-rust/src/hww.rs:72:19:
> called `Option::unwrap()` on a `None` value
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

sorry about that, fixed now, some unit-testing code was compiled in instead of real code.

File: src/rust/bitbox02-rust-c/src/lib.rs [hunk 3]
```diff
@@ -38,8 +38,7 @@ extern crate util;
 // Whenever execution reaches somewhere it isn't supposed to rust code will "panic". Our panic
 // handler will print the available information on the screen and over RTT. If we compile with
 // `panic=abort` this code will never get executed.
-#[cfg(not(test))]
-#[cfg(not(feature = "testing"))]
+#[cfg(not(any(test, feature = "testing", feature = "c-unit-testing")))]
```
benma: `test` is redundant and implied by `testing`. We introduced `testing` because `test` only applies to the current crate (or file?).

src/rust/bitbox02-rust/src/hww.rs
```diff
@@ -27,7 +27,7 @@ const OP_STATUS_FAILURE_UNINITIALIZED: u8 = 2;
 /// Must be called during the execution of a usb task. This sends out the response to the host and
 /// awaits the next request. If the request is not a valid noise encrypted protofbuf api request
 /// message, `Err(Error::InvalidInput)` is returned.
-#[cfg(not(feature = "testing"))]
+#[cfg(not(any(test, feature = "c-unit-testing")))]
```
benma: leave `feature="testing"` instead for consistency? Here and in more instances below.

src/rust/bitbox02-rust-c/src/lib.rs
NickeZ: since I couldn't run neither the rust nor the c unit test I haven't been running them and haven't thought too much about this. I'll fix it.

test/simulator-graphical/src/main.rs
```diff
@@ -0,0 +1,836 @@
...
+
+    info!("Sd card setup: success");
+
+    bitbox02::testing::mock_memory();
+    info!("Memory setup: success");
+
+    if preseed {
+        let mnemonic = "boring mistake dish oyster truth pigeon viable emerge sort crash wire portion cannon couple enact box walk height pull today solid off enable tide";
+        let seed = bitbox02_rust::bip39::mnemonic_to_seed(&mnemonic).unwrap();
+        bitbox02_rust::keystore::encrypt_and_store_seed(&seed, "").unwrap();
```
cedwies: <s>This should be a module path error and should not compile without custom aliases (or different crate structure?).

Had to change this to:

`
let seed = bitbox02_rust::bip39::mnemonic_to_seed(&mnemonic).unwrap();
bitbox02::keystore::_encrypt_and_store_seed(&seed, "").unwrap();
`
Would be interesting to see if that compiled on your machine anyway.</s>

cedwies (commented):
Built on Apple Silicon.

The simulator gets stuck at "Busy, won't accept new clients" even after completing the orientation screen interaction.

Workaround: Forcing `ACCEPTING_CONNECTIONS.store(true, Ordering::Relaxed)` immediately after spawning the orientation screen allows connections to work.

In `fn init_hww(preseed: bool) -> bool` in `test/simulator-graphical/src/main.rs`

```suggestions
    unsafe {
        rust_workflow_spawn_orientation_screen(
            Some(orientation_selected_cb),
            std::ptr::null_mut() as *mut _,
        )
    };
    // Workaround: Force accepting connections for debugging
    info!("DEBUG: Forcing ACCEPTING_CONNECTIONS to true");
    ACCEPTING_CONNECTIONS.store(true, Ordering::Relaxed);

    true
```

I am unsure about the root cause. May be Apple/Architecture Specific

NickeZ:
> Built on Apple Silicon.
> 
> The simulator gets stuck at "Busy, won't accept new clients" even after completing the orientation screen interaction.
> 
> Workaround: Forcing `ACCEPTING_CONNECTIONS.store(true, Ordering::Relaxed)` immediately after spawning the orientation screen allows connections to work.
> 
> In `fn init_hww(preseed: bool) -> bool` in `test/simulator-graphical/src/main.rs`
> 
> ```
>     unsafe {
>         rust_workflow_spawn_orientation_screen(
>             Some(orientation_selected_cb),
>             std::ptr::null_mut() as *mut _,
>         )
>     };
>     // Workaround: Force accepting connections for debugging
>     info!("DEBUG: Forcing ACCEPTING_CONNECTIONS to true");
>     ACCEPTING_CONNECTIONS.store(true, Ordering::Relaxed);
> 
>     true
> ```
> 
> I am unsure about the root cause. May be Apple/Architecture Specific

Can you check that you are on b1bfdc11fd2d1bb50cf3c1fa9aab694ae4c9719f? I think I fixed that issue and then force pushed. edit: and the reason I forced pushed was since I rebased on master to fix merge conflits.

src/rust/bitbox02-rust/src/hww.rs
NickeZ: that doesn't seem to work. what do you mean specifically?

NickeZ: the graphical simulator requires mocked hardware, so it sets the testing feature. Do you mean that is wrong?

NickeZ: I believe the "mocks" in this file is intended only for the rust unit tests and the c simulator. We don't want to mock communication in the graphical simulator.

README.md
```diff
@@ -59,6 +59,13 @@ messages. The encryption protocol used is Noise with out-of-band authentication.
 The instructions for setting up a development environment can be found in [BUILD.md](BUILD.md).
 Please also read our [CONTRIBUTING.md](CONTRIBUTING.md) before filing issues and pull requests.
 
+### Simulator
+
+To test without hardware there is a simulator.
+
+<p align="center"><img src="./doc/bb02-simulator-graphical.png" /></p>
+
+
```
benma, line 69: docs/BUILD.md has two sections about the old simulator, so they should be updated too, and maybe you could link to that here.

src/rust/bitbox02-rust-c/Cargo.toml
```diff
@@ -69,6 +69,15 @@ target-c-unit-tests = [
   "firmware",
   "c-unit-testing",
 ]
+target-simulator-graphical = [
+  "dep:bitcoin", # for rust_base58_encode_check()
```
benma: This line is not needed probably. See https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1682

src/rust/bitbox02/src/event.rs
```diff
@@ -0,0 +1,7 @@
+pub use bitbox02_sys::event_slider_data_t;
```
benma, line 14: License header missing here and all other new files.

src/rust/bitbox02/src/event.rs
```diff
@@ -0,0 +1,7 @@
...
+pub use bitbox02_sys::event_t;
+pub use bitbox02_sys::event_types;
+
+pub fn emit_event(event: *const event_t) {
```
benma: This is not a safe wrapper. You should be able to use `event: &event_t` instead or not?

src/rust/bitbox02/src/queue.rs
```diff
@@ -0,0 +1,3 @@
+pub fn pull_hww() -> *const u8 {
```
benma: Could be nice to have a Rusty wrapper instead of returning a pointer, i.e. return `Option<[u8; 64]>` or something like that.

src/rust/bitbox02/src/sd.rs
```diff
@@ -19,6 +19,11 @@ use alloc::vec::Vec;
 use crate::util::str_to_cstr_vec;
 use bitbox02_sys::SD_MAX_FILE_SIZE;
 
+#[cfg(feature = "simulator-graphical")]
+pub fn format() -> bool {
```
benma, line 23: Use this also in `test_list_write_read_erase`.

src/rust/bitbox02/src/usb_packet.rs
```diff
@@ -0,0 +1,5 @@
+pub use bitbox02_sys::USB_FRAME;
+
+pub fn process(packet: *const USB_FRAME) -> bool {
```
benma: same, this is not a safe wrapper, best to avoid pointers in the params.

src/ui/screen_process.h
```diff
@@ -49,6 +49,10 @@ void screen_process(void);
  * Period of screen updates.
  * The screen is refreshed every SCREEN_FRAME_RATE event loops cycles.
  */
-#define SCREEN_FRAME_RATE 30
+#if defined(TESTING)
+    #define SCREEN_FRAME_RATE 1
```
benma, line 55: A comment here would be helpful, not quite clear why it needs to be different in the simulator.

test/simulator-graphical/src/main.rs
```diff
@@ -0,0 +1,819 @@
...
+                    self.inbound_out = inbound_out;
+                }
+                // Send data to TCP Client
+                loop {
+                    let hww_data = bitbox02::queue::pull_hww();
+                    if hww_data != std::ptr::null() {
+                        if let Some(outbound_in) = &mut self.outbound_in {
+                            let mut data: [u8; 64] = [0; 64];
+                            let slice: &[u8] = unsafe { std::slice::from_raw_parts(hww_data, 64) };
+                            data.copy_from_slice(&slice[..]);
```
benma: Why this and not directly use the `slice` var only, and delete the `data` var?

src/rust/bitbox02-rust-c/Cargo.toml
NickeZ: fixed

src/rust/bitbox02/src/event.rs
NickeZ: fixed

src/rust/bitbox02/src/event.rs
NickeZ: fixed

src/rust/bitbox02/src/queue.rs
NickeZ: fixed

src/rust/bitbox02/src/sd.rs
NickeZ: fixed

src/rust/bitbox02/src/usb_packet.rs
NickeZ: fixed

src/ui/screen_process.h
NickeZ: fixed, I think in my latest "screen" work I also remove this define. We want to call screen_process with a fixed frame rate and not just divide the main loop by 30. So eventually this whole definition will go away.

test/simulator-graphical/src/main.rs
NickeZ: fixed, "slice" isn't safe to pass around since that is a pointer given to us by C and I'm not sure how that is allocated... I prefer to have a real rust object.

src/rust/bitbox02/src/queue.rs
```diff
@@ -1,3 +1,23 @@
...
+// limitations under the License.
+pub fn pull_hww() -> Option<[u8; 64]> {
+    let hww_data = unsafe { bitbox02_sys::queue_pull(bitbox02_sys::queue_hww_queue()) };
+    if hww_data.is_null() {
+        return None;
+    }
+    let mut data: [u8; 64] = [0; 64];
+    let slice: &[u8] = unsafe { std::slice::from_raw_parts(hww_data, 64) };
+    data.copy_from_slice(slice);
+    Some(data)
```
benma, line 21: `Some(slice.try_into().unwrap())` would also work if you prefer. Strange there is no from_raw_parts for arrays with const generics yet.

src/rust/bitbox02-rust-c/Cargo.toml
```diff
@@ -91,6 +99,8 @@ testing = ["bitbox02-rust/testing", "bitbox02/testing"]
 # Active when the Rust code is compiled to be linked into the C unit tests and simulator.
 c-unit-testing = ["bitbox02-rust/c-unit-testing", "bitbox02/c-unit-testing", "util/c-unit-testing"]
 
+simulator-graphical = ["bitbox02-rust/simulator-graphical"]
```
benma: should maybe also enable `bitbox02/simulator-graphical` explicitly, even though it's done via bitbox02-rust transitively. a bit unsure about that.

test/simulator-graphical/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "simulator-graphical"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+bitbox02-rust = { path="../../src/rust/bitbox02-rust" }
+bitbox02-rust-c = { path="../../src/rust/bitbox02-rust-c", features=["target-simulator-graphical"] }
+bitbox02 = { path="../../src/rust/bitbox02" }
```
benma: enable `simulator-graphical` explicitly in bitbox02-rust and bitbox02 as well? unsure what best practice is.

src/rust/bitbox02/src/queue.rs
NickeZ: Hmm, now that I think about it I even prefer to use `ptr::copy_nonoverlapping` and never even instantiate the "questionable" slice.

src/rust/bitbox02-rust-c/Cargo.toml
NickeZ: Maybe confusing if you want to remove it in the future and you have to remove it in multiple places?

test/simulator-graphical/Cargo.toml
NickeZ: I could just remove `target-simulator-graphical` in that case. I'll do that since we never build that target as a c static library.

src/rust/bitbox02-rust-c/Cargo.toml
NickeZ: I removed it and hoisted it up to the top crate for both `bitbox02-rust` and `bitbox02`.

PR #1629, opened 2025-10-10
Sd yield

benma:


NickeZ:
I tested it using iPhone, but the app never leaves the "setup your wallet" screen when I restore from sd-card.

edit: could reproduce the failure 2/2 times, so it seems consistent. App v4.48.2

PR #1631, opened 2025-10-12
orientation-screen: Migrate to rust

NickeZ:
create an async friendly hardware supported delayin C and wrap it in rust using Future.


src/rust/bitbox02-rust/src/workflow/orientation_screen.rs
```diff
@@ -0,0 +1,43 @@
...
+    let mut orientation_arrows = bitbox02::ui::orientation_arrows(|upside_down| {
+        if upside_down {
+            bitbox02::screen_rotate()
+        }
+        *result.borrow_mut() = Some(());
+    });
+    orientation_arrows.screen_stack_push();
+
+    // Wait until orientation has been chosen
+    option_no_screensaver(&result).await;
```
benma: I think you want `option` not `option_no_screensaver`. The screensaver should become active in the orientation screen. It may be the most likely to actually burn in the screen, e.g. when connected and left plugged in.

src/rust/bitbox02-rust/src/workflow/orientation_screen.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    // During this delay the bb02 logotype is shown
+    if let Ok(delay) = bitbox02::delay::Delay::from_ms(1300) {
+        delay.await;
+    }
+
+    // Switch to lockscreen that shows "See the bitbox app" and device name
+    bitbox02::ui::screen_process_waiting_switch_to_lockscreen();
+
+    orientation_selected_cb();
```
benma: Did a double take seeing a callback called at the end of an async function. Would it be possible to simply return whatever value you need and use it when polling is done (in  `if let Poll::Ready(result) = result { ... `)

NickeZ: I don't think so because this future isn't run in a traditional async runtime, it is run with our "glue layer". I could return the callback and then call it in the `rust_workflow_spin()` function instead. But not sure if that is great.

I was hoping it would be helpful if the orientation workflow was in rust for the simulator. But in the end it didn't play out so well, I still have to call the glue-layer because I don't have a proper async runtime in the simulator either.

NickeZ: (the rust version of the orientation selection is still more readable IMO than the c version with interrupts, so I still think it is a net win.)

benma (changes requested):
The orientation screen is not actually working, it's always the same orientation no matter which side you tap.

NickeZ:
> The orientation screen is not actually working, it's always the same orientation no matter which side you tap.

Right, I noticed now that we deliver multiple short taps so the screen is rotated back to normal. In the C-code we popped the screen in the event handler so the screen never sees more than one tap.

NickeZ:
> The orientation screen is not actually working, it's always the same orientation no matter which side you tap.

Last commit ensures that callback only is called once for the component.

src/ui/components/orientation_arrows.c
```diff
@@ -52,6 +52,7 @@ static void _flip(component_t* component)
     orientation_data_t* data = (orientation_data_t*)component->parent->data;
     if (data->enable_touch) {
         data->done_callback(true, data->cb_param);
+        data->enable_touch = false;
```
benma: there is `data->enable_touch = true;`  in the render function, and it seems it's called on every render? Or if not it's not obvious. For simplicity I'd prefer `if (data->enable_touch && data->done_callback != NULL) { ...; data->done_callback = NULL }`.

NickeZ: fixed

src/rust/bitbox02-rust-c/src/workflow.rs
```diff
@@ -131,6 +158,20 @@ pub unsafe extern "C" fn rust_workflow_confirm_poll(result_out: &mut bool) -> bo
     }
 }
 
+/// Returns true if there was a result.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn rust_workflow_orientation_screen_poll() -> bool {
```
benma: This function seems unused.

src/rust/bitbox02-rust-c/src/workflow.rs
```diff
@@ -92,6 +110,15 @@ pub unsafe extern "C" fn rust_workflow_spin() {
             }
             _ => (),
         }
+        match ORIENTATION_SCREEN_STATE {
+            TaskState::Running(ref mut task) => {
```
benma, line 106: I wonder why Rust is not erroring here as it's taking a mutable ref to a static mut :thinking:

src/rust/bitbox02-rust/src/workflow/orientation_screen.rs
```diff
@@ -0,0 +1,43 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use util::bb02_async::option;
+
+pub async fn create<CB>(orientation_selected_cb: CB)
```
benma: Not a must, but imho it would be be nice two split this into two functions:

    async fn choose_orientation() -> bool;

and

```
async fn process_orientation(on_done: CB) ... {
    let upside_down = choose_orientation().await;
    if upside_down {
            bitbox02::screen_rotate()
      } 
     // delay, update lockscreen, call callback
    
}
```

Then the actual orientation component is wrapped in a regular easy clean workflow that returns its result, and the other one is the middle-glue thing that is necessary as we call it from C.

src/rust/bitbox02/src/delay.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[cfg(not(any(feature = "testing", feature = "c-unit-testing")))]
+struct DelayInner {
+    bitbox02_delay: bitbox02_sys::delay_t,
+}
+
+#[cfg(any(feature = "testing", feature = "c-unit-testing"))]
+struct DelayInner {
+    thread_handle: Option<std::thread::JoinHandle<()>>,
+    done: std::sync::Arc<std::sync::atomic::AtomicBool>,
```
benma, line 27: why is this needed when the DelayState also has a Done arm? Can't you just move the state to `Done` directly?

src/rust/bitbox02-rust/src/workflow/orientation_screen.rs
```diff
@@ -0,0 +1,43 @@
...
+    option(&result).await;
+    drop(orientation_arrows);
+
+    // During this delay the bb02 logotype is shown
+    if let Ok(delay) = bitbox02::delay::Delay::from_ms(1300) {
+        delay.await;
+    }
+
+    // Switch to lockscreen that shows "See the bitbox app" and device name
+    bitbox02::ui::screen_process_waiting_switch_to_lockscreen();
```
benma, line 39: Any reason why you choose to wrap this one but not `usb_start()`, which is still in C (called by the callback)? 

I only noticed because in the original code, this function came after the BLE product change and usb_start(), not before.

src/rust/bitbox02/src/delay.rs
NickeZ: It is being set from another thread, so that wouldn't work.

src/rust/bitbox02-rust-c/src/workflow.rs
NickeZ: why would it error?

src/rust/bitbox02-rust-c/src/workflow.rs
NickeZ: Interesting. I guess I should use poll to get the result instead of a callback maybe. I'll try that.

src/rust/bitbox02-rust/src/workflow/orientation_screen.rs
NickeZ: yeah, usb_start is not needed in the simulator, whereas switch to lockscreen is universally required.

src/rust/bitbox02-rust-c/src/workflow.rs
benma: I don't think you have to or should. Sounds it might be more complicated actually :D

src/rust/bitbox02-rust-c/src/workflow.rs
benma: mutable refs to statics are disallowed, even in unsafe blocks:

https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html

NickeZ: we have `#![allow(static_mut_refs)]` at the top, and I just copy/pasted the other functions with similar code in the same file.

src/rust/bitbox02-rust/src/workflow/orientation_screen.rs
NickeZ: done

src/rust/bitbox02-rust-c/src/workflow.rs
NickeZ: It did work better, so I have done it like that now

src/firmware_main_loop.c
```diff
@@ -209,5 +196,22 @@ void firmware_main_loop(void)
...
+                    da14531_handler_current_product = (const uint8_t*)DEVICE_MODE;
+                    da14531_handler_current_product_len = sizeof(DEVICE_MODE) - 1;
+                    da14531_set_product(
+                        da14531_handler_current_product,
+                        da14531_handler_current_product_len,
+                        &uart_write_queue);
+                }
+                usb_start();
+            }
+        }
```
benma: consider moving this to a function or even a new file, to keep the main loop from growing too large.

or at least the part that was in the function before, sth like `handle_orientation(orientation);`

src/rust/bitbox02-rust-c/Cargo.toml
```diff
@@ -25,6 +25,7 @@ bitbox02-rust = { path = "../bitbox02-rust", optional = true }
 bitbox-aes = { path = "../bitbox-aes", optional = true }
 bitbox02 = { path = "../bitbox02", optional = true }
 bitbox02-noise = { path = "../bitbox02-noise", optional = true }
+cortex-m = { version = "0.7.7" }
```
benma: move it to src/rust/Cargo.toml and replace this and the one in src/rust/util/Cargo.toml with a workspace=true dep.

src/rust/bitbox02-rust/src/workflow/orientation_screen.rs
```diff
@@ -14,22 +14,21 @@
...
     });
     orientation_arrows.screen_stack_push();
-
     // Wait until orientation has been chosen
-    option(&result).await;
-    drop(orientation_arrows);
+    option(&result).await
+}
+
+pub async fn create() -> bool {
```
benma: don't know why that name, it does not mean anything to me. how about `process` or `choose_and_handle`?

src/firmware_main_loop.c
```diff
@@ -209,5 +196,22 @@ void firmware_main_loop(void)
...
+                if (memory_get_platform() == MEMORY_PLATFORM_BITBOX02_PLUS) {
+                    da14531_handler_current_product = (const uint8_t*)DEVICE_MODE;
+                    da14531_handler_current_product_len = sizeof(DEVICE_MODE) - 1;
+                    da14531_set_product(
+                        da14531_handler_current_product,
+                        da14531_handler_current_product_len,
+                        &uart_write_queue);
+                }
+                usb_start();
+            }
```
benma: Did you consider porting this part to Rust too? Then You wouldn't need to use/call the poll function and everything could be in orientation_screen.rs. (Could also be a separate PR)

src/rust/bitbox02-rust-c/src/lib.rs
```diff
@@ -46,5 +46,6 @@ fn panic(info: &core::panic::PanicInfo) -> ! {
     ::util::log::log!("{}", info);
     #[cfg(feature = "firmware")]
     bitbox02_rust::print_screen!(0, "Error: {}", info);
+    cortex_m::asm::bkpt();
```
benma, line 49: What's this about? Maybe should be a separate commit/PR with a commit msg.

NickeZ: Put it in a separate commit. It is a breakpoint so exeuction stops when there is a panic.

src/rust/bitbox02-rust/src/workflow/orientation_screen.rs
NickeZ: The C function is called `orientation_screen_create()`. I guess I could call it `orientation_screen` which is inline with `confirm`. But I do think it looks ugly to have `confirm::confirm` or `orientation_screen::orientation_screen`.

src/firmware_main_loop.c
NickeZ: Not sure if that is desirable as `usb_start` is target specific.

src/firmware_main_loop.c
NickeZ: factored it out

src/rust/bitbox02-rust/src/workflow/orientation_screen.rs
benma: Yeah the create functions there refer to creating component instances, but that does not transfer so well to higher level async Rust workflows.

f you don't like repeating words, the above function could be renamed to `choose` and this file to `orientation`, then it becomes `worklows::orientation::choose(...)`.

Could also merge the two funcs again, I asked to split them when you still called the C callback, but now the distinction is a bit fuzzier.

tACK regardless

NickeZ:
@benma I decided to simplify the delay, removed some state and abort if more than 10 concurrently are allocated. PTAL

src/delay.c
```diff
@@ -62,6 +61,10 @@ void delay_init_ms(delay_t* self, uint32_t ms)
         _tasks[i].timer.mode = TIMER_TASK_ONE_SHOT;
         timer_add_task(&TIMER_0, &_tasks[i].timer);
     }
+    CRITICAL_SECTION_LEAVE()
+    if (full) {
+        Abort("To many concurrent delays");
```
benma: To->Too

src/delay.c
```diff
@@ -74,3 +77,16 @@ bool delay_is_elapsed(const delay_t* self)
...
+void delay_cancel(const delay_t* self)
+{
+    ASSERT(self->id < COUNT_OF(_tasks));
+    // Check and remove task with disabled interrupts. Otherwise the interrupt may occur
+    // after checking the done flag and then task is removed twice (not allowed).
+    CRITICAL_SECTION_ENTER();
+    if (_tasks[self->id].timer.cb && !_tasks[self->id].done) {
+        timer_remove_task(&TIMER_0, &_tasks[self->id].timer);
+    }
+    CRITICAL_SECTION_LEAVE();
```
benma: Shouldn't this be one line down, the last thing in the function?

NickeZ: Yeah, if the api should be re-entrant it should be that. I can fix it.

PR #1636, opened 2025-10-13
make unlock-animation slower again like it used to be

benma:
When we switched from libwally to rust-bip39, bip39 unlock became twice as fast. It feels too fast now.

The animation speed is based on the frame render rate, which is not fixed. If the animation goes longer than the actual bip39 unlock computation, the animation afterwards is significantly faster as there is less work performed per mainloop iteration. In absence of fixed render frame rates, we just perform the bip39 unlock twice to maintain the same frame rate. Conincidentally, this results in an unlock speed which is basically the same as before with libwally, and we get a security check for free (repeat and double check the bip39 seed).

### DIFF
--- a/src/ui/components/unlock_animation.c
+++ b/src/ui/components/unlock_animation.c
@@ -26,13 +26,13 @@
 // This many iterations times the slowdown factor to render the whole animation.
 #define LOCK_ANIMATION_N_FRAMES (38)
 
-// Since BIP39 unlock takes 2048 iterations, and the screen frame rate is 30 (SCREEN_FRAME_RATE,
-// render is called only every 30th iteration), if we want both to finish at the same time, the
-// slowdown factor becomes the following:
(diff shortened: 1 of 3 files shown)


src/ui/components/unlock_animation.c
```diff
@@ -135,13 +135,17 @@ static const uint8_t LOCK_ANIMATION[LOCK_ANIMATION_ACTUAL_N_FRAMES][LOCK_ANIMATI
 /**
  * Gets a frame of the lock animation.
  */
+#include <screen.h>
 static const uint8_t* _get_frame(int frame_idx)
 {
-    if (frame_idx >= LOCK_ANIMATION_N_FRAMES) {
+    /* screen_sprintf_debug(200, "LOL %d", (int)frame_idx); */
```
NickeZ: unintentional debug logs?

benma: yep, removed

PR #1644, opened 2025-10-16
u2f: Do not turn off BLE when U2F is used over USB

NickeZ:
iOS devices support u2f over USB, so no need to disable BLE when only u2f is used.

iOS devices also send one packet of garbage when waking up, 64 null bytes, which also is ignored, so that the bitbox still can connect to the iOS device when it is awakened.

### DIFF
--- a/src/usb/usb_packet.c
+++ b/src/usb/usb_packet.c
@@ -67,7 +67,7 @@ bool usb_packet_process(const USB_FRAME* frame)
     switch (usb_frame_process(frame, &_in_state)) {
     case FRAME_ERR_IGNORE:
         // Ignore this frame, i.e. no response.
-        return false;
+        break;
     case FRAME_ERR_INVALID_SEQ:
         // Reset the state becuase this error indicates that there is a host application bug
         _reset_state();


src/usb/usb_packet.c
```diff
@@ -92,7 +92,7 @@ bool usb_packet_process(const USB_FRAME* frame)
                 ctx, _in_state.data, _in_state.len, _in_state.cmd, _in_state.cid)) {
             // Queue filled and will be sent during usb processing
             _reset_state();
-            return false;
+            return true;
```
benma, line 95: why this change?

src/usb/usb_packet.c
```diff
@@ -67,7 +67,7 @@ bool usb_packet_process(const USB_FRAME* frame)
     switch (usb_frame_process(frame, &_in_state)) {
     case FRAME_ERR_IGNORE:
         // Ignore this frame, i.e. no response.
-        break;
+        return false;
```
benma: Unnecessary? Previously break moves on to the `return false` after the switch block.

src/usb/usb_packet.h
```diff
@@ -42,7 +42,7 @@ typedef struct {
 /**
  * Processes an incoming USB packet.
  * @param[in] frame The frame that is to be processed.
- * @return true if we are waiting for more frames to complete a packet, false otherwise.
```
benma, line 45: This is still true, right? (returns true if more data is needed). The new comment makes that less clear. Or I don't understand the change.

src/usb/usb_packet.c
NickeZ: This change is to make the return code useful. returning whether "more data is needed" isn't useful anymore. It might've been before we moved the usb communication from interrupts to the main loop.

src/usb/usb_packet.h
NickeZ: Nope, it returns true if the packet was successfully parsed.

src/usb/usb_packet.c
NickeZ: why checking if packet was successfully parsed? Because then we can ignore spurious invalid messages, like those from iOS.

src/firmware_main_loop.c
```diff
@@ -112,30 +112,25 @@ void firmware_main_loop(void)
...
                 }
-                communication_mode_ble_disable();
+            } else {
+                util_log("usb_packet_process: invalid");
             }
         }
 #if APP_U2F == 1
         if (!u2f_data && hid_u2f_read(&u2f_frame[0])) {
             util_log("u2f data %s", util_dbg_hex((void*)u2f_frame, 16));
             u2f_packet_process((const USB_FRAME*)u2f_frame);
```
benma, line 133: u2f_packet_process was a copy/paste of usb_packet_process, and also returns true/false. Maybe there we can remove the return value completely then? (could be trickier as mock_hidapi.c seems to rely on it)

benma (approved):
utACK

PR #1645, opened 2025-10-16
U2F: remove option to cancel u2f flows

NickeZ:
In the U2F architecture it isn't possible for the device to "cancel" a user presence check because the browser can always just ask again. The only way to cancel it is from the OS/browser.

### DIFF
--- a/src/rust/bitbox02-rust-c/src/workflow.rs
+++ b/src/rust/bitbox02-rust-c/src/workflow.rs
@@ -62,6 +62,7 @@ pub unsafe extern "C" fn rust_workflow_spawn_confirm(
         CONFIRM_PARAMS = Some(confirm::Params {
             title: CONFIRM_TITLE.as_ref().unwrap(),
             body: CONFIRM_BODY.as_ref().unwrap(),
+            accept_only: true,
             ..Default::default()
         });
 


NickeZ:
> I don't understand. If a website asks "U2f register" or "U2F auth", and the user does not want that, shouldn't they be able to reject? The browser can ask, but do all of them do it again in an infinite loop? How is the user supposed to reject it?

A keyfob only has a single button, you reject by not pressing the button.

If you have multiple keyfobs, after confirming the one you want to confirm, the browser will stop asking and the rest of the devices will timeout after 30s I think.

The most obvious way to reject is of course to not put in the key you don't want to use. It is a hardware key after all.

edit: The U2F protocol is stateless, so the device also shouldn't store some sort of denylist of which sites that the user has cancelled. It also sounds complex to keep such a denylist around, like what if the user later wants to register with a site?

benma:
@NickeZ makes sense, but what's wrong with just leaving the reject button so user can reject instead of unplugging?

NickeZ:
> @NickeZ makes sense, but what's wrong with just leaving the reject button so user can reject instead of unplugging?

The reject button doesn't do anything, the dialog will pop up again within a second when the next request comes.

PR #1646, opened 2025-10-16
memory: Initialize noise private key and salt lazily

NickeZ:
If the device is powered off after erasing but before initializing those two fields the device is bricked. Initialize them on boot instead.

### DIFF
--- a/test/unit-test/test_memory.c
+++ b/test/unit-test/test_memory.c
@@ -19,6 +19,7 @@
 
 #include <memory/memory.h>
 #include <memory/memory_shared.h>
+#include <util.h>
 
 #include <stdint.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


src/memory/memory.c
```diff
@@ -399,13 +420,15 @@ bool memory_reset_hww(void)
             &random_bytes[sizeof(chunk_shared.fields.ble_identity_address)],
             sizeof(chunk_shared.fields.ble_identity_address));
 
-        // Two most significant bits must be set to indicate "public static address". See ch. 1.3.2
-        // in "Part B: Link Layer Specification" of bluetooth standard.
-        // ble_addr is stored in little endianness, so MSB is in the first byte.
+        // Two most significant bits must be set to indicate "public static address". See
```
benma: Should the BLE reset stuff also moved to execute on boot then?

Thinking further: what if power is cut while in the middle of erasing the chunks above, e.g. chunk 1 is erased, but not chunk 2?

Maybe it's better to set a memory flag somewhere "reset=true", which is set to false again at the end of `salt_root_empty()`, and if it still set on boot, it meanss the reset was interrupted and should be repeated.

NickeZ: > Should the BLE reset stuff also moved to execute on boot then?

BLE being reset isn't critical, so there is no risk of bricking the unit if it stays the same.

To make it 100% survive a power outage we would need a counter that we increase between every atomic operation (erased sector, written page). So that the sequence can be resumed when power comes back.

We should probably use the smart eeprom for that to avoid excessive wear.

benma: > To make it 100% survive a power outage we would need a counter that we increase between every atomic operation (erased sector, written page). So that the sequence can be resumed when power comes back.

Why not just repeat the whole memory_reset_hww() in that case? Does not seem worth the complexity to do it more granular.

NickeZ: I guess we could set a flag `reset_hww=true` and then erase and recreate keys during boot. As long as the flag is set we just do the whole reset again.

edit: and of course `reset_hww=false` at the end of the procedure

benma: That is exactly what I meant :smile:

PR #1660, opened 2025-10-30
rust/keystore: move bip39_unlock and ROOT_FINGERPRINT

benma:


src/rust/util/src/cell.rs
```diff
@@ -12,48 +12,44 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Same as `core::sync::SyncUnsafeCell`, which is still in nightly. Can remove once it is stable.
-pub struct SyncUnsafeCell<T: ?Sized> {
+// Needed because otherwise we have undefined references to _critical_section_1_0_acquire/release.
+// This is the official solution:
+// https://github.com/rust-embedded/critical-section/blob/bc0d1736c3b14e340c1b0c29042edb3e6bffe6b8/README.md#undefined-reference-errors
+#[cfg(not(any(feature = "testing", feature = "c-unit-testing")))]
+use cortex_m as _;
```
NickeZ, line 20: You could just write somethingg like "Link library to provide symbols necessary for critical section even if code isn't called explicitly."

src/rust/util/src/cell.rs
```diff
@@ -12,48 +12,44 @@
...
+
+// Inspired by as `core::sync::SyncUnsafeCell`, which is still in nightly. Critical section added so
```
NickeZ: Is it really inspired by SyncUnsafeCell? Doesn't seem to have much in common.

src/rust/util/src/cell.rs
```diff
@@ -12,48 +12,44 @@
...
     /// Reads the value from `self` without moving it. This leaves the
     /// memory in `self` unchanged.
-    ///
-    /// # Safety
-    ///
-    /// This is unsafe because it allows accessing the interior value without
-    /// synchronization. The caller must ensure no other code is currently
-    /// writing to this cell during the read operation.
-    pub unsafe fn read(&self) -> T
+    #[inline(never)]
```
NickeZ: Why inline never?

src/rust/util/src/cell.rs
benma: Done. Left the link b/c it could be useful.

src/rust/util/src/cell.rs
benma: Started out as, but yeah I removed that now.

src/rust/util/src/cell.rs
benma: Debug leftover, removed.

src/rust/util/src/cell.rs
```diff
@@ -12,48 +12,43 @@
...
+impl<T> SyncCell<T> {
     pub const fn new(val: T) -> Self {
-        SyncUnsafeCell {
+        SyncCell {
             value: core::cell::UnsafeCell::new(val),
         }
     }
 
     /// Reads the value from `self` without moving it. This leaves the
     /// memory in `self` unchanged.
```
NickeZ, line 38: Should we maybe demand that `T: Copy`? Making a clone of something that can be dropped seems scary?

benma: Great idea, added (also in write, to avoid running into misunderstandings that drop is not called on the old value).

PR #1661, opened 2025-10-31
keystore: port create_and_store_seed to Rust

benma:


src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -99,7 +99,31 @@ pub fn encrypt_and_store_seed(seed: &[u8], password: &str) -> Result<(), Error>
 /// bytes, resulting in 12 or 24 BIP39 recovery words.
 /// This also unlocks the keystore with the new seed.
 pub fn create_and_store_seed(password: &str, host_entropy: &[u8]) -> Result<(), Error> {
-    keystore::_create_and_store_seed(password, host_entropy)
+    if !matches!(host_entropy.len(), 16 | 32) {
+        return Err(Error::SeedSize);
+    }
+    let mut seed_vec = bitbox02::random::random_32_bytes();
+    let seed = &mut seed_vec[..host_entropy.len()];
```
NickeZ: what do you think about adding

```
let seed_len = host_entropy.len();
```

I think it would make L120 easier to read.

NickeZ (approved):
utACK, I assume there are tests that covers this?

benma:
> utACK, I assume there are tests that covers this?

Yes, in the same file, and also functional tests in other files.

src/rust/bitbox02-rust/src/keystore.rs
benma: Done

PR #1663, opened 2025-10-31
keystore: call rust keystore_unlock from C

benma:
Small fix: since keystore.rs::lock() does more than just call the C function since a recent commit, C invocations to lock must call the Rust function instead.


NickeZ (approved):
utACK

PR #1665, opened 2025-11-02
keystore: port _verify_seed to Rust

benma:


src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -46,6 +46,29 @@ pub fn is_locked() -> bool {
...
+        Err(_) => return false,
+    };
+    let decrypted = match bitbox_aes::decrypt_with_hmac(encryption_key, &cipher) {
+        Ok(decrypted) => decrypted,
+        Err(_) => return false,
+    };
+    if decrypted.len() != expected_seed.len() {
+        return false;
+    }
+    let mut diff = 0u8;
```
NickeZ: I'm surprised you don't just do `expected_seed == decrypted`? Does it need to be a constant time implementation?

`==` will be very likely lowered down to `memcmp`, which is what we did before. https://doc.rust-lang.org/src/core/slice/cmp.rs.html#139-155

NickeZ (commented):
left a comment

src/rust/bitbox02-rust/src/keystore.rs
benma: Yeah I was on the edge on this, the agent just did this proactively and I was wondering if I should change it back. I think I'll change it back, can't see how it helps here.

benma: Done

NickeZ (approved):
utACK

PR #1666, opened 2025-11-02
Port some keystore.c to Rust 

benma:


benma:
Will rebase this on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1669/ (HAL/Random) and use that instead of direct calls to bitbox02::random, getting rid of the ugly cfg(feature=testing) switch in the middle of the code.

NickeZ:
> Will rebase this on #1669 (HAL/Random) and use that instead of direct calls to bitbox02::random, getting rid of the ugly cfg(feature=testing) switch in the middle of the code.

You don't use stacked PRs? https://andrewlock.net/pushing-a-whole-stack-of-branches-with-a-single-git-command/

benma:
> > Will rebase this on #1669 (HAL/Random) and use that instead of direct calls to bitbox02::random, getting rid of the ugly cfg(feature=testing) switch in the middle of the code.
> 
> You don't use stacked PRs? https://andrewlock.net/pushing-a-whole-stack-of-branches-with-a-single-git-command/

Too long to read :smile: 

I rebased (will fix CI errors later) and could use it for bip39 unlock, but still temporarily need the inline mock inside retain_seed(), similar to the inline `#ifdef TESTING ... mock ... #endif` in keystore.c. Will be fixed soon enough though (in a separate PR).

benma:
This is ready, first commit is from https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1670

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -33,12 +33,84 @@ use bitcoin::hashes::{Hash, HashEngine, Hmac, HmacEngine, sha256, sha512};
 /// Length of a compressed secp256k1 pubkey.
 const EC_PUBLIC_KEY_LEN: usize = 33;
 
+/// aes256cbc-hmac cipher adds 16 bytes IV, 16 bytes padding, 32 bytes hmac.
+const ENCRYPTION_OVERHEAD: usize = 64;
+
+#[derive(Copy, Clone)]
+struct Buffer {
```
NickeZ: `Buffer` is very generic maybe `ReadOnlyBuffer`? Or some other name that tells me more about the purpose of it. I guess you wrote it because you wanted a "write once, read only" buffer?

NickeZ: If the only reason is to have a "fixed capacity inline vec" there are crates that implement that and also implements all the other "nice to haves" features of vecs. e.g. https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayVec.html

benma: Renamed it now.

I am hesitant  to add a dependency here, as our needs are very simple. I'll merge this for now, if you feel like the custom struct is not preferable, we can change it later.

PR #1668, opened 2025-11-03
ci: run unit tests on macos

NickeZ:


.github/workflows/ci-common.yml
```diff
@@ -108,6 +108,33 @@ jobs:
       - name: run ATECC608 unit tests
         run: (cd tools/atecc608; go test ./...)
 
+  unit-tests-macos:
```
benma: call it `rust-unit-tests-macos`?

.github/workflows/ci-common.yml
```diff
@@ -108,6 +108,33 @@ jobs:
...
+    steps:
+      - name: Clone the repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+          submodules: recursive
+
+      - name: Install system dependencies
+        run: brew install cbindgen bindgen cmocka hidapi
```
benma: hidapi should not be needed?

cmocka also not for Ruts unit tests?

NickeZ: yeah, cmocka shouldn't be a requirement. `hidapi` is needed to build the u2f unit tests. I suppose we could make `hidapi` an optional dependnecy, but then if `hidapi` is missing it will fail to build instead of fail to configure. Not sure if that is better...

NickeZ: pushed some commits to see if I can get rid of the cmocka dep at least.

benma (approved):

PR #1669, opened 2025-11-03
rust/tests: remove unneeded random::fake_reset calls and add Random to HAL

benma:


src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -33,10 +34,15 @@ pub trait Sd {
     async fn write_bin(&mut self, filename: &str, dir: &str, data: &[u8]) -> Result<(), ()>;
 }
 
+pub trait Random {
+    fn random_32_bytes(&mut self) -> Box<zeroize::Zeroizing<[u8; 32]>>;
```
NickeZ, line 38: Why boxing it up?

benma: because otherwise [u8;32] leaves copies on the stack everytime it moves. also it is just adopting the sig of the existig bitbox02::random::random_32_bytes function.

NickeZ: Ok, seems excessive to me, but if it already was like this we can keep it.

Maybe in the future we can refactor to something like this: https://docs.rs/rand/latest/rand/trait.Rng.html#method.fill

Feels wrong that the interface that generates random numbers also decides that it should be heap allocated and needs to be zeroized.

benma: Yeah but it's much more ergonomic this way :smile:

PR #1670, opened 2025-11-04
keystore: use HAL/Random in create_and_store_seed

benma:


benma:
Merged as part of https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1666/

PR #1672, opened 2025-11-06
rust: add zeroizing str to cstr util function

benma:
Not using CString as it did not look to me like it was actually preallocating for the null terminator, and reallocation would leave unzeroed copies.


NickeZ (approved):
utACK

PR #1676, opened 2025-11-07
Keystore unlock tests

benma:
Increase coverage


benma:
Merged, but GH didn't close this (I may have accidentally pushed a rebased commit or sth, but the diff is empty)

PR #1680, opened 2025-11-12
keystore: use slice::first_chunk

benma:
Small optimization, avoiding creating an intermediate subslice.

(saves 24 bytes of binary space too)

### DIFF
--- a/src/rust/bitbox02-rust/src/keystore.rs
+++ b/src/rust/bitbox02-rust/src/keystore.rs
@@ -86,8 +86,9 @@ impl RetainedEncryptedBuffer {
             &format!("{}_in", purpose),
             &format!("{}_out", purpose),
         )?;
-        let iv: [u8; 16] = random.random_32_bytes()[..16].try_into().unwrap();
-        let encrypted = bitbox_aes::encrypt_with_hmac(&iv, &encryption_key, data);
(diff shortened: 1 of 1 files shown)


src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -86,8 +86,9 @@ impl RetainedEncryptedBuffer {
             &format!("{}_in", purpose),
             &format!("{}_out", purpose),
         )?;
-        let iv: [u8; 16] = random.random_32_bytes()[..16].try_into().unwrap();
-        let encrypted = bitbox_aes::encrypt_with_hmac(&iv, &encryption_key, data);
+        let iv_rand = random.random_32_bytes();
+        let iv: &[u8; 16] = iv_rand.first_chunk::<16>().unwrap();
```
benma, line 90: Using 

    let iv: &[u8; 16] = &iv_rand[..16].try_into().unwrap();

also works, but results in a bigger binary than first_chunk, which is a bit perplexing :shrug:

NickeZ: I guess a subslice has to check bounds both left and right. whereas first chunk only checks the right one.

PR #1685, opened 2025-11-13
Port unlock

benma:


src/rust/bitbox02/src/keystore.rs
```diff
@@ -14,15 +14,10 @@
 
 extern crate alloc;
 
-use alloc::vec::Vec;
-
 use bitcoin::secp256k1::{All, Secp256k1};
 
-use bitbox02_sys::keystore_error_t;
```
NickeZ, line 21: Should keystore_error_t be removed from `build.rs` as well?

benma: Nice catch, removed. I wonder how many others are there that are not used anymore.

Issue #862, opened 2022-01-15
Feature: Hitting back during "keypad" input goes back to previous group of characters


hojdan:
Let's say I want to enter the letter a as a simple example.

1. I touch the group with a-h
2. I touch the next group that appears which is a-c
3. I then touch a

Let's say I click something wrong by accident at some point.

1. If I touch the next group i-q, touching the back button afterward will take me back to all the letters and then I can touch the correct group. No issue here.
2. If I touch the next group d-e, touching the back button afterward will take me all the way back to all the letters. Then I would need to first touch group a-h from all the letters then I can touch the correct group. I would prefer that touching the back button not take me back all the way but instead to just the previous group selection.
3. I click b by accident and thus b is entered. To correct my mistake I click the back button. No issue here.

benma:
Makes sense to me. @jadzeidan what do you think?

jadzeidan:
Yeah, also makes sense to me to go back to previous group. Ack from me.

c266:
Just wanted to bump this, would be a great improvement.

alessandro-opensource:
Hi! As mentioned in the PR I opened: https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1689#issue-3631273880

I built it in the container and tested it with the simulator, and it seems to behave as expected.

Would love any feedback!

PR #1689, opened 2025-11-16
src/ui/components: Added -Feature: Hitting back during "keypad" input goes back to previous group of characters #862-

alessandro-opensource:
Hi! I gave issue #862 a try.

I tested it by building in the container and using the graphical simulator, and it seems that now pressing the back button takes you to the previous group instead of all letters.

Unfortunately, I haven’t had a physical device yet, so I haven’t been able to try flashing it.

Would love your feedback! Thanks

### DIFF
--- a/src/ui/components/trinary_input_char.c
+++ b/src/ui/components/trinary_input_char.c
@@ -28,8 +28,17 @@ typedef struct {
 // excluding null terminator
 #define MAX_CHARS 33
 
+// Maximum depth for navigation stack
+#define MAX_NAVIGATION_DEPTH 4
+
 // Each of the three groups can occupy roughly a third of the width.
 static const UG_S16 _group_width = SCREEN_WIDTH / 3;
+
+// Stack entry for navigation history
+typedef struct {
+    char alphabet[MAX_CHARS + 1];
+    UG_S16 horiz_space;
+} navigation_stack_entry_t;
 
 typedef struct {
     void (*character_chosen_cb)(component_t*, char);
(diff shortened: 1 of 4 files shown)


NickeZ:
Tested, but I think I found an edge case. See the recording. It seems to jump back "two navigations" when I go back and forward.

edit: please use clang-format to format as well. I noticed CI failed on formatting.

https://github.com/user-attachments/assets/0a077828-ff2b-438f-adb6-d305acc679e8

alessandro-opensource:
Thank you for your feedback, I will definitely try to figure out what is wrong, will use clang-format as well :slightly_smiling_face:

alessandro-opensource:
Hi, I rebased the fork on master branch, added a commit that seems to fix the problem with the previous one and a clang-format commit as well, happy to receive your feedback :slightly_smiling_face:

NickeZ (approved):
Looks good to me, can you squash the commits?

@benma I think we want two approvals for external contributions right?

src/ui/components/trinary_input_char.c
```diff
@@ -86,8 +148,16 @@ static void _alphabet_selected(component_t* component, const char* alphabet)
...
+        /* Clear navigation stack when a character is chosen */
+        _navigation_stack_clear(data);
     } else {
         /* Select a sub-alphabet. */
+
+        /* Push current state to navigation stack before going deeper */
+        _navigation_stack_push(data, data->current_full_alphabet);
+
+        /* Store the new alphabet and navigate to it */
+        snprintf(data->current_full_alphabet, sizeof(data->current_full_alphabet), "%s", alphabet);
```
benma: This line is repeated inside the call below to trinary_input_char_set_alphabet, so it is not needed here, right?

Thinking further, maybe the `current_full_alphabet` var is not needed at all? Can't you just push the alphabet onto the stack directly in `trinary_input_char_set_alphabet()` instead of writing it to `current_full_alphabet`? (Not sure if possible, didn't think hard about it)

src/ui/components/trinary_input_char.c
```diff
@@ -344,10 +417,31 @@ component_t* trinary_input_char_create(
     component->dimension.height = SCREEN_HEIGHT;
 
     data->character_chosen_cb = character_chosen_cb;
+    // Initialize navigation stack
+    _navigation_stack_clear(data);
```
benma: Technically not needed right? `data` is already initialized to be all zeroes.

src/ui/components/trinary_input_char.c
```diff
@@ -40,9 +40,18 @@ typedef struct {
 // excluding null terminator
 #define MAX_CHARS 33
 
+// Maximum depth for navigation stack
+#define MAX_NAVIGATION_DEPTH 8
```
benma: 4 should be enough, it's the maximum number of taps needed to get to the char.

alessandro-opensource: ok, done

src/ui/components/trinary_input_char.c
alessandro-opensource: yes, removed in the new commit

src/ui/components/trinary_input_char.c
alessandro-opensource: yes, I think it is possible if we implement something like the `_rebuild_full_alphabet()` added in this last commit.

It's not exactly what you asked for, but it seemed fine this way: the pop from the stack is now only done in the `bool trinary_input_char_go_back(component_t* component)` and the push in the `static void _alphabet_selected(component_t* component, const char* alphabet)` only. In this way the new internal `_set_alphabet_internal` is dedicated to manipulate strings etc. and not pushing or popping from the stack nor clearing it up.

alessandro-opensource:
I squashed the commits as requested, but if it’s more convenient to review the recent additions in a separate commit, I can force-push that separate commit since I still have a local copy from before the squash. Thanks for the feedback again!

src/ui/components/trinary_input_char.c
alessandro-opensource: Overall, the bug you reported seems to be resolved with this approach.

src/ui/components/trinary_input_char.c
```diff
@@ -60,9 +69,98 @@ typedef struct {
...
+ * Returns false if stack is empty.
+ */
+static bool _navigation_stack_pop(data_t* data, char* out_alphabet, UG_S16* out_horiz_space)
+{
+    if (data->navigation_stack_size == 0) {
+        return false;
+    }
+
+    data->navigation_stack_size--;
+    navigation_stack_entry_t* entry = &data->navigation_stack[data->navigation_stack_size];
```
benma: can be `const`

src/ui/components/trinary_input_char.c
```diff
@@ -60,9 +69,98 @@ typedef struct {
...
+
+    to_copy = (m > total - off) ? (total - off) : m;
+    memcpy(out_buffer + off, data->middle_alphabet, to_copy);
+    off += to_copy;
+
+    to_copy = (r > total - off) ? (total - off) : r;
+    memcpy(out_buffer + off, data->right_alphabet, to_copy);
+    off += to_copy;
+
+    out_buffer[off] = '\0';
```
benma: This is a bit complicated. What would you think of this?

```C
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wformat-truncation"
    (void)snprintf(
        out_buffer,
        buffer_size,
        "%s%s%s",
        data->left_alphabet,
        data->middle_alphabet,
        data->right_alphabet);
#pragma GCC diagnostic pop
```

The warning is there because the three alphabets are all MAX_CHARS+1 in size, so the compiler thinks it will be truncated, but we know that the combination of all of them is still smaller then MAX_CHARS+1, so it's safe to ignore that warning.

benma:
@alessandro-opensource would you have time to address the [last remaining nit](https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1689#discussion_r2660839306)? Then after you sign your commit we can merge.

alessandro-opensource:
Hi @benma , sorry for the late reply, I must have missed the notification.
I will address this shortly, :slightly_smiling_face:  most likely over the weekend

alessandro-opensource:
Hi @benma, I updated the commit adding these latest requested changes and rebased on master. Looking forward to hearing from you

benma:
> Hi @benma, I updated the commit adding these latest requested changes and rebased on master. Looking forward to hearing from you

Hi @alessandro-opensource, thanks a lot, it's looking great :clap:  Could you gpg-sign the commit and add a CHANGELOG.md entry in the `[Unreleased]` section at the top? If there is trouble with gpg-signing or you'd prefer me to sign it, please let me know.

alessandro-opensource:
Hi @benma, :slightly_smiling_face:  I added the changelog entry and the gpg-signing. Is it common to sign it with a nickname? I can also sign with my complete name anyway.

benma:
> Hi @benma, 🙂 I added the changelog entry and the gpg-signing. Is it common to sign it with a nickname? I can also sign with my complete name anyway.

That's fine. Thanks a lot for this feature! And apologies that it took so long to process this PR, next time we'll try to do better :smile:

alessandro-opensource:
No worries at all, my pleasure! Hope to contribute again soon :slightly_smiling_face:

PR #1691, opened 2025-11-18
Memory functional tests

benma:


src/rust/bitbox02/src/memory.rs
```diff
@@ -326,4 +406,258 @@ mod tests {
...
+            *b"ffffffffffffffffffffffffffffffff",
+        ];
+        let names = ["name1", "name2", "name3", "name4", "name5", "name6"];
+
+        assert!(multisig_get_by_hash(&hashes[0]).is_none());
+
+        // set
+        assert!(multisig_set_by_hash(&hashes[0], names[0]).is_ok());
+        assert!(multisig_set_by_hash(&hashes[1], names[1]).is_ok());
+        assert!(multisig_set_by_hash(&hashes[1], names[1]).is_ok());
```
NickeZ, line 459: I think you want to keep the comment here that says "overwrite is possible", otherwise it looks like copy/paste error.

src/rust/bitbox02/src/memory.rs
```diff
@@ -326,4 +406,258 @@ mod tests {
...
+
+    unsafe extern "C" fn mcu_random_adapter(buf_out: *mut u8) {
+        let slice = unsafe { slice::from_raw_parts_mut(buf_out, 32) };
+        let array: &mut [u8; 32] = slice.try_into().unwrap();
+        random::mcu_32_bytes(array);
+    }
+
+    #[test]
+    fn test_memory_multisig() {
+        mock_memory();
```
benma, line 441: note to self, remove mock_memory or fake_memory_factoryreset, they are redundant and only one should be needed, here and elsewhere

NickeZ: I was thinking we could use some dependency for fixtures since you start almost every test the same way. But if it only is a single function call it probably doesn't make sense. https://crates.io/crates/rstest

src/rust/bitbox02/src/memory.rs
benma: Done

PR #1692, opened 2025-11-18
Add device password change functionality

cedwies:
Implements a new API endpoint to allow users to change their device password without re-entering the BIP39 passphrase.

Changes:
- Add ChangePasswordRequest message to proto definitions
- Implement change_password module in Rust API
- Refactor keystore to support password re-encryption:
  - Add re_encrypt_seed() function to change password while preserving BIP39 seed state
  - Extract encrypt_and_store_seed_internal() helper to share logic between initial setup and password change
- Add Python bindings for change_password workflow
- Add comprehensive tests including password verification
- Fix minor formatting issue in qtouch.c (clang-format)
- Regenerated protobuf files

The password change workflow:
1. Unlock device with current password
2. Prompt for new password (entered twice for confirmation)
3. Re-encrypt seed with new password while preserving BIP39 seed

This prevents users from having to re-enter their BIP39 passphrase after changing their password.

### DIFF
--- a/py/bitbox02/bitbox02/bitbox02/bitbox02.py
+++ b/py/bitbox02/bitbox02/bitbox02/bitbox02.py
@@ -212,22 +212,16 @@ class BitBox02(BitBoxCommonAPI):
             raise
         return True
 
-    def change_password(self) -> bool:
+    def change_password(self) -> None:
         """
-        Returns True if the user changes password successfully by
-        unlocking with old password and entering and confirming new password.
-        Returns False otherwise.
+        Changes the device password. The user must unlock with the old password
+        and enter and confirm the new password.
+        Raises a Bitbox02Exception on failure.
         """
         # pylint: disable=no-member
         request = hww.Request()
         request.change_password.CopyFrom(bitbox02_system.ChangePasswordRequest())
-        try:
-            self._msg_query(request, expected_response="success")
-        except Bitbox02Exception as err:
-            if err.code == ERR_GENERIC:
-                return False
-            raise
-        return True
+        self._msg_query(request, expected_response="success")
 
     def create_backup(self) -> bool:
         """
--- a/src/rust/bitbox02-rust/src/hww/api.rs
+++ b/src/rust/bitbox02-rust/src/hww/api.rs
@@ -139,7 +139,7 @@ fn can_call(request: &Request) -> bool {
         Request::SetPassword(_) | Request::RestoreBackup(_) | Request::RestoreFromMnemonic(_) => {
             matches!(state, State::Uninitialized | State::Seeded)
         }
-        Request::CreateBackup(_) | Request::ShowMnemonic(_) | Request::ChangePassword(_) => {
+        Request::CreateBackup(_) | Request::ShowMnemonic(_) => {
             matches!(state, State::Seeded | State::InitializedAndUnlocked)
         }
         Request::Fingerprint(_)
@@ -152,7 +152,8 @@ fn can_call(request: &Request) -> bool {
         | Request::Eth(_)
         | Request::Reset(_)
         | Request::Cardano(_)
-        | Request::Bip85(_) => {
+        | Request::Bip85(_)
+        | Request::ChangePassword(_) => {
             matches!(state, State::InitializedAndUnlocked)
         }
         // These are streamed asynchronously using the `next_request()` primitive in
--- a/src/rust/bitbox02-rust/src/hww/api/change_password.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/change_password.rs
@@ -5,22 +5,26 @@ use pb::response::Response;
 
 use crate::hal::Ui;
 use crate::keystore;
-use crate::workflow::{password, unlock};
+use crate::workflow::{confirm, password, unlock};
 
 pub async fn process(hal: &mut impl crate::hal::Hal) -> Result<Response, Error> {
-    // Must be initialized
-    if !bitbox02::memory::is_initialized() {
-        return Err(Error::Generic);
-    }
-
+    // Process confirmation and instruction for user
+    hal.ui()
+        .confirm(&confirm::Params {
+            title: "",
+            body: "Proceed to\nchange password?",
+            accept_is_nextarrow: true,
+            ..Default::default()
+        })
+        .await?;
     // Unlock with old password
     let seed = unlock::unlock_keystore(hal, "Unlock device", unlock::CanCancel::Yes).await?;
     // Enter and confirm new password
     let new_password = password::enter_twice(hal).await?;
 
     // Re-encrypt seed with new password
-    if keystore::re_encrypt_seed(hal, &seed, &new_password).is_err() {
-        hal.ui().status("Could not change\npassword", false).await;
+    if let Err(err) = keystore::re_encrypt_seed(hal, &seed, &new_password) {
+        hal.ui().status(&format!("Error\n{:?}", err), false).await;
         return Err(Error::Generic);
     }
 
@@ -73,7 +77,7 @@ mod tests {
             }
         }));
         // reset the chip counter
-        bitbox02::securechip::fake_event_counter_reset();
+        hal.securechip.event_counter_reset();
         // call process
         let result = block_on(process(&mut hal));
         // assert success
@@ -82,14 +86,21 @@ mod tests {
         let screens = hal.ui.screens.clone();
         assert_eq!(
             screens,
-            vec![Screen::Status {
-                title: "Success".into(),
-                success: true,
-            }]
+            vec![
+                Screen::Confirm {
+                    title: "".into(),
+                    body: "Proceed to\nchange password?".into(),
+                    longtouch: false,
+                },
+                Screen::Status {
+                    title: "Success".into(),
+                    success: true,
+                }
+            ]
         );
-        drop(hal);
 
-        let securechip_events = bitbox02::securechip::fake_event_counter();
+        let securechip_events = hal.securechip.get_event_counter();
+        drop(hal);
         // We expect 14 secure chip events. This is intentionally brittle to catch
         // unintended changes in the number of securechip operations during password change.
         // If this fails after a legitimate change, update the expected count.
@@ -113,18 +124,6 @@ mod tests {
         );
     }
 
-    // Test that we fail if the device is not initialized
-    #[test]
-    fn test_process_device_not_initialized() {
-        mock_memory();
-        let mut hal = TestingHal::new();
-        assert!(!bitbox02::memory::is_initialized());
-        bitbox02::securechip::fake_event_counter_reset();
-        assert_eq!(block_on(process(&mut hal)), Err(Error::Generic));
-        assert!(hal.ui.screens.is_empty());
-        assert_eq!(bitbox02::securechip::fake_event_counter(), 0);
-    }
-
     // Test that we fail if the unlock fails
     #[test]
     fn test_process_unlock_failure() {
@@ -146,19 +145,26 @@ mod tests {
             Ok("wrong_password".into())
         }));
 
-        bitbox02::securechip::fake_event_counter_reset();
+        hal.securechip.event_counter_reset();
         let result = block_on(process(&mut hal));
 
         assert_eq!(result, Err(Error::Generic));
         assert_eq!(
             hal.ui.screens,
-            vec![Screen::Status {
-                title: "Wrong password\n9 tries remain".into(),
-                success: false,
-            }]
+            vec![
+                Screen::Confirm {
+                    title: "".into(),
+                    body: "Proceed to\nchange password?".into(),
+                    longtouch: false,
+                },
+                Screen::Status {
+                    title: "Wrong password\n9 tries remain".into(),
+                    success: false,
+                }
+            ]
         );
         // We expect 5 secure chip events (sensitive to code changes)
-        assert_eq!(bitbox02::securechip::fake_event_counter(), 5);
+        assert_eq!(hal.securechip.get_event_counter(), 5);
 
         drop(hal);
         assert_eq!(prompt_counter, 1);
--- a/src/rust/bitbox02-rust/src/keystore.rs
+++ b/src/rust/bitbox02-rust/src/keystore.rs
@@ -42,6 +42,7 @@ const LONG_TIMEOUT: i16 = -70;
 
 #[derive(Debug)]
 pub enum Error {
+    InvalidState,
     CannotUnlockBIP39,
     IncorrectPassword,
     MaxAttemptsExceeded,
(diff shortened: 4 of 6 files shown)


py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -212,6 +212,23 @@ def set_password(self, entropy_size: int = 32) -> bool:
             raise
         return True
 
+    def change_password(self) -> bool:
+        """
+        Returns True if the user changes password successfully by
+        unlocking with old password and entering and confirming new password.
+        Returns False otherwise.
```
benma: Maybe better to simply return nothing on success and let the exception propagate on failure.

As right now, it returns False if they enter the wrong password, and an exception if cancelled or if there is another kind of error. Seems simpler to just do the same in all cases.

src/rust/bitbox02-rust/src/hww/api.rs
```diff
@@ -138,7 +139,7 @@ fn can_call(request: &Request) -> bool {
         Request::SetPassword(_) | Request::RestoreBackup(_) | Request::RestoreFromMnemonic(_) => {
             matches!(state, State::Uninitialized | State::Seeded)
         }
-        Request::CreateBackup(_) | Request::ShowMnemonic(_) => {
+        Request::CreateBackup(_) | Request::ShowMnemonic(_) | Request::ChangePassword(_) => {
```
benma: It should probably be in the arm below where it's only `InitializedAndUnlocked`. Here you also allow the call if it is not initialized (but seeded).

src/rust/bitbox02-rust/src/hww/api/change_password.rs
```diff
@@ -0,0 +1,223 @@
...
+
+use crate::hal::Ui;
+use crate::keystore;
+use crate::workflow::{password, unlock};
+
+pub async fn process(hal: &mut impl crate::hal::Hal) -> Result<Response, Error> {
+    // Must be initialized
+    if !bitbox02::memory::is_initialized() {
+        return Err(Error::Generic);
+    }
```
benma: Not needed if in api.rs you move the check in `can_call` so that it can only be called if initialized and unlocked.

src/rust/bitbox02-rust/src/hww/api/change_password.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    // Unlock with old password
+    let seed = unlock::unlock_keystore(hal, "Unlock device", unlock::CanCancel::Yes).await?;
+    // Enter and confirm new password
+    let new_password = password::enter_twice(hal).await?;
+
+    // Re-encrypt seed with new password
+    if keystore::re_encrypt_seed(hal, &seed, &new_password).is_err() {
+        hal.ui().status("Could not change\npassword", false).await;
```
benma: Elsewhere we use 


    hal.ui().status(&format!("Error\n{:?}", err), false).await;

Maybe that's better, it's not expected to ever be shown, but if it is, there will be some more info.

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -232,6 +230,48 @@ pub fn encrypt_and_store_seed(
...
+) -> Result<(), Error> {
+    if !bitbox02::memory::is_initialized() {
+        return Err(Error::Unseeded);
+    }
+
+    // Store the bip39 seed before re-encryption because:
+    // 1. The secure chip's internal keys are regenerated with the new password
+    // 2. encrypt_and_store_seed_internal calls lock() which clears RAM including BIP39 seed
+    // 3. We want to avoid forcing the user to re-enter their BIP39 passphrase
+    let bip39_seed = copy_bip39_seed().map_err(|_| Error::CannotUnlockBIP39)?;
```
benma: Wrong error, we are not in the process of unlocking bip39.

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -232,6 +230,48 @@ pub fn encrypt_and_store_seed(
...
+
+    encrypt_and_store_seed_internal(hal, seed, new_password)?;
+
+    // Re-retain the bip39 seed
+    RETAINED_BIP39_SEED.write(Some(RetainedEncryptedBuffer::from_buffer(
+        hal.random(),
+        bip39_seed.as_slice(),
+        "keystore_retained_bip39_seed_access",
+    )?));
```
benma: Best to make a `retain_bip39_seed` function similar to `retain_seed`

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -232,6 +230,48 @@ pub fn encrypt_and_store_seed(
...
+) -> Result<(), Error> {
+    if bitbox02::memory::is_initialized() {
+        return Err(Error::Memory);
+    }
+    // We are in setup phase, so we clear memory first to guarantee clean RAM
+    encrypt_and_store_seed_internal(hal, seed, password)
+}
+
+/// Re-encrypts the seed with a (new) password
+pub fn re_encrypt_seed(
```
benma, line 257: Would be nice to add some unit tests in this file too for this new function.

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -232,6 +230,48 @@ pub fn encrypt_and_store_seed(
...
+/// `password` is the password with which we encrypt the seed.
+pub fn encrypt_and_store_seed(
+    hal: &mut impl crate::hal::Hal,
+    seed: &[u8],
+    password: &str,
+) -> Result<(), Error> {
+    if bitbox02::memory::is_initialized() {
+        return Err(Error::Memory);
+    }
+    // We are in setup phase, so we clear memory first to guarantee clean RAM
```
benma: What does that mean?

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -232,6 +230,48 @@ pub fn encrypt_and_store_seed(
...
+
+/// Re-encrypts the seed with a (new) password
+pub fn re_encrypt_seed(
+    hal: &mut impl crate::hal::Hal,
+    seed: &[u8],
+    new_password: &str,
+) -> Result<(), Error> {
+    if !bitbox02::memory::is_initialized() {
+        return Err(Error::Unseeded);
+    }
```
benma: Fyi, seeded means an encrypted seed is stored, and initialized means it is seeded and a backup has been created.

In this case, `is_seeded()` is enough, because then there is an encrypted seed whose password we can change. And the `Unseeded` error will be correct too then :smile:

src/rust/bitbox02-rust/src/hww/api.rs
cedwies: As you mentioned already:

> In this case, is_seeded() is enough, because then there is an encrypted seed whose password we can change.

So we should be able to call `ChangePassword` when seeded, no?

src/rust/bitbox02-rust/src/keystore.rs
cedwies: Thanks, I'd suggest `Error::Memory` (?)

src/rust/bitbox02-rust/src/keystore.rs
cedwies: Thanks, was an outdated comment. In an earlier approach I only called `lock()` in `encrypt_and_store_seed(...)`. But I later moved it it into `encrypt_and_store_seed_internal(...)` so RAM is cleared both in `encrypt_and_store_seed(...)` and in `re_encrypt_seed(...)`

src/rust/bitbox02-rust/src/hww/api.rs
benma: Technically I guess you are right, but practically we will expose the functionality in the BitBoxApp under 'Manage device' after it's been initialized. I don't think we'll add a change password function right after setting the password, before creating a backup, in the setup wizard. If one wants a new password then, just reconnecting the device will do the trick too, as then the setup starts from scratch.

Long story short, imho you can enforce that it is initialized and unlocked here.

benma: As for

> In this case, is_seeded() is enough, be

That's true, in that case you could leave the check that it is initialized. It does not matter that much, as long as it's seeded and unlocked the function works (so is_seeded() is the minimum check). To me, it makes a bit more sense to keep the minimum check (seeded) in the keystore module, while doing the more strict check in the API here to decide when the API call can be made. No strong opinion though.

src/rust/bitbox02-rust/src/keystore.rs
benma: In other funcs we just use `copy_bip39_seed()?`, which propagates as Generic error.

py/bitbox02/bitbox02/bitbox02/bitbox02.py
cedwies: I suggest a similar approach compared to `show_mnemonic`:

```py
    def change_password(self) -> bool:
        """
        Returns True if the user changes password successfully by
        unlocking with old password and entering and confirming new password.
        Raises a Bitbox02Exception on failure.
        """
        # pylint: disable=no-member
        request = hww.Request()
        request.change_password.CopyFrom(bitbox02_system.ChangePasswordRequest())
```

src/rust/bitbox02-rust/src/hww/api/change_password.rs
```diff
@@ -0,0 +1,206 @@
...
+use crate::pb;
+
+use pb::response::Response;
+
+use crate::hal::Ui;
+use crate::keystore;
+use crate::workflow::{password, unlock};
+
+pub async fn process(hal: &mut impl crate::hal::Hal) -> Result<Response, Error> {
+    // Unlock with old password
```
benma, line 20: How about a "Change password" confirmation screen?

benma (changes requested):
tested now, found a small bug: after changing password, the root fingerprint api call (send_message entry number 3) does not work anymore. The reason is that it uses the cached `ROOT_FINGERPRINT`, but it is wiped on `lock()`.

Maybe the root fingerprint computation should move moved out of `bip39::derive_seed()` and done inside `retain_bip39_seed()` instead.

src/rust/bitbox02-rust/src/keystore.rs
benma: I'd stick to just using `?` here and not map to `Error::Memory`, because it is not clear there was an issue with the memory functions. By a quick look, `Memory` is not even a possibility here, but `Decrypt` and `Salt` errors are. `()` is preferable to me because it's not misleading.

cedwies:
> tested now, found a small bug: after changing password, the root fingerprint api call (send_message entry number 3) does not work anymore. The reason is that it uses the cached `ROOT_FINGERPRINT`, but it is wiped on `lock()`.
> 
> Maybe the root fingerprint computation should move moved out of `bip39::derive_seed()` and done inside `retain_bip39_seed()` instead.

But I am not sure if we should do this in this PR. This would require additional sensitive changes (including testing, etc). E.g. the fn `unlock_bip39()` in `keystore.rs` would need to be changed significantly.

Suggestion 1: Save the root_fingerprint similarly to the bip_39 seed.
<details>

```rust
pub fn re_encrypt_seed(
    hal: &mut impl crate::hal::Hal,
    seed: &[u8],
    new_password: &str,
) -> Result<(), Error> {
    if !bitbox02::memory::is_seeded() {
        return Err(Error::Unseeded);
    }

    // Store the bip39 seed before re-encryption because:
    // 1. The secure chip's internal keys are regenerated with the new password
    // 2. encrypt_and_store_seed_internal calls lock() which clears RAM including BIP39 seed
    // 3. We want to avoid forcing the user to re-enter their BIP39 passphrase
    let bip39_seed = copy_bip39_seed().map_err(|_| Error::Memory)?;
    let root_fingerprint = ROOT_FINGERPRINT.read().ok_or(()).map_err(|_| Error::Memory)?;

    encrypt_and_store_seed_internal(hal, seed, new_password)?;

    // Re-retain the bip39 seed and root fingerprint
    retain_bip39_seed(hal.random(), bip39_seed.as_slice())?;
    ROOT_FINGERPRINT.write(Some(root_fingerprint));

    Ok(())
}
```

</details>

Suggestion 2: **Additionally** calculate the `root_fingerprint` in `retain_bip39_seed()`. However, this would cause doing the calculation more often than needed (e.g. twice when we call `unlock_bip39()`)

I think going with option 1 is the best way, and then potentially a refactor of `bip39::derive_seed()` in a separate PR. Happy to hear your thoughts.

benma:
> I think going with option 1 is the best way,

sounds good :+1: (but let's not use the Memory error, that error should be for when there was an error reading memory, which is not the case here)

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -838,6 +883,193 @@ mod tests {
...
+                lock();
+                unlock(&mut mock_hal, prev_password).unwrap();
+            }
+
+            let mut random = crate::hal::testing::TestingRandom::new();
+            assert!(block_on(unlock_bip39(&mut random, &seed, "", async || {})).is_ok());
+            assert!(re_encrypt_seed(&mut mock_hal, &seed, new_password).is_ok());
+
+            // Seed should remain the same
+            assert_eq!(copy_seed().unwrap().as_slice(), seed_copy1.as_slice());
```
benma: Could you do this also for the bip39 seed and the root fingerprint?

Then you could also remove the `test_re_encrypt_seed_preserves_bip39_seed` test?

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -838,6 +883,193 @@ mod tests {
...
+        for (i, new_password) in ["password2", "password3", "password4"].iter().enumerate() {
+            // Need to be unlocked with BIP39 before re-encrypting
+            if i > 0 {
+                let prev_password = match i {
+                    1 => "password2",
+                    2 => "password3",
+                    _ => unreachable!(),
+                };
+                lock();
+                unlock(&mut mock_hal, prev_password).unwrap();
```
benma: Why this? Isn't it already unlocked?

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -838,6 +883,193 @@ mod tests {
...
+            }
+
+            let mut random = crate::hal::testing::TestingRandom::new();
+            assert!(block_on(unlock_bip39(&mut random, &seed, "", async || {})).is_ok());
```
benma: Feels like a test to change passwords multiple times should not manually lock/unlock/unlock-bip39 every time, or am I missing something?

benma (changes requested):

src/rust/bitbox02-rust/src/keystore.rs
cedwies: This is not necessary, however, it served as a testing purpose. It verified that you could actually decrypt with the previous password. But reconsidering, since we test it in a separate unit test (`test_re_encrypt_seed_changes_password()`), it should not be here.

src/rust/bitbox02-rust/src/keystore.rs
cedwies: No, you are correct.

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -838,6 +880,157 @@ mod tests {
...
+    fn test_re_encrypt_seed_preserves_seeds_and_fingerprint() {
+        mock_memory();
+        lock();
+
+        let mut mock_hal = TestingHal::new();
+        let seed = hex!("cb33c20cea62a5c277527e2002da82e6e2b37450a755143a540a54cea8da9044");
+
+        // Initial setup
+        assert!(encrypt_and_store_seed(&mut mock_hal, &seed, "password1").is_ok());
+        let seed_copy1 = copy_seed().unwrap();
```
benma: tiny nit: omve this down the the other two (bip39_seed and root_fingerprint) and name then consistently

cedwies: done :)

PR #1697, opened 2025-11-26
Port rest of component wrappers

NickeZ:
Port component wrappers to use Futures instead of callbacks


src/rust/bitbox02-rust/src/bb02_async.rs
```diff
@@ -12,12 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use core::cell::RefCell;
-
-/// Disables the screensaver while waiting for an option to contain a value. Afterwards, it returns that value
-pub async fn option_no_screensaver<O>(opt: &RefCell<Option<O>>) -> O {
+pub async fn screensaver_without<T>(fut: impl Future<Output = T>) -> T {
```
benma, line 15: This might case a binary blowup because of monomorphism, and dynamic dispatch might be better.

Or alternatively, there could be a "no screensaver" guard struct, that disables screensaver in `new()` and enables it on drop, and one can just do `let _no_screensaver = NoScreensaver::new()` to disable screensaver for the duration of the function.

NickeZ: It is a function with like 3 lines in it, so I don't think it has a big impact. I would probably also go with a "guard" that turns on the screensaver when dropped before going with dynamic dispatch. Dynamic dispatch requires the caller to box the future.

src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -48,116 +49,159 @@ impl Component<'_> {
...
+        waker: Option<Waker>,
+        result: Option<Result<zeroize::Zeroizing<String>, ()>>,
+    }
+    let shared_state = Arc::new(RefCell::new(SharedState {
+        waker: None,
+        result: None,
+    }));
+
+    unsafe extern "C" fn cancel_cb(user_data: *mut c_void) {
+        let shared_state: Arc<RefCell<SharedState>> = unsafe { Arc::from_raw(user_data as *mut _) };
```
benma, line 79: trinary_input_string.c enforces that confirm_cb can only be called once, but not for cancel_cb. That's why the removed/previous cancel_cb had `        let callback = user_data as *mut ContinueCancelCb;` and not `Box::from_raw()`, and the `on_drop` part dropped the data.

Here you don't cast but do `::from_raw()`, which must be paired with an `into_raw()`, so this assumes cancel_cb can only be called once. Seems fine because we drop the component on cancel, but the C code should enforce it for clarity.

src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -48,116 +49,159 @@ impl Component<'_> {
...
     let component = unsafe {
         bitbox02_sys::trinary_input_string_create(
             &params.to_c_params(&mut title_scratch).data, // title copied in C
-            Some(c_confirm_callback::<F>),
-            // passed to c_confirm_callback as `user_data`.
-            Box::into_raw(Box::new(confirm_callback)) as *mut _,
-            cancel_cb,
-            cancel_user_data,
+            Some(confirm_cb),
+            Arc::into_raw(Arc::clone(&shared_state)) as *mut _, // passed to confirm_cb as `user_data`.
```
benma, line 115: There are two `Arc::into_raw(Arc::clone())` here (one for cancel, one for confirm), but only one of them is brought back properly with a `from_raw()`, the other is leaked.

I guess the shared state should be leaked with `into_raw()` only once and passed to both callbacks, as only one of them will be called.

Or `trinary_input_string_create` could be changed to take one `user_data` arg and use it for both callbacks. That would probably be the easiest / clearest.

benma: Fixed in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1781

src/rust/bitbox02/src/ui/ui.rs
benma: Fixed in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1781

src/rust/bitbox02-rust/src/workflow/cancel.rs
```diff
@@ -12,55 +12,38 @@
...
-/// * `component` - component to process
-/// * `result_cell` - result var to synchronize the result on. Pass the same to `cancel` and
-///   `set_result`.
-pub async fn with_cancel<R>(
-    title: &str,
-    component: &mut bitbox02::ui::Component<'_>,
-    result_cell: &ResultCell<R>,
-) -> Result<R, Error> {
-    component.screen_stack_push();
+pub async fn with_cancel<F, T, E, GEN>(title: &str, future_generator: GEN) -> Result<T, Error>
```
benma, line 34: This rewrite is incorrect. Before, the cancel confirm prompt is added as a component on top of the still-running wrapped component (e.g. showing recovery words), so when rejecting the cancel prompt, you return where you were, e.g. at word number 22 of the recovery words.

With this rewrite, the underlying UI future is dropped and restarted, so customer starts back at the beginning.

benma: Fixed in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1788

benma:
I believe all of this has now been split into individual PRs and merged, so I'm closing this.

PR #1698, opened 2025-11-26
pip: fix editable imports for pip >= 25

cedwies:
Pip versions >=25 changed editable install behavior per PEP 660, causing relative imports like "from .bitbox02 import common" to resolve to the outer init.py instead of the inner subpackage, breaking dev workflows. 

Re-exporting symbols from the nested bitbox02 module fixes this resolution issue while preserving the existing package structure and API for downstream users. This ensures pip install -e . works reliably without flattening.

With pip 25 and before this change the following command (from repo root) does not work:

`pip uninstall -y bitbox02 && pip install -e py/bitbox02 && python py/send_message.py` will return an import error.

With pip 25 and after this change, the same command will not return an import error.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -385,6 +385,10 @@ docker exec -u 0 -it bitbox02-firmware-dev bash -c 'apt update && apt install -y
 
 There is a Python api library in `py/bitbox02`.
 
+> [!IMPORTANT]  
+> The Python scripts and editable installs require **pip ≥ 25**. Older pip versions will fail.  
+> For setup and usage instructions, see [`py/README.md`](py/README.md).
+
 ### BitBox02 CLI client
 
 Run `pip install -r py/requirements.txt` to install the deps (virtualenv recommended).
--- a/py/README.md
+++ b/py/README.md
@@ -1,9 +1,18 @@
 # Python scripts
 
-To use the scripts (`send_message.py`, `load_firmware.py` for example) go into the `bitbox02`
-directory and run `pip3 install .`.
+**Important: pip >= 25 is required.**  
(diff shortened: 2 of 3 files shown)


NickeZ:
Does this change also impact pip versions < 25 or did you find a way that works in both versions?

cedwies:
> Does this change also impact pip versions < 25 or did you find a way that works in both versions?

This change breaks compatibility with pip < 25. Unfortunately I was not able to find a solution to ensure compatibility with both pip < 25 and pip >= 25. I stopped before trying to make the Re-Export conditional, as I think the benefits are minor, while the complexity can increase drastically. Happy to discuss though, if you disagree.

NickeZ:
> This change breaks compatibility with pip < 25. Unfortunately I was not able to find a solution to ensure compatibility with both pip < 25 and pip >= 25. I stopped before trying to make the Re-Export conditional, as I think the benefits are minor, while the complexity can increase drastically. Happy to discuss though, if you disagree.

Sounds good to me, can you then add that as a note where we mention `pip install -e` in the docs? That we only support pip >= 25.

cedwies:
> > This change breaks compatibility with pip < 25. Unfortunately I was not able to find a solution to ensure compatibility with both pip < 25 and pip >= 25. I stopped before trying to make the Re-Export conditional, as I think the benefits are minor, while the complexity can increase drastically. Happy to discuss though, if you disagree.
> 
> Sounds good to me, can you then add that as a note where we mention `pip install -e` in the docs? That we only support pip >= 25.

Let me know if you agree to the modified `BUILD.md`

NickeZ:
> Let me know if you agree to the modified BUILD.md

Discussed over call that we should modify py/README.md as well.

PR #1700, opened 2025-11-29
Hal memory

benma:


src/rust/bitbox02-rust/src/hww.rs
```diff
@@ -600,6 +600,9 @@ mod tests {
                 }]
             );
 
+            // After factory reset, the device memory is reinitialized.
+            // TODO: can remove this once memory resetting is also in the memory HAL.
```
benma: Will fix this after https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1701

src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -419,11 +514,111 @@ pub mod testing {
...
+            Ok(())
+        }
+
+        fn get_encrypted_seed_and_hmac(&mut self) -> Result<alloc::vec::Vec<u8>, ()> {
+            self.encrypted_seed_and_hmac.clone().ok_or(())
+        }
+
+        fn set_encrypted_seed_and_hmac(&mut self, data: &[u8]) -> Result<(), ()> {
+            // 96 is the max space allocated in BitBox02's memory for this.
+            if data.len() > 96 {
```
cedwies, line 597: Nit: Depending on whether this could change in the future: should this be tied to a shared constant (e.g. exposed from bitbox02::memory) to avoid future drift?

src/rust/bitbox02-rust/src/hww/api/backup.rs
```diff
@@ -394,12 +386,18 @@ mod tests {
         );
 
         // Create another backup.
+
+        // Mock a reset, otherwise we can't make another backup, as it would ask for a password
+        // because the above backup creation set the initialized flag.
+        mock_hal.memory = crate::hal::testing::TestingMemory::new();
```
cedwies: Is there a reason you decided not to use HAL's `reset()` method?

src/rust/bitbox02-rust/src/hww/api/restore.rs
```diff
@@ -83,10 +83,10 @@ pub async fn from_file(
         let _ = bitbox02::securechip::u2f_counter_set(request.timestamp);
```
cedwies: Not scope of this PR: shouldn't this by now use HAL?

src/rust/bitbox02-rust/src/hww/api/set_mnemonic_passphrase_enabled.rs
```diff
@@ -72,7 +76,7 @@ mod tests {
             }],
         );
 
-        assert!(bitbox02::memory::is_mnemonic_passphrase_enabled());
+        assert!(mock_hal.memory.is_mnemonic_passphrase_enabled());
         // Disable:
         let mut mock_hal = TestingHal::new();
```
cedwies: We are testing **false -> false** here rather than a **true -> false** transition, no? Not wrong, given the implementation is fairly trivial, but if we want to mirror real behavior we could reuse the same TestingHal instance for enable‑then‑disable.

src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
```diff
@@ -90,9 +90,8 @@ mod tests {
         )
         .unwrap();
 
-        assert!(!bitbox02::memory::is_initialized());
-
         let mut mock_hal = TestingHal::new();
+        assert!(!mock_hal.memory.is_initialized());
```
cedwies, line 94: We assert `!mock_hal.memory.is_initialized()` to a different `TestingHal`. This just checks the default initialized state of a fresh test HAL (which may be your main goal here?) but we don't verify that `encrypt_and_store_seed()` does **not** set the `initialized` flag. Is this intended?

cedwies (commented):
Looks great! I don't see any major issues. Left some minor comments.

src/rust/bitbox02-rust/src/hww/api/restore.rs
benma: Indeed! Fixed in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1705

src/rust/bitbox02-rust/src/hal.rs
benma: Not sure, this crate will eventually uncouple from bitbox02, so TestingHal should not depend on it. Not quite sure what the best way is to handle it.

src/rust/bitbox02-rust/src/hww/api/show_mnemonic.rs
benma: Very good catch again! Fixed.

src/rust/bitbox02-rust/src/hww/api/set_mnemonic_passphrase_enabled.rs
benma: Good catch, fixed!

src/rust/bitbox02-rust/src/hww/api/backup.rs
benma: That method didn't exist yet at this commit, was added in a later commit. Changed now.

PR #1701, opened 2025-11-30
port reset_reset to Rust

benma:


src/rust/bitbox02-rust/src/reset.rs
```diff
@@ -0,0 +1,80 @@
...
+    crate::keystore::lock();
+
+    // Resetting takes longer than the default 500 ms watchdog. Bump the watchdog timeout to roughly
+    // 7 seconds (longer than needed) so we don't assume communication was lost mid-unlock.
+    const LONG_TIMEOUT: i16 = -70;
+    bitbox02::usb_processing::timeout_reset(LONG_TIMEOUT);
+
+    // Reset secure chip keys and U2F counter with retries, like in C reset_reset().
+    let mut reset_ok = false;
+    for _ in 0..5 {
```
cedwies, line 37: Even if translated from C `reset_reset()`, could we add a short comment why we try 5 times? I.e. that securechip ops can intermittently fail? (I assume that's the reason)

src/rust/bitbox02-rust/src/reset.rs
```diff
@@ -0,0 +1,80 @@
...
+/// - Resets MCU flash app memory.
+/// - Resets SmartEEPROM memory.
+/// - Shows a "Device reset" status message.
+///
+/// `status` selects whether the status message indicates success or failure (user invoked vs forced).
+pub(crate) async fn reset(hal: &mut impl crate::hal::Hal, status: bool) {
+    crate::keystore::lock();
+
+    // Resetting takes longer than the default 500 ms watchdog. Bump the watchdog timeout to roughly
+    // 7 seconds (longer than needed) so we don't assume communication was lost mid-unlock.
```
cedwies: "mid-unlock" or "mid-reset?"

src/rust/bitbox02-rust/src/reset.rs
```diff
@@ -0,0 +1,80 @@
...
+// limitations under the License.
+
+use crate::general::abort;
+use crate::hal::{SecureChip, Ui};
+
+/// Resets the device:
+/// - Updates secure chip KDF keys.
+/// - Resets the securechip EEPROM (U2F counter).
+/// - Resets MCU flash app memory.
+/// - Resets SmartEEPROM memory.
```
cedwies: "Disables SmartEEPROM memory"? or more detailed: "Disables SmartEEPROM memory so it will be erased on next boot"

src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -371,6 +383,16 @@ pub mod testing {
...
         }
+
+        fn reset_keys(&mut self) -> Result<(), ()> {
+            self.event_counter += 1;
+            Ok(())
+        }
+
+        #[cfg(feature = "app-u2f")]
+        fn u2f_counter_set(&mut self, _counter: u32) -> Result<(), ()> {
+            Ok(())
```
cedwies, line 418: why no `u2f_counter` like in `securechip_u2f_counter_set(uint32_t counter)`?

src/rust/bitbox02-rust/src/reset.rs
benma: Added :+1: 

This comment is AI slop :sob: I should have caught it and removed it. There is no point in mentioning C reset_reset() seeing it does not exist anymore :joy:

src/rust/bitbox02-rust/src/reset.rs
benma: Good catch. Typo as copied from keystore.rs. Reworded.

src/rust/bitbox02-rust/src/reset.rs
benma: Nice find! Fixed wording.

src/rust/bitbox02-rust/src/hal.rs
benma: Because there is no function there that reads/uses it yet.

benma:
@cedwies thanks for the comments. I addressed them, please check.

benma:
> Thanks for addressing the comments! Looks great.
> 
> Direct, focused tests of crate::reset::reset would be nice because this is a sensitive path (factory reset/auto wipe on too many failed attempts) and the control flow is new Rust glue even if underlying functions are old. However, since I see no regression versus C, I am happy to approve if we want to add tests in separate, focused PR.

Good idea, added. Please check.

src/rust/bitbox02-rust/src/reset.rs
```diff
@@ -0,0 +1,153 @@
...
+        bitbox02::memory::set_device_name("Custom name").unwrap();
+        assert!(!keystore::is_locked());
+        assert!(bitbox02::smarteeprom::is_enabled());
+
+        let mut hal = TestingHal::new();
+        // Make the reset keys call fail once, to test that it is retried.
+        hal.securechip.mock_reset_keys_fails();
+
+        // Simulate a non-zero U2F counter before reset.
+        SecureChip::u2f_counter_set(&mut hal.securechip, 42).unwrap();
```
cedwies, line 110: This is not gated behind `#[cfg(feature = "app-u2f")]`. Can we assume tests are always run with u2f enabled?
Same for `assert_eq!(hal.securechip.get_u2f_counter(), 0);`

src/rust/bitbox02-rust/src/hal.rs
cedwies: Can we fix the typo?

src/rust/bitbox02-rust/src/reset.rs
benma: Yes, in testing all features are enabled.

src/rust/bitbox02-rust/src/hal.rs
benma: Done, thanks.

PR #1703, opened 2025-12-01
bitbox02/memory: add unit test coverage

benma:
We will be adding all of these to a Memory HAL, so they will not be indirectly tested anymore by all the tests in bitbox02-rust. We add unit tests for them here, which serve as unit tests for the BitBox02 C memory functions.

### DIFF
--- a/src/rust/bitbox02/src/memory.rs
+++ b/src/rust/bitbox02/src/memory.rs
@@ -406,6 +406,172 @@ mod tests {
         set_salt_root(original.as_slice().try_into().unwrap()).unwrap();
     }
 
+    #[test]
+    fn test_get_set_device_name() {
+        mock_memory();
+
+        let original = get_device_name();
+        assert_eq!(original, "My BitBox");
+
+        let new_name = "Test device name";
+
+        set_device_name(new_name).unwrap();
+        assert_eq!(get_device_name(), new_name);
+
+        // A name with the maximum allowed length is accepted.
+        let max_len_name = "DeviceName_ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxy";
+        assert_eq!(max_len_name.len(), DEVICE_NAME_MAX_LEN);
+        set_device_name(max_len_name).unwrap();
+        assert_eq!(get_device_name(), max_len_name);
+
+        // Longer names are truncated to the maximum length.
+        let long_name = format!("{max_len_name}foobar");
+        set_device_name(&long_name).unwrap();
+        let stored = get_device_name();
+        assert_eq!(stored, max_len_name);
+
+        // Invalid names are rejected and must not change the stored value.
+        let invalid_names = [
+            "",         // empty
+            " name",    // leading space
+            "name ",    // trailing space
+            "foo\nbar", // control character
+            "Ä",        // non-ASCII
+            "漢字",     // non-ASCII
+        ];
+
+        for invalid in invalid_names {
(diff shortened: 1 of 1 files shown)


src/rust/bitbox02/src/memory.rs
```diff
@@ -406,6 +406,132 @@ mod tests {
...
+        assert!(!is_initialized());
+
+        set_initialized().unwrap();
+        assert!(is_initialized());
+    }
+
+    #[test]
+    fn test_mnemonic_passphrase_enabled_roundtrip() {
+        mock_memory();
+
```
cedwies, line 469: Very nitpicky but: We could `assert!(!is_mnemonic_passphrase_enabled());` after `mock_memory()` so in case the default changes to true (even though it should never), the test catches that we don't have a round trip anymore. `test_seed_birthdate_roundtrip()` also checks the default.

src/rust/bitbox02/src/memory.rs
```diff
@@ -406,6 +406,132 @@ mod tests {
...
+    #[test]
+    fn test_seed_birthdate_roundtrip() {
+        mock_memory();
+
+        assert_eq!(get_seed_birthdate(), 0);
+
+        let timestamp: u32 = 0xABCDEF11;
+        set_seed_birthdate(timestamp).unwrap();
+        assert_eq!(get_seed_birthdate(), timestamp);
```
cedwies, line 487: Very nitpicky: To make a roundtrip we would have to set back to 0 after setting to `0xABCDEF11`. 
Similarly to `test_get_set_device_name()` where we tested to overwrite an already written value.

src/rust/bitbox02/src/memory.rs
```diff
@@ -406,6 +406,132 @@ mod tests {
         set_salt_root(original.as_slice().try_into().unwrap()).unwrap();
     }
 
+    #[test]
+    fn test_get_set_device_name() {
```
cedwies, line 410: Should we test the behavior for overlong names and empty string? (i.e. expecting `Ok(())` for overlong and `Err(error)` for empty string). Maybe even non printable ASCII?

benma: I wonder a bit the checks (beside name length) should even be in the memory layer :shrug: 

Anyway, added tests for now.

src/rust/bitbox02/src/memory.rs
benma: Instead I chose to remove that part from `test_get_set_device_name()`. It serves no purpose to set it back to the original.

src/rust/bitbox02/src/memory.rs
benma: Done

benma:
Addressed all comments, please check.

src/rust/bitbox02/src/memory.rs
cedwies: one could now argue if `test_seed_birthdate_roundtrip()` is the correct function name here

src/rust/bitbox02/src/memory.rs
cedwies: Thanks, this looks great!

src/rust/bitbox02/src/memory.rs
benma: roundtrip refers to set->get and getting back the same value :thinking: at least that's how I interpret it. I can rename though if you feel strongly about it.

PR #1705, opened 2025-12-07
hal: use u2f_counter_set

benma:
Somehow was missed when it was added HAL.


src/rust/bitbox02-rust/src/reset.rs
cedwies: Unused import :)

PR #1707, opened 2025-12-09
bug: Prevent memory_reset_hww from masking flash write failures

cedwies:
memory_reset_hww combined the CHUNK_1 write result and the shared BLE chunk write using |=, so a later successful write could hide an earlier failure. Change both writes to be checked individually and return false if any of the two write operations fails, but ensuring both operations are attempted.

What could go wrong in the old code:

when `_write_chunk(CHUNK_1, chunk.bytes);` return `false` and `_write_to_address(FLASH_SHARED_DATA_START, 0, chunk_shared.bytes)` returns true, then `bool memory_reset_hww(void)` returns true due to the `|=` operator.

### DIFF
--- a/src/memory/memory.c
+++ b/src/memory/memory.c
@@ -371,7 +371,6 @@ bool memory_reset_hww(void)
     }
 
     // Initialize hww memory
-
     chunk_1_t chunk = {0};
     CLEANUP_CHUNK(chunk);
     _read_chunk(CHUNK_1, chunk_bytes);
@@ -405,7 +404,7 @@ bool memory_reset_hww(void)
         chunk_shared.fields.ble_identity_address[0] |= 0xc;
 
         memset(&chunk_shared.fields.ble_bond_db, 0xff, sizeof(chunk_shared.fields.ble_bond_db));
-        res |= _write_to_address(FLASH_SHARED_DATA_START, 0, chunk_shared.bytes);
+        res = _write_to_address(FLASH_SHARED_DATA_START, 0, chunk_shared.bytes) && res;
     }
 
     return res;


src/memory/memory.c
```diff
@@ -382,7 +381,9 @@ bool memory_reset_hww(void)
     // Set a new noise static private key.
     rust_noise_generate_static_private_key(rust_util_bytes_mut(
         chunk.fields.noise_static_private_key, sizeof(chunk.fields.noise_static_private_key)));
-    bool res = _write_chunk(CHUNK_1, chunk.bytes);
+    if (!_write_chunk(CHUNK_1, chunk.bytes)) {
```
benma: I guess the original intention was to run the below BLE stuff even if this fails, so to preserve this, just assign `bool res1` and `bool res2` and return `res1 && res2`?

benma (changes requested):
Nice catch!

NickeZ:
@codex review

src/memory/memory.c
```diff
@@ -405,7 +404,7 @@ bool memory_reset_hww(void)
         chunk_shared.fields.ble_identity_address[0] |= 0xc;
 
         memset(&chunk_shared.fields.ble_bond_db, 0xff, sizeof(chunk_shared.fields.ble_bond_db));
-        res |= _write_to_address(FLASH_SHARED_DATA_START, 0, chunk_shared.bytes);
+        res = _write_to_address(FLASH_SHARED_DATA_START, 0, chunk_shared.bytes) && res;
```
benma, line 407: relevant: https://c0x.shape-of-code.com/6.5.13.html 

first operand always executes

PR #1709, opened 2025-12-12
license: replace Shift Apache headers

benma:
Replace the existing Shift Crypto/Shift Devices Apache-2.0 boilerplate header blocks with the standard SPDX License Identifier tag: "SPDX-License-Identifier: Apache-2.0".

This is better than the previous multi-line text header because it is machine-readable, unambiguous, and widely understood by tooling (scanners, compliance automation), while keeping files smaller and reducing header churn.

### DIFF
--- a/messages/CMakeLists.txt
+++ b/messages/CMakeLists.txt
@@ -1,3 +1,5 @@
+# SPDX-License-Identifier: Apache-2.0
+
 set(PROTO_FILES
   hww.proto
   backup.proto
--- a/messages/keystore.proto
+++ b/messages/keystore.proto
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+
 // This function can be used to get an identifying xpub at the keypath m/4541509'/1112098098'"
 // The keypath argument has to be m/4541509'/1112098098'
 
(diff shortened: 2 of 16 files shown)


src/qtouch/qtouch.c
```diff
@@ -481,7 +481,7 @@ void qtouch_process(void)
         }
 
         /* Reset the flags for node_level_post_processing */
-        p_qtm_control->binding_layer_flags &= (uint8_t) ~(1U << node_pp_request);
+        p_qtm_control->binding_layer_flags &= (uint8_t)~(1U << node_pp_request);
```
cedwies: Something is off in this file

LICENSE
```diff
@@ -1,4 +1,5 @@
 Copyright 2019 Shift Cryptosecurity AG, Switzerland. All rights reserved.
+Copyright 2020-2025 Shift Crypto AG, Switzerland. All rights reserved.
```
cedwies, line 2: I assume the top line should be removed

benma: Don't think so, copyrights are extended, not replaced.

src/qtouch/qtouch.c
benma: Strange! Fixed

benma:
Rebased and fixed some new clang-tidy warnings, please check :pray: 

Also updated the copyright in the py/bitbox02 LICENSE

cedwies (approved):
ACK 👍

PR #1711, opened 2025-12-15
rust/util: add unit tests for rust_hmac_sha256/rust_hmac_sha512

benma:
And make sure overlaps are allowed explicitly, which we make use of for example in atecc.c/optiga.c.

### DIFF
--- a/src/rust/util/src/sha2.rs
+++ b/src/rust/util/src/sha2.rs
@@ -50,17 +50,23 @@ pub unsafe extern "C" fn rust_sha256_finish(ctx: *mut *mut c_void, out: *mut c_u
     unsafe { *ctx = core::ptr::null_mut() };
 }
 
-/// Safety: data must be a valid buffer for `len` bytes. `out` must be 32 bytes long.
+/// Safety: data must be valid buffer for `len` bytes. `out` must be 32 bytes long.
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn rust_sha256(data: *const c_void, len: usize, out: *mut c_uchar) {
+    let hash = {
+        let data = unsafe { core::slice::from_raw_parts(data as *const u8, len) };
+        Sha256::digest(data)
+    };
+
     let out = unsafe { core::slice::from_raw_parts_mut(out, 32) };
-    let data = unsafe { core::slice::from_raw_parts(data as *const u8, len) };
-    let hash = Sha256::digest(data);
     out.copy_from_slice(&hash[..]);
 }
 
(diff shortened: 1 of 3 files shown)


src/rust/util/src/sha2.rs
```diff
@@ -61,6 +61,9 @@ pub unsafe extern "C" fn rust_sha256(data: *const c_void, len: usize, out: *mut
 
 /// Safety: `key` and `data` must be a valid buffers of the corresponding sizes. `out` must be 32
```
cedwies: can we fix this _buffers_ typo? 

> _must be valid buffers_

src/rust/util/src/sha2.rs
```diff
@@ -70,19 +73,27 @@ pub unsafe extern "C" fn rust_hmac_sha256(
...
+        engine.input(data);
+        let hmac_result: Hmac<sha256::Hash> = Hmac::from_engine(engine);
+        hmac_result.to_byte_array()
+    };
+
+    let out = unsafe { core::slice::from_raw_parts_mut(out, 32) };
+    out.copy_from_slice(&result);
 }
 
 /// Safety: `key` and `data` must be a valid buffers of the corresponding sizes. `out` must be 64
```
cedwies, line 95: same here (_buffers_)

src/rust/util/src/sha2.rs
cedwies, line 55: `rust_sha256` has the same issue that was fixed for the `HMAC` functions. `&mut out` is created before `&data`, so they co-exist. The `test_overlapping` test (which already exists) for `rust_sha256` does this overlap case. Should `rust_sha256` be fixed with the same pattern in this PR or is that intentionally out of scope?

src/rust/util/src/sha2.rs
benma: Fixed

src/rust/util/src/sha2.rs
benma: Fixed

src/rust/util/src/sha2.rs
benma: Adjusted so it is consistent :+1:

benma:
>     1. I found `rust_hmac_sha256` overlap usage in `atecc.c/optiga.c`. Is there a similar use case for `rust_hmac_sha512`, or is this future proofing (for consistency)?

Basically for consistency so one does not have to worry about it.

>     2. > `out` may overlap with `data` (and/or `key`)

> Is `key` overlap actually used anywhere? If not, should the doc be narrowed/adjusted, or should a test be added? Otherwise this there might be a discrepancy between tests and docs, no?

Don't think it's used and seems rather unlikely it is used, but I added some tests for it anyway.

PR #1714, opened 2025-12-16
Optiga tests

benma:


src/optiga/optiga.c
```diff
@@ -1175,13 +923,13 @@ static int _factory_setup(void)
...
 
-    _util = optiga_util_create(OPTIGA_INSTANCE_ID_0, _optiga_lib_callback, NULL);
+    _util = optiga_util_create(OPTIGA_INSTANCE_ID_0, optiga_ops_callback, NULL);
     if (NULL == _util) {
         util_log("couldn't create optiga util");
         return SC_OPTIGA_ERR_CREATE;
     }
 
-    _crypt = optiga_crypt_create(OPTIGA_INSTANCE_ID_0, _optiga_lib_callback, NULL);
+    _crypt = optiga_crypt_create(OPTIGA_INSTANCE_ID_0, optiga_ops_callback, NULL);
```
NickeZ: I'm thinking `optiga_ops_callback` should be static and not global.

It seems we anyway aren't reading the result of the callback.

The callback must not be null, but could be a noop local function.

benma: The callback is setting the _optiga_lib_status static var, which is polled in the WAIT macro.

src/optiga/optiga_ops.c
```diff
@@ -0,0 +1,245 @@
...
+    uint16_t length)
+{
+    _optiga_lib_status = OPTIGA_LIB_BUSY;
+    optiga_lib_status_t res =
+        optiga_util_write_data(me, optiga_oid, write_type, offset, buffer, length);
+    _WAIT(res, _optiga_lib_status);
+    return res;
+}
+
+optiga_lib_status_t optiga_ops_util_read_metadata_sync(
```
NickeZ, line 83: `read_metadata_sync` is now always compiled in, previously only in factorysetup. is this intentional?

src/optiga/optiga_ops.c
```diff
@@ -0,0 +1,245 @@
...
+    uint8_t* buffer,
+    uint16_t* length)
+{
+    _optiga_lib_status = OPTIGA_LIB_BUSY;
+    optiga_lib_status_t res = optiga_util_read_metadata(me, optiga_oid, buffer, length);
+    _WAIT(res, _optiga_lib_status);
+    return res;
+}
+
+optiga_lib_status_t optiga_ops_util_write_metadata_sync(
```
NickeZ, line 95: previously only available in factorysetup

src/optiga/optiga.c
NickeZ: but the WAIT macro isn't avialable outside `optiga_ops.c`?

benma: Yes, but it sets a static var in optiga_ops.c, and the callback needs to be proveded here in optiga.c.

NickeZ: probably the two functions that aren't wrapped should be wrapped. then the callback can be local and you don't need any callback at all in this file.

PR #1715, opened 2025-12-16
readme: improve py/readme.md with install instructions

thisconnect:
Added copy / pasteable instructions so that users without deeper python knowledge can just use the python scripts.
i.e. users without knowledge about venv.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ compile_flags.txt
 bitbox02.egg-info
 /py/bitbox02/build
 /py/bitbox02/dist
+.venv


thisconnect:
The reason why I opened this PR is I only use these scripts every few month and run into the same venv issues again and again and need to ping others to help me. I'm a python noob.. 

This PR helps me next time (in a few month) :)

also I have no idea why the py/Makefile is used for.

py/README.md
```diff
@@ -1,18 +1,103 @@
...
 Older pip versions will fail due to changes in how editable installs are handled (PEP 660).
 
-To use the scripts (e.g. `send_message.py`, `load_firmware.py`):
+### Installing dependencies
+
+Install the required Python dependencies listed in `requirements.txt`:
+
 ```bash
-pip3 install py/bitbox02
+python3 -m venv .venv
```
benma, line 26: Probably best to put this .venv folder into .gitignore, or place the .venv folder outside of the repo.

py/README.md
```diff
@@ -1,18 +1,103 @@
...
+When connecting the first time to an initialized but unpaired BitBox, the device
+will prompt to unlock and continue to compare and confirm the Noise pairing key.
+This is a one-time action.
+
+
+## Flash the firmware.bin
+
+To install / flash the firmware binary the device must be in "Bootloader mode",
+then the following command will flash the firmware.bin onto the BitBox.
+This assumes you have a `./firmware.bin`.
```
benma: Mention that this is for unsigned firmware and a only works for dev-devices. For prod devices, the binary has to be signed and `--debug` needs to be skipped.

py/README.md
```diff
@@ -1,18 +1,103 @@
...
+```
+
+When connecting the first time to an initialized but unpaired BitBox, the device
+will prompt to unlock and continue to compare and confirm the Noise pairing key.
+This is a one-time action.
+
+
+## Flash the firmware.bin
+
+To install / flash the firmware binary the device must be in "Bootloader mode",
```
benma: It doesn't need to be in bootloader mode, load_firmware.py will prompt to enter the bootloader if it's booted into the regular firmware.

py/README.md
```diff
@@ -1,18 +1,103 @@
...
 ```bash
-pip install -e py/bitbox02
+pip install -e ./bitbox02
 ```
 
-It is highly recommended that you use the dockerized setup while developing, guide for setting it up
-can be found in [BUILD.md](../BUILD.md).
\ No newline at end of file
+It is highly recommended that you use the dockerized setup while developing,
+guide for setting it up can be found in [BUILD.md](../BUILD.md).
```
benma: For developing the Python sources, almost all of it can be done on the host easily. Only `make py` to regenerate protobufs is better in the Docker container.

I am not sure it's easily possible to connect a BitBox in the Docker container, to connect it to send_message. I never do that.

py/README.md
thisconnect: I checked with `git check-ignore -v .venv` and found out it was already ignored in my user git config
`~/.config/git/ignore`.

I'll update the .gitignore in the root as I think we always want to ignore .venv directories.

py/README.md
thisconnect: That was a missing piece, thanks. I think it makes sense to mention the simulator here for contributors that don't have a dev-device.

Also added a comment to mention production devices only accept newer firmware.

py/README.md
thisconnect: This text was just copied from the old version. I update with your recommendations.

thisconnect (commented):
@benma thanks for your review.

I added a separate commit to update the gitignore.

I'll fixup the readme updates once this is approved.

py/README.md
NickeZ: I also have it in my personal ignore file. Even if I don't think it is necessary I guess it doesn't hurt to have it in the project ignore file if we recommend people to use `.venv`.

thisconnect:
@benma I think I addressed all your comments PTAL

py/README.md
```diff
@@ -1,18 +1,115 @@
...
+will prompt to unlock and continue to compare and confirm the Noise pairing key.
+This is a one-time action.
+
+
+## Flash the firmware.bin
+
+Use the following script to flash the firmware.bin onto the BitBox.
+The script will prompt to enter the bootloader on the device before flashing.
+
+Production devices only accept `./firmware.bin` signed by BitBox.
```
benma: `firmware.signed.bin` - it's a bit silly but the script actually enforces `.signed.bin` in the name :sweat_smile:

thisconnect: changed to `firmware.signed.bin` for production devices

PR #1716, opened 2025-12-17
workflows/unlock: show remaning attempts before unlocking, not after

benma:
With a special warning and longtouch for the last attempt.

This way, the warning is clearer, as it can't be missed or forgotten. Another benefit is that it is shown in other workflows where unlocking is required, e.g. when showing mnemonic, etc.

### DIFF
--- a/src/rust/bitbox02-rust/src/workflow/unlock.rs
+++ b/src/rust/bitbox02-rust/src/workflow/unlock.rs
@@ -103,15 +103,14 @@ async fn maybe_confirm_remaining_unlock_attempts(
         .await
 }
 
-/// Prompts the user for the device password, and returns `Ok` if the
-/// keystore was successfully unlocked, or `Err` if the password was
-/// incorrect.
+/// Prompts the user for the device password, and returns `Ok` if the keystore was successfully
+/// unlocked, or `Err` if the password was incorrect.
 ///
 /// If there were failed password attempts already, a warning is shown before the password is
 /// entered.
 ///
-/// If they keystore is already unlocked, this function does not
-/// change the state and just checks the password.
+/// If the keystore is already unlocked, this function does not change the state and just checks the
+/// password.
 pub async fn unlock_keystore(
     hal: &mut impl crate::hal::Hal,
     title: &str,


NickeZ:
@codex review

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 - Fix bug that BLE was turned off when iOS device is unlocked
 - simulator-graphical: a new simulator with a graphical user interface
 - Add ability to change the device password after initial setup
+- Show remaning unlock attempts before unlocking instead of after a failed attempt
```
cedwies: Nit: "remaining"

src/rust/bitbox02-rust/src/workflow/unlock.rs
```diff
@@ -69,10 +69,39 @@ impl core::convert::From<password::EnterError> for UnlockError {
...
+async fn maybe_confirm_remaining_unlock_attempts(
+    hal: &mut impl crate::hal::Hal,
+    can_cancel: password::CanCancel,
+) -> Result<(), confirm::UserAbort> {
+    let remaining = crate::keystore::get_remaining_unlock_attempts();
+    if remaining >= bitbox02::memory::MAX_UNLOCK_ATTEMPTS {
+        return Ok(());
+    }
+
+    let body: alloc::string::String = if remaining == 1 {
```
cedwies: Nit: The higher level flow prevents `remaining == 0`, however, I think it could make sense to consider `remaining == 0` (even if it should be unreachable). Not a must though

src/rust/bitbox02-rust/src/workflow/unlock.rs
```diff
@@ -69,10 +69,39 @@ impl core::convert::From<password::EnterError> for UnlockError {
...
 /// Prompts the user for the device password, and returns `Ok` if the
 /// keystore was successfully unlocked, or `Err` if the password was
-/// incorrect. In that case, a status is displayed with how many
-/// attempts are remaining until the device resets.
+/// incorrect.
+///
+/// If there were failed password attempts already, a warning is shown before the password is
+/// entered.
 ///
 /// If they keystore is already unlocked, this function does not
```
cedwies: Nit: could we fix this comment? ("they")

CHANGELOG.md
benma: I need a spell checker in my emacs, I have too many typos :sob: 

Thanks, fixed.

src/rust/bitbox02-rust/src/workflow/unlock.rs
benma: :+1: fixed

benma:
> I want to share a thought: [...]

Very nice find! I agree with your conclusion, that the behavior is technically the same, just the UX is "worse" (though before it was also not ideal that it looked like one there was a chance unlock, but in reality there wasn't).

Imho it's such an extreme edge case that we don't need to worry about it. I added a comment about it in case it happened to someone anyway.

I guess the actual solution to this would involve resetting the device on boot if there are no more attempts (i.e. if a forced reset was supposed to happen but was interrupted), but that's out of scope here.

benma:
Rebased to resolve conflicts with the smarteeprom HAL change.

benma:
@cedwies the CI errors have to do with MAX_UNLOCK_ATTEMPTS in keystore.rs being exposed to C via rust.h, so it's redefined. The good news is that we can remove the define in C then :smile: I'll deal with it later today or tomorrow.

edit: rebased on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1720 which solves this

divyarajsinhrajput33-crypto:
Noo eat

PR #1717, opened 2025-12-18
HAL: refactor salt_root functions

cedwies:
Centralizes salt root access behind the HAL abstraction, enabling full test isolation without C memory dependencies.

TestingHal initializes with a default salt root to preserve existing test behavior without manual setup in each test.

The 2 rust_salt_hash_data tests keep using mock_memory() + C-backed salt root because the extern "C" shim internally constructs BitBox02Hal—these tests verify the FFI path works.

### DIFF
--- a/src/rust/bitbox02-rust/src/keystore.rs
+++ b/src/rust/bitbox02-rust/src/keystore.rs
@@ -192,9 +192,9 @@ fn verify_seed(
     decrypted.as_slice() == expected_seed
 }
 
-fn hash_seed(seed: &[u8]) -> Result<[u8; 32], Error> {
+fn hash_seed(hal: &mut impl crate::hal::Hal, seed: &[u8]) -> Result<[u8; 32], Error> {
     let salted_key =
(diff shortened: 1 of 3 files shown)


benma:
@codex review

src/rust/bitbox02-rust/src/keystore.rs
```diff
@@ -898,12 +899,10 @@ mod tests {
             .collect();
 
         for size in [16, 32] {
-            mock_memory();
-            bitbox02::random::fake_reset();
-            bitbox02::memory::set_salt_root(&mock_salt_root).unwrap();
-            lock();
-
             let mut hal = TestingHal::new();
+            let _ = hal.memory.reset_hww();
```
benma: Why is this call needed?

If needed, prefer `reset_hww().unwrap()` instead of ignoring the output.

edit: tested, tests also pass without this line. TestingHal::new() inits the fields to the same values after all.

benma (approved):
utACK, nice! (Left a small comment, please take a look at that before merging).

PR #1718, opened 2025-12-18
rust: update toolchain to 1.92

benma:
There is a breaking change about abort_immediate_panic, see CMakeLists.txt.


NickeZ:
Can you do `cargo update --precise` with the libc version used in rust stdlib so that we avoid vendoring two versions?

benma:
> Can you do `cargo update --precise` with the libc version used in rust stdlib so that we avoid vendoring two versions?

Good catch, done. It only saved <100 bytes in the fw binary though.

PR #1719, opened 2025-12-18
remove keystore.c

benma:
All functions in it are unrelated to the keystore.

### DIFF
--- a/src/rust/bitbox02-rust/src/secp256k1.rs
+++ b/src/rust/bitbox02-rust/src/secp256k1.rs
@@ -70,7 +70,7 @@ impl Deref for GlobalContext {
 ///
 /// # Returns
 /// * `Ok(SignResult)` containing signature in compact format and recoverable id on success
-/// * `Err(())` if the keystore is locked
+/// * `Err(())` on error.
 pub fn secp256k1_sign(
     private_key: &[u8; 32],
     msg: &[u8; 32],


src/rust/bitbox02/src/secp256k1.rs
```diff
@@ -58,19 +72,32 @@ pub fn _secp256k1_nonce_commit(
     msg: &[u8; 32],
     host_commitment: &[u8; 32],
 ) -> Result<[u8; EC_PUBLIC_KEY_LEN], ()> {
-    let mut signer_commitment = [0u8; EC_PUBLIC_KEY_LEN];
-    match unsafe {
-        bitbox02_sys::keystore_secp256k1_nonce_commit(
+    let mut signer_commitment = MaybeUninit::<bitbox02_sys::secp256k1_ecdsa_s2c_opening>::uninit();
```
NickeZ, line 73: why do you prefer uninitialized bytes here instead of `zeroed()` (like earlier)?

src/keystore.h
```diff
@@ -17,68 +17,9 @@
...
-#include <stdbool.h>
-#include <stddef.h>
-#include <stdint.h>
-
-#include <secp256k1.h>
-
 #define KEYSTORE_U2F_SEED_LENGTH SHA256_LEN
 
 // Max. length of an xpub string, including the null terminator.
 #define XPUB_ENCODED_LEN 113
```
NickeZ: `XPUB_ENCODED_LEN` is only used in rust, migrate as well?

src/keystore.h
```diff
@@ -17,68 +17,9 @@
 
 #include "compiler_util.h"
 
-#include <stdbool.h>
-#include <stddef.h>
-#include <stdint.h>
-
-#include <secp256k1.h>
-
 #define KEYSTORE_U2F_SEED_LENGTH SHA256_LEN
```
NickeZ: move `KEYSTORE_U2F_SEED_LENGTH` to rust or `u2f.c`?

src/rust/bitbox02/src/secp256k1.rs
benma: It's transferred from the C impl, where there was a `= {0}` in the C version for this one and not the other. I don't think it matters, should I change the other to `uninit()` as well?

src/keystore.h
benma: It's a bit confusing, SHA256_LEN is defined in u2f.c, so shouldn't be available here? :thinking: 

I'll delete it anyway, the only use is in SHA256 functions, so we can just use 32 (like we often do in hashing), or `SHA256_LEN` in u2f.c.

src/keystore.h
benma: Indeed, can be removed, it's unused. Will do.

benma: Done, removed it and keystore.h

src/keystore.h
benma: Done, removed keystore.h

src/rust/bitbox02/src/secp256k1.rs
benma: Changed the other to uninit()

src/rust/bitbox02-rust/src/secp256k1.rs
```diff
@@ -43,3 +45,115 @@ impl Deref for GlobalContext {
         })
     }
 }
+
+/// Length of a compressed secp256k1 pubkey.
+const EC_PUBLIC_KEY_LEN: usize = 33;
```
cedwies: why are you not using `bitbox02::secp256k1::EC_PUBLIC_KEY_LEN` here? Why did you choose the duplicate value?

src/rust/bitbox02-rust/src/secp256k1.rs
```diff
@@ -43,3 +45,115 @@ impl Deref for GlobalContext {
...
+///
+/// # Arguments
+/// * `private_key` - 32 byte private key
+/// * `msg` - 32 byte message to sign
+/// * `host_nonce` - 32 byte nonce contribution. Cannot be NULL.
+///   Intended to be a contribution by the host. If there is none available, use 32 zero bytes.
+///
+/// # Returns
+/// * `Ok(SignResult)` containing signature in compact format and recoverable id on success
+/// * `Err(())` if the keystore is locked
```
cedwies: This doc was moved, but I think this line is incorrect. Does this fn even look at "keystore locked"? Isn't this checked **before** this fn runs?

src/rust/bitbox02-rust/src/secp256k1.rs
benma: :+1:  Removed the constant completely, see added commit.

src/rust/bitbox02-rust/src/secp256k1.rs
benma: Fixed, see added commit. This used to be true but stopped being true after some refactor in the past.

PR #1722, opened 2025-12-22
Move "get_attestation_*" functions to HAL

cedwies:
Also, `attestation.rs` was expanded with tests

### DIFF
--- a/src/rust/bitbox02-rust/src/attestation.rs
+++ b/src/rust/bitbox02-rust/src/attestation.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::hal::SecureChip;
+use crate::hal::{Memory, SecureChip};
 use sha2::{Digest, Sha256};
 
 pub struct Data {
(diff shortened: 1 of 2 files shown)


src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -658,6 +682,22 @@ pub mod testing {
...
             }
         }
+
+        fn get_attestation_pubkey_and_certificate(
+            &mut self,
+            pubkey_out: &mut [u8; 64],
+            certificate_out: &mut [u8; 64],
+            root_pubkey_identifier_out: &mut [u8; 32],
+        ) -> Result<(), ()> {
+            *pubkey_out = [0u8; 64];
```
benma: Seems like attestation.rs could now be extended with some unit tests using HAL only - should be an easy job for an agent. What do you think?

cedwies: unit tests are always great, so you recommend the proper implementation for TestingHal?

benma: Yeah I think it's not avoidable. You could add a set_attestation_certificate helper on TestingMemory similar to [this](https://github.com/BitBoxSwiss/bitbox02-firmware/blob/f5942961a7f4b6cc39a444ed5b5f7dafae13fe79/src/rust/bitbox02/src/memory.rs#L324) so you can mock the values in the test.

src/rust/bitbox02-rust/src/attestation.rs
```diff
@@ -31,14 +31,66 @@ pub fn perform(hal: &mut impl crate::hal::Hal, host_challenge: [u8; 32]) -> Resu
...
     )?;
+    result.bootloader_hash = hal.memory().get_attestation_bootloader_hash();
     let hash: [u8; 32] = Sha256::digest(host_challenge).into();
-    result.bootloader_hash = bitbox02::memory::get_attestation_bootloader_hash();
     hal.securechip()
         .attestation_sign(&hash, &mut result.challenge_signature)?;
     Ok(result)
 }
+
+#[cfg(all(test, feature = "testing"))]
```
benma: `cfg(test)` is enough, same like everywhere else. Fyi, the `testing` feature exists because `cfg(test)` is only active in the currently tested crate, not in deps, so with the `testing` feature one can skip compiling test-only functions in deps.

src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -432,6 +462,8 @@ pub mod testing {
                 reset_keys_fail_once: false,
                 #[cfg(feature = "app-u2f")]
                 u2f_counter: 0,
+                attestation_signature: [0u8; 64],
```
benma: I slightly prefer this (and the set function) be named `mock_attestation_signature` so it's clear it's not a real sig.

src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -562,6 +608,21 @@ pub mod testing {
         pub fn set_salt_root(&mut self, salt_root: &[u8; 32]) {
             self.salt_root = *salt_root;
         }
+
+        pub fn set_attestation(
```
benma: rename to `set_attestation_certificate`? Consistent with `memory_set_attestation_certificate`.

src/rust/bitbox02-rust/src/attestation.rs
cedwies: Thank you for the clarification. Will fix!

src/rust/bitbox02-rust/src/hal.rs
cedwies: Will fix. However, other similar fields in `TestingMemory` (e.g. `salt_root`, `attestation_certificate`, ...) are not using the `mock_` prefix. Wouldn't it make sense there as well (even if beyond the scope of this PR)?

src/rust/bitbox02-rust/src/hal.rs
cedwies: Thanks!

src/rust/bitbox02-rust/src/hal.rs
benma: I'd make a distinction between fields needed to fake the implementation of the trait, and fields used to inject/ mock specific values. For example `device_name`  is not manipulated to mock specific values, it's simply what is stored by the `set_device_name()` trait function. `attestation_signature` on the other hand is used to mock the return value of a function.

benma (approved):
utACK

PR #1723, opened 2025-12-22
da14531: port da14531_protocol_format to Rust

benma:
C unit tests for the C impl are added.

Then the C function is ported to Rust, while still passing the C tests.

Then the C tests are ported.

This way we can be reasonably sure that the Rust port is correct.


benma:
> Very nice, but I think the naming is wrong,

Renamed the crate

> There are no negative tests (which isn't a blocker), so no invalid sequences are tested.

There can be no negative tests, right? It's encoding, not decoding - one can encode any payload.

> I would also prefer if we could handle failure by resetting the communication stack rather than panicking. But this means that we need to return `Result` instead of panic.

Can we postpone that to another PR?  For this PR I'd like to keep it refactor-only, not change functionality.

src/rust/bitbox-framed-serial-link/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+///
+/// For example, 64 bytes require 140 byte buffer worst case.
+///
+/// Returns number of formatted bytes
+pub fn protocol_format(out: &mut [u8], packet_type: ProtocolPacketType, payload: &[u8]) -> usize {
+    let payload_len = u16::try_from(payload.len()).expect("payload too large");
+    let len_bytes = payload_len.to_le_bytes();
+
+    let mut idx = 0usize;
+    assert!(idx < out.len());
```
NickeZ, line 59: asf4 `ASSERT` is the same as rust `debug_assert`.

NickeZ: Indexing in rust is checked, so you don't need out-of-bounds checks.

NickeZ:
> There can be no negative tests, right? It's encoding, not decoding - one can encode any payload.

Can check what happens if out-buf is to small.

> 
> > I would also prefer if we could handle failure by resetting the communication stack rather than panicking. But this means that we need to return `Result` instead of panic.
> 
> Can we postpone that to another PR? For this PR I'd like to keep it refactor-only, not change functionality.

👍

benma:
> Can check what happens if out-buf is to small.

It will run into the assert, not sure how to test that in C. In Rust, I could add a test to check if it panics, but I don't see how that is helpful.

NickeZ:
> It will run into the assert, not sure how to test that in C. In Rust, I could add a test to check if it panics, but I don't see how that is helpful.

I didn't even see the c-tests initially when I reviewed.

benma:
> > It will run into the assert, not sure how to test that in C. In Rust, I could add a test to check if it panics, but I don't see how that is helpful.
> 
> I didn't even see the c-tests initially when I reviewed.

Should have read the PR comment :nerd_face: :smile: 

https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1723#issue-3753489690

NickeZ:
> > > It will run into the assert, not sure how to test that in C. In Rust, I could add a test to check if it panics, but I don't see how that is helpful.
> > 
> > 
> > I didn't even see the c-tests initially when I reviewed.
> 
> Should have read the PR comment 🤓 😄
> 
> [#1723 (comment)](https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1723#issue-3753489690)

Yeah, I was very confused when I only saw rust unit tests :P

benma:
Split first commit into separate PR so we can see full CI passing on it: https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1726

benma:
First commit was merged in #1726. Second commit is split into a new PR #1750, to see CI pass with the Rust port against the previously added C unit tests.

src/rust/bitbox-framed-serial-link/src/lib.rs
NickeZ: @benma did you see this comment?

NickeZ: actually, I guess this will be handled in the other PR

NickeZ:
After the dependency is merged I'll approve this. Unit tests successfully ran on my machine.

PR #1724, opened 2025-12-22
Move get_datetime/format_datetime to HAL

cedwies:
A new trait "Time" was added to HAL, as the
functions do not fit well into an existing trait.

No new tests were added, my current opinion is that
the tests which already exist cover well.


cedwies:
`get_datetime` and `format_datetime` should not be part of the hardware abstraction layer. Therefore, this PR can be closed.

PR #1728, opened 2025-12-25
C aliasing

benma:


src/da14531/da14531_protocol.c
```diff
@@ -276,8 +276,18 @@ static struct da14531_protocol_frame* _serial_link_in_poll(
...
+            util_log(
+                "da14531: ERROR, short frame len %u, dropped frame", (unsigned)self->frame_len);
+            self->state = SERIAL_LINK_STATE_READING;
+            self->frame_len = 0;
+            return NULL;
+        }
+
         // bytes with index 1-2 are the length
-        uint16_t len = *((uint16_t*)&self->frame[1]);
+        uint16_t len = (uint16_t)self->frame[1] | ((uint16_t)self->frame[2] << 8);
```
NickeZ: why not memcpy here?

src/da14531/da14531_protocol.c
```diff
@@ -292,7 +302,8 @@ static struct da14531_protocol_frame* _serial_link_in_poll(
 
         // CRC in frame
         // bytes with index n-2 and n-1 are the crc
-        uint16_t crc_frame = *(uint16_t*)&self->frame[3 + len];
+        uint16_t crc_frame =
+            (uint16_t)self->frame[3 + len] | ((uint16_t)self->frame[3 + len + 1] << 8);
```
NickeZ: why not memcpy here?

src/factorysetup.c
```diff
@@ -275,7 +275,7 @@ static bool _rtt_receive(uint8_t* msg_out, size_t* len_out)
             return false;
         }
         // util_log("read %s", util_dbg_hex(buffer, read));
-        uint16_t len = *((uint16_t*)buffer);
+        uint16_t len = ((uint16_t)buffer[0] << 8) | (uint16_t)buffer[1];
```
NickeZ: why not memcpy here?

NickeZ (commented):
Looks good in general, left a few comments, I will need to test as well.

src/da14531/da14531_protocol.c
benma: Could do, but the agent thought it was better to be explicit about endianness. Do you prefer to change it to memcpy?

NickeZ: But why doesn't that argument apply to the other places? where in one case there even is a comment about endianess.

I don't see any reasons for different locations to use different solutions and just wanted to point out the inconsistency. `memcpy` seems more readable/maintainable to me. I don't really care what the agent thinks, I care about what you think is more readable and maintainable. The end result is probably the same instructions.

benma: I really don't have any opinion, so I'll change it to memcpy. There are no big endian systems we would use anyway right? 

But I can't find any comment about it (grepped "little endian"), which code are you referring too? Though I assume there must be some more places where we implicitly assume little endian.

benma: Changed memcpy.

src/da14531/da14531_protocol.c
benma: Changed memcpy.

src/factorysetup.c
benma: Changed memcpy.

src/da14531/da14531_protocol.c
NickeZ: https://github.com/benma/bitbox02-firmware/blob/c5b04de209450c0e13c529ef95bc6d9989940528/src/bootloader/bootloader.c#L560

NickeZ: Since we used little endian from the start and this is about two arms communicating with each other, I think we can say that in this protocol we use little endian.

benma: Funnily enough the line you link to was also changed in this PR, but the agent here did not think of parsing it like in the other instances :sweat_smile:

NickeZ:
@benma please rebase after merging https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1750

benma:
> @benma please rebase after merging #1750

Done. I will also test this PR manually again in all affected builds.

NickeZ (approved):
tACK

Tested only using iOS:

Tested bluetooth pairing (with bootloader and firmware), enabling/disabling passphrase, signing btc tx

PR #1729, opened 2025-12-26
Restore test

benma:


src/rust/bitbox02-rust/src/workflow/mnemonic.rs
```diff
@@ -300,12 +296,11 @@ async fn get_12th_18th_word(
     }
 }
 
-/// Retrieve a BIP39 mnemonic sentence of 12, 18 or 24 words from the user.
```
cedwies, line 303: Just so I am not missing anything: The comment was wrong anyway, right?. Did we even support 18 words?

src/rust/bitbox02-rust/src/workflow.rs
```diff
@@ -3,7 +3,10 @@
 pub mod cancel;
 pub mod confirm;
 pub mod menu;
-#[cfg_attr(feature = "c-unit-testing", path = "workflow/mnemonic_c_unit_tests.rs")]
+#[cfg_attr(
```
cedwies, line 6: With `cfg_attr(all(feature = "c-unit-testing", not(feature = "testing")), ...)`, if someone (accidentally) enables both `c-unit-testing` and `testing` at the same time, the code silently uses the "real" `mnemonic.rs.` Would it be safer to add a `compile_error!` to catch this?

src/rust/bitbox02-rust/src/workflow/mnemonic.rs
benma: We used to until recently.

src/rust/bitbox02-rust/src/workflow.rs
benma: In Rust unit tests, both features are active (all features are active in Rust unit tests), and the "real" mnemonic.rs should be used. That's why I did this change. Or am I missing something?

cedwies: ah no I was missing something, thanks for clarifying

benma:
> The test mnemonic **boring mistake dish oyster...** appears in both `testing.rs` and `mnemonic_c_unit_tests.rs`. If someone updates one but forgets the other, tests could silently pass with different data. Would a shared constant be better, or is the duplication intentional to keep the Rust tests and C tests completely isolated?

The mnemonic in mnemonic_c_unit_tests.rs is used by the (non-graphical) simulator and shouldn't change, as probably integration tests in various repos depend on it. (The same mnemonic also exists once more in graphical simulator main.rs).

The new one in testing.rs is merely for Rust unit tests, and could be a different one, or be replaced by some sort of mocking facility later, so I don't think there should be a shared constant.

PR #1731, opened 2025-12-29
bitbox-secp256k1

benma:
edit (nd): to test you need to run `git submodule update --init` because libsecp has been moved to a rust crate.

### DIFF
--- a/src/rust/bitbox-secp256k1/src/lib.rs
+++ b/src/rust/bitbox-secp256k1/src/lib.rs
@@ -249,18 +249,18 @@ pub fn anti_exfil_host_verify(
     signer_commitment: &[u8; 33],
 ) -> Result<(), ()> {
     let mut opening = MaybeUninit::<ffi::secp256k1_ecdsa_s2c_opening>::uninit();
-    let parse_res = unsafe {
+    if unsafe {
         ffi::secp256k1_ecdsa_s2c_opening_parse(
             secp.ctx().as_ptr(),
             opening.as_mut_ptr(),
             signer_commitment.as_ptr(),
         )
-    };
-    if parse_res != 1 {
+    } != 1
+    {
         return Err(());
     }
     let opening = unsafe { opening.assume_init() };
-    let verify_res = unsafe {
+    if unsafe {
         ffi::secp256k1_anti_exfil_host_verify(
             secp.ctx().as_ptr(),
             signature.as_c_ptr(),
@@ -269,8 +269,12 @@ pub fn anti_exfil_host_verify(
             host_nonce.as_ptr(),
             &opening,
         )
-    };
-    if verify_res == 1 { Ok(()) } else { Err(()) }
+    } == 1
+    {
+        Ok(())
+    } else {
+        Err(())
+    }
 }
 
 pub fn dleq_prove(
@@ -282,7 +286,7 @@ pub fn dleq_prove(
 ) -> Result<Vec<u8>, ()> {
     let mut s = [0u8; 32];
     let mut e = [0u8; 32];
-    let result = unsafe {
+    if unsafe {
         ffi::bitbox_secp256k1_dleq_prove(
             secp.ctx().as_ptr(),
             s.as_mut_ptr(),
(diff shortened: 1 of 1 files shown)


src/rust/bitbox02-rust-c/src/lib.rs
```diff
@@ -13,6 +13,8 @@ mod alloc;
 pub mod async_usb;
 #[cfg(feature = "firmware")]
 mod der;
+#[cfg(feature = "target-factory-setup")]
```
NickeZ: This is the first time we would use a "meta" feature directly in code (feature that exists to activate other features). Should we have a "factory-setup" feature instead?

src/rust/bitbox-secp256k1/src/lib.rs
```diff
@@ -0,0 +1,425 @@
...
+) -> Result<(), ()> {
+    let mut opening = MaybeUninit::<ffi::secp256k1_ecdsa_s2c_opening>::uninit();
+    let parse_res = unsafe {
+        ffi::secp256k1_ecdsa_s2c_opening_parse(
+            secp.ctx().as_ptr(),
+            opening.as_mut_ptr(),
+            signer_commitment.as_ptr(),
+        )
+    };
+    if parse_res != 1 {
```
NickeZ: In the functions previous to this you don't use a temporary (`parse_res`) to store the result you just compare to 1 directly. Any reason why it is different?

src/rust/bitbox-secp256k1/src/lib.rs
```diff
@@ -0,0 +1,425 @@
...
+        ffi::bitbox_secp256k1_dleq_verify(
+            secp.ctx().as_ptr(),
+            proof[..32].as_ptr(),
+            proof[32..].as_ptr(),
+            p1.as_c_ptr(),
+            gen2.as_c_ptr(),
+            p2.as_c_ptr(),
+        )
+    };
+    if result == 1 { Ok(()) } else { Err(()) }
```
NickeZ: I think skipping the `result` variable, like you do in other functions, is better.

NickeZ:
Looks good, left a few non-blocking comments. Need to test as well.

src/rust/bitbox-secp256k1/src/lib.rs
benma: In this PR the code is just moved, parse_res and the other vars are on master: https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/src/rust/bitbox02/src/secp256k1.rs#L113

As for why it was like this: when the agent generates code, it might be slightly different to how I would do it manually, but I don't correct it in such cases as I don't mind either style.

I added a commit to skip these vars.

src/rust/bitbox-secp256k1/src/lib.rs
benma: Done

src/rust/bitbox02-rust-c/src/lib.rs
benma: :+1: added factory-setup feature

.gitmodules
```diff
@@ -11,5 +11,5 @@
 	path = external/embedded-swd
 	url = https://github.com/BitBoxSwiss/embedded-swd
 [submodule "external/secp256k1-zkp"]
```
NickeZ:
```suggestion
[submodule "src/rust/bitbox-secp256k1/depend/secp256k1-zkp"]
```

NickeZ:
The `external/secp256k1-zkp` directory should be removed

.gitmodules
benma: Fixed. Interestingly, this was not changed automatically, and it seems this is just a name can be anything. Updated now to match the path.

benma:
> The `external/secp256k1-zkp` directory should be removed

This directory is now untracked by git, so if it is still there, you should see it as untracked and delete it manually.

NickeZ:
After rebasing it failed on bb03 simulator: 

```
make[4]: *** No rule to make target 'test/simulator-graphical-bb03/secp256k1', needed by 'test/simulator-graphical-bb03/dummy'.  Stop.
```

benma:
> After rebasing it failed on bb03 simulator:
> 
> ```
> make[4]: *** No rule to make target 'test/simulator-graphical-bb03/secp256k1', needed by 'test/simulator-graphical-bb03/dummy'.  Stop.
> ```

Fixed in fixup commit, please check. Will squash it into the right commit before merging.

PR #1736, opened 2026-01-07
Rust miniscript v13.0.0

benma:


src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
```diff
@@ -1615,6 +1633,118 @@ mod tests {
...
+        let Descriptor::Tr(tr) = derived else {
+            panic!("expected tr");
+        };
+
+        // Internal key not present in any leaf script.
+        let internal_key_bytes: [u8; 33] = tr.inner.internal_key().inner.serialize();
+        assert_eq!(tr.get_leaf_hash_by_pubkey(&internal_key_bytes), None);
+
+        // Leaf key is found and returns the corresponding leaf hash.
+        let leaf_pk_1: [u8; 33] =
```
cedwies: Non-blocking: Would it make sense to add comments to the magic constants to hint how these values were obtained?

src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
```diff
@@ -210,10 +211,14 @@ impl Tr<bitcoin::PublicKey> {
     /// Note that we assume that each pubkey is unique according to BIP-388 and validated by
     /// `validate_keys()`, so the leaf is unique.
     fn get_leaf_hash_by_pubkey(&self, pk: &[u8; 33]) -> Option<TapLeafHash> {
-        for (_, ms) in self.inner.iter_scripts() {
-            if ms.iter_pk().any(|pk2| *pk == pk2.inner.serialize()) {
+        for leaf in self.inner.leaves() {
```
cedwies, line 214: Non-blocking: Would it make sense to add a small debug-only check here that panics if we ever find the same pubkey in more than one leaf? That would catch any future violation of this assumption early, and make the code’s security assumption explicit.

src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
```diff
@@ -1615,6 +1633,118 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_get_leaf_hash_by_pubkey() {
```
cedwies, line 1637: `test_taproot_spend_info` looks great for the “happy path” (internal key -> key spend, leaf key -> script spend). How about one small negative test that calls `taproot_spend_info` with a keypath that does not correspond to either the internal key or any leaf key? IMO that would document the expected error behaviour clearly (what error we return, and that we dont accidentally match the wrong spend path).

src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
```diff
@@ -1615,6 +1633,118 @@ mod tests {
...
+        );
+
+        let mut hal = crate::hal::testing::TestingHal::new();
+        let parsed_policy = parse(&mut hal, &policy, coin).unwrap();
+
+        const ADDRESS_INDEX: u32 = 5;
+        let mut xpub_cache = Bip32XpubCache::new(crate::xpubcache::Compute::Once);
+
+        // Internal key results in a key path spend.
+        let keypath_internal: Vec<u32> = KEYPATH_ACCOUNT
```
cedwies, line 1709: Very minor: could you add a short comment here explaining how
`KEYPATH_ACCOUNT + [0, ADDRESS_INDEX]` selects the internal key branch
and `KEYPATH_ACCOUNT + [2, ADDRESS_INDEX] `selects the leaf branch, given the policy string `"tr(@0/<0;1>/*,pk(@0/<2;3>/*))`? Thef mapping is clear when you think it through, but I think a oneline comment would make the test much easier to follow.

src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
benma: Maybe, but I prefer not to, as there is almost no use of debug_assert in our codebase, and it seems exceedingly unlikely it would ever trigger - no one is running debug firmware executing this specific code, and the unit tests already cover duplicate keys.

src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
benma: I changed it to not use hex literals for the leaf pks at all, but just extract them from the descriptor.

src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
benma: Added negative test :+1:

src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
benma: Added comments explaining this :+1:

benma:
@cedwies also added a commit fixing an unrelated test, please check

src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
```diff
@@ -1615,6 +1633,129 @@ mod tests {
...
+                assert_eq!(
+                    tweak.to_byte_array(),
+                    hex!("369416353930adfe3345578e43fc31a8c222905d9d134f3511ea4f3a4ac041a5"),
+                );
+            }
+            _ => panic!("expected key spend"),
+        }
+
+        // Leaf key results in a script path spend. The leaf key is ` @1/<2;3>/*`, so `/2/5`
+        // selects that one as `2` matches the first multipath index of that key.
```
cedwies, line 1728: shouldn't it be `@0/<2;3>/*` instead of `@1/<2;3>/*`?

benma: Dang, yes good catch :smile: Fixed

cedwies (approved):
utACK

PR #1737, opened 2026-01-07
simulator: fix simulator by fixing Nova memory fake

benma:
Simulator's memory made the simulator think it used ATECC, but the fake securechip impl assumes Optiga, so the stretching algo vor ATECC (V0) was passed to the fake securechip, which expects the V1 algo.


NickeZ (approved):
utACK

PR #1738, opened 2026-01-07
workflow/status: do not delay in simulator

benma:
3f0a5df626a1c1104080d36f42079e7e2a898511 made the simulator slow by actually waiting for 2s on each status, while before it was screen counter based and hence instant in the simulator. This removes the delay so the simulator is fast again.

### DIFF
--- a/src/rust/bitbox02/src/delay.rs
+++ b/src/rust/bitbox02/src/delay.rs
@@ -121,6 +121,8 @@ impl Drop for Delay {
     }
 }
 
-pub fn delay_for(duration: Duration) -> Delay {
-    Delay::from_ms(duration.as_millis() as u32)
+pub async fn delay_for(#[allow(unused)] duration: Duration) {
+    // Do not delay in (non-graphical) simulator and Rust unit tests.
+    #[cfg(not(any(feature = "c-unit-testing", feature = "testing")))]
+    Delay::from_ms(duration.as_millis() as u32).await
 }


NickeZ:
shouldn't all delays be ignored in the c-simulator? I mean modifying `delay.rs` instead of `status.rs`.

benma:
> shouldn't all delays be ignored in the c-simulator? I mean modifying `delay.rs` instead of `status.rs`.

Good point. Changed the commit to do that.

test/simulator-graphical/Cargo.lock
```diff
@@ -3110,6 +3110,7 @@ dependencies = [
  "hex",
  "num-bigint",
  "sha2",
+ "zeroize",
```
NickeZ, line 3113: seems unrelated

benma: Yeah it appears when building the simulator, not sure why. Should maybe have a CI check that checks lock files don't become dirty after building.

Anyway, moved it to a separate commit.

NickeZ: I see, util is a path dependency, so we probably made some change to that crate and didn't rebuild the simulator to get the update. Maybe this commit:

https://github.com/BitBoxSwiss/bitbox02-firmware/commit/daec715d342b8feb381de6f54dda0dc673ce3453#diff-e6780e742752a59fab9e0dc0a28ccf57b5407890c4a5452fcb95fdebe46dba73

NickeZ: created an issue for it

PR #1739, opened 2026-01-07
Eth stream

Tomasvrba:


src/rust/bitbox02-rust/src/hww/api/ethereum/sighash.rs
```diff
@@ -77,3979 +234,405 @@ fn hash_u64<W: Write>(writer: &mut W, value: u64) {
...
-                },
-                expected_sighash: *b"\xbf\xf8\x98\xd5\xb1\x3f\x54\x0a\xec\x09\xd5\x79\x0c\xfa\x5f\x49\xa0\xe2\x3a\x5b\xde\xa9\xe2\x5f\x4f\x8b\x77\x7e\xcc\xc4\x02\x2e",
-            },
-            Test {
-                params: ParamsEIP1559 {
-                    chain_id: 200,
-                    nonce: b"\x1d\x49",
-                    max_priority_fee_per_gas: b"\x08\x10\xe5\x68\x97\x5f\x7e\xc9\x63",
-                    max_fee_per_gas:
```
benma: Here and elsehwere, just use `block_on`, it's imported.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -237,11 +274,20 @@ async fn verify_standard_transaction(
 ) -> Result<(), Error> {
     let recipient = parse_recipient(request.recipient())?;
 
-    if !request.data().is_empty() {
+    let data_length = match request {
```
benma: Should data_length be bounded in any way? E.g. 100mB makes no sense right?

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -255,16 +301,30 @@ async fn verify_standard_transaction(
...
-                body: &hex::encode(request.data()),
-                scrollable: true,
-                display_size: request.data().len(),
-                accept_is_nextarrow: true,
-                ..Default::default()
-            })
-            .await?;
+        if data_length > 0 {
+            // Streaming mode: data is too large to display, show size instead
+            // TODO: Add option to scroll through streamed data chunks
```
benma: Please remove the TODO - TODOs tend to pollute the codebase and never get resolved. For stuff that should actually be done at some point, the issue tracker is better.

src/rust/bitbox02-rust/src/hww/api/ethereum/sighash.rs
```diff
@@ -5,29 +5,155 @@ use sha3::{Digest, Keccak256};
...
+        let response = super::next_request(super::pb::eth_response::Response::DataChunkRequest(
+            super::pb::EthSignDataRequestChunkResponse {
+                offset: self.offset as u32,
+                length: chunk_length,
+            },
+        ))
+        .await?;
+
+        #[cfg(test)]
+        let response = tests::mock_next_request(self.offset as u32, chunk_length)?;
```
benma: Such guards in the middle of prod code are a bit scary. Please check if you can make use of `crate::hww::MOCK_NEXT_REQUEST`, for example like here:

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs#L1670

src/rust/bitbox02-rust/src/hww/api/ethereum/sighash.rs
```diff
@@ -77,3979 +234,405 @@ fn hash_u64<W: Write>(writer: &mut W, value: u64) {
...
-                    AccessList: accessList,
-                }
-                tx := types.NewTx(txData)
+    fn decode_hex(s: &str) -> Vec<u8> {
+        hex::decode(s).unwrap()
+    }
 
-                outputtx(nets[rand.Intn(len(nets))], tx)
-            }
+    fn extract_json_string<'a>(json: &'a str, key: &str) -> &'a str {
```
benma: Please use serde_json with nice structs that represent the test data instead of manual JSON parsing. Please check this example:

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/src/rust/streaming-silent-payments/tests/table_test.rs#L107-L109

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
Tomasvrba: I guess we could put an arbitrary limit on it, like 1MB?

src/rust/bitbox02-rust/src/hww/api/ethereum/sighash.rs
```diff
@@ -77,3979 +231,226 @@ fn hash_u64<W: Write>(writer: &mut W, value: u64) {
...
     hash_header(&mut hasher, RLP_SMALL_TAG, RLP_LARGE_TAG, counter.0 as u16);
-    hash_params_eip1559(&mut hasher, params);
+    hash_params_eip1559(&mut hasher, params).await?;
     Ok(hasher.0.finalize().into())
 }
 
-#[cfg(test)]
-mod tests {
+#[cfg(any(test, feature = "testing"))]
+pub mod tests {
```
benma, line 322: Why this change? Looks like it was good before.

src/rust/bitbox02-rust/src/hww/api/ethereum/sighash.rs
```diff
@@ -77,3979 +231,226 @@ fn hash_u64<W: Write>(writer: &mut W, value: u64) {
...
-            fmt.Printf("value: %s,\n", encbytes(tx.Value().Bytes()))
-            fmt.Printf("data: %s,\n", encbytes(tx.Data()))
-            fmt.Printf("},\n")
-            signer := types.MakeSigner(net, net.LondonBlock)
-            fmt.Printf("expected_sighash: *%s,\n", encbytes(signer.Hash(tx).Bytes()))
-            fmt.Printf("},\n")
-        }
+    pub fn decode_hex(s: &str) -> Vec<u8> {
+        hex::decode(s).unwrap()
+    }
```
benma: unused?

src/rust/bitbox02-rust/src/hww/api.rs
```diff
@@ -5,7 +5,7 @@ use crate::pb;
 pub(super) mod error;
 
 #[cfg(feature = "app-ethereum")]
-mod ethereum;
+pub mod ethereum;
```
benma: why?

src/rust/bitbox02-rust/src/hww/api/ethereum.rs
```diff
@@ -10,7 +10,7 @@ mod amount;
 mod keypath;
 mod params;
 mod pubrequest;
-mod sighash;
+pub mod sighash;
```
benma: why?

src/rust/bitbox02-rust/Cargo.toml
```diff
@@ -97,3 +97,7 @@ c-unit-testing = []
 simulator-graphical = []
 
 firmware = []
+
+[dev-dependencies]
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
```
benma: Move this to src/rust/Cargo.toml and use `serde = { workspace = true }` here and in the streaming-silent-payments' Cargo.toml. This way we have only one declaration of the dep for all workspace crates, which helps keeping the versions in sync.

src/rust/bitbox02-rust/src/hww/api/ethereum/sighash.rs
Tomasvrba: used in sighash_tests.rs https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1739/changes/342edb37dd77d31b3afab990dfd41db30e19464c#diff-8ea92aa3796804f5d7ad75e20246d05bbf28e6b492eb208b3a09993700562540 
sorry that was the test file i forgot to add 

i moved the decode_hex to the test file

src/rust/bitbox02-rust/src/hww/api.rs
Tomasvrba: adding serde_json as a dep directly to the crate caused a bunch of type inference issues in existing and unrelated tests ( For example, existing tests like:
  assert_eq!(some_slice, &[]);
  become ambiguous because the compiler can't determine if you mean &[u8], &[u32], or a serde_json::Value.)
  
  having the tests separately fixes that but then those modules need to be public for sighash_tests.rs to access; idk how to get around that, if you have a better solution happy to hear it

src/rust/bitbox02-rust/src/hww/api/ethereum/sighash.rs
Tomasvrba: same reason as below
 #[cfg(test)] only compiles the module when running unit tests in the specific crate so this is now needed since the json parsing moved to sighash_tests.rs

src/rust/bitbox02-rust/src/hww/api/ethereum.rs
Tomasvrba: see above

src/rust/bitbox02-rust/src/hww/api.rs
benma: The type inferences issues are easily solved. The downsides of moving the tests to a separate file are too large. It's less discoverable, all the visibility modifications and importing/exporting become a mess, etc.

See https://github.com/BitBoxSwiss/bitbox02-firmware/commit/3adcde3be6eb895d6117a66d477754c565552aec.

Could you move them back to sighash::tests?

It is quite bad of serde though to be causing these issues :shrug:

src/rust/bitbox02-rust/tests/sighash_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+struct Eip1559TestCase {
+    chain_id: u64,
+    nonce: String,
+    max_priority_fee: String,
+    max_fee_per_gas: String,
+    gas_limit: String,
+    recipient: String,
+    value: String,
+    data: String,
+    expected_sighash: String,
```
benma: edit: if you feel like it, the current way is fine too of course:

For hex strings, you can make it `Vec<u8>`, and annotate it with

    #[serde(deserialize_with = "hex::serde::deserialize")]

(might need to activate the serde feature, please check if that can be done only when used as a dev-dependency).

Tomasvrba: idk, it works, but having `#[serde(deserialize_with = "hex::serde::deserialize")]` on each of the struct fields makes them really ugly, I think i prefer the current `decode_hex` approach

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
Tomasvrba: ok, this is in the fixup commit https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1739/changes/b5a3f4214eef515dde472904739698a5cbd35cf4

src/rust/bitbox02-rust/src/hww/api.rs
Tomasvrba: ok, moved the tests back to sighash, changes fixed up into https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1739/changes/c6763757f4684e221ef168a68616028aa6583f6b already

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -237,11 +274,25 @@ async fn verify_standard_transaction(
...
-    if !request.data().is_empty() {
+    let data_length = match request {
+        Transaction::Legacy(legacy) => legacy.data_length,
+        Transaction::Eip1559(eip1559) => eip1559.data_length,
+    };
+
+    const MAX_DATA_LENGTH: u32 = 1024 * 1024; // 1MB
+    if data_length > MAX_DATA_LENGTH {
+        return Err(Error::InvalidInput);
+    }
```
benma: There is a section in `fn _process` at `// Size limits.` -> move this check there?

messages/eth.proto
```diff
@@ -151,5 +165,6 @@ message ETHResponse {
     ETHSignResponse sign = 2;
     AntiKleptoSignerCommitment antiklepto_signer_commitment = 3;
     ETHTypedMessageValueResponse typed_msg_value = 4;
+    ETHSignDataRequestChunkResponse data_chunk_request = 5;
```
benma: For consistency, this should be called `data_request_chunk`

messages/eth.proto
```diff
@@ -142,6 +155,7 @@ message ETHRequest {
     ETHSignTypedMessageRequest sign_typed_msg = 5;
     ETHTypedMessageValueRequest typed_msg_value = 6;
     ETHSignEIP1559Request sign_eip1559 = 7;
+    ETHSignDataResponseChunkRequest data_chunk = 8;
```
benma: for consistency, this should be named `data_response_chunk`, or similar to bluetooth.rs, drop the `Response` out of the message name.

src/rust/bitbox02-rust/src/hww/api/ethereum.rs
```diff
@@ -77,5 +77,6 @@ pub async fn process_api(
         Request::AntikleptoSignature(_) => Err(Error::InvalidInput),
         Request::SignTypedMsg(request) => sign_typed_msg::process(hal, request).await,
         Request::TypedMsgValue(_) => Err(Error::InvalidInput),
+        Request::DataChunk(_) => Err(Error::InvalidInput),
```
benma: In bitcoin, we return `InvalidState` for these 🤔 Not sure which makes more sense, but for consistency maybe better to make it like in bitcoin, like here:

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/e89eafafd9bb66043d0595881b49603fb4d26bb3/src/rust/bitbox02-rust/src/hww/api/bitcoin.rs#L310-L316

Copy the comment minus the mention of bitcoin/signtx.rs, and group `AntikleptoSignature`, `TypedMsgValue` and `DataChunk` here, retrning `Error::InvalidState`.

src/rust/bitbox02-rust/src/hww/api/ethereum/sighash.rs
```diff
@@ -5,29 +5,152 @@ use sha3::{Digest, Keccak256};
...
+            Ok(Some(self.0.to_vec()))
+        } else {
+            Ok(None)
+        }
+    }
+}
+
+pub struct ChunkingProducer {
+    total_length: usize,
+    offset: usize,
```
benma: make these `u32` instead if `usize`? Avoids all the casting.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
benma: Also add a check to return InvalidInput if `data_length > 0 && !data.is_empty()`, i.e. if client tries to use both ways to send data at once.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -155,33 +155,70 @@ fn parse_fee<'a>(request: &Transaction<'_>, params: &'a Params) -> Amount<'a> {
...
-        gas_limit: &request.gas_limit,
-        recipient: &request.recipient,
-        value: &request.value,
-        data: &request.data,
-        chain_id,
-    })
-    .map_err(|_| Error::InvalidInput)?;
-    Ok(hash)
+async fn hash_legacy(chain_id: u64, request: &pb::EthSignRequest) -> Result<[u8; 32], Error> {
+    if request.data_length > 0 {
```
benma, line 165: Both branches are the same except for `data: ...`, so just do `data: if request.data_length > 0 { ... } else { ... }`?

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -155,33 +155,70 @@ fn parse_fee<'a>(request: &Transaction<'_>, params: &'a Params) -> Amount<'a> {
...
-        max_fee_per_gas: &request.max_fee_per_gas,
-        gas_limit: &request.gas_limit,
-        recipient: &request.recipient,
-        value: &request.value,
-        data: &request.data,
-    })
-    .map_err(|_| Error::InvalidInput)?;
-    Ok(hash)
+async fn hash_eip1559(request: &pb::EthSignEip1559Request) -> Result<[u8; 32], Error> {
+    if request.data_length > 0 {
```
benma, line 197: Same, inline the if/else thing in `data: if { ... } else { ... }` to not repeat all other lines.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -1340,4 +1422,196 @@ mod tests {
...
         );
     }
+
+    #[test]
+    pub fn test_streaming_equivalence_legacy() {
+        const KEYPATH: &[u32] = &[44 + HARDENED, 60 + HARDENED, 0 + HARDENED, 0, 0];
+        let test_data: Vec<u8> = (0..4000u32).map(|i| (i % 256) as u8).collect();
+
+        mock_unlocked();
+        let mut mock_hal_traditional = TestingHal::new();
```
benma: traditional -> nonstreaming? To not confuse it with `legacy` in the test title and generally being more descriptive.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -1340,4 +1422,196 @@ mod tests {
...
+                address_case: pb::EthAddressCase::Mixed as _,
+                data_length: 10000,
+            }),
+        ));
+        clear_chunk_responder();
+
+        assert!(result.is_ok());
+        match result {
+            Ok(Response::Sign(sig)) => {
+                assert_eq!(sig.signature.len(), 65);
```
benma: Would be good to compare the sig here against the concrete value, similar to `test_process_standard_transaction_with_data`.

Also check the expected `mock_hal.ui.screens`, as in `test_process_standard_transaction_with_data`, as it shows different screens in that case.

src/rust/bitbox02-rust/src/hww/api/ethereum/testdata/gen_eip1559_tests.go
```diff
@@ -58,6 +58,16 @@ func randbytes(n int) []byte {
 	return res
 }
 
+func generateDataSize(i int) int {
+	// 80% small (0-1024 bytes), 20% large (5KB, 8KB, 12KB, 15KB)
+	if i%5 == 0 {
```
benma, line 63: This function does not know `i` is looping to 100 -> if you increase the number of test cases, it's easy to miss this. Pass `n` as a param or move the number of cases to a global const.

Same for gen_legacy_tests.go

src/rust/bitbox02-rust/src/hww/api/ethereum/sighash.rs
```diff
@@ -318,3912 +319,273 @@ pub async fn compute_eip1559<D: DataProducer<Error = Error>>(
...
+        nonce: String,
+        gas_price: String,
+        gas_limit: String,
+        recipient: String,
+        value: String,
+        data: String,
+        expected_sighash: String,
+    }
+
+    const DATA_THRESHOLD: usize = 6144;
```
benma, line 383: the Go generator file creates 5kB data fields in the 20% that creates "large" data sizes. These would not be using the streamer in the tests here. Is that deliberate? If so please document it, otherwise lower this 3kB or something like that.

py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -905,6 +951,9 @@ def handle_antiklepto(request: eth.ETHRequest) -> bytes:
                 raise Exception(
                     f"chainID argument ({chain_id}) does not match chainID encoded in transaction ({decoded_chain_id_int})"
                 )
+
+            use_streaming = len(data) > streaming_threshold
+
```
benma, line 958: add `if use_streaming: self._require_atleast(semver.VersionInfo(9, 26, 0))`

maybe also rename `use_streaming` to `require_streaming`, as beyond this threshold one can't use the old way.

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -313,14 +375,21 @@ pub async fn _process(
     }
 
     // Size limits.
+    const MAX_DATA_LENGTH: u32 = 1024 * 1024;
     if request.nonce().len() > 16
         || request.gas_limit().len() > 16
         || request.value().len() > 32
         || request.data().len() > 6144
+        || request.data_length() > MAX_DATA_LENGTH
```
benma: For clarity, please add two consts MAX_STREAMING_DATA_LENGTH and MAX_NONSTREAMING_DATA_LENGTH? A bit confusing seeing these two lines for someone who doesn't know that data_length is not data().len() 😄

src/rust/bitbox02-rust/src/hww/api/ethereum/testdata/gen_eip1559_tests.go
benma: I was stupid - `i % 5 == 0` comes out to 20% regardless of the number of tests. I notice this now that you added `n int` but are not actually using it. Please remove the `n int` arg again :see_no_evil:

src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
```diff
@@ -1340,4 +1427,268 @@ mod tests {
...
+                Screen::Confirm {
+                    title: "Unknown\ncontract".into(),
+                    body: "Only proceed if you\nunderstand exactly\nwhat the data means.".into(),
+                    longtouch: false,
+                },
+                Screen::Confirm {
+                    title: "Transaction\ndata".into(),
+                    body: "12000 bytes\n(too large to\ndisplay)".into(),
+                    longtouch: false,
+                },
```
benma, line 1682: Probably it doesn't make sense to say 'Only proceed if you understand exactly what the data means', and then not show the data 🤔 

Leave as-is or reword?

Tomasvrba: fixed

benma (approved):
Nice work!

PR #1740, opened 2026-01-07
simulator-graphical: simulate Nova, same as the old simulator

benma:
It also fixes wallet creation, same issue as 5a6fea4d053ff7150e8423bbcbd4a826f80b868f.


NickeZ (approved):
tACK

PR #1742, opened 2026-01-07
device-info: change pw stretching algo strings

benma:
One-indexed is better than zero-indexed, as these strings are user visible.


NickeZ (approved):
lgtm

PR #1743, opened 2026-01-08
Refactor multisig_get/set_by_hash into HAL

cedwies:
Also adds logic to TestingHal in order to make
bitbox02-rust tests use TestingHal instead of
mock_memory

### DIFF
--- a/src/rust/bitbox02-rust/src/hal.rs
+++ b/src/rust/bitbox02-rust/src/hal.rs
@@ -85,6 +85,12 @@ pub trait Memory {
         root_pubkey_identifier_out: &mut [u8; 32],
     ) -> Result<(), ()>;
     fn get_attestation_bootloader_hash(&mut self) -> [u8; 32];
+    fn multisig_set_by_hash(
+        &mut self,
+        hash: &[u8; 32],
+        name: &str,
+    ) -> Result<(), bitbox02::memory::MemoryError>;
+    fn multisig_get_by_hash(&self, hash: &[u8; 32]) -> Option<String>;
 }
 
 /// Hardware abstraction layer for BitBox devices.
@@ -295,6 +301,18 @@ impl Memory for BitBox02Memory {
     fn get_attestation_bootloader_hash(&mut self) -> [u8; 32] {
         bitbox02::memory::get_attestation_bootloader_hash()
     }
+
+    fn multisig_set_by_hash(
+        &mut self,
+        hash: &[u8; 32],
+        name: &str,
+    ) -> Result<(), bitbox02::memory::MemoryError> {
+        bitbox02::memory::multisig_set_by_hash(hash, name)
+    }
+
+    fn multisig_get_by_hash(&self, hash: &[u8; 32]) -> Option<String> {
+        bitbox02::memory::multisig_get_by_hash(hash)
+    }
 }
 
 pub struct BitBox02Hal {
@@ -463,6 +481,7 @@ pub mod testing {
         attestation_certificate: Option<[u8; 64]>,
         attestation_root_pubkey_identifier: Option<[u8; 32]>,
         attestation_bootloader_hash: [u8; 32],
+        multisig_entries: Vec<([u8; 32], String)>,
     }
 
     impl TestingSecureChip {
@@ -612,6 +631,8 @@ pub mod testing {
         }
     }
 
+    // Same as MEMORY_MULTISIG_NUM_ENTRIES in memory.h.
+    const MULTISIG_LIMIT: usize = 25;
     impl TestingMemory {
         pub fn new() -> Self {
             Self {
@@ -629,6 +650,7 @@ pub mod testing {
                 attestation_certificate: None,
                 attestation_root_pubkey_identifier: None,
                 attestation_bootloader_hash: [0; 32],
+                multisig_entries: Vec::new(),
             }
         }
 
@@ -742,6 +764,7 @@ pub mod testing {
             self.seed_birthdate = 0;
             self.encrypted_seed_and_hmac = None;
             self.device_name = None;
+            self.multisig_entries = Vec::new();
             Ok(())
         }
 
@@ -788,6 +811,53 @@ pub mod testing {
 
         fn get_attestation_bootloader_hash(&mut self) -> [u8; 32] {
             self.attestation_bootloader_hash
+        }
+
+        fn multisig_set_by_hash(
+            &mut self,
+            hash: &[u8; 32],
+            name: &str,
+        ) -> Result<(), bitbox02::memory::MemoryError> {
+            // Validate input
+            if name.is_empty() {
+                return Err(bitbox02::memory::MemoryError::MEMORY_ERR_INVALID_INPUT);
+            }
+            // Check for duplicate name with different hash
+            for (existing_hash, existing_name) in &self.multisig_entries {
+                if existing_name == name {
+                    if existing_hash != hash {
+                        // Mirror bitbox02::memory multisig_set_by_hash semantics (duplicate-name / full-table),
+                        // even if these branches are not currently exercised in bitbox02-rust tests.
+                        return Err(bitbox02::memory::MemoryError::MEMORY_ERR_DUPLICATE_NAME);
+                    }
+                    // same name, same hash (already stored)
+                    return Ok(());
+                }
+            }
+            // Try to find existing entry with same hash
+            if let Some((_, existing_name)) = self
+                .multisig_entries
+                .iter_mut()
+                .find(|(existing_hash, _)| existing_hash == hash)
+            {
+                // rename: same hash, new name
+                *existing_name = String::from(name);
+                return Ok(());
+            }
+            if self.multisig_entries.len() >= MULTISIG_LIMIT {
+                // See comment above about mirroring bitbox02::memory semantics.
+                return Err(bitbox02::memory::MemoryError::MEMORY_ERR_FULL);
+            }
+            // Insert new entry
+            self.multisig_entries.push((*hash, String::from(name)));
+            Ok(())
+        }
+
+        fn multisig_get_by_hash(&self, hash: &[u8; 32]) -> Option<String> {
+            self.multisig_entries
+                .iter()
+                .find(|(existing_hash, _)| existing_hash == hash)
+                .map(|(_, name)| name.clone())
         }
     }
 
--- a/src/rust/bitbox02-rust/src/hww/api/bitcoin.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/bitcoin.rs
@@ -196,7 +196,7 @@ pub async fn address_multisig(
         .or(Err(Error::InvalidInput))?;
     let account_keypath = &keypath[..keypath.len() - 2];
     multisig::validate(hal, multisig, account_keypath)?;
-    let name = match multisig::get_name(coin, multisig, account_keypath)? {
+    let name = match multisig::get_name(hal, coin, multisig, account_keypath)? {
         Some(name) => name,
         None => return Err(Error::InvalidInput),
     };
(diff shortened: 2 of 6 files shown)


py/bitbox02/bitbox02/communication/generated/antiklepto_pb2.pyi
```diff
@@ -1,9 +1,7 @@
-# SPDX-License-Identifier: Apache-2.0
-
 """
 @generated by mypy-protobuf.  Do not edit manually!
 isort:skip_file
-"""
+SPDX-License-Identifier: Apache-2.0"""
```
benma, line 4: This is unrelated, and also a weird formatting bug, the license header should stay on one line with `# ` at the top. Could you check what causes this and how to fix it?

src/rust/bitbox02-rust/src/hww/api/bitcoin.rs
```diff
@@ -997,17 +997,18 @@ mod tests {
...
-                &multisig::get_hash(
-                    test.coin,
-                    &multisig,
-                    multisig::SortXpubs::Yes,
-                    &test.keypath[..test.keypath.len() - 2],
-                )
-                .unwrap(),
-                name,
+            let mut mock_hal = TestingHal::new();
+            use core::convert::TryInto;
```
benma: please put imports at the top of the module

src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -789,6 +811,55 @@ pub mod testing {
...
+        fn multisig_set_by_hash(
+            &mut self,
+            hash: &[u8; 32],
+            name: &str,
+        ) -> Result<(), bitbox02::memory::MemoryError> {
+            // Validate input
+            if name.is_empty() {
+                return Err(bitbox02::memory::MemoryError::MEMORY_ERR_INVALID_INPUT);
+            }
+            let empty_hash = [0xFFu8; 32];
```
benma: 0xFF is the erased value in mcu flash, it has no relevance here, right?

src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -463,6 +481,7 @@ pub mod testing {
         attestation_certificate: Option<[u8; 64]>,
         attestation_root_pubkey_identifier: Option<[u8; 32]>,
         attestation_bootloader_hash: [u8; 32],
+        multisig_entries: Vec<([u8; 32], String)>,
```
benma, line 484: clear this in reset_hww()?

src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -789,6 +811,55 @@ pub mod testing {
...
+            let empty_hash = [0xFFu8; 32];
+            if hash == &empty_hash {
+                return Err(bitbox02::memory::MemoryError::MEMORY_ERR_INVALID_INPUT);
+            }
+            // Check for duplicate name with different hash
+            for (existing_hash, existing_name) in &self.multisig_entries {
+                if existing_name == name {
+                    if existing_hash != hash {
+                        return Err(bitbox02::memory::MemoryError::MEMORY_ERR_DUPLICATE_NAME);
+                    } else {
```
benma: No else after return please :D I am surprised Clippy does not error on this.

src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -789,6 +811,55 @@ pub mod testing {
...
+            }
+            let empty_hash = [0xFFu8; 32];
+            if hash == &empty_hash {
+                return Err(bitbox02::memory::MemoryError::MEMORY_ERR_INVALID_INPUT);
+            }
+            // Check for duplicate name with different hash
+            for (existing_hash, existing_name) in &self.multisig_entries {
+                if existing_name == name {
+                    if existing_hash != hash {
+                        return Err(bitbox02::memory::MemoryError::MEMORY_ERR_DUPLICATE_NAME);
```
benma, line 831: DUPLICATE_NAME and FULL are not used in bitbox02-rust unit tests right now, so this is technically dead code at the moment. Okay to leave it anyway.

benma (changes requested):

cedwies:
> Haven't finished yet, but why did you change the hash type from `&[u8]` to `&[u8; 32]`? If you want to do that, consider also changing `policies::get_hash()` and `multisig::get_hash()`, to remove a bunch of the ugly `hash32 := ...` vars.

I changed it to `&[u8; 32]` because these functions only ever operate on 32‑byte SHA‑256 hashes (the C `memory_multisig_*` APis also require exactly 32 bytes), and I wanted that invariant in the type system instead of as a runtime length check.
You’re right that this now creates extra hash32 conversions from `Vec<u8>`. I can follow up by changing `multisig::get_hash()` and `policies::get_hash()` to return `[u8; 32]` directly so the call sites stay clean, or revert to `&[u8]` if we decide to prefer the original, more permissive signature. Let me know what you think.

src/rust/bitbox02-rust/src/hww/api/bitcoin.rs
cedwies: This is a left over from before my clean up. The import isn't necessary anyway :D Missed removing it.

src/rust/bitbox02-rust/src/hal.rs
cedwies: Yes. I wanted to also check for `[0xFFu8; `32]` in the higher level crate. After re-thinking,  I agree that this is a check which should be in bitbox02/memory and not in the hal implementation. Thanks!

src/rust/bitbox02-rust/src/hal.rs
cedwies: Nice, thanks!

src/rust/bitbox02-rust/src/hal.rs
cedwies: This was due to still being a draft, I plan to deal with it, thanks :)

Edit: I was leaning towards leaving them, with a comment as a hint.

benma:
> > Haven't finished yet, but why did you change the hash type from `&[u8]` to `&[u8; 32]`? If you want to do that, consider also changing `policies::get_hash()` and `multisig::get_hash()`, to remove a bunch of the ugly `hash32 := ...` vars.
> 
> I changed it to `&[u8; 32]` because these functions only ever operate on 32‑byte SHA‑256 hashes (the C `memory_multisig_*` APis also require exactly 32 bytes), and I wanted that invariant in the type system instead of as a runtime length check. You’re right that this now creates extra hash32 conversions from `Vec<u8>`. I can follow up by changing `multisig::get_hash()` and `policies::get_hash()` to return `[u8; 32]` directly so the call sites stay clean, or revert to `&[u8]` if we decide to prefer the original, more permissive signature. Let me know what you think.

Changing it sounds good. In the future, I'd prefer separate commits for clarity, but no need to split it into multiple commits now :+1:

cedwies:
Rebased and addressed the comments:

1. Removed unncessary `use core::convert::TryInto;` import
2. Removed the check for empty entry in `hal.rs`
3. Adjusted `reset_hww()`
4. Changed the return types of `multisig::get_hash()` and `policies::get_hash()`
5. Swapped `assert()` orders (was necessary since the return types of `get_hash()` changed)

Ready to review again.

src/rust/bitbox02-rust/src/hww/api/bitcoin/multisig.rs
```diff
@@ -366,24 +364,24 @@ mod tests {
...
-            get_hash(BtcCoin::Ltc, &multisig, SortXpubs::No, keypath).unwrap(),
             hex::decode("6cf181d3e131eafefd4258084e5e48366a32d59be80a0afb13345589294ccf2d")
                 .unwrap(),
+            get_hash(BtcCoin::Ltc, &multisig, SortXpubs::No, keypath).unwrap(),
         );
         assert_eq!(
-            get_hash(BtcCoin::Tltc, &multisig, SortXpubs::No, keypath).unwrap(),
             hex::decode("0e5ee1d18a74d22cf7e3255a3529b9a453e9b080005ca0bd886f6decf9e4b845")
                 .unwrap(),
+            get_hash(BtcCoin::Tltc, &multisig, SortXpubs::No, keypath).unwrap(),
```
benma, line 380: What is the point of switching all the assert_eq! args here? :grin:

cedwies: Since `get_hash()` now returns a fixed array `[u8; 32]` and `hex::decode()` returns a `Vec` the order had to be swapped (since Rust only knows how to compare `Vec<u8> == [u8; 32]`, not `[u8; 32] == Vec<u8>`). I decided it's much simpler to just swap the asserts, rather than doing another conversion.

benma: Aha I see, interesting. In that case, I recommend switching to using the hex!() macro over `hex::decode()`. `hex!("aabbcc....")` creates an array too, and is compile-time not runtime. We only recently added the hex! dep, which is why in many cases you will still see `hex::decode`. or `b"\xaa\xbb..."`

cedwies:
Reverted the order again and swapped `hex::decode()` with `hex!()`

src/rust/bitbox02-rust/src/hww/api/bitcoin.rs
```diff
@@ -322,7 +322,7 @@ mod tests {
     use super::*;
 
     use crate::bip32::parse_xpub;
-    use crate::hal::testing::TestingHal;
+    use crate::hal::{Memory, testing::TestingHal};
```
benma, line 325: Not 100% sure but I don't think this import is needed, because you are not using the Memory trait in the tests but the actual TestingMemory instance.

src/rust/bitbox02-rust/src/hww/api/bitcoin.rs
```diff
@@ -1170,14 +1167,11 @@ mod tests {
...
-            mock_memory();
             let name = "some name";
-            bitbox02::memory::multisig_set_by_hash(
-                &policies::get_hash(test.coin, &policy).unwrap(),
-                name,
-            )
-            .unwrap();
-
+            let mut mock_hal = TestingHal::new();
+            use core::convert::TryInto;
```
benma: not needed right?

src/rust/bitbox02-rust/src/hww/api/bitcoin/registration.rs
```diff
@@ -182,12 +184,15 @@ mod tests {
     use bitbox02::testing::mock_memory;
     use util::bip32::HARDENED;
 
+    use crate::hal::{Memory, testing::TestingHal};
```
benma: Same as above, probably the Memory import is not needed

src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs
```diff
@@ -1288,7 +1288,7 @@ pub async fn process(
 mod tests {
     use super::*;
     use crate::bip32::parse_xpub;
-    use crate::hal::testing::TestingHal;
+    use crate::hal::{Memory, testing::TestingHal};
```
benma, line 1291: Same

src/rust/bitbox02-rust/src/hww/api/bitcoin.rs
cedwies: We need this import since we are calling a trait method from `Memory`, not just using `TestingHal`.

src/rust/bitbox02-rust/src/hww/api/bitcoin/registration.rs
cedwies: Here you are correct, since we already import `use crate::hal::{Memory, Ui};` at the top of the file. So in this file `Memory` is imported twice.

src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs
cedwies: Here we need it again, since use `crate::hal::Memory;` is not imported at the top of the file.

cedwies:
Squashed the commits, no code change.

PR #1747, opened 2026-01-13
releases: add v9.25.0

benma:
Since 011c8aa893dcf48e47268de0f52a839ee8fbc4eb we don't tag and release the bitcoin-only variant separately anymore, so in the releases folder we also consolidate both into one folder for consistency. This way we can link to one folder from the release page, which now contains both variants.


.ci/check-release-sigs
```diff
@@ -6,8 +6,14 @@ set -x
 # Import all keys
 find ./releases/pubkeys -name '*.asc' -exec gpg --quiet --import {} \;
 
-# Verify all sigs.
-find ./releases -name '*.sig' |
-    while read file; do
-        gpg --verify $file $(dirname $file)/assertion.txt 2>/dev/null
+# Verify all sigs. Find all assertion files - both old formats (just assertio.txt until v9.24.0) and
```
Beerosagos: typo: assertio.txt

PR #1749, opened 2026-01-22
Initial bb03 simulator

NickeZ:
New simulator with prototoype device image as background and prototype screen resolution. Removing bb02 specific code to be done.


cedwies:
Not done yet, started testing. `--preseed` path panics with SecureChip(-7) since simulator memory config leaves `securechip_type` at default -> Atecc. Just to check: We intentionally consider this option non-functional for this PR or do you intend to fix this before merging?

test/simulator-graphical-bb03/src/main.rs
```diff
@@ -0,0 +1,682 @@
...
+    inbound_in: mpsc::Sender<[u8; 64]>,
+    counter: Arc<AtomicU32>,
+) {
+    let mut buf = [0; 1024];
+    while let Ok(len) = stream.read(&mut buf) {
+        if len == 0 {
+            // Client disconnected
+            break;
+        }
+        for i in 0..len / 64 {
```
cedwies, line 67: is it just me or is this discarding left-over bytes? with `len = 130`, `i = 0` and `i = 1`, but then 2 bytes are left?

NickeZ:
> Not done yet, started testing. `--preseed` path panics with SecureChip(-7) since simulator memory config leaves `securechip_type` at default -> Atecc. Just to check: We intentionally consider this option non-functional for this PR or do you intend to fix this before merging?

Thanks for catching that! I basically did the initial simulator before we fixed this issue in the bb02 simulator.

Pushed a fix

test/simulator-graphical-bb03/src/main.rs
```diff
@@ -0,0 +1,683 @@
...
+        let screen_id = canvas
+            .create_image(
+                ImageSource::try_from(&*SCREEN_FB.lock().unwrap()).unwrap(),
+                ImageFlags::NEAREST,
+            )
+            .unwrap();
+
+        canvas.save();
+        canvas.reset();
+        canvas.restore();
```
cedwies: Why do we need this block?

test/simulator-graphical-bb03/src/main.rs
```diff
@@ -0,0 +1,683 @@
...
+                            }
+                        }
+                    }
+                }
+                if !disconnected {
+                    self.inbound_out = inbound_out;
+                }
+                // Send data to TCP Client
+                loop {
+                    if let Some(data) = bitbox02::queue::pull_hww() {
```
cedwies, line 512: If no client is connected (`self.outbound_in` is `None`), `pull_hww()` is still called and still  removes packets from the firmware queue but then drop them since there is not `self.outbound_in` (data loss?)
Maybe that is even okay but we could also change to only drain the queue if there is a connected client?

like:
**if** `outbound_in.is_some()` **then** _pull+send in a loop_
**else** _leave queue untouched._

test/simulator-graphical-bb03/src/main.rs
```diff
@@ -0,0 +1,683 @@
...
+        let el_proxy = event_loop.create_proxy();
+        let Ok(listener) = TcpListener::bind("0.0.0.0:15423") else {
+            return Err(Box::new(AppError::new("Failed to bind to address!")));
+        };
+        move || {
+            // Use this counter to ensure that we only connect to a single client.
+            let counter = Arc::new(AtomicU32::new(0));
+
+            for stream in listener.incoming() {
+                if ACCEPTING_CONNECTIONS.load(Ordering::Relaxed)
```
cedwies, line 623: If `stream` is `Err(e)` but we are **not** accepting yet (or busy) then we go into `else` branch and print `Busy, won't accept new clients` when in reality we did not get a valid connection in the first place.

test/simulator-graphical-bb03/src/main.rs
NickeZ: The protocol for communicating with the bb02 is defined to use 64 byte packets due to usb generic hid using that.

test/simulator-graphical-bb03/src/main.rs
NickeZ: It seems we don't, thanks for finding that! Probably some left over from some copied example.

test/simulator-graphical-bb03/src/main.rs
NickeZ: I think I prefer that a newly connected client doesn't get eventual "garbage" from the previous connected one. With real hardware I think it wouldn't stop sending just because the App has died.

test/simulator-graphical-bb03/src/main.rs
NickeZ: What you are basically saying is that I forgot to clear the counter `counter`?

NickeZ:
@cedwies fixed comments and rebased on latest master

test/simulator-graphical-bb03/src/main.rs
cedwies: Optional nit:

In `UserEvent::NewConnection`, we could call a `queue_clear()` to guarantee a clean slate when a new client attaches. Behavior would be the same as right now, but it would be more explicit.
Or leave comment saying that if no client is connected, we still pull and drop HWW queue frames, to mimic hardware behavior where unsent frames are not buffered across host reconnections.

test/simulator-graphical-bb03/src/main.rs
cedwies: Yes, exactly. Thanks for the fix.

Remaining issue is just that the “Busy” log message does not differentiate between:
“We are not accepting yet / already have a client” and “There was an I/O error from listener.incoming()”
But after re-checking I think it's fine to leave as is.

test/simulator-graphical-bb03/src/main.rs
cedwies: Yes but TCP does not guarantee message boundaries. A safer pattern would be to buffer leftovers across reads. The OS networking stack is allowed to: Combine multiple `send()` calls into one `recv()` delivery or
split one `send()` across multiple `recv()` calls, no? But I saw BB02 Simulator uses the same pattern, and since this is a simulator, it should be fine.

NickeZ: Technically correct (the best kind of correct), but I think in practice it will never be a problem if both apps are on the same host.

PR #1751, opened 2026-01-27
HAL: add reboot_to_bootloader()

cedwies:


src/rust/bitbox02-rust/src/hal.rs
```diff
@@ -861,12 +879,27 @@ pub mod testing {
...
+
+    impl TestingSystem {
+        pub fn new() -> Self {
+            Self
+        }
+    }
+
+    impl super::System for TestingSystem {
+        fn reboot_to_bootloader(&mut self) -> ! {
+            panic!("reboot_to_bootloader called")
```
benma: This works with the existing tests, but now that it's in HAL, we can improve the style of the tests to not use panics and catch_unwind.

You could set a `self.reboot_to_bootloader_called = true` var and check it is set in the unit tests, or maybe make it a counter in case we want to check it was only called once.

benma:
Also this should be now unused and can be removed, right?

in bitbox02/src/lib.rs:

```Rust
#[cfg(feature = "testing")]
pub fn reboot_to_bootloader() -> ! {
    panic!("reboot_to_bootloader called")
}
```

src/rust/bitbox02-rust/src/hal.rs
cedwies: I agree it would be nice, but this clashes with the return type. If we don't panic, we hang since we would have to loop forever. But I will look into this.

cedwies:
> Also this should be now unused and can be removed, right?
> 
> in bitbox02/src/lib.rs:
> 
> ```rust
> #[cfg(feature = "testing")]
> pub fn reboot_to_bootloader() -> ! {
>     panic!("reboot_to_bootloader called")
> }
> ```

Yes, if we remove the `cfg` from the other `reboot_to_bootloader()` function

cedwies:
Changes:
- TestingSystem now uses a counter which counts how many times `reboot_to_bootloader()` was called
- Remove conditional compilation from `bitbox02::reboot_to_bootloader()`
  (no longer needed since tests use `TestingHal`, not the `bitbox02` crate directly)
- Replace panic-catching test pattern with counter-based verification
 
Important: The `bitbox02::reboot_to_bootloader()` is now unconditional and therefore always compiled. I think this is okay, since tests uses TestingHal and production code uses `bitbox02` call. It would just be a problem if a test uses `bitbox02::reboot_to_bootloader()`, but in the worst case, a test gets stuck in a loop.

src/rust/bitbox02-rust/src/hal.rs
benma: I think your instinct was right and my idea was wrong, I did not consider the return type would have to change and that it would lead to needing to send a "fake" success response that is actually not used in reality. It seems worthwhile to use the correct return type.

Sorry for pointing you in the wrong direction. Could you revert it back to ` -> !` and panic in TestingSystem?

cedwies: No problem, I think it was worth to test it out. I am a bit unsure about this. The observable behavior of the tests is much nicer now, reboot behavior did not change. But I understand that the type signature lacks and we have a weaker compile-time guarantee, so theoretically:
```rs
hal.system().reboot_to_bootloader();
do_something_important();
```
is possible.

When reverting back it would be nice to get rid of `catch_unwind` and just use `should_panic`. In tests, can't we assert e.g. `screens` before rebooting?

NickeZ: Testing diverging functions seems conceptually quite hard. Almost like you need to run the test as a separate process and ensure that it exits with some exit code...

benma: I think I did the tests using panic+catch_unwind for this reason.

NickeZ: Maybe not something we need to change now, but I think it looks a bit strange in the output whenever we use panics in test:
```
test hww::api::system::tests::test_reboot_to_bootloader ...
thread 'hww::api::system::tests::test_reboot_to_bootloader' (4105721) panicked at bitbox02/src/lib.rs:231:5:
reboot_to_bootloader called
ok
```

It doesn't really scream confidence :P

Worse is:
```

test async_usb::tests::test_full_cycle ...
thread 'async_usb::tests::test_full_cycle' (4105516) panicked at bitbox02-rust/src/async_usb.rs:95:14:
spawn: wrong state

thread 'async_usb::tests::test_full_cycle' (4105516) panicked at bitbox02-rust/src/async_usb.rs:95:14:
spawn: wrong state

thread 'async_usb::tests::test_full_cycle' (4105516) panicked at bitbox02-rust/src/async_usb.rs:199:16:
range end index 4 out of range for slice of length 1

thread 'async_usb::tests::test_full_cycle' (4105516) panicked at bitbox02-rust/src/async_usb.rs:95:14:
spawn: wrong state

thread 'async_usb::tests::test_full_cycle' (4105516) panicked at bitbox02-rust/src/async_usb.rs:95:14:
spawn: wrong state

thread 'async_usb::tests::test_full_cycle' (4105516) panicked at bitbox02-rust/src/async_usb.rs:199:16:
range end index 4 out of range for slice of length 1

thread 'async_usb::tests::test_full_cycle' (4105516) panicked at bitbox02-rust/src/async_usb.rs:95:14:
spawn: wrong state

thread 'async_usb::tests::test_full_cycle' (4105516) panicked at bitbox02-rust/src/async_usb.rs:95:14:
spawn: wrong state

thread 'async_usb::tests::test_full_cycle' (4105516) panicked at bitbox02-rust/src/async_usb.rs:199:16:
range end index 4 out of range for slice of length 1
ok
```

cedwies: I suggest reverting back to the previous commit.
I would also look into suppressing prints for the expected panics to scream more confidence :D But (if this even works how I imagine) I open a separate PR.

cedwies:
Reverted back to my first commit.

Also removed the testing `reboot_to_bootloader()` function in `lib.rs` since the testing functions now go through testing hal.

PR #1752, opened 2026-01-28
simulator: persist securechip type in fake memory

bznein:
Ensure FAKE_MEMORY_FILEPATH mode writes the securechip type to the shared fake memory file so the simulator doesn't default to ATECC.

### DIFF
--- a/test/hardware-fakes/src/fake_memory.c
+++ b/test/hardware-fakes/src/fake_memory.c
@@ -68,9 +68,13 @@ void fake_memory_factoryreset(void)
 
 bool fake_memory_nova(void)
 {
-    chunk_shared_t* shared_ptr = (chunk_shared_t*)&_memory_shared_data[0];
-    shared_ptr->fields.platform = MEMORY_PLATFORM_BITBOX02_PLUS;
-    shared_ptr->fields.securechip_type = MEMORY_SECURECHIP_TYPE_OPTIGA;
+    chunk_shared_t shared = {0};
+    memory_read_shared_bootdata_fake(shared.bytes);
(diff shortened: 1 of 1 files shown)


test/hardware-fakes/src/fake_memory.c
```diff
@@ -72,6 +78,11 @@ bool fake_memory_nova(void)
     shared_ptr->fields.platform = MEMORY_PLATFORM_BITBOX02_PLUS;
     shared_ptr->fields.securechip_type = MEMORY_SECURECHIP_TYPE_OPTIGA;
 
+    const char* base_path = getenv(FAKE_MEMORY_ENV_VAR);
+    if (base_path) {
+        _write_file_chunk(SHARED_SUFFIX, 0, _memory_shared_data, FLASH_SHARED_DATA_LEN);
+    }
+
```
benma: I am not sure, but it seems that if the memory is file based, the static `_memory_shared_data` var should not be used? Here you overwrite the file with whatever is in `_memory_shared_data`, but it was not read from file before, right? So it would be overwriting the persisted shared memory every time the simulator launches.

The right way that works with both RAM or file backend could be to:

```
chunk_shared_t shared = {0}; 
memory_read_shared_bootdata_fake(&shared_ptr.bytes);
// set shared chunk platform and securechip type fields
if (!memory_write_to_address_fake(FLASH_SHARED_DATA_START, 0, &shared.bytes)) return false;
```

Wdyt?

bznein: Yeah you are right, my solution only worked on the first launch! I tried with your proposal and confirmed that it works when also restarting the simulator again! Both with file backend (seed is persisted) and RAM backend (each new launch is a "new" wallet)

PR #1753, opened 2026-01-29
build: Avoid hard coding path to compiler

NickeZ:
This commit allows developers to choose compiler by modifying PATH

why: beacuse I don't want to install the compiler to `/usr/local/bin` because the arm toolchain installer wants to put it in `/Applications`

### DIFF
--- a/arm.cmake
+++ b/arm.cmake
@@ -1,10 +1,7 @@
 set(CMAKE_SYSTEM_NAME "Generic")
 set(CMAKE_SYSTEM_PROCESSOR "arm")
 
-set(TOOLCHAIN_PREFIX_PREFIX "arm-none-eabi")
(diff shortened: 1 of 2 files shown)


arm.cmake
```diff
@@ -3,8 +3,7 @@ set(CMAKE_SYSTEM_PROCESSOR "arm")
 
 set(TOOLCHAIN_PREFIX_PREFIX "arm-none-eabi")
 set(TOOLCHAIN_PREFIX "${TOOLCHAIN_PREFIX_PREFIX}-")
```
benma: If you like, also inline `TOOLCHAIN_PREFIX_PREFIX` here and remove the `TOOLCHAIN_PREFIX_PREFIX` var to simplify, it's a lot of prefixing otherwise :grin:

NickeZ: removed both variables, opted to do runtime detection of the target triple as well.

PR #1754, opened 2026-02-01
Hal: Add screen_saver_enable/disable to HAL

cedwies:
`screen_saver_disable()` and `screen_saver_enable()` were called from `option_no_screensaver()` via the `bitbox02` crate. Passing **Hal** through to `option_no_screensaver()` to reach these functions causes borrow conflicts (especially when **Hal** is also needed somewhere else in the same async flow).

To avoid this, screensaver control is exposed on the Workflows trait and `option_no_screensaver()` now works with `&mut hal_ui` instead of needing to reach into `bitbox02`. This keeps bitbox02-rust screensaver handling behind HAL/Ui boundary.

`unlock_animation::animate()` is a special, as it runs in parallel with `keystore::unlock_bip39()` inside `unlock_bip39()`. If `animate()` takes `Ui`, there are two borrows of Hal/Ui in the `zip()`. To work around this, we move the screensaver disable/enable responsibility to `unlock_bip39()`

Note that for review this is split into 2 commits (first commit is the change itself, second commit is regeneration of protobuf messages)


cedwies:
This PR is closed to be postponed after a bigger refactor.

PR #1755, opened 2026-02-02
py/bitbox02: list_backups returns tz aware timestamps

NickeZ:
Naive timestamps without timezone are by many python functions
interpreted as being in the local timezone. The backups are timestamped
with UTC so we return timmezone aware timestamps instead so that python
tooling correctly interpret them as UTC.

Fixes a bug in send_message.py that it would display incorrect
timestamps.

### DIFF
--- a/py/bitbox02/CHANGELOG.md
+++ b/py/bitbox02/CHANGELOG.md
@@ -1,10 +1,10 @@
 # Changelog
 
 ## [Unreleased]
-
-# 7.1.0
 - Add `btc_xpubs()` to fetch multiple xpubs at once
 - Bitcoin: add support for OP_RETURN outputs
+- Add `change_password()`
+- Backups: return non-naive timestamps with UTC timezone
 
 # 7.0.0
 - get_info: add optional device initialized boolean to returned tuple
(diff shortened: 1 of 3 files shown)


NickeZ:
@benma like so?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ## Firmware
 
 ### [Unreleased]
+- send_message.py: Fix incorrect timestamps of backups
```
benma: I think this should be removed, because it's not a firmware change and does not impact the firmware release.

py/bitbox02/CHANGELOG.md
```diff
@@ -1,10 +1,12 @@
 # Changelog
 
 ## [Unreleased]
+- Backups: return non-naive timestamps with UTC timezone
```
benma: This should also go into the 7.1.0 section as 7.1.0 is not released yet.

NickeZ: Can we use the same workflow in both of the changelog files? Can I remove 7.1.0 until we release it?

NickeZ:
@benma updated again, force pushed this time as I rewrote the commit message.

py/bitbox02/CHANGELOG.md
benma: Okay, or alternatively I could just release it right after this is merged. Let me know.

PR #1756, opened 2026-02-02
py: handle busy BitBox connection errors

strmci:
Catch hid open_path failures in send_message and show a user-friendly message instead of a traceback.


NickeZ:
Tested and merged, thanks 🙏

PR #1757, opened 2026-02-02
rust: Avoid modification to lockfile during build

NickeZ:
In CI we want the build to fail if the lockfile needs to be updated so that we don't forget to commit the new lockfiles when the cargo config is changed.

Proof that it worked before I updated the lockfiles: https://github.com/BitBoxSwiss/bitbox02-firmware/actions/runs/21599837976/job/62242509527

### DIFF
--- a/.github/workflows/ci-common.yml
+++ b/.github/workflows/ci-common.yml
@@ -14,6 +14,7 @@ on:
         required: true
         type: string
 env:
+  CARGOFLAGS: "--locked"
   RUSTFLAGS: "-Dwarnings"
   CFLAGS: "-Werror"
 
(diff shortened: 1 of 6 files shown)


NickeZ:
> How does one then update the lockfile if one changes deps?
> 
> A possible alternative could be to fail CI if it detects a dirty lockfile after a build/test.

It only fails on CI you can try locally with

```
make clean
CARGOFLAGS=--locked make simulator-graphical
```

Important to clean as only environment flags set before cmake makes the build directory counts.

---

So during normal development the lockfile is updated as before.

---

edit:

The whole purpose of the `--locked` flag is CI. https://doc.rust-lang.org/cargo/commands/cargo-build.html#option-cargo-build---locked

benma (approved):

PR #1758, opened 2026-02-02
Nickez/py format

NickeZ:
CI never ran on a PR that modified send_message 🤷


NickeZ:
force pushed to trigger CI now that github actions is up again

PR #1759, opened 2026-02-03
Bug: Fix BLE identity address offset in memory_reset_hww()

cedwies:
When generating a new ble IRK and a new ble identity address, we generate 32 random bytes and want to use random[0-16) for IRK and random[16,21) for identity address.

The current implementation has a bug which causes the identity address to be random[6,12), resulting in an overlap. This means that the ble identity address exposes 6 bytes of our ble identity resolving key.


NickeZ:
Very nice catch! I'll test it

cedwies:
Squashed, no change.

PR #1760, opened 2026-02-03
fatfs wrapper

NickeZ:
Simplify CMake files by linking with cargo/rust instead.


NickeZ:
@benma ready for review

src/rust/bitbox02-rust/Cargo.toml
```diff
@@ -37,6 +37,7 @@ minicbor = { version = "0.24.0", default-features = false, features = ["alloc"],
 crc = { workspace = true, optional = true }
 ed25519-dalek = { version = "2.1.1", default-features = false, features = ["hazmat", "digest"], optional = true }
 hmac = { workspace = true }
+fatfs-sys = { path = "../fatfs-sys"}
```
benma, line 40: Okay here for now, but I think we will move this (and the linking part in lib.rs) to a new crate `bitbox-sd` at some point soon, which will also contain the implementation of the `hal::Sd` trait, which then can be used by bitbox02 and bitbox03 HAL implementations - I assume both will use the same sdcard C code.



(bitbox02-rust is using HAL to access sd-card functionality, and while the BitBox02Hal instance is currently also in this crate, it will eventually move to another crate, I assume to the bitbox02 crate).

NickeZ: I assume they can share code that calls fatfs symbols. But they will for sure have differnet `diskio` implementations. (STM HAL vs ASF HAL)

PR #1762, opened 2026-02-03
Rust-wrapped ringbuffer/set_product/set_name/power_down

benma:


src/rust/bitbox02/src/da14531.rs
```diff
@@ -0,0 +1,49 @@
...
+        let mut queue = RingBuffer::new(&mut buf);
+
+        set_product(product, &mut queue);
+
+        let actual = drain(&mut queue);
+
+        let mut payload = vec![CTRL_CMD_PRODUCT_STRING];
+        payload.extend_from_slice(product.as_bytes());
+        let mut expected = vec![0u8; 140];
+        let expected_len = protocol_format(&mut expected, ProtocolPacketType::CtrlData, &payload);
```
NickeZ, line 59: Actually, I think I would prefer `expected` to be hard coded and not generated with `protocol_format`. Seems like the test uses code it tests?

benma:
> tACK, works with iOS

How did you test that, seeing that these functions are not yet used? :smile: 

They are used in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1772, did you test that one?

src/rust/bitbox02/src/da14531.rs
benma: Makes sense, will fix.

benma: Then again, that would unit test protocol_format once more in a way, which is already tested :smile: 

I now added the hard coded literal comparison, but left the existing one too. I figure it's a good reference for when we port it to Rust.

NickeZ:
> > tACK, works with iOS
> 
> How did you test that, seeing that these functions are not yet used? 😄
> 
> They are used in #1772, did you test that one?

haha, good point, I was very tired and thought they were used in the fw... :|

src/rust/bitbox02/src/da14531.rs
NickeZ: hmm, yeah, fair point.. it probably is best to use `protocol_format` to avoid testing the formatting. but wouldn't it be enough to check like queue length > 0 if you really didn't want to test the formatting?

benma: Imho it's fine as it is, the test is easier to understand and maintain too by using protocol_format.

PR #1766, opened 2026-02-03
maintenance: constify some read only things

benma:
One of the commits of https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1690/


benma:
Will merge after https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1728 as there is a conflict with it

PR #1769, opened 2026-02-04
Avoid string literal hex's

NickeZ:
Starting with GCC-15 there is a warning if a static string isn't null terminated. Use array literals intead of strings to avoid the warning.


src/factorysetup.c
```diff
@@ -110,106 +111,806 @@ static uint8_t _root_pubkey_bytes[][ROOT_PUBKEY_SIZE] = {
...
-    "\x04\xd6\x58\x31\x02\x9a\xe2\x1c\x55\x4c\x2b\xa4\x42\xa5\x5b\x92\x80\x2a\x82\x84\x06\x80\xb8\xb9\x5c\xf4\xb9\x39\xf7\x97\x2f\x01\x38\x83\xba\x99\x81\x39\xaa\xe8\x96\xaf\xf4\x1b\xb2\xeb\x73\xe1\xa2\xfb\x1b\xfa\xaf\x19\x58\x82\x3b\x87\xa2\xb1\x83\x08\x96\x98\x1b",
-    "\x04\x8b\x86\xb0\x98\x4f\xcc\x63\x43\x4b\x12\xdc\x57\xf2\xc3\x89\x6e\x78\x51\xcd\xb7\xed\x90\x7b\x18\x78\x0f\xf7\x5b\x47\xef\xf7\x38\x35\x5f\xc2\xb0\x28\xdc\x2e\x16\xae\xdc\x63\xf3\xdd\x8d\x86\x4f\xc4\x50\xe1\x41\xcc\x0f\x21\xb4\x16\x67\x6c\x69\x3c\xf5\x7f\xfd",
-    "\x04\x78\x70\xa2\x56\xd9\xac\x4e\x25\x3c\x24\x7d\xc7\x1a\xd4\x04\xc3\x60\xf1\x91\xcb\x40\xec\xbb\x81\x00\x4b\x6c\x8a\x80\xfb\xf6\x10\x20\x1b\xb5\x7d\xed\x75\x87\x46\x20\xa8\xaa\x1e\xa8\xcd\xcd\x60\x04\xe8\x42\x42\x81\x0f\xe4\xc0\x3c\x0a\xc8\x3d\xd4\x2e\x57\x0a",
-    "\x04\xc6\x15\x92\x83\x41\x2e\xf6\xcf\xa7\x46\x60\x56\x5a\xa5\x1d\x3f\xee\x4e\x7f\x85\x2d\xa6\x2f\x9b\x33\x63\x3f\x62\x6c\xda\x4d\xfa\x8c\x39\x37\x49\x40\x64\x0b\x94\xfc\x07\x5d\x15\x75\x54\x5c\xc7\x68\x5c\x96\x91\x37\xb2\x7e\x9d\x0d\x17\x8f\xeb\x2c\xa0\x06\x23",
+    {
```
benma, line 113: Remove the clang-format off/on markers around this array and run clang-format for consistent number of elements per line.

scripts/reformat_hex_strings.py
```diff
@@ -0,0 +1,326 @@
+#!/usr/bin/env python3
```
benma: Imho this is not something to be added to the repo, for these reasons:

- The number of fixes are small, especially going forward - having a fairly complicated script to maintain for that seems overkill to me
- Can just ask an agent to fix instances should they arise, or do it manually, as it's not difficult

src/optiga/optiga.c
```diff
@@ -1823,7 +1823,8 @@ bool optiga_gen_attestation_key(uint8_t* pubkey_out)
     // Parse DER "BIT STRING", see Solution Reference Manual 6.2.2,
     // example for ECC NIST-P256.
     // The 64 byte X/Y values are the last 64 bytes.
-    if (pubkey_der_size != 68 || !MEMEQ(pubkey_der, "\x03\x42\x00\x04", 4)) {
+    uint8_t magic[] = {0x03, 0x42, 0x00, 0x04};
```
benma: add `const`?

PR #1777, opened 2026-02-06
inital async-engine, rebased

benma:
ports orientation screen and delay to show that waking works

Rebase of https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1695


src/rust/bitbox02/src/delay.rs
```diff
@@ -1,128 +1,117 @@
...
-}
+pub async fn delay_for(duration: Duration) {
+    use core::cell::RefCell;
+    use core::ffi::c_void;
+    // Shared between the async context and the c callback
 
-pub struct Delay {
-    inner: DelayInner,
-}
+    let mut bitbox02_delay = bitbox02_sys::delay_t { id: 0 };
```
benma, line 23: In https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1695, this was inside the shared state struct - I moved it out as it is not shared with C callback.

benma:
@NickeZ fyi this PR adds ~3kB to the binary.

src/rust/bitbox02/src/delay.rs
```diff
@@ -1,128 +1,117 @@
...
-}
 
-pub async fn delay_for(#[allow(unused)] duration: Duration) {
-    // Do not delay in (non-graphical) simulator and Rust unit tests.
-    #[cfg(not(any(feature = "c-unit-testing", feature = "testing")))]
-    Delay::from_ms(duration.as_millis() as u32).await
+    let shared_state = Arc::new(Mutex::new(SharedState {
+        waker: None,
+        result: None,
+        handle: None,
```
NickeZ: The `handle` is also never shared right?

src/rust/bitbox02/src/delay.rs
```diff
@@ -83,7 +83,10 @@ pub async fn delay_for(duration: Duration) {
     let mut handle: Option<std::thread::JoinHandle<()>> = Some(std::thread::spawn({
         let shared_state = Arc::clone(&shared_state);
         move || {
+            // Do not delay in (non-graphical) simulator and Rust unit tests.
+            #[cfg(not(any(feature = "c-unit-testing", feature = "testing")))]
```
NickeZ: seems unnecessary to even spawn a thread if we aren't going to wait for it. Why not just put:

```rust
#[cfg(any(feature = "c-unit-testing", feature = "testing"))]
return;
```

at the beginning of `delay_for`?

src/rust/bitbox02/src/delay.rs
benma: Fixed

src/rust/bitbox02/src/delay.rs
benma: Right. I moved it to a separate empty function -I didn't like nested `cfg` and thought clippy would probably complain about all the "dead code" after the return

src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -469,30 +472,54 @@ where
...
-        // this C callback is guaranteed to be called only once.
-        let on_done = unsafe { Box::from_raw(param as *mut F2) };
-        on_done(upside_down);
+pub async fn choose_orientation() -> bool {
+    // Shared between the async context and the c callback
+    struct SharedState {
+        waker: Option<Waker>,
+        result: Option<bool>,
+    }
+    let shared_state = Arc::new(RefCell::new(SharedState {
```
NickeZ: Both the async context and the c callback are executed on the same thread. So unless the compiler complains, you can switch `Arc` to `Rc`.

benma: Nice, it works. See fixup commit.

src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -469,30 +472,54 @@ where
...
-        // this C callback is guaranteed to be called only once.
-        let on_done = unsafe { Box::from_raw(param as *mut F2) };
-        on_done(upside_down);
+pub async fn choose_orientation() -> bool {
+    // Shared between the async context and the c callback
+    struct SharedState {
+        waker: Option<Waker>,
+        result: Option<bool>,
+    }
+    let shared_state = Rc::new(RefCell::new(SharedState {
```
NickeZ: `Box`?

src/rust/bitbox02/src/delay.rs
```diff
@@ -1,128 +1,120 @@
...
-                }
-            }));
-            (handle, done)
-        };
-        Delay {
-            inner: DelayInner {
-                thread_handle,
-                done,
-            },
+    let shared_state = Arc::new(RefCell::new(SharedState {
```
NickeZ: `Box`?

src/rust/bitbox02/src/ui/ui.rs
benma: Done

src/rust/bitbox02/src/delay.rs
benma: Done. In the simulator version with threads, codex also calls `join` to avoid dangling pointers in the thread that might still run after the future is cancelled - see fixup commit msg body.

benma: Reverted the simulator change

src/rust/bitbox02/src/delay.rs
```diff
@@ -1,128 +1,121 @@
...
-pub async fn delay_for(#[allow(unused)] duration: Duration) {
-    // Do not delay in (non-graphical) simulator and Rust unit tests.
-    #[cfg(not(any(feature = "c-unit-testing", feature = "testing")))]
-    Delay::from_ms(duration.as_millis() as u32).await
+    let shared_state = Arc::new(Mutex::new(SharedState {
+        waker: None,
+        result: None,
+    }));
+
+    let mut handle: Option<std::thread::JoinHandle<()>> = Some(std::thread::spawn({
```
NickeZ, line 91: Does the `handle` really need to be an `Option`?

benma: Seems so, since we call `.join()` manually which consumes the handle

PR #1779, opened 2026-02-06
ui: port confirm component to be awaitable, reduce callback soup

benma:
Snippet to test it in the main_loop.rs using EXECUTOR:

```

            EXECUTOR
                .spawn(async {
                    let result = bitbox02::ui::confirm(&bitbox02::ui::ConfirmParams {
                        title: "Confirm",
                        body: "Dummy",
                        ..Default::default()
                    })
                    .await;
                    crate::workflow::status::status(&format!("Result: {:?}", result), true).await;
                })
                .detach();
```

### DIFF
--- a/src/rust/bitbox02/src/ui/ui.rs
+++ b/src/rust/bitbox02/src/ui/ui.rs
@@ -139,13 +139,29 @@ pub async fn confirm(params: &ConfirmParams<'_>) -> bool {
         }
     }
 
-    let mut title_scratch = Vec::new();
-    let mut body_scratch = Vec::new();
+    // We truncate at a bit higher than MAX_LABEL_SIZE, so the label component will correctly
+    // truncate and append '...'.
+    const TRUNCATE_SIZE: usize = bitbox02_sys::MAX_LABEL_SIZE as usize + 1;
+    let title =
+        util::strings::str_to_cstr_vec(util::strings::truncate_str(params.title, TRUNCATE_SIZE))
+            .unwrap();
+    let body =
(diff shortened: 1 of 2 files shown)


src/rust/bitbox02/src/screen_saver.rs
```diff
@@ -13,3 +13,20 @@ pub fn screen_saver_disable() {
         bitbox02_sys::screen_saver_disable();
     }
 }
+
+pub struct ScreensaverGuard {
```
NickeZ: Would it be OK to call it `ScreensaverInhibitor`? And a comment on the struct that it will inhibit the screensaver until it is dropped?

benma: Done now.

src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -469,30 +512,54 @@ where
...
+            waker.wake_by_ref();
+        }
+    }
+
     let component = unsafe {
         bitbox02_sys::orientation_arrows_create(
-            Some(c_on_done::<F>),
-            Box::into_raw(Box::new(on_done)) as *mut _, // passed to c_on_done as `param`.
+            Some(callback),
+            Rc::into_raw(Rc::clone(&shared_state)) as *mut _, // passed to callback as `user_data`.
```
benma: If the confirm future is dropped before the callback is called, then this is a leak. 

The same issue exists already on master with Box::leak, so it's not a regression, but I'll see if we can fix it now that we are touching this.

benma: fixed in fixup commit

NickeZ (approved):
tACK

PR #1785, opened 2026-02-11
Add coin purchase memo support

cedwies:
Gate CoinPurchase handling behind a new app-swap feature to limit to BitBox Multi.

One could decide to not gate behind a feature (also including it in BTC only version) since this adds negligible binary size, introduces no new attack surface (receiving a`CoinPurchaseMemo` requires a valid signature from a known identity), and the protobuf type cannot be cfg-gated anyway since it is generated code.

I chose the strict way and gated behind a new feature, but it may be discussed.


benma:
@codex review

messages/btc.proto
```diff
@@ -246,8 +246,14 @@ message BTCPaymentRequestRequest {
     message TextMemo {
       string note = 1;
     }
+    message CoinPurchaseMemo {
+      uint32 coin_type = 1;
+      string amount = 2;
+      string address = 3;
```
benma: We need to add a field here about the address derivation details, so we can verify the address belongs to the BitBox. It should be a `oneof`, as the message will depend on the coin_type.

Maybe as a start, we can add one oneof variant for ETH, which I think only needs the keypath.

cedwies:
Added address derivation field (only eth for now).
I used a dummy derivation path in the tests, so they verify that the field is excluded for the sighash

src/rust/bitbox02-rust/src/hww/api/bitcoin/payment_request.rs
```diff
@@ -204,6 +216,28 @@ mod tests {
...
+        Memo {
+            memo: Some(memo::Memo::CoinPurchaseMemo(memo::CoinPurchaseMemo {
+                coin_type,
+                amount: amount.into(),
+                address: address.into(),
+                address_derivation: Some(memo::coin_purchase_memo::AddressDerivation::Eth(
+                    memo::coin_purchase_memo::EthAddressDerivation {
+                        keypath: vec![
+                            44 + util::bip32::HARDENED,
+                            60 + util::bip32::HARDENED,
```
benma: Maybe `coinType + ` instead of `60 +` here?

benma:
> Added address derivation field (only eth for now). I used a dummy derivation path in the tests, so they verify that the field is excluded for the sighash

Nice! How about adding validation of it to `fn validate` in payment_request.rs?

src/rust/bitbox02-rust/src/hww/api/bitcoin/payment_request.rs
benma: Maybe `coin_type +` here instead of `60 +`?

cedwies:
To validate the eth keypath I made `is_valid_keypath_address(keypath: &[u32])` available. Modified `Cargo.toml` because app-swap should only be available with app-ethereum. Otherwise, if only app-swap is enabled and not app-ethereum, the code will not compile.

src/rust/bitbox02-rust/src/hww/api/bitcoin/payment_request.rs
```diff
@@ -202,6 +202,21 @@ pub fn validate(
...
+        if let Memo {
+            memo: Some(memo::Memo::CoinPurchaseMemo(coin_purchase_memo)),
+        } = memo
+        {
+            if let Some(memo::coin_purchase_memo::AddressDerivation::Eth(eth)) =
+                &coin_purchase_memo.address_derivation
+            {
+                if !super::super::ethereum::is_valid_keypath_address(&eth.keypath) {
+                    return Err(ValidationError::Other);
+                }
```
benma: You also need to verify the address provided in the SLIP24 signed payload matches the address at that keypath.

I suggest to not make `is_valid_keypath_address` public, but instead split a part of `ethereum::pubrequest::process_address()`  into a helper function `fn derive_address(keypath: ...) -> Result<String, Error>`, and use that here.

src/rust/bitbox02-rust/src/hww/api/bitcoin/payment_request.rs
```diff
@@ -190,6 +202,21 @@ pub fn validate(
     if payment_request.memos.len() > MAX_MEMOS_NUM {
         return Err(ValidationError::Other);
     }
+    #[cfg(feature = "app-swap")]
+    for memo in payment_request.memos.iter() {
+        if let Memo {
+            memo: Some(memo::Memo::CoinPurchaseMemo(coin_purchase_memo)),
+        } = memo
+        {
+            if let Some(memo::coin_purchase_memo::AddressDerivation::Eth(eth)) =
```
benma: Maybe just gate this with `cfg(feature="app-ethereum")` (return `Error::Disabled` otherwise)?

src/rust/bitbox02-rust/Cargo.toml
```diff
@@ -79,6 +79,8 @@ app-litecoin = [
   "bitbox02/app-litecoin",
 ]
 
+app-swap = ["app-ethereum"]
```
benma: I think you enabled Ethereum functions in the BTC-only edition by doing this 😅 

Imo best to not have indirect activation of the coins like this, and use the app-ethereum guard where appropriate (see my other comment).

benma: Could think about a CI job that checks that the btconly firmware binary is free of altcoin/u2f stuff, and that the multi firmware binary includes it :thinking:

cedwies:
@codex review

src/rust/bitbox02-rust/src/hww/api/bitcoin/payment_request.rs
```diff
@@ -191,20 +203,77 @@ pub fn validate(
...
+                    {
+                        let derived_address =
+                            super::super::ethereum::derive_address(_hal, &_eth.keypath)
+                                .map_err(|_| ValidationError::Other)?;
+                        if derived_address != coin_purchase_memo.address {
+                            return Err(ValidationError::Other);
+                        }
+                    }
+                    #[cfg(not(feature = "app-ethereum"))]
+                    return Err(ValidationError::Other);
```
benma: Maybe add `::Disabled` and map it to `Error::Disabled` in signtx.rs?

src/rust/bitbox02-rust/src/hww/api/ethereum.rs
```diff
@@ -23,6 +23,19 @@ use pb::eth_response::Response;
 
 use core::convert::TryInto;
 
+pub(crate) fn derive_address(
```
benma, line 26: Add unit tests please.

src/rust/bitbox02-rust/src/hww/api/bitcoin/payment_request.rs
```diff
@@ -174,6 +185,7 @@ fn ecdsa_verify(sig64: &[u8], msg32: &[u8], pubkey33: &[u8]) -> Result<(), Valid
 
 /// Validate the payment request: amount, signature, etc.
 pub fn validate(
+    _hal: &mut impl crate::hal::Hal,
```
benma: Not sure if `_hal` is the common pattern to suppress unused warnings, I think it's used more for actually unused vars that still need to be bound 🤔 

Maybe better: `#[allow(unused)]` right above this line and name it `hal: ...`.

src/rust/bitbox02-rust/src/hww/api/bitcoin/payment_request.rs
```diff
@@ -262,11 +383,107 @@ mod tests {
...
+            tst_sign_payment_request(coin_params, &mut payment_request, value, address);
+            assert!(matches!(
+                validate(
+                    &mut TestingHal::new(),
+                    coin_params,
+                    &payment_request,
+                    value,
+                    address
+                ),
+                Err(ValidationError::Other)
```
benma: You could add a dedicated error enum arm for address verification errors so the unit tests can check for it specifically.

cedwies:
- Added 2 unit tests for `derive_address()`
- Introduced `AddressMismatch` and `Disabled` error.
- switched from `_hal` to `unused_variables` (not for `_eth`)

Ready to review again

src/rust/bitbox02-rust/src/hww/api/ethereum.rs
```diff
@@ -23,6 +23,19 @@ use pb::eth_response::Response;
...
+    hal: &mut impl crate::hal::Hal,
+    keypath: &[u32],
+) -> Result<alloc::string::String, Error> {
+    if !keypath::is_valid_keypath_address(keypath) {
+        return Err(Error::InvalidInput);
+    }
+    let pubkey = crate::keystore::get_xpub_twice(hal, keypath)
```
benma, line 33: Mabye not worth optimizing now, but to verify the address in a payment request, deriving the xpub once is enough, as a derivation error (bit flip) will be caught and not have any bad consequence.

src/rust/bitbox02-rust/src/hww/api/bitcoin/payment_request.rs
```diff
@@ -94,6 +94,10 @@ pub async fn user_verify(
 pub enum ValidationError {
     UnknownRecipient,
     InvalidSignature,
+    #[cfg(feature = "app-ethereum")]
```
benma, line 97: Is there a compiler warning/error without this guard here and below? You can leave it, but imho it's fine to also not guard, as it will change whenever a new coin is supported in the coinpurchase memo.

cedwies: CI complains because when `app-ethereum` is disabled and `AddressMismatch` is never constructed.

cedwies:
Squashed, no change.

PR #1787, opened 2026-02-11
ui: port sdcard component to async

benma:
Same kind of refactor as:

- https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1781
- https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1779

### DIFF
--- a/src/rust/bitbox02-rust/src/workflow/sdcard.rs
+++ b/src/rust/bitbox02-rust/src/workflow/sdcard.rs
@@ -1,19 +1,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::bb02_async::option_no_screensaver;
-use core::cell::RefCell;
-
 pub struct UserAbort;
 
 pub async fn sdcard() -> Result<(), UserAbort> {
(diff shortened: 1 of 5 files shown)


src/rust/bitbox02-rust/src/workflow/sdcard.rs
```diff
@@ -1,19 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::bb02_async::option_no_screensaver;
-use core::cell::RefCell;
-
 pub struct UserAbort;
 
 pub async fn sdcard() -> Result<(), UserAbort> {
```
NickeZ, line 5: Same comment as on other PR, can we return

```rust
enum UserResponse {
    Yes,
    No,
}
```
instead of `Result`?

benma: or `bool` like in `confirm()`? Or rather make `confirm()` also return `UserResponse`?

benma: Fixed

PR #1788, opened 2026-02-11
ui: port menu component to async

benma:
- The `cancel` field in `MenuParams` is deleted, as all callers set it to true
- The `with_cancel` primitive that works with `Component` is inlined into `bitbox02::ui::menu` - it's the only component with a cancel prompt. If there will be more, we can think of an abstraction again.

### DIFF
--- a/src/rust/bitbox02-rust/src/workflow/cancel.rs
+++ b/src/rust/bitbox02-rust/src/workflow/cancel.rs
@@ -1,54 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::bb02_async::option_no_screensaver;
-use core::cell::RefCell;
-
-use super::confirm;
-
 #[derive(Debug)]
 pub enum Error {
     Cancelled,
-}
-
-pub type ResultCell<R> = RefCell<Option<Result<R, Error>>>;
-
-/// Resolves the `with_cancel` future as cancelled.
-pub fn cancel<R>(result_cell: &ResultCell<R>) {
-    *result_cell.borrow_mut() = Some(Err(Error::Cancelled));
-}
-
-/// Resolves the `with_cancel` future with the given result.
-pub fn set_result<R>(result_cell: &ResultCell<R>, result: R) {
-    *result_cell.borrow_mut() = Some(Ok(result));
-}
-
-/// Blocks on showing/running a component until `cancel` or `result` is
-/// called on the same `result_cell`.
-/// In the former, a prompt with the given title to confirm cancellation is shown.
-///
-/// * `title` - title to show in the cancel confirm prompt.
-/// * `component` - component to process
-/// * `result_cell` - result var to synchronize the result on. Pass the same to `cancel` and
-///   `set_result`.
-pub async fn with_cancel<R>(
-    title: &str,
-    component: &mut bitbox02::ui::Component<'_>,
-    result_cell: &ResultCell<R>,
-) -> Result<R, Error> {
-    component.screen_stack_push();
-    loop {
-        let result = option_no_screensaver(result_cell).await;
-        if let Err(Error::Cancelled) = result {
-            let params = confirm::Params {
-                title,
-                body: "Do you really\nwant to cancel?",
-                ..Default::default()
-            };
-
-            if let Err(confirm::UserAbort) = confirm::confirm(&params).await {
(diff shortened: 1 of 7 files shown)


src/rust/bitbox02-rust/src/workflow/cancel.rs
```diff
@@ -1,54 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::bb02_async::option_no_screensaver;
-use core::cell::RefCell;
-
-use super::confirm;
-
 #[derive(Debug)]
 pub enum Error {
```
NickeZ, line 4: if `enum Error` is the only thing that is left of the `cancel.rs` file I think we should replace that type with an

```rust
enum Response<T> {
    Approve(T),
    Reject,
}
```

or

```rust
enum Response {
    Yes,
    No,
}
```

src/rust/bitbox02/src/ui/types.rs
```diff
@@ -66,12 +66,21 @@ pub struct TrinaryInputStringParams<'a> {
 pub type SelectWordCb<'a> = Box<dyn FnMut(u8) + 'a>;
 pub type ContinueCancelCb<'a> = Box<dyn FnMut() + 'a>;
 
+pub enum MenuResult {
```
NickeZ: instead of `MenuResult` could we call it `MenuResponse` or something that indicates it is user input?

I don't understand what `ContinueOnLast` means.

Would something like this work?

```rust
enum MenuResponse {
    SelectWord(u8),
    ContinueOnLast,
    Cancel,
}
```

src/rust/bitbox02/src/ui/types.rs
```diff
@@ -66,12 +66,21 @@ pub struct TrinaryInputStringParams<'a> {
...
+#[derive(Debug)]
+pub struct CancelError;
+
 pub struct MenuParams<'a> {
     pub words: &'a [&'a str],
     pub title: Option<&'a str>,
-    pub select_word_cb: Option<SelectWordCb<'a>>,
-    pub continue_on_last_cb: Option<ContinueCancelCb<'a>>,
-    pub cancel_cb: Option<ContinueCancelCb<'a>>,
+    pub select_word: bool,
```
NickeZ, line 83: Can you also comment what `select_word` and `continue_on_last` means?

src/rust/bitbox02/src/ui/ui.rs
```diff
@@ -299,52 +299,81 @@ where
...
+    // Shared between the async context and the c callback
+    struct SharedState {
+        waker: Option<Waker>,
+        result: Option<Result<MenuResult, CancelError>>,
+    }
+    let shared_state = Box::new(RefCell::new(SharedState {
+        waker: None,
+        result: None,
+    }));
+    let shared_state_ptr = shared_state.as_ref() as *const RefCell<SharedState> as *mut c_void;
```
NickeZ, line 315: We did the same in the previous PRs. But this reads a bit wrong to me. Shouldn't it be

```diff
-shared_state.as_ref() as *const RefCell<SharedState> as *mut c_void;
+shared_state.as_mut() as *mut RefCell<SharedState> as *mut c_void;
```

src/rust/bitbox02/src/ui/types.rs
benma: Done

src/rust/bitbox02-rust/src/workflow/cancel.rs
benma: Done (but not with a generic, keeping it in the MenuResponse for now)

src/rust/bitbox02/src/ui/types.rs
benma: Done

src/rust/bitbox02/src/ui/ui.rs
benma: We want a pointer to the data, and RefCell provides interior mutability - we don't want/need "mut RefCell".

NickeZ: I guess the most correct thing would be to take `const *void` in the callback and force interior mutability there. I'll resolve it for now.

NickeZ (approved):
utACK

PR #1790, opened 2026-02-12
UI address

benma:


src/rust/bitbox02/src/ui/types.rs
```diff
@@ -33,6 +33,11 @@ pub enum SdcardResponse {
     Cancelled,
 }
 
+pub enum ApproveCancelResponse {
```
NickeZ: could it be `ConfirmResponse` or is it used for more components?

benma: Three components in total:
- confirm
- confirm_transaction_address
- confirm_transaction_fee

But all start with confirm, so I agree it's a better name :)

benma: Renamed now

NickeZ (approved):
utACK

PR #1795, opened 2026-02-13
docker: Add target for cortex-m33

NickeZ:
* Add target for cortex-m33
* Update rustc to latest (reduces firmware size by about 4k)
* Update llvm to match rust
* Vendor new deps due to new toolchain
* Update Arm GNU toolchain to 15 (further reduces by about 2k)

edit: measured:
```
1.92.0, gcc-14: 700424
1.93.1, gcc-15: 694016
```

- [x] Test that old firmwares and bootloaders work


benma (approved):
utACK

PR #1796, opened 2026-02-14
Hal securechip/memory

benma:
We want to make HAL self-contained and not depend on the bitbox02 crate.

The plan is to move the HAL traits to a new crate bitbox-hal, and the hal::bitbox02 implementation of it to the bitbox02 crate.

### DIFF
--- a/src/rust/bitbox02-rust/src/hal/bitbox02/memory.rs
+++ b/src/rust/bitbox02-rust/src/hal/bitbox02/memory.rs
@@ -4,24 +4,77 @@ use alloc::string::String;
 use alloc::vec::Vec;
 
 use crate::hal::Memory;
+use crate::hal::memory::{Error, PasswordStretchAlgo, Platform, SecurechipType};
 
 pub(crate) struct BitBox02Memory;
 
+fn to_hal_securechip_type(securechip_type: bitbox02::memory::SecurechipType) -> SecurechipType {
+    match securechip_type {
+        bitbox02::memory::SecurechipType::Atecc => SecurechipType::Atecc,
+        bitbox02::memory::SecurechipType::Optiga => SecurechipType::Optiga,
+    }
+}
+
+fn to_hal_platform(platform: bitbox02::memory::Platform) -> Platform {
+    match platform {
+        bitbox02::memory::Platform::BitBox02 => Platform::BitBox02,
+        bitbox02::memory::Platform::BitBox02Plus => Platform::BitBox02Plus,
+    }
+}
+
+fn to_hal_password_stretch_algo(
+    algo: bitbox02::memory::PasswordStretchAlgo,
+) -> PasswordStretchAlgo {
+    match algo {
+        bitbox02::memory::PasswordStretchAlgo::MEMORY_PASSWORD_STRETCH_ALGO_V0 => {
+            PasswordStretchAlgo::V0
+        }
+        bitbox02::memory::PasswordStretchAlgo::MEMORY_PASSWORD_STRETCH_ALGO_V1 => {
+            PasswordStretchAlgo::V1
+        }
+    }
+}
+
+fn to_hal_error(error: bitbox02::memory::MemoryError) -> Error {
+    match error {
+        bitbox02::memory::MemoryError::MEMORY_OK => {
+            unreachable!("MEMORY_OK must not be converted to hal::memory::Error")
+        }
+        bitbox02::memory::MemoryError::MEMORY_ERR_INVALID_INPUT => Error::InvalidInput,
+        bitbox02::memory::MemoryError::MEMORY_ERR_FULL => Error::Full,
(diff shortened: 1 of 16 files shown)


src/rust/bitbox02-rust/src/hww/api/bluetooth.rs
```diff
@@ -10,7 +10,7 @@ use pb::bluetooth_response::Response;
 
 use sha2::{Digest, Sha256};
 
-use crate::hal::{Memory, Ui};
+use crate::hal::{Memory, Ui, memory as hal_memory};
```
cedwies, line 13: Why are you using an alias here and in `device_info.rs` (`hal_memory`), but e.g. in `communication_mode.rs` and `reset.rs` not?

src/rust/bitbox02-rust/src/hal/bitbox02/memory.rs
```diff
@@ -4,24 +4,77 @@ use alloc::string::String;
...
+            unreachable!("MEMORY_OK must not be converted to hal::memory::Error")
+        }
+        bitbox02::memory::MemoryError::MEMORY_ERR_INVALID_INPUT => Error::InvalidInput,
+        bitbox02::memory::MemoryError::MEMORY_ERR_FULL => Error::Full,
+        bitbox02::memory::MemoryError::MEMORY_ERR_DUPLICATE_NAME => Error::DuplicateName,
+        bitbox02::memory::MemoryError::MEMORY_ERR_UNKNOWN => Error::Unknown,
+    }
+}
+
+fn to_bitbox02_password_stretch_algo(
```
cedwies: Why do we have this function twice? We also have it in `src/hal/bitbox02/securechip.rs` line 70.

src/rust/bitbox02-rust/src/hww/api/bluetooth.rs
benma: Because here we have another import of bitbox02::memory, so something needs to be done to disambiguate them.

src/rust/bitbox02-rust/src/hal/bitbox02/memory.rs
benma: Accident, removed one of them. Good catch!

benma:
> Nice!
> 
> Regarding the mappings (model type, error type, algo): This is a very good architectural change, but now correctness depends on this mapping staying 1:1 forever. I dont see tests that lock this down. If one mapping arm is changed later (by accident or whatever), it still compiles but runtime behavior changes. This can even happen silently. Or e.g. we use the wrong algorithm tag. I think adding tests would be in the scope of this PR.
> 
> Not regarding this PR: I noticed in `src/securechip/securechip.h` we have:
> 
> ```c
> SC_ERR_MEMORY = -8,
> ```
> 
> But in Rust, in `SECURECHIP_ERRORS`, `SC_ERR_MEMORY` is missing. So we get `Error::Status(-8)` instead of `Error::SecureChip(SecureChipError::SC_ERR_MEMORY`. Should we add `SecureChipError::SC_ERR_MEMORY` to `SECURECHIP_ERRORS` in `src/rust/bitbox02/src/securechip.rs` ?

I added a commit to add the missing error - nice find!

I am unsure of the value of unit tests - wouldn't they just replicate the mapping again? Hard to imagine how they would catch a regression, but they add a lot of boilerplate :thinking:

cedwies:
> I am unsure of the value of unit tests - wouldn't they just replicate the mapping again? Hard to imagine how they would catch a regression, but they add a lot of boilerplate 🤔

I see your point, though I disagree.

A test would look roughly like:

```rust
assert_eq!(to_hal_model(bitbox02::securechip::Model::ATECC_ATECC608A), Model::Atecc608A);
assert_eq!(to_hal_model(bitbox02::securechip::Model::ATECC_ATECC608B), Model::Atecc608B);
assert_eq!(to_hal_model(bitbox02::securechip::Model::OPTIGA_TRUST_M_V3), Model::OptigaTrustM3);
```

And yes, for intentional changes we would then change the tests too, but IMO the tests are worth it to catch accidental changes (merge conflicts? reorders? new variants?). I agree that it is hard to imagine a scenario where this happens, but since this is a security critical path, I think it's worth to make sure we catch unintentional changes in case they somehow occur in the near or distant future.

E.g. in `keystore.rs` there is
```rust
securechip::Error::SecureChip(sc_err) => Error::SecureChip(sc_err as i32),
```

If someone changed `IncorrectPassword` from `-6` to `-5`, the code still compiles, but the wrong error code propagates to the app layer.

A test like
```rust
assert_eq!(SecureChipError::IncorrectPassword as i32, -6);
```
would catch that immediately. I am especially careful because this can all happen silently if we don't have tests.

What do you think?

cedwies (approved):
utACK, thanks!

PR #1864, opened 2026-03-20
Decode uniswap ETH to token swaps

jadzeidan:
Add decoding of Ethereum mainnet to ERC20 token swaps for 
Uniswap V2/V3 so amounts are verifiable on device.

Note, decoding of ERC20 token amounts limited the
ones listed in the firmware.

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/ethereum/sign.rs
@@ -11,12 +11,14 @@ use crate::keystore;
 use crate::hal::Ui;
 use crate::workflow::transaction;
 
+use alloc::string::String;
 use alloc::vec::Vec;
 use hex_lit::hex;
 use pb::eth_response::Response;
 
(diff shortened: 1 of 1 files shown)


jadzeidan:
@Tomasvrba have a look if you would like to test :)

Note, only native ETH to ERC20 tokens (listed in firmware) are decoded. Did not manage to get ERC20 to ETH or ERC20 to ERC20 swaps to work (seems to be limited by the protocol since token “Swap from / Swap to min” amounts are not in the call data?)

Tomasvrba:
will take a look

benma:
Please report how many bytes the firmware binary is increased by this PR.

Tomasvrba:
haven't tested yet, but code should definitely be moved out of ethereum/sign.rs, probably should have a separate file for each contract we would add for clear signing or the eth signing file will explode

Tomasvrba:
Can scrap this PR in favor of adopting ERC-7730 for clear signing

PR #1915, opened 2026-04-08
scripts/format: only format git-tracked files

benma:
It also formatted all headers in src/rust/target etc., slowing it down.


NickeZ (approved):
Lgtm

PR #1917, opened 2026-04-09
ui: add swap component

benma:
Similar to confirm_transction.c, dedicated to the swap screen.

Screenshot_2026-04-09_12-04-35

(real screen shows "Swap" not "SWAP")

Tested manually by @bznein


src/ui/components/confirm_swap.c
```diff
@@ -0,0 +1,118 @@
...
+    if (strlen(from) > BIG_FONT_MAX_CHARS) {
+        from_font = &font_font_a_9X9;
+    }
+    const UG_FONT* to_font = NULL;
+    if (strlen(to) > BIG_FONT_MAX_CHARS) {
+        to_font = &font_font_a_9X9;
+    }
+
+    ui_util_add_sub_component(
+        confirm, label_create_offset(from, from_font, CENTER_TOP, 0, 17, confirm));
```
cedwies, line 113: Can we be sure that "from" fits on screen? What happens when `from` does not fit on the screen?

src/rust/bitbox02-rust/src/hww/api/payment_request.rs
```diff
@@ -110,6 +113,32 @@ fn parse_coin_purchase_amount(amount: &str) -> Result<(&str, &str), Error> {
     Ok((destination_amount, destination_unit))
 }
 
+/// Formats a coin purchase amount for display on the swap screen.
+///
+/// This matches the truncation budget used by Ethereum amount formatting, but only truncates after
+/// the decimal point so integer digits are always preserved.
+fn format_coin_purchase_amount_for_display(amount: &str) -> Result<String, Error> {
```
cedwies, line 120: Could you explain this design choice? Why do we "fit it to 13 only when that truncates fractional digits, not integer digits."? Can't the 14th decimal place be even more important than the 16th integer digit? In general it also feels weird to truncate information for a user verification flow, no? Is it just because we believe this to be an estimate anyway?

benma: First of all, 16 integer digits are quite unrealistic in coin amounts, it's a huge number.

But if you had a big number, it's probably not good to truncate before the decimal point, it would be a totally different numer. Truncating after 13 decimal places is basically negligible.

Not sure what you mean, how could the 14th decimal place be more important than the 16th integer digit?

src/ui/components/confirm_swap.c
benma: I didn't mean to make the `to` label scrollable, so I undid that, to match confirm_transaction.c.

It's the same labels and font size logc as in confirm_transaction.c, so it should work for all currently supported amounts.

src/rust/bitbox02-rust/src/hww/api/payment_request.rs
cedwies: > But if you had a big number, it's probably not good to truncate before the decimal point, it would be a totally different numer.

Jup, missed that.

Then:
> Can't the 14th decimal place be even more important than the 16th integer digit?

cannot really happen, thanks

cedwies (approved):
utACK, nice!

PR #1921, opened 2026-04-13
Async monotonic counter

benma:


src/rust/bitbox-securechip/src/optiga/ops.rs
```diff
@@ -0,0 +1,336 @@
...
+    // completed yet. The old command may still be reading/writing the shared static buffers, so a
+    // new call must not reuse them yet.
+    Detached,
+}
+
+// The Optiga callback exposes only one shared status variable and one wakeup hook, so the Rust
+// wrappers can support only one in-flight async operation at a time. The extra Detached state is
+// needed because cancellation must not free the slot until the callback has really finished using
+// the shared static buffers.
+static OPTIGA_ASYNC_OP_STATE: SyncCell<AsyncOpState> = SyncCell::new(AsyncOpState::Idle);
```
NickeZ: I think this can just be called "STATE". It isn't exported and there is no benefit in copying it's path into its name.

src/rust/bitbox-securechip/src/optiga/ops.rs
```diff
@@ -0,0 +1,336 @@
...
+static OPTIGA_ASYNC_OP_WAKER: WakerCell = WakerCell::new();
```
NickeZ: I think this can just be called "WAKER". It isn't exported and there is no benefit in copying it's path into its name.

src/rust/bitbox-securechip/src/optiga/ops.rs
```diff
@@ -0,0 +1,336 @@
...
+
+use crate::{Error, SecureChipError};
+use core::cell::UnsafeCell;
+use core::future::poll_fn;
+use core::task::{Poll, Waker};
+use util::cell::SyncCell;
+use zeroize::Zeroize;
+
+// This is the biggest buffer we want to move through the async data-object helpers.
+const ARBITRARY_DATA_OBJECT_TYPE_3_MAX_SIZE: usize =
```
NickeZ, line 11: "ARBITRARY_DATA_OBJECT_TYPE_3_MAX_SIZE" can be a local const in the function where it is used. And it doesn't need to have such a long name. MAX_SIZE is enough.

src/rust/bitbox-securechip/src/optiga/ops.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// The Optiga callback exposes only one shared status variable and one wakeup hook, so the Rust
+// wrappers can support only one in-flight async operation at a time. The extra Detached state is
+// needed because cancellation must not free the slot until the callback has really finished using
+// the shared static buffers.
+static OPTIGA_ASYNC_OP_STATE: SyncCell<AsyncOpState> = SyncCell::new(AsyncOpState::Idle);
+static OPTIGA_ASYNC_OP_WAKER: WakerCell = WakerCell::new();
+
+struct WakerCell {
```
NickeZ, line 36: Isn't this very similar to `SyncCell<Option<Waker>>`?

src/rust/bitbox-securechip/src/optiga/ops.rs
```diff
@@ -0,0 +1,336 @@
...
+    fn clear(&self) {
+        drop(self.take());
+    }
+}
+
+// The Optiga C API retains raw pointers to these statics until its async callback completes,
+// potentially after the Rust future has been dropped. The surrounding async-op state machine
+// ensures only one Rust wrapper owns the buffers at a time, so these helpers expose only raw
+// pointers and copy operations and never hand out Rust references to the static storage.
+struct StaticBytes<const N: usize> {
```
NickeZ: Isn't this very similar to `GroundedArrayCell`?

src/rust/bitbox-securechip/src/optiga/ops.rs
```diff
@@ -0,0 +1,336 @@
...
+    })
+    .await
+}
+
+pub(super) async fn util_read_data(oid: u16, offset: u16, out: &mut [u8]) -> Result<(), Error> {
+    // Static because the Optiga library keeps raw pointers to this buffer and length until the
+    // async callback completes, and the Rust future may be dropped before that happens.
+    static BUF: StaticBytes<ARBITRARY_DATA_OBJECT_TYPE_3_MAX_SIZE> =
+        StaticBytes::new([0; ARBITRARY_DATA_OBJECT_TYPE_3_MAX_SIZE]);
+    static LEN: StaticValue<u16> = StaticValue::new(0);
```
NickeZ: Couldn't `LEN` just be a `GroundedCell`?

src/rust/bitbox-securechip/src/optiga/ops.rs
```diff
@@ -0,0 +1,336 @@
...
+    if out.len() > ARBITRARY_DATA_OBJECT_TYPE_3_MAX_SIZE {
+        panic!("optiga async read larger than max supported size");
+    }
+    let requested_len = out.len();
```
NickeZ: Maybe this (or similar) here?
```rust
let requested_len: u16 = out.len().try_into().unwrap();
```

src/rust/bitbox-securechip/src/optiga/ops.rs
```diff
@@ -0,0 +1,336 @@
...
+
+    let util = unsafe { bitbox_securechip_sys::optiga_util_instance() };
+    if util.is_null() {
```
NickeZ: `optiga_util_read_data` already has a null check.

src/optiga/optiga_ops.c
```diff
@@ -5,6 +5,8 @@
 #include <securechip/securechip.h>
 #include <util.h>
 
+extern void rust_optiga_callback_wake(void);
```
NickeZ: why isn't this in `rust.h`?

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: It will be used later when the u2f functions are ported - u2f uses arbitrary data and its size is the biggest size we need.

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: Don't think so, SyncCell has no drop:

Codex:

---

• Conceptually yes, but not with the current SyncCell API.

  WakerCell in src/rust/bitbox-securechip/src/optiga/ops.rs:34 is basically a specialized critical-section-protected Option<Waker>. But SyncCell in src/rust/util/src/cell.rs:11 only exposes read() and write()
  for T: Copy, and Option<Waker> is not Copy.

  The more important semantic difference is drop behavior:

  - SyncCell::write() does a raw overwrite and explicitly does not drop the old value.
  - For Option<Waker>, dropping the old stored Waker matters.
  - WakerCell::register() uses normal assignment, so an old Some(waker) gets dropped correctly.
  - WakerCell::take() is also essential, and SyncCell has no equivalent.

  So:

  - Similar idea: yes.
  - Direct replacement with SyncCell<Option<Waker>>: no.
  - If you added a different utility cell with replace()/take() semantics for non-Copy types, then WakerCell could likely disappear.

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: What do you mean?

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: Renamed

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: Renamed

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: Changed

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: Done

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: Removed

src/optiga/optiga_ops.c
benma: This crate is not in bitbox02-cbindgen.toml, and bitbox02-rust-c would need to start depending on this crate to make sure it's working like all the other crates. Seemed like overkill for one little trivial declaration only used here.

Do you want me to change it anyway?

src/rust/bitbox-securechip/src/optiga/ops.rs
NickeZ: I mean that we can try converting `out.len()` to `u16` as soon as possible instead of extending variables to `usize` in other places. And also panicking or returning error if out.len() doesn't fit in a u16.

src/optiga/optiga_ops.c
NickeZ: If you don't want to auto generate it I suppose we should have a comment to keep it in sync. If they get out of sync it is **very dangerous** and will silently just corrupt the stack.

src/rust/bitbox-securechip/src/optiga/ops.rs
NickeZ: ok, but still rename it to something like `MAX_SIZE` or `ASYNC_BUF_MAX_SIZE`?

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: Can do, but the check above is enough?

```
    if out.len() > ARBITRARY_DATA_OBJECT_TYPE_3_MAX_SIZE {
        panic!("optiga async read larger than max supported size");
    }
```

okay you need to believe that const is smaller than max u16, so i'll change it to try_into...

src/optiga/optiga_ops.c
benma: Good point! Went with rust.h then after all.

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: Fixed

src/rust/bitbox-securechip/src/optiga/ops.rs
benma: Added, `ASYNC_BUF_MAX_SIZE = ARBITRARY_DATA_OBJECT_TYPE_3_MAX_SIZE` - the latter const will be used later and is the one that determines what the max size is for the async buffer.

NickeZ (approved):
tACK

PR #1930, opened 2026-04-15
Move protobuf bindings into bitbox-proto

NickeZ:
Add a committed bitbox-proto crate for the generated Rust protobuf bindings.

Switch regeneration to a Cargo-driven flow, add a make shortcut and CI drift check, and keep the generator copy-on-change so no-op regen runs do not churn timestamps or invalidate build caches.


src/rust/bitbox-proto/build.rs
```diff
@@ -0,0 +1,46 @@
...
+        "messages/backup_commands.proto",
+        "messages/bitbox02_system.proto",
+        "messages/btc.proto",
+        "messages/cardano.proto",
+        "messages/eth.proto",
+        "messages/mnemonic.proto",
+        "messages/system.proto",
+        "messages/perform_attestation.proto",
+        "messages/keystore.proto",
+        "messages/antiklepto.proto",
```
benma: Can just do `messages`, rerun-if-changed consider director changes too afaik

Makefile
```diff
@@ -174,6 +174,8 @@ dockerdev:
 	./scripts/dockerenv.sh
 dockerrel:
 	./scripts/dockerenv.sh release
+generate-protobufs:
```
benma, line 180: When you rebase, careful that this does not end up twice in the Makefile, as master already has an entry of the same name in a different location

src/rust/bitbox-proto/build.rs
```diff
@@ -0,0 +1,46 @@
...
+        "tools/prost-build-proto/src/main.rs",
+    ] {
+        println!(
+            "cargo:rerun-if-changed={}",
+            repo_root.join(filename).display()
+        );
+    }
+
+    let status = Command::new("prost-build-proto")
```
benma: Why not generate-protoubf-rust.sh? I think like this modified generated files are created because they are not formatted

Also this now assumes it's available on the host, but the Docker image contains this too. This becomes confusing as to which version to use, especially if there are changes to the tool. Any smart way to resolve this? Or do we rely on CI using the binary of the docker image and check if it produces a different result?

NickeZ:
@benma Made some changes based on feedback

* no changes at all to prost-build-proto
* Put logic for copy-on-change in shell script
* Call shell script in cargo build script

.github/workflows/ci-common.yml
```diff
@@ -141,6 +155,9 @@ jobs:
         with:
           rust-src-dir: src/rust
 
+      - name: install prost-build-proto
+        run: cargo install --path tools/prost-build-proto --locked
```
benma, line 159: Why not use the one in the Docker image? It's a bit wasteful to rebuild/install it on every CI run.

If you prefer to keep it this way anyway, then you should probably remove it from the Dockerfile.

benma (changes requested):

.github/workflows/ci-common.yml
NickeZ: this is to test macos. I don't think we have docker images for macos? Like, I don't think the concept of macos in a container exists?

benma: Ah the diff was folded such that I didn't see it was in the macOS job.

benma (approved):
utACK

PR #1932, opened 2026-04-16
Async passwords

benma:
The unit tests in optiga.rs are a 1:1 port of test_optiga.c, so we can be sure that the v0 and v1 stretching and setting a new password work the same way as before. 

To test v0 password algo manually:

- reset device
- load firmware v9.24.0
- restore or create wallet, set password
- load firmware of this branch
- password works and device info reports V2 as the stretching algo


src/optiga/optiga.h
```diff
@@ -92,20 +91,11 @@ typedef struct optiga_crypt optiga_crypt_t;
...
-USE_RESULT int optiga_stretch_password(
-    const char* password,
-    securechip_password_stretch_algo_t password_stretch_algo,
-    uint8_t* stretched_out);
-USE_RESULT bool optiga_reset_keys(void);
 USE_RESULT bool optiga_gen_attestation_key(uint8_t* pubkey_out);
 USE_RESULT bool optiga_attestation_sign(const uint8_t* challenge, uint8_t* signature_out);
 USE_RESULT optiga_util_t* optiga_util_instance(void);
 USE_RESULT optiga_crypt_t* optiga_crypt_instance(void);
+USE_RESULT bool optiga_ifs_random_32_bytes(uint8_t* rand_out);
```
benma, line 98: This is temporary and all uses of it will be removed soon when optiga_random is ported to Rust, so we can call it directly from Rust instead of going through C.

benma:
Rebased and added small commit to improve param types.

NickeZ (approved):
tACK

PR #1934, opened 2026-04-17
Merge 9.26 -> master

benma:


NickeZ:
~Should this wait until release is out? in case there are fixes?~

NickeZ (approved):
utACK, I suppose all of this already was reviewed when merged into the release branch.

We should think one more time if we really want to do CI for every commit in a PR or what we should do.

PR #1937, opened 2026-04-18
Fix timer interrupt race 

benma:


src/rust/bitbox02/src/hal/timer.rs
```diff
@@ -26,17 +26,25 @@ impl bitbox_hal::timer::Timer for BitBox02Timer {
...
+        let shared_state_ptr =
+            shared_state.as_ref() as *const GroundedCell<SharedState> as *mut c_void;
         unsafe extern "C" fn callback(user_data: *mut c_void) {
-            let shared_state = unsafe { &*(user_data as *mut RefCell<SharedState>) };
-            let mut shared_state = shared_state.borrow_mut();
-            shared_state.result = Some(());
-            if let Some(waker) = shared_state.waker.as_ref() {
-                waker.wake_by_ref();
+            let shared_state = unsafe { &*(user_data as *mut GroundedCell<SharedState>) };
+            let waker = critical_section::with(|_| unsafe {
```
NickeZ: critical section probably not needed here. AFAIK we don't use interrupt priorities, and an interrupt with the same priority cannot interrupt an interrupt.

NickeZ:
Did you consider using an AtomicWaker from futures-core and AtomicBool for the result instead of turning off interrupts?

benma:
> Did you consider using an AtomicWaker from futures-core and AtomicBool for the result instead of turning off interrupts?

Nice idea - it seems to work, PTAL

NickeZ (approved):
utACK

PR #1940, opened 2026-04-19
releases: add v9.26.1

benma:
Includes a little power-coded script that generates these files from the published release.


NickeZ (approved):
reproduced ACK

PR #1941, opened 2026-04-19
bitbox02_rust: remove mock_memory

benma:
All tests use HAL now.


NickeZ (approved):
utACK, love a net negative line PR :)

PR #1942, opened 2026-04-19
ui: don't show waiting screen during password stretch

benma:
(the effect described below is only really visible after https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1932, where pw stretching becomes async)

Before the securechip operations became async, the UI did not process during keystore unlock and froze at the last screen of the password confirmation, and UI continued with the BIP39 unlock animation.

Now that they are async, the password screen is popped and the waiting screen would be shown for roughly 1s before the unlock animation started. To improve UI there, we start showing first screen of the unlock animation (the locked lock) as the filler screen.
The downside is that when entering the wrong password, it is also briefly shown before the wrong-password error message, but it's acceptable.

There are other workflows where the waiting screen is also shown during the stretch, not addresed by this commit: show mnemonic, create backup, and change password (any call to
`workflows::unlock::unlock_keystore()` to check the password).

### DIFF
--- a/src/rust/bitbox03/src/ui.rs
+++ b/src/rust/bitbox03/src/ui.rs
@@ -49,6 +49,7 @@ impl<Timer: bitbox_hal::timer::Timer> hal::ui::Ui for BitBox03Ui<Timer> {
     type Progress = BitBox03UiProgress;
 
     type Empty = BitBox03UiEmpty;
+    type UnlockAnimation = BitBox03UiEmpty;
 
     async fn confirm(
         &mut self,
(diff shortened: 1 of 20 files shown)


src/rust/bitbox03/src/ui.rs
```diff
@@ -122,6 +123,10 @@ impl<Timer: bitbox_hal::timer::Timer> hal::ui::Ui for BitBox03Ui<Timer> {
         todo!()
     }
 
+    fn unlock_animation_create(&mut self) -> Self::UnlockAnimation {
+        todo!()
```
NickeZ: let it just be an empty function so simulator doesn't crash here.

NickeZ (approved):
tACK

PR #1946, opened 2026-04-21
optiga: dedup busy polling

benma:
Share the busy-wait polling logic between reclaim_detached_op() and wait() in the async Optiga wrapper.

This keeps the async behavior unchanged while shaving a small amount of future state-machine code from the firmware.


NickeZ (approved):
tACK

PR #1949, opened 2026-04-22
Async optiga

benma:
Builds on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1948

Contains a commit to manually test the attestation signing, to be dropped before merge.


NickeZ (approved):
tACK

PR #1952, opened 2026-04-22
Shrink mbedtls AES tables

benma:
Enable MBEDTLS_AES_FEWER_TABLES in the Optiga Trust M mbedTLS user config.

This keeps the same AES/CCM functionality, but trades a small amount of AES speed for smaller ROM usage by deriving the extra AES lookup tables from FT0/RT0 instead of keeping all eight tables in flash.

- savings: 6208 bytes

The main win is that six live 1024-byte AES tables drop out of the final image: FT1, FT2, FT3, RT1, RT2, and RT3. FT0 and RT0 remain, and the AES encrypt/decrypt functions shrink slightly as well.

No functionality changes are intended; this is a ROM-for-speed tradeoff in the mbedTLS AES implementation used by the Optiga Trust M integration.

### DIFF
--- a/external/mbedtls_config.h
+++ b/external/mbedtls_config.h
@@ -65,6 +65,7 @@
 /* Save RAM at the expense of ROM */
 #define MBEDTLS_AES_ROM_TABLES
 #define MBEDTLS_AES_FEWER_TABLES
+#define MBEDTLS_SHA256_SMALLER
 
 /* Save some RAM by adjusting to your exact needs */
 #define MBEDTLS_PSK_MAX_LEN    16 /* 128-bits keys are generally enough */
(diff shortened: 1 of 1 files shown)


NickeZ:
Did you try removing the ROM tables entirely?

```diff
-#define MBEDTLS_AES_ROM_TABLES
```

benma:
> Did you try removing the ROM tables entirely?
> 
> ```diff
> -#define MBEDTLS_AES_ROM_TABLES
> ```

I did now! It seems to work and shaves off another 2.1kB. The previous `#define MBEDTLS_AES_FEWER_TABLES ` commit still works even in combination with this.

NickeZ:
You could probably get rid of the AES module entirely if you use the SAMD hardware module instead?

benma:
> You could probably get rid of the AES module entirely if you use the SAMD hardware module instead?

We will re-use this in BB03, so I don't want to conditionally have different code paths here, at least for now.

NickeZ (approved):
tACK

PR #1954, opened 2026-04-22
Enable blake2 size_opt

benma:
The blake2 crate exposes a size_opt feature that trades aggressive inlining for a smaller implementation. In this firmware build the Blake2b compression path was still one of the largest Rust code contributors, so this is a direct ROM for speed trade that matches the current goal.

Enable size_opt for the existing optional dependency. The algorithm, test vectors, and call sites stay the same; only the crate internal code generation strategy changes.

In the resulting image, the large Blake2b compression routine is split into much smaller outlined helpers instead of one heavily inlined body. That is where the ROM reduction comes from.

Saves 11576 bytes in firmware.bin.


NickeZ (approved):
utACK

PR #1956, opened 2026-04-23
Disable Rust Debug formatting

benma:
Pass `-Zfmt-debug=none` in the firmware Rust build.

According to
clippy::no_debug (https://rust-lang.github.io/rust-clippy/master/index.html?search=use_debug):

> It should not be used in user-facing output

This tells rustc not to emit normal `Debug` formatter implementations into the cross-compiled firmware image. The option also makes `{:?}` print nothing, so app-visible error status messages must not rely on derived `Debug`.

Keep the live keystore and backup error status screens explicit by formatting those errors through small match-based helpers instead of `Debug`. This preserves the visible diagnostic strings while still allowing derived Debug formatting to be compiled out of the firmware.

There are still live library `Debug` formatting paths in a no-flag build even when our callers map those errors to `InvalidInput`. Protobuf and Miniscript parsers construct detailed errors internally before the caller discards them, for example `WireType` decode errors and Miniscript `NonTopLevel(format!("{:?}", ...))` errors. `fmt-debug=none` prevents those detailed `Debug` formatters and their generic formatting helpers from being linked.

Savings: 6464 bytes

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -219,7 +219,7 @@ if(CMAKE_CROSSCOMPILING)
   set(RUST_TARGET_ARCH_DIR ${RUST_TARGET_ARCH})
   set(RUST_TARGET_ARCH_ARG --target ${RUST_TARGET_ARCH})
   set(RUST_CARGO_FLAGS ${RUST_CARGO_FLAGS} -Zbuild-std=core,alloc -Zbuild-std-features=optimize_for_size)
-  string(JOIN " " RUSTFLAGS "${RUSTFLAGS}" "-Zunstable-options -Cpanic=immediate-abort")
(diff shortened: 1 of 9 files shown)


src/CMakeLists.txt
```diff
@@ -259,7 +259,9 @@ if(NOT CMAKE_CROSSCOMPILING)
           --target-dir ${RUST_BINARY_DIR}/clippy
           --release
           --tests
-          -- # disabled linters:
+          -- # enabled linters:
+            -D clippy::use_debug
```
NickeZ: should be a warning (`-W`) so it is allowed locally bit denied in CI

benma: Fixed

NickeZ (approved):

PR #1957, opened 2026-04-23
Rm f64

benma:


NickeZ:
very good! does removing the use of u128 also reduce size more? I suppose u128 is entirely software implemneted.

benma:
> very good! does removing the use of u128 also reduce size more? I suppose u128 is entirely software implemneted.

Codex actually made an accounting error and the use of u128 undid the previous commit's gains.

The code to replace it (adde as a wip commit) is quite ugly and harder to review.

An alternative would be to use the same as ETH, which uses BigUint:

https://github.com/BitBoxSwiss/bitbox02-firmware/blob/a830ced8e1f7b1c171ad4a300886dd4c86b609e4/src/rust/bitbox02-rust/src/hww/api/ethereum/amount.rs#L39

It would be reusing existing clean code, but the big-int dep would have to be then enabled for BTC-only too, adding ~8kB to that binary (but saving bytes in the Multi binary as that one already uses big-ints).

NickeZ:
Do we have to support larger values than 2.1e15? (all satoshis that ever will be available?)

benma:
> Do we have to support larger values than 2.1e15? (all satoshis that ever will be available?)

BTC is not the only user of this function, Cardano does too. 

I split out the first easy commit (and CI check) to https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1959

NickeZ:
Have two functions then? In multi it is strictly better to use BigUints, so there it isn't a question. In BTC we only use 60% of the fw size so there it is more of a ideological question.

PR #1959, opened 2026-04-23
Use integer progress fractions

benma:


NickeZ (approved):
tACK

src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs
```diff
@@ -359,15 +359,15 @@ async fn handle_prevtx(
...
-        progress_component.set({
-            let step = 1f32 / (num_inputs as f32);
-            let subprogress: f32 = (prevtx_input_index as f32)
-                / (prevtx_init.num_inputs + prevtx_init.num_outputs) as f32;
-            (input_index as f32 + subprogress) * step
-        });
+        progress_component.set_fraction(
+            input_index * prevtx_total_ios + prevtx_input_index,
+            num_inputs * prevtx_total_ios,
+        );
```
benma, line 382: @NickeZ did you even test/check that?

NickeZ: I tested signing a regular tx with `send_message.py`. Are there any larger ones I can test?

benma: Doesn't seem like there is out of the box. I assigned @cedwies to verify this is the same calculation :smile:

benma: I checked that this part is correct.

benma:
Added two more commits with small fixes, PTAL @cedwies

src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs
```diff
@@ -360,15 +360,15 @@ async fn handle_prevtx(
...
     for prevtx_input_index in 0..prevtx_init.num_inputs {
         // Update progress.
-        progress_component.set({
-            let step = 1f32 / (num_inputs as f32);
-            let subprogress: f32 = (prevtx_input_index as f32)
-                / (prevtx_init.num_inputs + prevtx_init.num_outputs) as f32;
-            (input_index as f32 + subprogress) * step
-        });
+        progress_component.set_fraction(
+            input_index * prevtx_total_ios + prevtx_input_index,
```
cedwies: You already fixed the wrap around in C by using int64 instead of int32. What about here in rust?

cedwies (changes requested):

src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs
benma: If we change params from u32 to u64 here, then in C it could overflow uint64_t :joy: 

Since no one is expected to run into this, I converted it to checked arithmetic that fails on overflow. See new commit.

cedwies: > If we change params from u32 to u64 here, then in C it could overflow uint64_t 

🥲 

Checked arithmetic is great 👍

cedwies (approved):
utACK

Non blocking: a small regression test with oversized prevtx input/output counts would make this harder to accidentally undo later

PR #1961, opened 2026-04-23
Async atecc

benma:
The second commit (to be dropped before merge) has a manual testsuite that runs after choosing orientation. **It resets keys, creates a new password, so a reset/restore of the device is needed afterwards**


NickeZ:
do we need to test factory setup of the atecc?

benma:
> do we need to test factory setup of the atecc?

Yes, also of Optiga actually.

benma:
Rebased.

NickeZ:
- Low: production async ownership and cancellation logic is excluded from unit tests. The #[cfg(test)] switch replaces the real implementation with ops_fake.rs (src/rust/bitbox-securechip/src/atecc.rs:12). Consequently, the Idle/Running/Detached state machine, dropped-future recovery, shared C command-buffer reuse, and FFI response handling are only typechecked. Add tests using a controllable FFI backend that cancel while busy, start a subsequent command, and verify the old context is fully reclaimed before reuse.

benma:
> * Low: production async ownership and cancellation logic is excluded from unit tests. The #[cfg(test)] switch replaces the real implementation with ops_fake.rs (src/rust/bitbox-securechip/src/atecc.rs:12). Consequently, the Idle/Running/Detached state machine, dropped-future recovery, shared C command-buffer reuse, and FFI response handling are only typechecked. Add tests using a controllable FFI backend that cancel while busy, start a subsequent command, and verify the old context is fully reclaimed before reuse.

Instead of splitting ops.rs/ops_fake.rs, there is now on ops.rs with raw.rs and raw_fake.rs for only the FFI stuff. Added unit tests in a new commit. Also added C unit tests for the ATECC C part in another commit.

PR #1964, opened 2026-04-26
optiga: fix detached buffer reuse

benma:
The Optiga C API keeps raw pointers to the Rust wrapper's static buffers until the asynchronous callback completes. The Rust wrapper has a Detached state for futures that were dropped after launching a command, but the individual wrappers initialized their per-operation static buffers before calling begin_async_op().

That meant a repeated call to the same wrapper could clear or overwrite the exact static storage still owned by the detached C command before begin_async_op() noticed the Detached state and waited for the old callback.

Move begin_async_op().await before every static buffer or length-cell mutation. Detached operations are now reclaimed before any static storage is reused, and live concurrent operations still panic before their buffers can be corrupted.


NickeZ (approved):
tACK

PR #1965, opened 2026-04-27
AGENTS: run make commands serially

benma:
Otherwise the dev-machine can easily lock up.


NickeZ:
This has never been an issue on my dev machine. Are you sure we should slow down development for everyone or can you have a local AGENTS.md?

NickeZ (approved):
utACK

PR #1966, opened 2026-04-28
da14531: restore BLE UART backpressure

benma:
ListBackups returns one protobuf containing all SD card backups. With many backups, the response spans many 64-byte HWW/U2FHID reports. USB sends those reports directly over HID, but BLE first serial-frames each report for the DA14531 and queues the escaped bytes in the fixed-size UART ByteQueue.

The Rust ByteQueue port kept the panicking single-byte put() API and the BLE poller used it for every framed byte. That removed the previous v9.25 backpressure behavior: da14531_protocol_poll() consumed the pending HWW report before knowing whether the whole framed BLE packet fit in the 2048-byte UART queue. If ListBackups produced data faster than UART drained it, put() eventually hit capacity and panicked with bytequeue overflow. USB does not use this UART queue, which is why the same request worked there.

Add ByteQueue::try_put_slice() plus an extern C wrapper that takes util::Bytes and performs an all-or-nothing enqueue. The BLE HWW path now formats the 64-byte report into a temporary serial frame and only clears hww_data after try_put_slice() succeeds. If the queue is full, it returns without writing any bytes, preserving frame boundaries and leaving the same report pending for the next poll after UART has drained.

Keep the existing panicking put() behavior for the other bytequeue callers so the change stays scoped to the actual backpressure boundary. The separate Rust and extern-wrapper tests cover successful writes and failed overflow attempts being atomic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ## Firmware
 
 ### [Unreleased]
+- Fixed a crash when listing many backups over Bluetooth
 
 ### v9.26.1
 - Fix a payment request validation issue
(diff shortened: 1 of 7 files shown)


NickeZ (approved):
utACK

PR #1967, opened 2026-04-29
Refactor agent instructions

NickeZ:
Make AGENTS.md a concise entry point for agent guidance and move detailed repository, command, coding, testing, and review instructions into doc/agent.

Preserve the existing repository rules while tightening command guidance and linking to OpenAI's AGENTS.md and Codex best-practices documentation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,126 +1,56 @@
-# Repository Guidelines for bitbox02-firmware
+# AGENTS.md
 
-## Project Structure & Module Organization
-Core firmware and bootloader code sits in `src/`, grouped by subsystem (`bootloader`, `usb`, `ui`,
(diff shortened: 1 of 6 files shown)


NickeZ:
@coderabbitai review

benma:
@NickeZ coderabbit is not set up for this repo yet. there is also no CODEOWNERS etc here yet.

Are you sure just referencing other files is ideal? This would potentially add tool calls to every session. This approach seems similar to SKILLS, which has the benefit of reducing context overall, so maybe we should use skills instead with a lean AGENTS.md?

NickeZ:
I have no idea, just asked codex to rewrite the file using best practices it could find on openai's website.

PR #1968, opened 2026-04-30
reduce securechip ops in payment request validation

benma:


NickeZ (approved):
tested swap and it feels a bit faster, still shows the background briefly, so we might need to push an empty screen as well?

benma:
> tested swap and it feels a bit faster, still shows the background briefly, so we might need to push an empty screen as well?

if so, different PR. the solution might also be to revert to the old behavior to freeze the previous screen for this duration.

PR #1970, opened 2026-05-06
lvgl-sys: remove cmake build

NickeZ:
Build LVGL directly with cc from an explicit source list and drop the cmake crate dependency. Use the shared bitbox-lvgl-sys lv_conf.h for the BB03 graphical simulator and remove the simulator-specific duplicate config.


benma (approved):
utACK

PR #1971, opened 2026-05-06
Nickez/bb03 binaries

NickeZ:
Some terminology

Platform - Everything that is common to the stm32u5 platform (basically the stm32u5 HAL)
Board - Everything that is specific to a single board and MCU, for example the STM32U5A9J-DK dev-kit, the bb03 "test board" or the bb03 "production board". Every board has at least a unique pinout and memory layout.


scripts/copy_st_drivers.py
```diff
@@ -0,0 +1,459 @@
...
+        raise ValueError(
+            "board must match {} (letters, digits, '.', '_' and '-', no slashes)".format(
+                BOARD_RE.pattern
+            )
+        )
+    if board in RESERVED_BOARD_NAMES:
+        raise ValueError("board name '{}' is reserved".format(board))
+
+
+def hal_driver_directories(source):
```
cedwies: for `HAL_CONF` we have an exact path and name, but for the `HAL_DRIVER` folder we go for a loose/generic way. If possible, I would prefer consistency here. Are we really expecting multiple HAL drivers when we just expect a single `stm32u5xx_hal_conf.h`?

scripts/copy_st_drivers.py
```diff
@@ -0,0 +1,459 @@
...
+    Path("Core/Src"),
+    Path("Drivers/CMSIS"),
+)
+REQUIRED_FILES = (HAL_CONF,)
+OPTIONAL_SHARED_DIRECTORIES = (
+    Path("USBX/App"),
+    Path("USBX/Target"),
+)
+OPTIONAL_MIDDLEWARE_DIRECTORIES = (Path("Middlewares/ST/usbx"),)
+RESERVED_BOARD_NAMES = frozenset(("Core", "Common", "Drivers"))
```
cedwies, line 60: shouldn't "USBX" and "Middlewares" be part of that set too?

scripts/copy_st_drivers.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+def copy_board_directory(source, temp_root, board_dirs):
+    board_dir = temp_root / "board"
+    for directory in board_dirs:
+        src = source / directory
+        if directory.parts[0] == "Core":
+            dst = board_dir / directory.relative_to("Core")
+        else:
+            dst = board_dir / directory
```
cedwies: Does the else statement make sense here? How can it not have "Core" as the `parts[0]` but still be valid?

scripts/copy_st_drivers.py
```diff
@@ -0,0 +1,459 @@
...
+
+BOARD_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]*$")
+HAL_CONF = Path("Core/Inc/stm32u5xx_hal_conf.h")
+REQUIRED_DIRECTORIES = (
+    Path("Core/Inc"),
+    Path("Core/Src"),
+    Path("Drivers/CMSIS"),
+)
+REQUIRED_FILES = (HAL_CONF,)
+OPTIONAL_SHARED_DIRECTORIES = (
```
cedwies: Regarding optional directories: If a later CubeMX project does not have e.g. USBX (for whatever reason), then the USBX folder can become stale. For USBX that might not be a problem but what's your opinion on potentially stale code?

src/rust/bitbox-platform-stm32u5-sys/build.rs
```diff
@@ -0,0 +1,141 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use std::env;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+const ST_SOURCES: &[&str] = &[
```
cedwies, line 7: These sources are not minimal. The build succeeds, even without:

- stm32u5xx_hal_dma.c
- stm32u5xx_hal_dma_ex.c
- stm32u5xx_hal_rcc_ex.c
- stm32u5xx_hal_gpio.c
- stm32u5xx_hal_exti.c
- stm32u5xx_hal_pwr.c
- stm32u5xx_hal_icache.c

Is this intentionally chosen as a base for upcoming work? Or should we keep this minimal for now?

src/rust/bitbox-platform-stm32u5-sys/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#![no_std]
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+#![allow(clippy::all)]
+
+#[cfg(all(target_arch = "arm", target_os = "none"))]
```
cedwies, line 24: in `src/rust/bitbox-platform-stm32u5-sys/build.rs` we use 

```rs
target.starts_with("thumb")
```

which is a bit narrower (I think), can we use that here too?

src/rust/bitbox-platform-stm32u5-sys/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+
+// These CMSIS clock symbols mirror ST's STM32U5 system template definitions in
+// `external/ST/Drivers/CMSIS/Device/ST/STM32U5xx/Source/Templates/system_stm32u5xx.c`.
```
cedwies: Why is this folder mentioned? I don't find it in the repo. The matching file seems to be in

`external/ST/stm32u5a9j-dk/Src/system_stm32u5xx.c`?

Also, we intentionally do not compile `system_stm32u5xx.c`, but provide its CMSIS globals from Rust and port the needed `SystemInit()` logic via `#[pre_init]`. Could we add a one or two sentences explaining that split? Otherwise it might look mysterious and fragile.

src/rust/bitbox-platform-stm32u5-sys/build.rs
```diff
@@ -0,0 +1,141 @@
...
+    Ok(())
+}
+
+fn is_release_profile() -> bool {
+    env::var("PROFILE").expect("PROFILE not set") == "release"
+}
+
+fn main() -> Result<(), &'static str> {
+    println!("cargo::rerun-if-changed=wrapper.h");
+    println!("cargo::rerun-if-env-changed=PROFILE");
```
cedwies, line 64: since `run_bindgen()` is just a normal command, we should add something like:

```rs
println!("cargo::rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS");
println!("cargo::rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS_{target}");
```
Maybe also
```rs
println!("cargo::rerun-if-env-changed=CFLAGS");
println!("cargo::rerun-if-env-changed=CFLAGS_{target}");
```
But I think `cc::build `prints its own `rerun-if-env-changed=...` lines internally (?)

src/rust/bitbox-board-stm32u5a9j-dk-sys/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#![no_std]
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+#![allow(clippy::all)]
+
+#[cfg(all(target_arch = "arm", target_os = "none"))]
```
cedwies, line 9: Similar to previous comment (`target.starts_with("thumb")`?)

src/rust/bitbox-platform-stm32u5/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#![no_std]
+
+#[cfg(all(target_arch = "arm", target_os = "none"))]
```
cedwies, line 8: same

src/rust/bitbox-platform-stm32u5/src/inner.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    // Reset the RCC clock configuration to the default reset state.
+    rcc.cr().write(|w| w.msison().enabled());
+    rcc.cfgr1().write(|w| unsafe { w.bits(0) });
+    rcc.cfgr2().write(|w| unsafe { w.bits(0) });
+    rcc.cfgr3().write(|w| unsafe { w.bits(0) });
+    rcc.cr().modify(|r, w| unsafe {
+        w.bits(
+            r.bits()
+                & !((1_u32 << 16) | (1_u32 << 19) | (1_u32 << 24) | (1_u32 << 26) | (1_u32 << 28)),
```
cedwies: This is hard to read. Can we make these named constants? Or a comment:
`// Clear HSEON, CSSON, PLL1ON, PLL2ON, PLL3ON.`?

src/rust/bitbox-platform-stm32u5/src/inner.rs
```diff
@@ -0,0 +1,40 @@
...
+        ffi::HAL_IncTick();
+    }
+}
+
+/// Rust port of the ST `SystemInit()` startup routine.
+///
+/// This intentionally keeps only the CMSIS/HAL compatibility work that must
+/// happen before regular Rust runtime startup.
+#[pre_init]
+unsafe fn system_init() {
```
cedwies, line 34: Could we document which parts of ST `SystemInit()` are intentionally left out here because `cortex-m-rt` handles them?

src/rust/bitbox-platform-stm32u5/bitbox03-common.ld
```diff
@@ -0,0 +1,294 @@
+/* # Developer notes
```
cedwies, line 1: This matches the layout but while the linker script is derived from `cortex-m-rt`, the .got assertion and interrupt vector presence checks are removed. Why? Was this intentional?

src/rust/bitbox-board-stm32u5a9j-dk-sys/wrapper.h
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#include <board.h>
+#include <flash.h>
```
cedwies, line 4: `flash.c` and `flash.h` are not present in this commit yet, so it's not buildable on it's own. Should we change this?

src/rust/bitbox-board-stm32u5a9j-dk/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+    #[allow(unused_imports)]
+    pub use bitbox_platform_stm32u5_sys::*;
+}
+
+pub mod memory;
+
+#[cfg(all(target_arch = "arm", target_os = "none"))]
+pub fn init() {
+    unsafe {
+        let _ = ffi::HAL_Init();
```
cedwies: We ignore a potential HAL_ERROR?

src/rust/bitbox-board-stm32u5a9j-dk-sys/build.rs
```diff
@@ -0,0 +1,138 @@
...
+    Ok(())
+}
+
+fn is_release_profile() -> bool {
+    env::var("PROFILE").expect("PROFILE not set") == "release"
+}
+
+fn main() -> Result<(), &'static str> {
+    println!("cargo::rerun-if-changed=wrapper.h");
+    println!("cargo::rerun-if-env-changed=PROFILE");
```
cedwies, line 69: Similar to previous comment: this does not track:

`BINDGEN_EXTRA_CLANG_ARGS` and `BINDGEN_EXTRA_CLANG_ARGS_<target>`

src/rust/bitbox-board-stm32u5a9j-dk/src/memory.rs
```diff
@@ -0,0 +1,28 @@
...
+pub const BOOT_ARGS_ADDR: usize = 0x2000_0000;
+pub const BOOT_ARGS_LEN: usize = 512;
+pub const RAM_ADDR: usize = BOOT_ARGS_ADDR + BOOT_ARGS_LEN;
+pub const RAM_LEN: usize = 2496 * 1024 - BOOT_ARGS_LEN;
+pub const SRAM4_ADDR: usize = 0x2800_0000;
+pub const SRAM4_LEN: usize = 16 * 1024;
+pub const GRAM_ADDR: usize = 0xa000_0000;
+pub const GRAM_LEN: usize = 64 * 1024 * 1024;
+
+pub const BOOT0_ADDR: usize = 0x0800_2000;
```
cedwies, line 21: what exactly lives before `BOOT0`? shouldn't boot0 also be immutable? so why the gap between `FLASH_BASE_NS` and `BOOT0_ADDR`?

src/rust/bitbox-board-stm32u5a9j-dk/build.rs
```diff
@@ -0,0 +1,99 @@
...
+    out
+}
+
+fn generate_memory_x(out_dir: &Path, manifest_dir: &Path) {
+    let source = manifest_dir.join("src/memory.rs");
+    println!("cargo::rerun-if-changed={}", source.display());
+    std::fs::write(out_dir.join("memory.x"), memory_x()).expect("write memory layout script");
+}
+
+fn main() {
```
cedwies, line 92: This looks good. But I did not find anything checking that this will stay good in the future. Can we somehow have compile-time assertions, which check overlaps/boundaries so future changes fail immediately if regions overlap or exceed boundaries?

src/rust/bitbox-debug/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+                mode: rtt_target::ChannelMode::NoBlockSkip,
+                name: "API Request",
+                section: ".segger_rtt_buf",
+            }
+        }
+        section_cb: ".segger_rtt"
+        reuse_if_initialized: true
+    };
+    rtt_target::set_print_channel(channels.up.0);
+    rtt_target::init_logger_with_level(log::LevelFilter::Trace);
```
cedwies, line 28: Does this PR make sure that this does not accidentally end up in release/production?

src/rust/bitbox-debug/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "bitbox-debug"
+version = "0.1.0"
+edition = "2024"
+
+[features]
+rtt = ["dep:log", "dep:rtt-target"]
+
+[dependencies]
+log = { version = "0.4.22", default-features = false, optional = true }
+rtt-target = { version = "0.6.1", features = ["log"], optional = true }
```
cedwies: should it really depend on version `0.6.1` instead of `0.6.2`?

src/rust/bitbox-boot-utils/src/image_header.rs
```diff
@@ -0,0 +1,30 @@
...
+pub struct ImageHeader {
+    pub magic: [u8; 4],
+    pub header_len: u32,
+    pub code_size: u32,
+}
+
+impl ImageHeader {
+    pub const LEN: usize = 4 + 4 + 4;
+
+    pub fn from_bytes(bytes: &[u8; Self::LEN]) -> Self {
```
cedwies: could verification make sense here? e.g. that magic is known, len is sane and/or code size fits in slot?

src/rust/bitbox-boot-utils/src/image_header.rs
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pub const IMAGE_HEADER_LEN: usize = 1024;
```
cedwies, line 4: this is duplicate truth (together with `bitbox03_image_header.py`. A test which makes sure that these stay in sync would be nice

scripts/bitbox03_image_header.py
```diff
@@ -0,0 +1,209 @@
...
+    finalize.add_argument("--output", type=Path, required=True)
+
+    finalize_elf = subparsers.add_parser("finalize-elf")
+    finalize_elf.add_argument("elf", type=Path)
+    return parser
+
+
+def main() -> int:
+    args = build_parser().parse_args()
+    if args.command in ("render-dev-header", "render-release-header"):
```
cedwies: `render-dev-header` and `render-release-header` are identical. Is this intentional future proofing?

scripts/bitbox03_image_header.py
```diff
@@ -0,0 +1,209 @@
...
+
+def _elf_section(elf: Path, section_name: str) -> tuple[int, int, int]:
+    data = elf.read_bytes()
+    _validate_elf32_le(data, elf)
+
+    section_header_offset = _read_u32(data, 32)
+    section_header_size = _read_u16(data, 46)
+    section_count = _read_u16(data, 48)
+    section_names_index = _read_u16(data, 50)
+    if section_header_size < 40:
```
cedwies, line 75: Why not `==`? Could a longer `section_header_size` actually be valid?

scripts/bitbox03_image_header.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def _elf_flash_payload_len(elf: Path, payload_address: int) -> int:
+    data = elf.read_bytes()
+    _validate_elf32_le(data, elf)
+
+    program_header_offset = _read_u32(data, 28)
+    program_header_size = _read_u16(data, 42)
+    program_count = _read_u16(data, 44)
+    if program_header_size < 32:
```
cedwies, line 112: Similar as before: why `<` instead of `==`?

scripts/bitbox03_image_header.py
```diff
@@ -0,0 +1,209 @@
...
+        segment_address = _read_u32(header, 12)
+        segment_file_size = _read_u32(header, 16)
+        if segment_type != 1 or segment_file_size == 0:
+            continue
+        # RAM LOAD segments must not turn the image into a sparse flash range.
+        if (segment_address ^ payload_address) & 0xFF00_0000:
+            continue
+        segment_end = segment_address + segment_file_size
+        if segment_end > payload_address:
+            payload_end = max(payload_end, segment_end)
```
cedwies, line 132: `_elf_flash_payload_len()` treats every `PT_LOAD` segment with the same top address byte as the image payload range. That excludes RAM which is good, but I wonder if it would also include any unrelated flash `LOAD` segment in the same 0x08...... region and inflate `payload_size`. Is this possible? Or do can we say that BOOT1/FW ELFs never contain loadable flash segments outside their assigned slot?

scripts/bitbox03_image_header.py
```diff
@@ -0,0 +1,209 @@
...
+    if len(header_bytes) != HEADER_LEN:
+        raise ValueError("header must be exactly 1024 bytes")
+    if not 0 <= code_size <= 0xFFFF_FFFF:
+        raise ValueError("code_size must be a u32")
+    updated = bytearray(header_bytes)
+    updated[8:12] = code_size.to_bytes(4, "little")
+    return bytes(updated)
+
+
+def finalize_code_size(header_bytes: bytes, payload: bytes) -> bytes:
```
cedwies: This wrapper here seems a bit unnecessary?

src/rust/bitbox-boot-utils/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+    expected_magic: [u8; 4],
+) -> Result<*const u32, ()> {
+    let bytes = unsafe { &*(slot_address as *const [u8; ImageHeader::LEN]) };
+    let header = ImageHeader::from_bytes(bytes);
+
+    if header.magic != expected_magic {
+        return Err(());
+    }
+    let header_len = header.header_len as usize;
+    if header_len < ImageHeader::LEN || !header_len.is_multiple_of(4) {
```
cedwies: Why not go for a check which is as strict as possible? (`header_len == 1024`)?

src/rust/bitbox-boot-utils/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+/// `vector_table` must point to readable memory containing at least the first
+/// two Cortex-M vector table entries.
+pub unsafe fn vector_table_is_valid(
+    vector_table: *const u32,
+    code_address: usize,
+    code_len: usize,
+) -> bool {
+    let initial_stack_pointer = unsafe { core::ptr::read_volatile(vector_table) };
+    let reset_vector = unsafe { core::ptr::read_volatile(vector_table.add(1)) };
+    if initial_stack_pointer == u32::MAX || reset_vector == u32::MAX {
```
cedwies, line 261: This checks if `initial_stack_pointer` is erased, which is good. The `fn` does not currently allow it, but would it make sense to add a stronger check here that the stack pointer points to valid RAM and is properly aligned? We would need valid RAM range passed to this `fn` though.

src/rust/bitbox-boot-utils/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+    cortex_m::asm::dsb();
+    cortex_m::asm::isb();
+
+    let reset_vector = unsafe { core::ptr::read_volatile(vector_table.add(1)) };
+    if reset_vector == u32::MAX {
+        halt();
+    }
+
+    unsafe {
+        cortex_m::interrupt::enable();
```
cedwies, line 322: Why don't we let the next image enable interrupts?

src/rust/bitbox-boot-utils/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+    cortex_m::asm::dsb();
+    cortex_m::asm::isb();
+
+    unsafe {
+        (*SCB::PTR).vtor.write(vector_table as u32);
+    }
+    cortex_m::asm::dsb();
+    cortex_m::asm::isb();
+
+    let reset_vector = unsafe { core::ptr::read_volatile(vector_table.add(1)) };
```
cedwies, line 316: `cortex_m::asm::bootload()` will load both both reset vector and initial stack pointer. Why not check stack pointer here too?

src/rust/bitbox-boot-utils/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+
+        assert_eq!(ImageHeader::LEN, 12);
+        assert_eq!(header.magic, IMAGE_HEADER_MAGIC_FIRMWARE);
+        assert_eq!(header.header_len, IMAGE_HEADER_LEN as u32);
+        assert_eq!(header.code_size, 0x1234);
+    }
+
+    #[test]
+    #[cfg(target_pointer_width = "32")]
+    fn test_vector_table_from_image_header() {
```
cedwies: Unhappy paths would be nice here.

src/rust/bitbox-boot-utils/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "bitbox-boot-utils"
+version = "0.1.0"
+edition = "2024"
+
+[lib]
+test = false
```
cedwies: why don't we run the unit tests? `cargo test -p bitbox-boot-utils` does not run the unit tests in `lib.rs`

Note: `cargo test -p bitbox-boot-utils --lib` does run the tests

scripts/bitbox03_image_header.py
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env python3
```
cedwies, line 1: This is crucial logic but AFIK we don't test it properly, do we?

scripts/bootstrap-cargo-config
```diff
@@ -98,11 +116,36 @@ write_cc_env() {
         args="${args} -mcpu=cortex-m4 -mfloat-abi=softfp -mfpu=fpv4-sp-d16"
     fi
 
+    if [[ "${target}" == thumbv8* ]]; then
+        args="${args} -mcpu=cortex-m33 -mthumb -flto"
```
cedwies: `-mthumb` is redunant here.

scripts/flash-bitbox03-boot0-openocd.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+image="${1:-$repo_root/src/rust/target/thumbv8m.main-none-eabihf/debug/bitbox03-boot0}"
```
cedwies, line 5: I understand a default can be convenient and it's fine for dev kits, but later, flashing a wrong boot0 could make a device unusable or require recovery tooling right? Is this script intended to also flash the prod boot0 later?

src/rust/bins/bitbox03-boot1/bitbox03-boot1.ld
```diff
@@ -0,0 +1,27 @@
...
+} INSERT BEFORE .vector_table;
+
+PROVIDE(_flash_start = ORIGIN(FLASH) + IMAGE_HEADER_LEN);
+
+INCLUDE bitbox03-common.ld
+
+ASSERT(ADDR(.image_header) == ORIGIN(FLASH), "
+ERROR(cortex-m-rt): The image header must start at the beginning of the FLASH memory.");
+
+ASSERT(SIZEOF(.image_header) <= IMAGE_HEADER_LEN, "
```
cedwies: Why allow smaller than 1024?

src/rust/bins/bitbox03-boot1/bitbox03-boot1.ld
```diff
@@ -0,0 +1,27 @@
+INCLUDE memory.x
+REGION_ALIAS("FLASH", BOOT1_FLASH);
+
+IMAGE_HEADER_LEN = 1024;
```
cedwies, line 4: We again have a source of truth here. What would be the best way to ensure they stay in sync?

src/rust/bins/bitbox03-boot1/build.rs
```diff
@@ -0,0 +1,82 @@
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+const IMAGE_HEADER_LEN: usize = 1024;
```
cedwies: here again

src/rust/bins/bitbox03-boot1/build.rs
```diff
@@ -0,0 +1,82 @@
...
+    let header_bin = out_dir.join("bitbox03-boot1-header.bin");
+    let header_object = out_dir.join("bitbox03-boot1-header.o");
+
+    println!("cargo::rerun-if-changed={}", script.display());
+    println!("cargo::rerun-if-changed={}", header_manifest.display());
+
+    run_command(
+        Command::new("python3")
+            .arg(&script)
+            .arg("render-dev-header")
```
cedwies: As mentioned before, `render-release-header` does the exact same thing. If that ever changes, note that this `build.rs` always runs `render-dev-header`, even if we build for release.

src/rust/bins/bitbox03-boot1/build.rs
```diff
@@ -0,0 +1,82 @@
...
+
+        println!("cargo::rustc-link-search={}", out_dir.display());
+        println!("cargo::rustc-link-arg=-Tbitbox03-boot1.ld");
+        println!(
+            "cargo::rustc-link-arg=-Map={}",
+            out_dir.join("bitbox03-boot1.map").display()
+        );
+    }
+
+    println!("cargo::rustc-env=BITBOX03_IMAGE_HEADER_LEN={IMAGE_HEADER_LEN}");
```
cedwies: Is `BITBOX03_IMAGE_HEADER_LEN` intended for later build binaries? I do not see it used in this commit.

src/rust/bins/bitbox03-boot1/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[features]
+board-stm32u5a9j-dk = ["dep:bitbox-board-stm32u5a9j-dk"]
+rtt = ["bitbox-debug/rtt"]
+
+[dependencies]
+bitbox-boot-utils = { path = "../../bitbox-boot-utils" }
+bitbox-board-stm32u5a9j-dk = { path = "../../bitbox-board-stm32u5a9j-dk", optional = true }
+
+[target.'cfg(all(target_arch = "arm", target_os = "none"))'.dependencies]
```
cedwies: `Boot0` had these deps unconditionally. I think this (how it's done for `Boot1`) is an improvement, but should we then do it for `Boot0` as well (also for consistency)?

src/rust/bins/bitbox03-firmware/build.rs
```diff
@@ -0,0 +1,84 @@
...
+            .arg(".data=.image_header,alloc,load,readonly,data,contents")
+            .arg(&header_bin)
+            .arg(&header_object),
+        "convert firmware image header to object",
+    );
+
+    println!("cargo::rustc-link-arg={}", header_object.display());
+}
+
+fn main() {
```
cedwies, line 54: `Boot0`/`Boot1` fail if no board feature is selected. Why don't we have the same behavior here?

src/rust/bins/bitbox03-factorysetup/src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+use bitbox_board_stm32u5a9j_dk as board;
+use bitbox_mcu_stm32u5 as _;
+use bitbox_platform_stm32u5 as _;
+use bitbox_platform_stm32u5::flash::{self, BootAddressConfig};
+#[cfg(feature = "board-stm32u5a9j-dk")]
+use board::ffi;
+use core::panic::PanicInfo;
+use cortex_m_rt::entry;
+
+const BOOT0_ADDR: u32 = 0x0800_2000;
```
cedwies: Why not use `board::memory::BOOT0_ADDR`?

src/rust/bins/bitbox03-factorysetup/memory.x
```diff
@@ -0,0 +1,6 @@
+MEMORY
```
cedwies: This duplicates memory assumptions. Should we make sure they don't accidentally drift?

scripts/copy_st_drivers.py
NickeZ: removed for now

scripts/copy_st_drivers.py
NickeZ: Removed for now

src/rust/bitbox-platform-stm32u5-sys/src/lib.rs
NickeZ: unfortunately the "cfg" variables do not contain the full platform. But I also think you cannot do string matching and stuff in "cfg". So I think this is as close as we can get.

src/rust/bitbox-platform-stm32u5-sys/build.rs
NickeZ: yeah, CC already outputs what it needs.

src/rust/bitbox-platform-stm32u5/bitbox03-common.ld
NickeZ: I've tried to make the linker scripts as similar as possible and commented otherwise.

src/rust/bitbox-board-stm32u5a9j-dk/src/memory.rs
NickeZ: Take a look at the ADR for the memory.

src/rust/bitbox-boot-utils/src/image_header.rs
NickeZ: I can remove the unwraps, but I think checking that values are appropriate should be done by the consumers, because it might be different based on the kind of image.

src/rust/bitbox-boot-utils/src/image_header.rs
NickeZ: I'll add a comment

scripts/bitbox03_image_header.py
NickeZ: It seems the ELF spec does allow larger header sizes.

scripts/bitbox03_image_header.py
NickeZ: Same here, the ELF spec does allow larger sizes.

scripts/bitbox03_image_header.py
NickeZ: It seems correct to me, a target has multiple PT_LOAD segments like the vector table, executable text and read only data.

src/rust/bitbox-boot-utils/src/lib.rs
NickeZ: future images with extended headers must be parsable by the current bootloader.

src/rust/bitbox-boot-utils/src/lib.rs
NickeZ: We reset interrupts to their default state. So I think that is reasonable.

src/rust/bitbox-boot-utils/src/lib.rs
NickeZ: maybe this is an outdated comment? there are tests of unhappy paths using `test_vector_table`.

src/rust/bitbox-boot-utils/Cargo.toml
NickeZ: fixed

scripts/bitbox03_image_header.py
NickeZ: We should have a rust crate that handles everything regarding the image header. It might be that all that code is in `bitbox-boot-utils` right now. We can always improve on that later.

scripts/flash-bitbox03-boot0-openocd.sh
NickeZ: factory setup will be a completely different beast. As long as you have "debug probe access" you can flash what you want wherever you want.

Once the chip is locked (debug access disabled) it is not as easy. I don't think this PR even adds any code that would lock the chip.

src/rust/bins/bitbox03-boot1/Cargo.toml
NickeZ: I don't think the conditional makes sense for binary targets. They cannot compile for any other platform. So I removed it

src/rust/bitbox-platform-stm32u5-sys/src/lib.rs
cedwies: ACK. Custom config would be too much.

src/rust/bitbox-platform-stm32u5/bitbox03-common.ld
cedwies: Makes sense. Since all binaries share `bitbox03-common.ld` we should re-enable this check (conditionally or per binary) once Boot1 or firmware start using device interrupts.

src/rust/bitbox-board-stm32u5a9j-dk-sys/wrapper.h
cedwies: In `src/rust/bitbox-board-stm32u5a9j-dk-sys/build.rs` there was also:

```rust
const BOARD_SOURCES: &[&str] = &["stm32u5a9j-dk/Src/board.c", "stm32u5a9j-dk/Src/flash.c"];
```

and `board.c` has 

```rs
#include "flash.h"
```

Why not just move `flash.c` and `flash.h` to when you start referencing them?

src/rust/bitbox-board-stm32u5a9j-dk/src/lib.rs
cedwies: You fixed it with a debug assert. Why only a debug assert?

scripts/bitbox03_image_header.py
cedwies: Yes, multiple `PT_LOAD` segments are expected. But this does not address my concern. My concern was whether the filter is too broad. After re-checking, the current linker layout likely guarantees that ordinary generated segments remain inside the selected slot, but this guarantee is implicit. The failure would probably be safe though, since the bootloader checks that `header_len + code_size` fits the slot and would reject the image. But it could still turn a successful build into an unbootable image. Fine by me.

src/rust/bitbox-boot-utils/src/lib.rs
cedwies: I just noticed that `ImageHeader::LEN` is `12`. I mixed it up with `IMAGE_HEADER_LEN`, which is `1024`. I understand why we want to accept `header_len > IMAGE_HEADER_LEN` for future extended headers. But that does not explain accepting headers between 12` `and `1024` bytes. Is accepting shorter headers intentional? Otherwise, should the minimum be `IMAGE_HEADER_LEN`?

src/rust/bitbox-boot-utils/src/lib.rs
cedwies: The RAM range check is good. But you left out the alignment part, why? `RAM_ADDR + 3` still passes if it's in RAM range.

Also, while adding RAM validation, reset handler validation was weakened (It had to point inside the declared code payload, now it only has to point somewhere inside the entire flash slot, which may include the header or bytes after the declared `code_size`).

src/rust/bitbox-boot-utils/src/lib.rs
cedwies: But still, there is a small window after interrupts are enabled but before we jump to the new image. An interrupt could fire while we are still running with the old image  stack/state, even though VTOR already points to the new image. Wouldn’t it be safer for the new image to enable interrupts once it is ready?

src/rust/bitbox-boot-utils/src/lib.rs
cedwies: I meant unhappy paths of `vector_table_from_image_header()` itself: wrong magic, invalid/misaligned header length, invalid code size, image exceeding the slot, etc. (In case I am not missing sth)

src/rust/bitbox-board-stm32u5a9j-dk-sys/wrapper.h
NickeZ: done, let's see after rebase if it worked :D

src/rust/bitbox-board-stm32u5a9j-dk/src/lib.rs
NickeZ: because we are so early that we don't have any other means of communication. No hardware is initialized, no screen, no nothing. What we can do is break the debugger in debug mode. If we halt there will be no message about why we halted to the user.

src/rust/bitbox-boot-utils/src/lib.rs
NickeZ: The room for the image header is 1024 because there are limitations from the MCU where you can place the interrupt vectors. Basically 1024 is the smallest we can pick and then we can pick multiples of 1024.

`LEN` is the actually used length.

The image header is also far from done with this PR. Included is basically the minimal necessary fields for now.

src/rust/bitbox-boot-utils/src/lib.rs
NickeZ: My gut feeling is that I don't want to trust the payload for the acceptable region.

NickeZ: regarding alignment, I think the last bit in the address signifies if it is thumb/arm mode.

NickeZ: I'll add some alignment checks, 1024 of the vectors and 64 of the SP.

NickeZ: done

src/rust/bitbox-boot-utils/src/lib.rs
NickeZ: since it is the default that interrupts already are enabled we must write our interrupt handlers such that they work wherever the firmware is in its execution. I don't want to stray further away from the default than necessary as it can lead to unexpected complications later.

src/rust/bitbox-boot-utils/src/lib.rs
NickeZ: OK, I've added some tests. We cannot test code that depends on 32 bit behavior and I'm not inclined to complicate the code just to make it testable. Those tests would anyway not be applicable to the actual firmware.

src/rust/bitbox-boot-utils/src/lib.rs
```diff
@@ -0,0 +1,618 @@
...
+/// payload that fits into the slot. The vector table is also checked before it
+/// is returned.
+pub fn vector_table_from_image_header(
+    slot_address: usize,
+    slot_len: usize,
+    ram_base: usize,
+    ram_len: usize,
+    expected_magic: [u8; 4],
+) -> Result<*const u32, ()> {
+    let bytes = unsafe { &*(slot_address as *const [u8; core::mem::size_of::<ImageHeader>()]) };
```
cedwies, line 213: `vector_table_from_image_header()` is a safe function, but it dereferences the caller-provided `slot_address` without being able to verify that it points to readable memory. Could we make `vector_table_from_image_header()` `unsafe`? Otherwise we have a broken promise.

cedwies (changes requested):

src/rust/bitbox-boot-utils/src/lib.rs
NickeZ: good catch, pushed a fix!

cedwies (approved):
tACK.

PR #1973, opened 2026-05-13
Nickez/bb02 lvgl font

NickeZ:
It should look exactly the same, I even applied our manual fixes to the 9px Arial. You can test unicode characters in message signing. For example, try signing `Zürich`.

I propose that we fix the truncation issues by migrating the label component to rust: https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1974

The new font format is more efficient, so the Bootloader is about 700B smaller with the same character set. Firmware size is increased by 7kB


NickeZ:
Updates to the container are necessary, reopning from another branch

PR #1975, opened 2026-05-18
Nickez/ugui no inverse

NickeZ:
I want to remove inverted because multi-byte encodings like utf8 are not trivial to traverse backwards.

remove inverted, transparent and null checks in release mode.

* inverted is only used once (orientation screen)
* transparent is never used
* no need to do null-checks in release builds. Especially not confusing "silently fail" ones.

These all are things that we have added and are not part of the original ugui:

https://github.com/achimdoebler/UGUI/blob/ce0bccb5b7d4877c42081419fccadf7aa5727303/ugui.c#L60


benma (approved):
I'd prefer keeping the NULL checks, they don't really hurt right? And if a refactor or sth smuggles in a NULL it would help.

Fyi: the rotated orientation screen label is not exactly the same as before, but moved by one pixel:

```
  Measured ink bounding boxes:

  before top:    x=35..95, y=1..10
  before bottom: x=33..93, y=54..63

  after top:     x=34..94, y=0..9
  after bottom:  x=33..93, y=54..63
```


[orientation_tap_before_323f_parent.bmp](https://github.com/user-attachments/files/27963610/orientation_tap_before_323f_parent.bmp)
[orientation_tap_after_323f.bmp](https://github.com/user-attachments/files/27963609/orientation_tap_after_323f.bmp)

I guess that is better than before, now it actually looks the same when viewed from either side.

tACK

PR #1980, opened 2026-06-01
Nickez/new firmware targets

NickeZ:


src/CMakeLists.txt
```diff
@@ -535,6 +537,14 @@ if(CMAKE_CROSSCOMPILING)
   target_compile_definitions(firmware-btc.elf PRIVATE PRODUCT_BITBOX_BTCONLY "APP_U2F=0")
   target_sources(firmware-btc.elf PRIVATE ${PLATFORM-BITBOX02-SOURCES})
 
+  target_sources(firmware-nova.elf PRIVATE firmware.c)
+  target_compile_definitions(firmware-nova.elf PRIVATE PRODUCT_BITBOX_MULTI "APP_U2F=1")
+  target_sources(firmware-nova.elf PRIVATE ${FIRMWARE-U2F-SOURCES} ${FIRMWARE-U2F-DRIVER-SOURCES} ${PLATFORM-BITBOX02-SOURCES})
+
+  target_sources(firmware-nova-btc.elf PRIVATE firmware.c)
+  target_compile_definitions(firmware-nova-btc.elf PRIVATE PRODUCT_BITBOX_BTCONLY "APP_U2F=0")
```
benma: Shouldn't this be PRODUCT_BITBOX_NOVA_BTCONLY?

benma (changes requested):

py/bitbox02/bitbox02/bitbox02/bootloader.py
```diff
@@ -17,8 +17,8 @@
     parse_device_version,
     BB02MULTI_BOOTLOADER,
     BB02BTC_BOOTLOADER,
-    BITBOX02PLUS_MULTI_BOOTLOADER,
-    BITBOX02PLUS_BTC_BOOTLOADER,
+    BITBOX02NOVA_MULTI_BOOTLOADER,
+    BITBOX02NOVA_BTC_BOOTLOADER,
```
benma: Are these and the other py lib changes part of the public API? If so, it would be helpful to keep the old names as aliases, maybe with a deprecation warning. Otherwise bump major version and add CHANGELOG entry?

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -202,7 +202,7 @@ class Platform(enum.Enum):
     """Available hardware platforms"""
 
     BITBOX02 = "bitbox02"
-    BITBOX02PLUS = "bitbox02-plus"
+    BITBOX02NOVA = "bitbox02-plus"
```
benma: Breaking API change -> please bump major version and add CHANGELOG entry

benma (changes requested):

py/bitbox02/bitbox02/bitbox02/bootloader.py
NickeZ: reverted, don't want to introduce extra complexity in this PR

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
NickeZ: reverted, don't want to introduce extra complexity in this PR

src/CMakeLists.txt

benma:
https://github.com/BitBoxSwiss/bitbox02-firmware/blob/master/releases/README.md?plain=1#L8

This sentence needs to be changed to say that it was the same until v9.26.1, but different afterwards.

src/usb/class/usb_desc_bitbox02nova.h
```diff
@@ -35,7 +35,7 @@
             0x03, /* bDescriptorType */                                                          \
             'B', 0, 'i', 0, 't', 0, 'B', 0, 'o', 0, 'x', 0, '0', 0, '2', 0, ' ', 0, 'N', 0, 'o', \
             0, 'v', 0, 'a', 0, ' ', 0, 'F', 0, 'a', 0, 'c', 0, 't', 0, 'o', 0, 'r', 0, 'y', 0,
-#elif PRODUCT_BITBOX_BTCONLY == 1
+#elif PRODUCT_BITBOX_NOVA_BTCONLY == 1
```
benma, line 38: We should at some point do a compile time selection of this file instead of dynamic based on the memory flag. (This file is also compiled for BB02, so the BB02 btconly binary will now contain 'multi' string inside 😅

src/platform/platform_init.c
```diff
@@ -53,12 +53,9 @@ void platform_init(void)
             #error "unknown product"
         #endif
     #else
-        // Currently we have one firmware for both BB02 and BB02_NOVA, and only the
```
benma, line 56: Maybe leave a comment to explain that the below is only needed for Nova, but since BLE stuff is still compiled for BB02, it needs to be defined anyway. (Is this true?)

src/rust/bitbox02-sys/build.rs
benma, line 1: Same file on line 347:

```
vec!["-DTESTING", "-D_UNIT_TEST_", "-DPRODUCT_BITBOX_MULTI=1"]
```

Should this maybe be PRODUCT_BITBOX_NOVA_MULTI?

PR #1981, opened 2026-06-02
bootloader: reduce snprintf

benma:
Move the BLE dev menu version formatting to the fixed-format bootloader formatter and add unit coverage.

Extend the unwanted-symbol check to regular development and production bootloader outputs.

Reduces binary size.


NickeZ (approved):
utACK

PR #1984, opened 2026-06-03
Bump Rust toolchain to 1.96

NickeZ:
Update the pinned Rust toolchain to 1.96.0 and adjust the BTC signtx tests for clippy's collapsible_match lint on the new toolchain.


cedwies (approved):
utACK

PR #1985, opened 2026-06-03
Nickez/sysroot wrappers attempt2

NickeZ:
Alternative to https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1982

You need to run `make bootstrap` or `./scripts/bootstrap-cargo-config` once after you checkout the project.

@benma said he didn't have a strong opinion.


scripts/bootstrap-cargo-config
```diff
@@ -0,0 +1,127 @@
...
+{
+    echo "# Generated by scripts/bootstrap-cargo-config."
+    echo "# Host-specific bindgen and cc configuration. Re-run after changing ARM toolchains."
+    echo
+    echo "[env]"
+    if [[ -n "${sysroot}" ]]; then
+        write_cc_env "thumbv7em-none-eabi"
+        write_cc_env "thumbv8m.main-none-eabihf"
+    fi
+    write_bindgen_env "thumbv7em-none-eabi"
```
cedwies, line 118: There is some discrepancy. We allow writing `bindgen_env` with empty `sysroot`, but do not allow this for `cc_env`.

src/rust/bitbox-lvgl-sys/build.rs
```diff
@@ -271,3 +272,9 @@ fn main() -> Result<(), &'static str> {
     fonts.compile("lvgl_fonts");
     run_bindgen(&wrapper, &out_path, &cflags)
 }
+
+fn emit_bindgen_env_rerun_if_changed() {
+    let target = env::var("TARGET").expect("TARGET not set");
+    println!("cargo::rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS");
+    println!("cargo::rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS_{target}");
```
cedwies, line 279: Would it make sense here to check for `CFLAGS` as well here?

cedwies (commented):
This looks good. Left 2 comments.

One command failed on my machine:

`cmake --build build-pr1985-review-arm --target rust-docs -j12`. I consider this non-blocking, but want to mention it here. Consequently, make `rust-docs` also does not work.

src/rust/bitbox-lvgl-sys/build.rs
NickeZ: I'm pretty sure CC emits what is necessary for those variables. https://docs.rs/cc/latest/cc/#external-configuration-via-environment-variables

NickeZ:
Fixed according to review, also removed some optimization flags for bindgen. Bindgen never compiles any code, so optimization flags have no impact. `-fomit-function-pointers` is also enabled by default, so it can be removed.

scripts/bootstrap-cargo-config
```diff
@@ -92,10 +88,14 @@ write_bindgen_env() {
...
-    local args="--sysroot=${sysroot}"
+    local args="-fshort-enums -mthumb"
+
+    if [[ -n "${sysroot}" ]]; then
+        args=" --sysroot=${sysroot}"
+    fi
 
     if [[ "${target}" == thumbv7* ]]; then
-        args="${args} -mcpu=cortex-m4 -mthumb -mfloat-abi=softfp -mfpu=fpv4-sp-d16 -fomit-frame-pointer"
+        args="${args}-mcpu=cortex-m4 -mfloat-abi=softfp -mfpu=fpv4-sp-d16"
```
cedwies: There is a missing space, resulting in `--sysroot=/usr/local/arm-none-eabi-mcpu=cortex-m4`

scripts/bootstrap-cargo-config
```diff
@@ -92,10 +88,14 @@ write_bindgen_env() {
 
 write_cc_env() {
     local target="$1"
-    local args="--sysroot=${sysroot}"
+    local args="-fshort-enums -mthumb"
+
+    if [[ -n "${sysroot}" ]]; then
+        args=" --sysroot=${sysroot}"
```
cedwies: if `sysroot` is not empty, we lose `-fshort-enums -mthumb`

scripts/bootstrap-cargo-config
```diff
@@ -113,10 +113,8 @@ trap 'rm -f "${tmp}"' EXIT
     echo "# Host-specific bindgen and cc configuration. Re-run after changing ARM toolchains."
     echo
     echo "[env]"
-    if [[ -n "${sysroot}" ]]; then
-        write_cc_env "thumbv7em-none-eabi"
-        write_cc_env "thumbv8m.main-none-eabihf"
-    fi
+    write_cc_env "thumbv7em-none-eabi"
```
cedwies, line 116: we now write `cc_env` and `bindgen_env` even if `sysroot` is empty. This solves the discrepancy, which is good, but I wonder why we shouldn't just fail if `sysroot` is empty (i.e. require `sysroot`).

cedwies (changes requested):

scripts/bootstrap-cargo-config
NickeZ: If you only plan to build host things, like the simulator, you don't need these variables. So the MacOS simulator CI build for example would fail.

scripts/bootstrap-cargo-config
NickeZ: dang

cedwies (approved):
tACK

PR #1987, opened 2026-06-03
[wip] add firmware translations

sutterseba:
WIP


src/rust/bitbox-i18n/build.rs
benma, line 1: If possible, it's probably better to move i18n to a new crate `bitbox-i18n`. It's easier that a build.rs is scoped to one task.

PR #1988, opened 2026-06-04
bb03: fix --preseed in simulator

jadzeidan:
fixes --preseed in simulator so it actually
loads with preseeded test wallet.


jadzeidan:
@NickeZ PTAL

NickeZ (approved):
utACK, please sign commit so we can merge.

PR #1992, opened 2026-06-12
Nickez/config time variable

NickeZ:
Build systems are typically split into "config time" options and "build time" options. All of our options has been build time and thus doubling the output targets for every new option we add. Here is a proposal to simplify the build system by bringing most of the options to config time.


adam2k:
Hi @NickeZ, I had a read through this and I like the direction. Config-time options instead of a target per combination seems like a clear win.

Are you planning to rebase this?  It's been sitting a couple of months and master has moved on a fair bit since.  CI is also red at the moment, though it looks like a single real failure with the rest cancelled by fail-fast.

Happy to do a proper review once it's rebased, if that's useful to you.  I have some notes already but they'd be more useful against a current base than against the June state of the branch.

NickeZ:
> Hi @NickeZ, I had a read through this and I like the direction. Config-time options instead of a target per combination seems like a clear win.
> 
> Are you planning to rebase this? It's been sitting a couple of months and master has moved on a fair bit since. CI is also red at the moment, though it looks like a single real failure with the rest cancelled by fail-fast.
> 
> Happy to do a proper review once it's rebased, if that's useful to you. I have some notes already but they'd be more useful against a current base than against the June state of the branch.

This is just a vibe coded PoC. Parts of it might be picked up in future smaller scoped PRs.

PR #1993, opened 2026-06-15
Read chunks via typed pointers

NickeZ:
Use typed pointers for read-only memory accessors so firmware builds can read fields directly from flash instead of copying whole chunks first. Keep test builds on the existing fake read hooks.

### DIFF
--- a/src/memory/memory.c
+++ b/src/memory/memory.c
@@ -233,6 +233,66 @@ static void _read_chunk(uint32_t chunk_num, uint8_t* chunk_out)
 #endif
 }
 
+static const chunk_0_t* _chunk_0(void)
+{
+#ifdef TESTING
+    static chunk_0_t chunk;
+    util_zero(&chunk, sizeof(chunk));
+    _read_chunk(CHUNK_0_PERMANENT, chunk.bytes);
+    return &chunk;
+#else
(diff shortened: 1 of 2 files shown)


benma (changes requested):
CI is failing.

edit: otherwise the change looks great!

benma:
There is also a new warning in host builds:

```

/../../src/memory/memory.c:332:62: warning: '%s' directive output may be truncated writing up to 7963 bytes into a region
  of size 64 [-Wformat-truncation=]


Yes, the warning is new in this commit.

  It appears only in the host/simulator build path:
  make simulator-graphical on bd7712e4b warns in memory_get_device_name().

  The parent commit edcac8b70 builds simulator-graphical without that warning. The firmware target does not emit it on either commit; I rebuilt firmware after checking out both commits, and ARM GCC only showed the
  usual linker/newlib warnings.

  Cause: after the refactor, host GCC sees chunk->fields.device_name as a pointer into the full 8KB chunk, so fortified snprintf("%s", ...) warns it may read up to 7963 bytes. Before, it was a local
  chunk.fields.device_name array, so GCC kept the tighter object bound.


```

benma (approved):
I pushed a commit to fix the warning / CI. Please check and squash & merge if okay.

PR #1994, opened 2026-06-15
bb03 UI: placeholder UI for BTC signing workflows

jadzeidan:
Implement placeholder UI for BTC signing demos
listed in send_message.py.


NickeZ (approved):
utACK

PR #1997, opened 2026-06-22
bb03 ui: navigation icon buttons (Back/Next/Confirm/Cancel)

jadzeidan:
- adds navigation buttons for back, next, confirm, and cancel, replacing existing text buttons.
The icons are PNG bitmaps whereas the button borders are vector drawn.

- on press, the frame fills white and the icon color inverts.

- Adds bitbox-lvgl support (LvStyleTransition, ObjExt add_state/add_flag/
child, LvObjFlag),

- icon assets are located in bitbox03/icons/

- adds a demo screen (ui/demo.rs, reachable in the simulator via the set_device_name `__demo_nav__` sentinel)


jadzeidan:
Screenshots

Demo page of navigation buttons.

image


Example of new buttons in context.

image

py/send_message.py
```diff
@@ -307,6 +307,17 @@ def _reboot(self) -> None:
     def _check_sd_presence(self) -> None:
         print(f"SD Card inserted: {self._device.check_sdcard()}")
 
+    def _demo_nav_buttons(self) -> None:
```
benma: Can just use the "set device name" send message entry and delete this.

For the future, it may be easier to have a flags of the simulator to launch it into specific workflows to increase iteration speed.

src/rust/bitbox03/examples/render_nav_buttons.rs
```diff
@@ -0,0 +1,194 @@
+// SPDX-License-Identifier: Apache-2.0
```
benma: This seems to write screenshots to files or similar? Wouldn't it be better to run the workflows/screens interactively in the simulator? (See my other comment - but could be in another PR)

NickeZ:
I was also going to ask how useful the single screen examples are. because it is quite a lot of code and the screens can be tested in other ways.

If it only was useful to claude while developing, we can delete them now. Claude can generate the specific examples it needs while working.

py/send_message.py
NickeZ: I can imaging that it is easier for an LLM to implement screens if it can produce bitmaps. But we don't need to keep all that intermediate code around.

jadzeidan:
> I was also going to ask how useful the single screen examples are. because it is quite a lot of code and the screens can be tested in other ways.
> 
> If it only was useful to claude while developing, we can delete them now. Claude can generate the specific examples it needs while working.

I find it quite useful for testing in the simulator (actually asked Claude to make it).

Regarding lines of code, can we keep it in the repo but exclude it when making binaries so it doesn't contribute to overall binary size?

NickeZ:
> I find it quite useful for testing in the simulator (actually asked Claude to make it).

The code to generate single screens is not the simulator. Every example file is kind of it's own small simulator that only simulates a single screen.

> Regarding lines of code, can we keep it in the repo but exclude it when making binaries so it doesn't contribute to overall binary size?

It already does not contribute to binary size. The problem is code size, more code always means more to work to maintain.

---

Examples like these are also not representative of any real screen so you might also miss details that you otherwise would care about. (Basically the more similar to prod you test is always better).


> Benma: For the future, it may be easier to have a flags of the simulator to launch it into specific workflows to increase iteration speed.

I foresee complications because our workflows involve multiple request/response rounds.

I propose that we:
1. have an `--unlocked` flag that opens it unlocked. That way you can immediately invoke a workflow from `./send_message.py`.
2. Improve `./send_message.py` to take an argument for the workflow to start. For example `./send_message.py -c '9,1'` for "sign normal btc tx".

@jadzeidan If you agree, can you remove the "examples" from this PR and make another PR with my proposal?

py/send_message.py
jadzeidan: The demo screen is useful for testing buttons (I intentionally added it here), as not all workflows have the same buttons (so need to enter multiple anyways). Also I can test button characteristics (like on-press animations) more thoroughly since it doesn't change the screen or exit the workflow.

jadzeidan:
Removed `examples` as they are not needed for testing. PTAL @NickeZ

py/send_message.py
NickeZ: I think what @benma means is that you can use the existing python command.

Image

NickeZ:
> Removed `examples` as they are not needed for testing. PTAL @NickeZ

Remove the changes to `send_message.py` as well.

jadzeidan:
> > Removed `examples` as they are not needed for testing. PTAL @NickeZ
> 
> Remove the changes to `send_message.py` as well.

Removed now

NickeZ (approved):

PR #1998, opened 2026-06-23
Add IO key backup scripts

NickeZ:
Add J-Link helper scripts to back up and restore the secure-chip IO/auth/encryption key flash blocks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /build/
 /build-*/
+/bitbox02-flash-data-backup-*/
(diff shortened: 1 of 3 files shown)


scripts/securechip-io-keys-backup
benma, line 1: Rename, as it saves more than just the keys, it stores and restores the whole erase block? Why not extend it to cover the full app data area, 64kB instead of 8kB?

benma (approved):
utACK

PR #1999, opened 2026-06-23
Add dev bootloader chip erase

NickeZ:
Add a development-only bootloader command that requests a chip erase after the USB response has been sent.

Expose the command through the Python bootloader client and defer execution until outgoing transport writes are idle.

---

It was useful to restore debug access during experiments with bootloader.

### DIFF
--- a/scripts/template-bootloader.jlink
+++ b/scripts/template-bootloader.jlink
@@ -1,3 +1,15 @@
-loadfile $file
+// Reset and halt the device
 r
+
+// Temporarily disable bootloader protection before flashing.
+// NVMCTRL_CTRLB = NVMCTRL_CTRLB_CMDEX_KEY | NVMCTRL_CTRLB_CMD_SBPDIS.
+w2 0x41004004 0xA51A
+
+// Flash
+loadfile $file 0 noreset
+
+// Reset
+r
+
+// Quit
 q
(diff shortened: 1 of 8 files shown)


scripts/template-bootloader.jlink
```diff
@@ -1,3 +1,15 @@
-loadfile $file
+// Reset and halt the device
 r
+
+// Temporarily disable bootloader protection before flashing.
```
benma, line 4: Why? Doesn't the chip erase bootprot too?

src/bootloader/startup.c
```diff
@@ -156,6 +156,9 @@ int main(void)
                 // screen_sprintf_debug(1000, "got frame");
                 da14531_handler(frame, uart_write_queue);
             }
+            if (!hww_data && rust_bytequeue_num(uart_write_queue) == 0) {
+                bootloader_process_pending_action();
```
benma, line 160: Why all the effort to wait until the response is done and not do it right away when handling OP_CHIP_ERASE? We do it too for the hww reset and reboot API calls in the firmware - the client simply knows to not expect a response in these cases.

NickeZ: Right, we are pretty unique in having a request/reponse protocol and allow some requests to not have responses, so I totally forgot about it.

scripts/template-bootloader.jlink
NickeZ: No, the user row is not reset by chip erase.

PR #2000, opened 2026-06-23
Nickez/bb02 lvgl font

NickeZ:
depends on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2065


tools/ttf2lvgl/Makefile
```diff
@@ -0,0 +1,36 @@
...
+endif
+
+ifeq ($(FREETYPE_LIBS),)
+FREETYPE_LIBS := -L/usr/local/lib -L/opt/homebrew/lib -lfreetype
+endif
+
+ifeq ($(ICU_LIBS),)
+ICU_LIBS := -licuuc -licudata
+endif
+
```
cedwies, line 20: why not something similar for `ICU_CFLAGS`?

tools/ttf2lvgl/ttf2lvgl.c
```diff
@@ -0,0 +1,1136 @@
...
+        snprintf(out, len, "%s.h", font_name);
+        return out;
+    }
+
+    char* out = xstrdup(output_file);
+    char* slash = strrchr(out, '/');
+    char* base = slash == NULL ? out : slash + 1;
+    char* dot = strrchr(base, '.');
+    if (dot != NULL) {
+        strcpy(dot, ".h");
```
cedwies, line 237: Nit: for `foo.` it can write past the allocated string.

tools/ttf2lvgl/ttf2lvgl.c
```diff
@@ -0,0 +1,1136 @@
...
+{
+    for (size_t i = 0; i < len; i++) {
+        if (i % 12 == 0) {
+            fprintf(out, "    ");
+        }
+        fprintf(out, "0x%02X,", bytes[i]);
+        if (i + 1 < len) {
+            fprintf(out, " ");
+        }
+        if (i % 12 == 11 || i + 1 == len) {
```
cedwies, line 688: Nit: That creates trailing whitespace at the end of every full 12-byte line. Fix and regenerate?

tools/ttf2lvgl/ttf2lvgl.c
```diff
@@ -0,0 +1,1136 @@
...
+        fputc((char)(0x80 | (codepoint & 0x3F)), out);
+    } else {
+        fputc((char)(0xF0 | (codepoint >> 18)), out);
+        fputc((char)(0x80 | ((codepoint >> 12) & 0x3F)), out);
+        fputc((char)(0x80 | ((codepoint >> 6) & 0x3F)), out);
+        fputc((char)(0x80 | (codepoint & 0x3F)), out);
+    }
+}
+
+static void print_glyph_comment(FILE* out, UChar32 codepoint)
```
cedwies, line 713: It also does that for non-breaking space and soft hyphen, so we get:

```c
/* U+00A0 " " */
0x00,

/* U+00AD "­" */
0xE0,
```

This can look confusing because e.g. for `U+00A0` it looks like a normal space, when it's not. The function already excludes normal control chars (e.g. newline), but should we escape invisible Unicode too?

src/ui/fonts/password_12.h
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
```
cedwies, line 1: No include guards?

src/ui/fonts/password_9.h
```diff
@@ -0,0 +1,3 @@
+#include <ugui.h>
```
cedwies, line 1: No include guards? And license header is not required?

src/rust/bitbox02-sys/build.rs
```diff
@@ -321,6 +323,13 @@ pub fn main() -> BuildResult<()> {
     emit_rerun_if_changed("../../../src/bootloader/bootloader_version.h.tmpl");
     emit_rerun_if_changed("../../../scripts/generate_version_headers.py");
     emit_rerun_if_changed("../../../scripts/generate_rust_header.sh");
+    emit_rerun_if_changed("../bitbox-lvgl-sys/lv_conf.h");
+    emit_rerun_if_changed("../../../src/ui/fonts/arial_11.h");
+    emit_rerun_if_changed("../../../src/ui/fonts/arial_12.h");
+    emit_rerun_if_changed("../../../src/ui/fonts/arial_9.h");
+    emit_rerun_if_changed("../../../src/ui/fonts/monogram_16.h");
+    emit_rerun_if_changed("../../../src/ui/fonts/password_12.h");
+    emit_rerun_if_changed("../../../src/ui/fonts/password_9.h");
```
cedwies, line 332: doesn't it also depend on e.g. `src/ui/ugui/ugui.h`, `src/ui/fonts/lvgl.h` and lvgl font header under `external/lvgl`?

src/ui/components/trinary_input_char.c
```diff
@@ -34,6 +34,13 @@ typedef struct {
 // Each of the three groups can occupy roughly a third of the width.
 static const UG_S16 _group_width = SCREEN_WIDTH / 3;
 
+static UG_U16 _char_width(const UG_FONT* font, char chr)
+{
+    UG_U16 width = 0;
+    UG_GetCharWidth(font, (uint8_t)chr, &width);
```
cedwies, line 40: would something like `ASSERT(found)` make sense so it does not silently return 0?

src/ui/ugui/ugui.c
```diff
@@ -90,99 +90,132 @@ static void _UG_PSet(UG_S16 x, UG_S16 y, UG_COLOR c)
...
+
+/**
+ * Decodes the next UTF-8 codepoint and returns the byte after it.
+ *
+ * The leading byte determines the sequence length. ASCII bytes are returned
+ * directly, valid 2-, 3-, and 4-byte sequences are assembled from their payload
+ * bits, and malformed sequences fall back to returning the leading byte so
+ * iteration always advances.
+ */
+static const char* _UG_NextCodepoint(const char* str, uint32_t* codepoint)
```
cedwies, line 106: The encoding rules are too loose. This function mainly checks the shape rules, but e.g.

`E0 80 AF` will result in `U+002F` which is `/`.

So e.g. when the first byte is `E0`, second byte must be `A0..BF`

I think your two byte sequence is correct (because it starts at `0xC2` instead of `0xC0`), but the 3 byte and 4 byte parsing needs correction.

src/ui/ugui/ugui.c
```diff
@@ -193,39 +226,38 @@ static void _UG_PutString( UG_S16 x, UG_S16 y, UG_S16 *xout, UG_S16 *yout, const
...
 
-    const int str_length = strlens(str);
+    const char* cursor = str;
 
-    for (int i = 0; i < str_length; i++) {
-        chr = str[i];
-        if (chr != '\n' && (chr < gui->font.start_char || chr > gui->font.end_char)) {
+    while (*cursor != '\0') {
+        cursor = _UG_NextCodepoint(cursor, &codepoint);
+        if (codepoint != '\n' && !UG_GetCharWidth(gui->font, codepoint, &cw)) {
```
cedwies, line 240: Are confident that we can just silently skip? Shouldn't we fail more loudly for a security UI? Or a placeholder which clearly shows that a character was skipped?

src/ui/ugui/ugui.c
cedwies, line 659: This looks risky since it assumes we are still in "character = 1 byte"-world. We could cut a UTF8 character in half. Suppose we have `C3 BC` (`ü`), and we cut after `C3` (which puts `00` at the end). `_UG_NextCodepoint()` then treats the bad leading byte as its own code point (which is `Ã`)

src/ui/ugui/ugui.h
```diff
@@ -64,7 +50,7 @@ typedef struct {
     void (*pset)(UG_S16, UG_S16, UG_COLOR);
     UG_S16 x_dim;
     UG_S16 y_dim;
-    UG_FONT font;
+    const UG_FONT *font;
```
cedwies, line 53: Could we document that the life time contract changed and that callers must never pass a stack/local font? This could be a future dangling pointer trap.

src/rust/bitbox02-rust/src/workflow/verify_message.rs
```diff
@@ -18,10 +20,33 @@ impl core::convert::From<crate::hal::ui::UserAbort> for Error {
     }
 }
 
+fn is_displayable_with_default_font(ui: &impl Ui, bytes: &[u8]) -> bool {
+    let Ok(msg) = core::str::from_utf8(bytes) else {
+        return false;
+    };
+    msg.chars()
+        .all(|c| c == '\n' || (!c.is_control() && ui.has_glyph(Font::Default, c)))
```
cedwies, line 28: `has_glyph` IMO is not the same as "safe to show for signing", since `has_glyph` allows visually ambiguous characters (e.g. `0020` and `00A0` will look (almost) the same, but represent different bytes).

src/rust/bitbox02-rust/src/workflow/verify_message.rs
```diff
@@ -18,10 +20,33 @@ impl core::convert::From<crate::hal::ui::UserAbort> for Error {
...
+}
+
+fn truncate_message_page(page: &str) -> Cow<'_, str> {
```
cedwies, line 31: Although this was present in the old behavior: What is a malicious host puts meaningful content behind byte 640?

src/rust/bitbox02-rust/src/workflow/verify_message.rs
```diff
@@ -36,15 +61,20 @@ pub async fn verify(
...
-        // The message is all ascii and printable.
+    let is_displayable = {
+        let ui = hal.ui();
+        is_displayable_with_default_font(&*ui, msg)
+    };
+
+    if is_displayable {
         let msg = core::str::from_utf8(msg).unwrap();
 
         let pages: Vec<&str> = msg.split('\n').filter(|line| !line.is_empty()).collect();
```
cedwies, line 72: Not introduced in this PR but still:

`pay\nalice` and `pay\n\nalice` will be displayed equally despite being a different byte sequence. Probably out of scope for this PR.

src/rust/bitbox02-rust/src/workflow/verify_message.rs
```diff
@@ -53,7 +83,7 @@ pub async fn verify(
             };
             let params = ConfirmParams {
```
cedwies, line 84: here we do not show `display_size` (in UTF-8 mode we show it). Should we think about showing it here too, especially when trucation happened? Probably out of scope for this PR though.

cedwies (changes requested):

src/rust/bitbox02-rust/src/workflow/verify_message.rs
NickeZ: Maybe there could be a warning for visually similar character? I think customers expect to be able to sign any message they are provided.

tools/ttf2lvgl/ttf2lvgl.c
NickeZ: non-breaking space is the same as space except a word processor will not introduce a line break there.

The fact that `00AD` is there means that the font had a glyph for it. I wonder how it is rendered on the device... maybe as a hyphen?

tools/ttf2lvgl/Makefile
NickeZ: LIBS is a common name for libraries added at linking time comapared to general flags added to the compilation (CFLAGS).

src/ui/ugui/ugui.c
NickeZ: I'll harden it

tools/ttf2lvgl/Makefile
cedwies: Yes, maybe my question was not clear: I was asking about the missing fallback for `ICU_CFLAGS`.

tools/ttf2lvgl/ttf2lvgl.c
cedwies: I don't know how it is rendered on the device. My point is, it would make the generated C source code easier to read. Right now, someone reading the source cannot visually tell which character is there without interpreting `U+00A0`.

If we escape this explicitly (e.g. Newline and tab become `"\n"` and `"\t"`), this would IMO make it easier.

```c
/* U+00A0 "\u00A0" */
0x00,
```

No blocker, though.

cedwies (changes requested):
Resolved all conversations except for two (see follow ups).

I was able to reproduce a bug (cosmetic, stored name remains intact):
In the "See the BitBoxApp" screen, we show the name of the device at the bottom. If the name is too long to be shown, we cut and replace it with "...". Previously, e.g. the letter "ü" was not supported to be the device name. Now, the following is a valid name: 
`ThisIsANiceValidNameWhichIsMoreThan48ButLessThan63ü` (the bug can be reproduced by using this name with the firmware API, not the BitBox App)

Since the bytes of `ü` will be cut incorrectly (by `_truncate_to_fit()` in `lockscreen.c`), the device name will vanish completely.`_truncate_to_fit()` should shorten the name to the previous valid UTF-8 character boundary

Additionally, the issue with undistinguishable characters (e.g. `Õ` and `Ő` are drawn in the same way, just like capital `i` and lowercase `l`, even though this collision pre-dates this PR) still remains. One way would be to manually make them visually different (haven't tested this yet). A user has to know what message is being signed, or else be warned in some way like you suggested (but I assume lots of message contain capital `i` or lowercase `l`, and the warning could be confusing). Also, showing the hex bytes might be confusing too.

NickeZ:
> Resolved all conversations except for two (see follow ups).
> 
> I was able to reproduce a bug (cosmetic, stored name remains intact): In the "See the BitBoxApp" screen, we show the name of the device at the bottom. If the name is too long to be shown, we cut and replace it with "...". Previously, e.g. the letter "ü" was not supported to be the device name. Now, the following is a valid name: `ThisIsANiceValidNameWhichIsMoreThan48ButLessThan63ü` (the bug can be reproduced by using this name with the firmware API, not the BitBox App)
> 
> Since the bytes of `ü` will be cut incorrectly (by `_truncate_to_fit()` in `lockscreen.c`), the device name will vanish completely.`_truncate_to_fit()` should shorten the name to the previous valid UTF-8 character boundary

I'll take a closer look.

> 
> Additionally, the issue with undistinguishable characters (e.g. `Õ` and `Ő` are drawn in the same way, just like capital `i` and lowercase `l`, even though this collision pre-dates this PR) still remains. One way would be to manually make them visually different (haven't tested this yet). A user has to know what message is being signed, or else be warned in some way like you suggested (but I assume lots of message contain capital `i` or lowercase `l`, and the warning could be confusing). Also, showing the hex bytes might be confusing too.

If the character is used in your native language (like [dotless i](https://en.wikipedia.org/wiki/Dotless_I)) then it would be bad UX to warn about it.

I think as long as we check that for example addresses only use supported characters we have checked enough. When you sign a message in real life you don't really care about the unicode code point of every character. You sign the "intent" of the message.

PR #2001, opened 2026-06-29
Add unified BitBox02 CLI

NickeZ:
Introduce bitbox02_cli.py as the canonical Python CLI with explicit interactive mode and non-interactive subcommands for device, wallet, bootloader, and firmware workflows.

Keep send_message.py and load_firmware.py as compatibility wrappers, and update docs plus Python linting to point at the new entrypoint.

---

This is vibe coded so far. I'll need to make a self-review as well. Looking for concept ack from @benma and @cedwies


cedwies:
concept ACK

benma:
Maybe rename to `bitbox_cli` so it works with future BitBoxes too?

scripts/lint-python
```diff
@@ -23,7 +23,7 @@ ARGS=($(find py releases -name '*.py' | grep -v -e generated -e old))
 # behavior).
 # We must typecheck the whole `py` directory even if only a few files are modified so that mypy
 # sees all types
-${MYPY} --implicit-reexport --namespace-packages --ignore-missing-imports --strict py/send_message.py py/bitbox02/bitbox02 releases/describe_signed_firmware.py
+${MYPY} --implicit-reexport --namespace-packages --ignore-missing-imports --strict py/bitbox02_cli.py py/bitbox02/bitbox02 releases/describe_signed_firmware.py
```
benma, line 26: keep send_message.py? Can't hurt to check it while it exists.

py/README.md
```diff
@@ -47,15 +47,19 @@ firmware manually read the next section or install with the official BitBoxApp.
 
 Connect your BitBox and "tap this side".
 
-List and execute all available commands by running:
+List available commands by running:
 
 ```bash
-python ./send_message.py
+python ./bitbox02_cli.py
```
benma, line 53: Add non-interactive examples? I am confused how to use it, what are the args to the API calls when it's not interactive, like signmsg?

benma (commented):
The commit msg does not sufficiently explain why a non interactive tool is needed. What do we need it for? Does it make sense to have the interactive and non-interactive tools be unified in one script?

Also this was a chance to split the monster send_message.py into multiple files 😁

py/README.md
NickeZ: `--help` ;).

It is all vibe coded so far, so mostly looking for concept ack.

benma (approved):
concept ACK. approved because i can't remove myself as reviewer 😄

PR #2002, opened 2026-06-30
Speed up firmware updates

NickeZ:
Approximately halfs the current flashing time to 55s.

---

Erase the full firmware area on OP_ERASE, then flash 4 KiB firmware chunks.

### DIFF
--- a/src/bootloader/bootloader.c
+++ b/src/bootloader/bootloader.c
@@ -125,6 +125,7 @@ COMPILER_PACK_RESET()
 static bool _firmware_block_cache_valid = false;
 static uint8_t _firmware_cached_block = 0;
 static uint8_t _firmware_next_chunk = 0;
+static bool _firmware_app_erased = false;
 // Indicates whether the whole app flash contains only 0xFF.
 // This controls bootloader text messages on the screen.
 // The value is computed at bootloader enter.
@@ -487,6 +488,9 @@ static uint8_t _firmware_erase_block(uint32_t addr)
         addr % FIRMWARE_BLOCK_LEN != 0) {
         return OP_STATUS_ERR_ERASE;
     }
+    if (_is_erased((const void*)addr, FIRMWARE_BLOCK_LEN)) {
+        return OP_STATUS_OK;
+    }
     if (flash_erase(&FLASH_0, addr, FLASH_ERASE_PAGE_NUM) != ERR_NONE) {
         return OP_STATUS_ERR_ERASE;
     }
@@ -508,7 +512,7 @@ static uint8_t _firmware_block_cache_flush(void)
         return OP_STATUS_OK;
     }
 
-    if (_is_erased((const void*)addr, FIRMWARE_BLOCK_LEN)) {
+    if (_firmware_app_erased || _is_erased((const void*)addr, FIRMWARE_BLOCK_LEN)) {
         if (flash_append(&FLASH_0, addr, _firmware_block_cache, FIRMWARE_BLOCK_LEN) != ERR_NONE) {
             return OP_STATUS_ERR_WRITE;
         }
@@ -581,6 +585,7 @@ static size_t _api_firmware_erase(uint8_t firmware_num_chunks, uint8_t* output)
     }
     _loading_ready = false;
     _firmware_num_chunks = 0;
+    _firmware_app_erased = false;
     _firmware_block_cache_reset();
     for (uint32_t i = 0; i < (uint32_t)FLASH_APP_PAGE_NUM; i += FLASH_REGION_PAGE_NUM) {
         if (flash_unlock(&FLASH_0, FLASH_APP_START + i * FLASH_PAGE_SIZE, FLASH_REGION_PAGE_NUM) !=
@@ -596,6 +601,7 @@ static size_t _api_firmware_erase(uint8_t firmware_num_chunks, uint8_t* output)
             return _report_status(status, output);
         }
     }
+    _firmware_app_erased = true;
 
     if (firmware_num_chunks > 0) {
         _firmware_num_chunks = firmware_num_chunks;
--- a/src/bootloader/bootloader.c
+++ b/src/bootloader/bootloader.c
@@ -44,16 +44,14 @@
 
 // Bootloader API command op codes
 //
-// OP_ERASE - Receives the total number of firmware chunks to write,
-// then pads any left over FLASH that would be empty by 0xFFs;
-// FLASH areas that will contain firmware are erased when writing the
-// firmware chunk to FLASH.
+// OP_ERASE - Erases the whole firmware flash area. Receives the total number
+// of firmware chunks to write. If non-zero, starts a sequential firmware
+// loading session.
 #define OP_ERASE ((uint8_t)'e') /* 0x65 */
 // OP_REBOOT - Reboot the MCU and clear the auto_enter flag.
 #define OP_REBOOT ((uint8_t)'r') /* 0x72 */
-// OP_WRITE - Write a firmware chunk; the binary must be streamed in chunks.
-// The command must include the chunk number in order to be written to the
-// correct FLASH location.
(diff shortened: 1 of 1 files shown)


benma (commented):
Seems overly complicated, I think the cache/flush is not needed?

Why not just erase in OP_ERASE, and just use flash_append with 4kB chunks during flashing?

> Also reject signature data before all declared firmware chunks have been received and flushed.

Why? One can't successfully do it anyway beforehand, as that API call rejects on invalid signatures. Can't hurt but seems not needed.

src/bootloader/bootloader.c
```diff
@@ -451,6 +464,69 @@ static size_t _report_status(uint8_t status, uint8_t* output)
...
+        return OP_STATUS_OK;
+    }
+
+    const uint32_t addr = FLASH_APP_START + _firmware_cached_block * FIRMWARE_BLOCK_LEN;
+    if (MEMEQ((const void*)addr, _firmware_block_cache, FIRMWARE_BLOCK_LEN)) {
+        _firmware_block_cache_valid = false;
+        return OP_STATUS_OK;
+    }
+
+    if (_firmware_app_erased || _is_erased((const void*)addr, FIRMWARE_BLOCK_LEN)) {
```
adam2k: This branch looks unreachable. `_firmware_app_erased` is cleared only at 588, immediately after `_loading_ready = false` (586), and set true at 604 before `_loading_ready = true` (608). `_api_write_chunk requires` `_loading_ready` on entry and only re-arms it on success, so `_loading_ready` implies `_firmware_app_erased`. Both the `_is_erased()` call and the `flash_write` branch below are dead.

src/bootloader/bootloader.c
```diff
@@ -451,6 +464,69 @@ static size_t _report_status(uint8_t status, uint8_t* output)
...
+}
+
+static uint8_t _firmware_block_cache_flush(void)
+{
+    if (!_firmware_block_cache_valid) {
+        return OP_STATUS_OK;
+    }
+
+    const uint32_t addr = FLASH_APP_START + _firmware_cached_block * FIRMWARE_BLOCK_LEN;
+    if (MEMEQ((const void*)addr, _firmware_block_cache, FIRMWARE_BLOCK_LEN)) {
```
adam2k: Near-dead: post-erase the block is `0xFF` and each block is written once, so this only hits on an all-`0xFF` chunk pair, at the cost of an 8 kB compare per block. The old per-chunk MEMEQ paid for itself because flash wasn't pre-erased.

src/bootloader/bootloader.c
```diff
@@ -462,35 +538,31 @@ static size_t _api_write_chunk(const uint8_t* buf, uint8_t chunknum, uint8_t* ou
         return _report_status(OP_STATUS_ERR_MACRO, output);
     }
 
-    // The second is redundant, as _firmware_num_chunks <=
-    // FIRMWARE_MAX_NUM_CHUNKS.
-    if (chunknum > _firmware_num_chunks - 1 || chunknum > FIRMWARE_MAX_NUM_CHUNKS - 1) {
+    // The second is redundant, as _firmware_num_chunks <= FIRMWARE_MAX_NUM_CHUNKS.
+    if (chunknum > _firmware_num_chunks - 1 || chunknum > FIRMWARE_MAX_NUM_CHUNKS - 1 ||
+        chunknum != _firmware_next_chunk) {
```
adam2k: This makes out-of-order writes a hard error, tightening the host protocol permanently in the bootloader. `bootloader.py:191` is sequential so the bundled client is fine, but worth confirming BitBoxApp is too, and that it's intended as a contract change.

src/bootloader/bootloader.c
```diff
@@ -514,28 +584,25 @@ static size_t _api_firmware_erase(uint8_t firmware_num_chunks, uint8_t* output)
...
-        if (MEMEQ((const void*)addr, empty_page, sizeof(empty_page))) {
-            continue;
-        }
-        if (flash_erase(&FLASH_0, addr, FLASH_ERASE_PAGE_NUM) != ERR_NONE) {
-            return _report_status(OP_STATUS_ERR_ERASE, output);
-        }
-        if (!MEMEQ((const void*)addr, empty_page, sizeof(empty_page))) {
-            return _report_status(OP_STATUS_ERR_CHECK, output);
+
+    for (uint32_t addr = FLASH_APP_START; addr < FLASH_APP_START + FLASH_APP_LEN;
```
adam2k, line 478: This erases 864 kB (108 blocks) in one blocking call, where previously a full-size firmware erased almost nothing here.  Any host-side timeout on `OP_ERASE`?  `_render_progress(0)` is also drawn before the loop, so the screen sits at 0% throughout.  Disclaimer:  I do _not_ have a dev device, so I can't measure it.

src/bootloader/bootloader.c
```diff
@@ -894,7 +965,11 @@ static size_t _api_command(const uint8_t* input, uint8_t* output, const size_t m
         if (output[1] != OP_STATUS_OK) {
             bootloader_render_default_screen();
         } else {
-            _render_progress((float)chunk_num / (float)(_firmware_num_chunks - 1));
+            float progress = 1;
```
adam2k, line 853: Real fix rather than cleanup: at `_firmware_num_chunks == 1` the old expression is `0/0 = NaN`. Unrelated to the caching change, so maybe its own commit.

adam2k (commented):
I'm new here and disclaimer up front I do _not_ have a dev device, so I am doing static analysis only.

@benma asked whether the cache is needed.  I believe it is not, and the HAL explains the speedup: `flash_write` -> `_flash_write` (`hpl_nvmctrl.c:158`) is read-modify-write per 8 kB block, so two 4 kB chunks cost two erase-and-reprogram cycles. `flash_append` -> `_flash_append` (`:204`) programs page by page with no erase. Chunks sit at `FLASH_APP_START + n*4096` with `FLASH_APP_START == 0x10000`, so they're always page-aligned. After the upfront erase, plain `flash_append` per chunk should be enough.

Two branches in the current shape already look dead as a result.  I've noted those inline.

adam2k:
> Why not just erase in OP_ERASE, and just use flash_append with 4kB chunks during flashing?

+1 for the above comment.  `_api_write_sig_data` -> `_set_firmware_data` -> `_firmware_verified_jump()` verifies over what's actually in flash, so an incomplete firmware fails there regardless.

src/bootloader/bootloader.c
NickeZ: thanks, not applicable anymore.

src/bootloader/bootloader.c
NickeZ: thanks, not applicable anymore.

src/bootloader/bootloader.c
NickeZ: thanks, not applicable anymore.

src/bootloader/bootloader.c
NickeZ: I haven't noticed any issues in testing.

src/bootloader/bootloader.c
NickeZ: fixed

NickeZ:
Simplified PR. Still flashes in 55 seconds. It is on "Upgrading" screen while erasing for about 4-5 seconds. It doesn't stand still at 0%.

src/bootloader/bootloader.c
NickeZ: That call to `_render_progress(0)` was probably removed, so now it sits at "Upgrading" instead.

NickeZ:
> Simplified PR. Still flashes in 55 seconds. It is on "Upgrading" screen while erasing for about 4-5 seconds. It doesn't stand still at 0%.

Actually flashing only takes about 60 seconds. So there is more like a 10% upgrade. I guess we already removed a lot of double work in some earlier PR.

PR #2003, opened 2026-07-06
Fix LVGL transition style clearing

NickeZ:
Wrap the transition style setter around LvStyleTransition so callers use the safe wrapper instead of raw descriptors.

Remove the local transition style property when None is passed, matching the behavior of other optional style setters.


jadzeidan (approved):
tAck. Tested UI and still works as expected.

PR #2007, opened 2026-07-09
Warn before truncated value displays

benma-agent:
Show the existing large-value warning before value confirmation bodies that exceed the UI label limit.

Apply it to shared message verification, Ethereum transaction and typed-data values, and BTC policy name/policy displays.


cedwies (approved):
utACK.

Nit: `bitbox02-firmware/src/ui/components/label.h` still has the stale comment `// Keep this in sync with src/rust/bitbox02-rust/src/hww/api/ethereum.rs:MAX_CONFIRM_BODY_SIZE`

Nit: BTC policy display now uses `confirm_value()` but there is no direct policy-specific long-policy test. But I leave that up to you if you want to add another test for this.

PR #2010, opened 2026-07-09
eth: loosen EIP-712 identifier validation

Tomasvrba:
Allow ':' in EIP-712 type names used by some dapps as a namespace separator


Tomasvrba:
closing in favor of https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2011

PR #2011, opened 2026-07-09
eth: loosen EIP-712 identifier validation

benma:
Allow ':' in EIP-712 type names used by some dapps as a namespace separator


Tomasvrba (approved):
tACK

PR #2013, opened 2026-07-14
ui: keep frame during async processing

benma-agent:
Async operations can leave the screen stack temporarily empty between
two UI calls. This caused the waiting screen to flash, unlike the old
synchronous securechip behavior.

Treat locally processing USB requests as active UI work and keep the last
frame when no component is stacked. Host waits, screensavers, and explicit
components continue to render normally.

### DIFF
--- a/src/rust/bitbox02-rust/src/async_usb.rs
+++ b/src/rust/bitbox02-rust/src/async_usb.rs
@@ -109,6 +109,17 @@ pub fn is_idle() -> bool {
     matches!(*USB_TASK_STATE.0.borrow(), UsbTaskState::Nothing)
 }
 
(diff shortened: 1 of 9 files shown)


NickeZ (approved):
tACK,

One time I definitely saw the keyboard again after unlocking the device. But I can't reproduce it.. 🤷

benma:
> One time I definitely saw the keyboard again after unlocking the device. But I can't reproduce it.. 🤷

Interesting. Sol xhigh traced the execution and rendering and couldn't figure out how the keyboard (letters) would be rendered after the arrows touch - arrows touching would be the last frame and linger until the next UI screen.

PR #2014, opened 2026-07-19
usb: block HWW during U2F workflows

benma-agent:
Track detached U2F workflow lifetimes independently of their public
result state. Return HWW_RSP_BUSY before starting an HWW request while a
U2F workflow can still own the shared UI.

This prevents HWW session reset and cancellation paths from invalidating
live U2F UI objects. Add regressions for the lifetime guard and busy
response.


benma:
@NickeZ I successfully tested this. HWW in send_messae.py gets a busy response and does not start a workflow while a U2F screen is displayed. Before this commit, HWW would launch into action overlaying the U2F screen.

NickeZ (approved):
utACK

I would've preferred a solution where all workflows are blocked in case any workflow is running. But this is still an improvement.

PR #2017, opened 2026-07-21
Add shared Bitcoin transaction vectors

benma-agent:
Add a readable source of Bitcoin transaction signing vectors, authored
as PSBTs with the metadata needed by signing APIs. Derive firmware
requests in memory, exercise them in signtx.rs, and remove overlapping
bespoke tests.

Advance the development firmware version to v9.27.0 and add versioned
coverage for the all-change fee confirmation flow introduced there.
Older simulator entries are skipped only where their stdout omitted
address and fee screens; genuine older behavior remains versioned.

For successful vectors, populate the PSBT with the firmware responses
and check that the completed transaction and its signatures are valid.

The generated fixtures are meant to be consumed by client libraries so
firmware and clients can cover the same transactions and versioned
expectations without repo-specific vector sources. They will also be
useful when adding the Bitcoin API to the new bitbox-api-ts library.

Keep the generated JSON synchronized with its Rust source through a
drift test.

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs
@@ -1361,6 +1361,8 @@ mod tests {
     use crate::hal::{Memory, testing::TestingHal};
     use crate::keystore::testing::{mock_unlocked, mock_unlocked_using_mnemonic};
     use alloc::boxed::Box;
+    use alloc::collections::{BTreeMap, BTreeSet};
+    use bitbox_test_vectors::btc_transaction as btc_test_vectors;
     use hex_lit::hex;
     use pb::btc_payment_request_request::{Memo, memo};
     use util::bip32::HARDENED;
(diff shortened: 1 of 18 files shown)


cedwies:
Concept ACK, looks good. Here are some points I'd like to raise for continuing:

1. Manual copying is reasonable for proving the concept. Do you want an automated update/hash check for production?
2. For some test vectors you check the shape (correct signature kind, valid looking encoding, ...) and not whether the signature is cryptographically correct. Will the finished vector-based firmware tests independently verify all returned signatures, and will client tests keep equivalent validation for every case they support? Or how do we prevent all clients from agreeing with a wrong answer?
3. I think skipping cases should be made explicit and visible in order to avoid silent skipping while CI remains green. Like: `These exact eight vectors are intentionally unsupported.
If that list changes, fail (and require a human decision?).`

benma:
> Concept ACK, looks good. Here are some points I'd like to raise for continuing:

Thanks!

 
>     1. Manual copying is reasonable for proving the concept. Do you want an automated update/hash check for production?

I don't think it's needed, at least not for the time being. Similar to protobuf messages, they are also copied to the client libs on demand. Maybe a small utility tool or Makefile entry to make it easier, but it's only copying one file :shrug: 
 
>     2. For some test vectors you check the shape (correct signature kind, valid looking encoding, ...) and not whether the signature is cryptographically correct. Will the finished vector-based firmware tests independently verify all returned signatures, and will client tests keep equivalent validation for every case they support? Or how do we prevent all clients from agreeing with a wrong answer?

The signatures are created inside the table driven unit tests, and their validity (full tx verification including sigs) is checked in the client libs. I thought this was also done in the vector tests in this repo, but I was wrong, so it slipped through - will add it here too.

> 
>     3. I think skipping cases should be made explicit and visible in order to avoid silent skipping while CI remains green. Like: `These exact eight vectors are intentionally unsupported. If that list changes, fail (and require a human decision?).`

Skipping happens when the vector is not supported, and it is already explicit. Afaik in the firmware, all vectors are used. In the Rust lib, the vectors with silent_payments / payment_requests  are skipped as not supported yet:

```Rust
fn skip_reason(vector: &TestVector) -> Option<&'static str> {
    if !vector.psbt.options.outputs.is_empty() || !vector.psbt.options.payment_requests.is_empty() {
        return Some("btc_sign_psbt does not expose per-output options or payment requests");
    }
    None
}
```

The Go lib also supports all.

Did I miss something?

benma:
> The signatures are created inside the table driven unit tests, and their validity (full tx verification including sigs) is checked in the client libs. I thought this was also done in the vector tests in this repo, but I was wrong, so it slipped through - will add it here too.

@cedwies added the validity/sig-check now in a 2nd commit

cedwies:
Nice, regarding 1:

ACK, makes sense.

regarding 3:

The rule is explicit, but my concern was slightly different: The exact amount of skipped coverage is not fixed or enforced. If today 8 vectors match the skip condition, and another vector is added and also skipped, this can be overlooked. But this would be a small follow-up at most, and not relevant to the concept ACK.

benma:
> regarding 3:
> 
> The rule is explicit, but my concern was slightly different: The exact amount of skipped coverage is not fixed or enforced. If today 8 vectors match the skip condition, and another vector is added and also skipped, this can be overlooked. But this would be a small follow-up at most, and not relevant to the concept ACK.

Makes sense to help reviewers see which test inclusions/exclusions happened, but I agree it's not urgent. 

I lifted this PR out of draft, ready for actual review (along with the PRs in the other repos)

src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs
```diff
@@ -1740,6 +1743,756 @@ mod tests {
...
+                    });
+                    observations.borrow_mut().observe(&response);
+                    {
+                        let observations = observations.borrow();
+                        assert_vector_observations(vector, &sign_request, &observations);
+                        assert_vector_transaction_valid(vector, &observations);
+                    }
+                    assert_vector_pubkeys(&mut mock_hal, vector, &sign_request).await;
+                }
+                btc_test_vectors::Outcome::InvalidInput => {
```
cedwies, line 2575: Nonblocking: For `InvalidInput` vectors that are intended to fail during pre-sign validation, would it make sense to also assert that `observations.signatures ` is empty? Before, this was not checked either, so this would not be a regression. It would protect against a regression where the firmware returns a signature for Invalid Input.

cedwies (approved):
utACK, please see my comment

benma-agent:
Please take [PR #2042](https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2042) into account when rebasing this PR before merging. It adds an input-relative high-fee warning for all-change and OP_RETURN-only transactions, with tests currently added directly in `signtx.rs`.

Since this PR replaces overlapping bespoke `signtx` tests with shared vectors, the vector cases/screens likely need to be updated or extended to cover the new high-fee warning screen. Once equivalent vector coverage exists, the manual `test_high_fee_warning_total_inputs` and `test_no_high_fee_warning_total_inputs_below_threshold` additions from #2042 may be removable during conflict resolution.

benma:
Rebased and migrated the new fee tests from https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2042 to the vectors table. Also removed all the separate <v9.20 cases that didn't add any coverage and just existed because the simulators of these versions didn't print some of the screens.

PR #2020, opened 2026-07-28
usb: fix watchdog aborting Bluetooth requests

### LINKED ISSUE #2019: BLE: EIP-712 signing NACKs when a request needs a device confirmation (same request signs over USB)
## Summary

On a BitBox02 Nova (main firmware **9.26.4**, Bluetooth firmware **2**), an `ETHSignTypedMessage` request whose `EIP712Domain` has no `chainId` fails over **BLE** with `hwwRspNack` (0x03) after the "Typed data has no chain ID" warning is confirmed.

The **identical request over USB signs correctly** on the same device, and a request that carries `chainId` signs correctly over BLE. Both firmware versions are reported as up to date by the BitBoxApp.

Reproduced using only `bitbox02-api-go` plus a generic BLE library — the repro below contains no vendor-specific transport code beyond ~60 lines of GATT plumbing.

## Results

Same device, same message, four combinations:

| Transport | Stack | domain without `chainId` | domain with `chainId` |
|---|---|---|---|
| USB | `bitbox02-api-go` + `karalabe/hid` | **signs** | signs |
| BLE | `bitbox02-api-go` + `tinygo.org/x/bluetooth` | **NACK** | signs |

The USB row used the same 13-member message as the BLE row, so payload size and content are not the variable — the transport is.

## Why the domain matters here

Missing `chainId` is what makes `sign_typed_msg.rs` display a confirmation before the device can answer:

```rust
let cached_chain_id = validate_chain_id(request).await?;
if cached_chain_id.is_none() {
    hal.ui().confirm(&ConfirmParams {
        title: "Warning",
        body: "Typed data has no chain ID. Message is valid for every chain.",
        ..Default::default()
    }).await?;
}
```

That confirmation forces the host into the `hwwRspNotready` retry loop. With `chainId` present there is no warning, the device answers immediately (`polls=0`), and the retry loop is never entered — which is why that variant succeeds over BLE.

So the trigger appears to be **a multi-frame request held across a poll cycle over BLE**, not the chain-ID validation itself. Consistent with that, a *small* request in the retry loop also succeeds over BLE; only multi-frame + polling fails.

## Frame trace (BLE, failing case)

```
> write 64 bytes: ff000000c1000101…    # hwwReqRetry
< read  64 bytes: ff000000c1000101…    # hwwRspNotready   (repeats)
> write 64 bytes: ff000000c1000101…
< read  64 bytes: ff000000c1000103…    # hwwRspNack
```

`rawQueryV7` turns this into `unexpected NACK response` (`api/firmware/query.go:104`). It is not a user abort — an abort returns a decrypted `Response_Error` with code 104.

Timing is not the factor: failures occur anywhere from ~2 s to ~10 s of dwell on the warning, and longer waits on other confirmations in the same session succeed.

## Reproduction

Go program using `bitbox02-api-go` for all protocol work (noise, U2FHID framing, protobuf) and `tinygo.org/x/bluetooth` for transport. The only local code is an `io.ReadWriteCloser` that writes 64-byte frames to the writer characteristic and reads them from notifications on the reader characteristic.

GATT UUIDs used:

```
service  E1511A45-F3DB-44C0-82B8-6C880790D1F1
writer   799D485C-D354-4ED0-B577-F8EE79EC275A
reader   419572A5-9F53-4EB1-8DB7-61BCAB928867
product  9D1C9A77-8B03-4E49-8053-3955CDA7DA93
```

```go
type bleDevice struct {
    writer bluetooth.DeviceCharacteristic
    frames chan []byte
}

func (d *bleDevice) Write(p []byte) (int, error) {
    for off := 0; off < len(p); off += 64 {
        end := min(off+64, len(p))
        if _, err := d.writer.WriteWithoutResponse(p[off:end]); err != nil {
            return off, err
        }
    }
    return len(p), nil
}

func (d *bleDevice) Read(p []byte) (int, error) {
    select {
    case frame := <-d.frames:
        return copy(p, frame), nil
    case <-time.After(90 * time.Second):
        return 0, fmt.Errorf("BLE read timed out")
    }
}
func (d *bleDevice) Close() error { return nil }

// …then, exactly as in cmd/playground:
comm := u2fhid.NewCommunication(bd, 0x80+0x40+0x01)
device := firmware.NewDevice(nil, nil, &mocks.Config{}, comm, &mocks.Logger{})
device.Init()
sig, err := device.ETHSignTypedMessage(1, keypath, []byte(msgNoChainID))
```

The message is a 13-member struct with a `{name, version}` domain; the control is the same message with `chainId` added to the domain and types. Happy to share the full program or run further tests on this hardware.

## Environment

- BitBox02 Nova Multi, `bb02p-multi`
- Main firmware 9.26.4 (BitBoxApp: up to date)
- Bluetooth firmware 2 (BitBoxApp: up to date)
- Secure chip OPTIGA_TRUST_M_V3
- Host: macOS, `bitbox02-api-go` `v0.0.0-20250819132928-6ab3141dbffa`

## Note

I opened #2018 earlier today from an incomplete diagnosis and closed it — that report claimed the chain-ID warning could not be dismissed, which is wrong: over USB it is dismissable and signing proceeds exactly as the source and its unit test describe. This issue is the corrected finding, with the USB leg as control.

TaprootFreak:
Signing an EIP-712 typed message whose `EIP712Domain` carries no `chainId` fails over Bluetooth
with `hwwRspNack` shortly after the "Typed data has no chain ID" warning is confirmed on the
device. The same message signs correctly over USB, and the same message with a `chainId` signs
correctly over Bluetooth too.

## Cause

`hwwRspNack` in reply to an `HWW_REQ_RETRY` has one source: `_maybe_write_response()` in
`src/hww.c` falls through to `HWW_RSP_NACK` when `rust_async_usb_copy_response()` reports
`UsbResponseNack`, which `take_response()` only returns for `UsbTaskState::Nothing` — no task is
running any more. After an `HWW_REQ_NEW` the state is `Running`, and short of the host collecting
the final result, the only way out of it is `async_usb::cancel()`. Of its two call sites, the one
in `hww/api/bluetooth.rs` runs only after a successful Bluetooth firmware upgrade and is
unreachable for an Ethereum request, which leaves `hww_abort_outstanding_op()` called from
`_check_lock_timeout()` in `src/usb/usb_processing.c`.

That watchdog aborts an outstanding operation once more than `USB_OUTSTANDING_OP_TIMEOUT_MS`
(500 ms) pass between two fully processed host packets. The host polls a pending operation every
200 ms (`py/bitbox02/bitbox02/communication/bitbox_api_protocol.py`), so roughly 300 ms are left
for the complete round trip. Over USB that is ample. Over Bluetooth the round trip additionally
carries the connection interval in each direction plus the UART relay to the DA14531, so the
window is occasionally missed while the operation is perfectly healthy.

This explains every observed combination:

* With a `chainId` the device answers immediately, the retry loop is never entered, and the
  watchdog cannot fire.
* Without a `chainId` the warning holds the operation open across many poll cycles, and a single
  slow cycle is enough — which also matches the reported spread of 2 s to 10 s before failure.
* Over USB the remaining headroom is never consumed, so the same message signs.

Multi-frame framing is not involved: U2FHID reassembly completes before the task is spawned, and a
truncated request would surface as a frame error or a stall, not as a well-formed NACK.

## Change

Keep the 500 ms window for USB and use 3000 ms while Bluetooth is the active transport. 3000 ms is
fifteen poll intervals, so a healthy operation survives several consecutive slow round trips, while
a client that disappears still frees the stack quickly.

The timeout counter is shared between the HWW and the U2F stack, so the long window is limited to
the HWW context. U2F is always USB, because the Bluetooth transport only relays HWW frames.

`hww/transport.rs` carries the same constant and is scaled alongside it, so the behaviour does not
regress once it replaces the C implementation. A few existing comments named the fixed 500 ms
figure and are adjusted, since the window now depends on the transport.

## Tests

`test/unit-test/test_usb_processing.c` is new and covers the resulting matrix for both stacks: at
the threshold the operation survives, one tick above it is aborted, and U2F is aborted at the USB
threshold even while Bluetooth is active. It drives the public API only and observes the abort
through linker wraps on `hww_abort_outstanding_op` / `u2f_abort_outstanding_op`.

## What is not verified

This change is derived from the code, not from a measurement on a device, and we could not close
that gap ourselves:

* A retail device only runs firmware signed by BitBox, and a customer cannot replace the bootloader,
  so flashing this branch requires a device that ships with a development bootloader.
* The simulator cannot stand in for it. Its watchdog timer is compiled out (`_register_timer()` sits
  behind `#if !defined(BOOTLOADER) && !defined(TESTING)`), so the counter never advances, and it
  communicates over a TCP socket, so the Bluetooth latency that triggers the bug does not exist
  there.

What is verified is the mechanism: the unit test drives `_check_lock_timeout()` through
`usb_processing_process()` for both stacks and both sides of each threshold, and it fails if the
comparison is changed to `>=` or if the Bluetooth window is removed.

What remains open is whether a retry interval above 500 ms actually occurred in the reported
failures. `rawQueryV7` in bitbox02-api-go already logs the last ten retry intervals when it hits the
NACK; the reproduction in #2019 discards them by passing `&mocks.Logger{}`. With a real logger those
intervals would turn the argument above into a measurement, without needing to flash anything.

## Note on the simulator

`test/simulator-graphical` presents itself as a Nova and never sees a USB request, so
`ble_enabled()` stays true there and it consistently gets the Bluetooth window. That follows from
the documented meaning of the communication mode ("Bluetooth is active until the first USB request
is seen") rather than being a side effect of this change, so the simulator is left untouched.

fixes #2019

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ### [Unreleased]
 - Fixed a crash when listing many backups over Bluetooth
 - Add support for BitBoxSync
+- Bluetooth: fix requests being aborted while confirming on the device
 
 ### v9.26.4
 - Ethereum: allow ':' in EIP-712 type names, used by some dapps as a namespace separator
--- a/src/rust/bitbox-hal/src/system.rs
+++ b/src/rust/bitbox-hal/src/system.rs
@@ -14,8 +14,10 @@ pub trait System {
     /// This watchdog tracks the amount of time to wait before an outstanding
     /// operation times out (for example, if the client closes).
     /// Use this for long running operations that are expected to take longer
-    /// than about 300ms (a bit less than the 500ms timeout before a task is
-    /// cancelled).
+    /// than the timeout window, which depends on the active transport.
+    ///
+    /// `value` counts in 100ms units: `0` restarts the normal window, a
+    /// negative value extends it by `abs(value)` units.
     fn communication_timeout_reset(&mut self, value: i16);
 
     fn is_btconly(&mut self) -> bool;
--- a/src/rust/bitbox02-rust/src/async_usb.rs
+++ b/src/rust/bitbox02-rust/src/async_usb.rs
@@ -90,8 +90,8 @@ where
             *state = UsbTaskState::Running(Some(task), WaitingForNextRequestState::Idle);
         }
         // This panic could happen e.g. if someone reconnects to the BitBox while a task is running,
-        // before the 500ms timeout cancels the task. The proper way to handle would be to let the
-        // host know we are busy so the host can re-retry after some time.
+        // before the outstanding-operation timeout cancels the task. The proper way to handle
+        // would be to let the host know we are busy so the host can re-retry after some time.
         _ => panic!("spawn: wrong state"),
     }
 }
--- a/src/rust/bitbox02-rust/src/communication_mode.rs
+++ b/src/rust/bitbox02-rust/src/communication_mode.rs
@@ -35,15 +35,16 @@ fn has_ble(hal: &mut impl crate::hal::Hal) -> bool {
 }
 
 #[cfg(test)]
+pub(crate) fn reset_for_testing() {
+    USB_HWW_REQUEST_SEEN.write(false);
+    HAS_BLE.write(None);
+}
+
+#[cfg(test)]
 mod tests {
     use super::*;
 
     use crate::hal::testing::TestingHal;
-
-    fn reset_for_testing() {
-        USB_HWW_REQUEST_SEEN.write(false);
-        HAS_BLE.write(None);
-    }
 
     #[test]
     fn test_ble_disabled_on_non_plus() {
--- a/src/rust/bitbox02-rust/src/hww/transport.rs
+++ b/src/rust/bitbox02-rust/src/hww/transport.rs
@@ -16,7 +16,10 @@ const HWW_RSP_ACK: u8 = 0;
 const HWW_RSP_NOT_READY: u8 = 1;
 const HWW_RSP_BUSY: u8 = 2;
 const HWW_RSP_NACK: u8 = 3;
+// Bluetooth needs a larger window than USB, because its round trip eats into the time the host
+// has to poll.
 const USB_OUTSTANDING_OP_TIMEOUT_MS: u64 = 500;
+const USB_OUTSTANDING_OP_TIMEOUT_BLE_MS: u64 = 3000;
 
 pub type HwwTransport<H> = U2fHid<HwwVendorHandler<H>>;
 
@@ -80,12 +83,19 @@ impl<H> HwwVendorHandler<H> {
             deadline_ms: None,
         }
     }
+}
 
+impl<H: Hal> HwwVendorHandler<H> {
     fn refresh_timeout(&mut self, now_ms: u64) {
         self.deadline_ms = if crate::async_usb::is_idle() {
             None
         } else {
-            Some(now_ms.saturating_add(USB_OUTSTANDING_OP_TIMEOUT_MS))
+            let timeout_ms = if crate::communication_mode::ble_enabled(&mut self.hal) {
+                USB_OUTSTANDING_OP_TIMEOUT_BLE_MS
+            } else {
+                USB_OUTSTANDING_OP_TIMEOUT_MS
+            };
+            Some(now_ms.saturating_add(timeout_ms))
         };
     }
 }
@@ -207,6 +217,8 @@ mod tests {
     fn test_guard() -> MutexGuard<'static, ()> {
         let guard = TEST_LOCK.lock().unwrap();
         crate::async_usb::cancel();
+        // Reset communication-mode statics so BLE state cannot leak between tests.
+        crate::communication_mode::reset_for_testing();
         guard
     }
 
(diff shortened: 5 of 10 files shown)


benma:
Thanks. However, a confirm dialog being open is not normally a cause for a timeout/NACK response, we use the same confirm dialog in many workflows where the host app polls. 

I suspect the underlying issue may have already been fixed in `master` and has to do with the address derivation being blocked on securechip operations that take too long.

NickeZ (changes requested):
nACK, we can't have a 3 second timeout.

Danswar:
Ran the measurement proposed under "What is not verified" — real logger, no flashing.

**Setup.** BitBox02 Nova (`bb02p-multi`), main firmware v9.26.4, Bluetooth firmware 2, over BLE from an iOS host. `bitbox02-api-go` driven through `bitbox_flutter`, which passes `&mocks.Logger{}` at `go/api/api.go:109` and `:152` — the same discard as the repro in #2019. Replaced with a logger that appends to a file, since Go's stdout does not reach `devicectl --console` on iOS. Signed the chainId-less domain, i.e. the failing case.

**Result.** Seven consecutive NACKs, 56 retry intervals:

| run | intervals | max | gap preceding the NACK |
|---|---|---|---|
| 1 | 10 | 253.9 ms | 239.5 ms |
| 2 | 8 | 306.8 ms | 240.7 ms |
| 3 | 10 | 254.3 ms | 240.5 ms |
| 4 | 10 | 254.1 ms | 239.5 ms |
| 5 | 5 | 301.7 ms | 252.4 ms |
| 6 | 10 | 258.6 ms | 240.6 ms |
| 7 | 3 | 303.2 ms | 239.7 ms |

Overall range 238.8–306.8 ms. Nothing above 500 ms.

Sample from run 1:

```
unexpected NACK response; last successful retry query intervals (newest first) were:
239.525292ms; 240.647ms; 239.576208ms; 241.864917ms; 253.855833ms;
239.596ms; 240.595708ms; 239.251625ms; 240.435334ms; 239.5685ms;
```

Two things stand out. The gap preceding the NACK is the most stable value in the set (239.5–252.4 ms across all seven). And runs 5 and 7 failed after only 5 and 3 polls, about one second of dwell, with normal timing throughout.

**Caveat.** These are host-side query starts, so they differ from the device's fully-processed-packet clock by BLE latency each way — not by the ~250 ms needed to reach the threshold.

Happy to re-run with different instrumentation if something more specific would help.

NickeZ:
We believe the issue is with the secure chip communication which was blocking but has been made non blocking in master.

benma:
Thanks for everone's help. I am going to close the PR, as increasing the timeout is not the proper fix for the symptom, and because the issue is already fixed in master.

PR #2021, opened 2026-07-28
Limit recursive signing policy depth

benma-agent:
## Summary

- bound recursive EIP-712 struct and array traversal
- reject WSH Miniscript policies and Taproot leaves beyond the firmware encoding limit
- cover the depth boundaries and representative deep policy shapes with tests

### DIFF
--- a/src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
+++ b/src/rust/bitbox02-rust/src/hww/api/bitcoin/policies.rs
@@ -27,6 +27,24 @@ use sha2::{Digest, Sha256};
 // Arbitrary limit of keys that can be present in a policy.
 const MAX_KEYS: usize = 20;
 
+// Conservative rather than exact stack threshold. Miniscript encoding uses relatively small
+// synchronous recursion frames, so 64 preserves useful policy depth while bounding stack use.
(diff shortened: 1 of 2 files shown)


cedwies (approved):
ACK

Non-blockers:

1. Why was 16 and 64 chosen? Could we document why those limits are safe?
Currently it just says:
```rs
// Bound recursion to the available firmware stack.
const MAX_RECURSION_DEPTH: usize = 16;
```
In the future, someone might not be able to tell whether 16 was based on the actual stack measurement, chosen as a conservative estimate, ... ?

2. Did you avoid a full `process()` test on purpose? Couldn't this protect against future regressions e.g. where the validation call gets moved in a refactor?

benma:
> ACK
> 
> Non-blockers:
> 
>     1. Why was 16 and 64 chosen? Could we document why those limits are safe?
>        Currently it just says:
> 
> 
> ```rust
> // Bound recursion to the available firmware stack.
> const MAX_RECURSION_DEPTH: usize = 16;
> ```
> 
> In the future, someone might not be able to tell whether 16 was based on the actual stack measurement, chosen as a conservative estimate, ... ?
> 
>     2. Did you avoid a full `process()` test on purpose? Couldn't this protect against future regressions e.g. where the validation call gets moved in a refactor?

Thanks, good points. Addressed both.

PR #2022, opened 2026-07-28
Disable BitBoxSync API by default and prepare v9.26.5

benma-agent:
Gate the BitBoxSync API behind a disabled-by-default Cargo feature while API changes are pending.

Requests now return the standard disabled error unless the feature is explicitly enabled.


NickeZ (approved):
utACK

PR #2023, opened 2026-07-28
py: extract bootloader connection

benma:
Move the USB bootloader fallback into a nested helper to satisfy pylint's return-count limit without changing behavior.


NickeZ (approved):
utACK

PR #2025, opened 2026-07-31
factorysetup: validate RTT message length

benma-agent:
Reject RTT frames whose declared payload exceeds the bytes returned by the transport.

This prevents copying uninitialized stack data into factory commands when a frame is partial.


NickeZ (approved):
lgtm

PR #2026, opened 2026-07-31
rust: initialize C output buffers

benma-agent:
Rust slices require every byte in their backing storage to be initialized,
even when a function only intends to overwrite the bytes. Several C callers
passed uninitialized output buffers through BytesMut, and SHA/HMAC created
mutable slices directly from such pointers. This was formal UB and could
allow optimizer-dependent behavior.

In practice, these paths only wrote their outputs before C consumed them, and
current builds emitted the expected stores. No secret disclosure, corrupted
output, or control-flow impact was observed. This is hardening rather than a
fix for a demonstrated exploit.

Initialize C storage before constructing BytesMut. Keep output-only helpers on
raw pointers so they can legally initialize storage, and document the FFI
invariant. Raw-pointer APIs that already used direct stores remain valid.

### DIFF
--- a/src/bootloader/bootloader_format.c
+++ b/src/bootloader/bootloader_format.c
@@ -10,12 +10,14 @@
 void bootloader_format_pairing_code(char* out, size_t out_len, uint32_t pairing_code)
 {
     ASSERT(out_len >= sizeof("000000"));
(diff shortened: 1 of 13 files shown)


NickeZ (approved):
utACK

PR #2027, opened 2026-07-31
Fix GroundedCell result lifetime

benma-agent:
Read the confirmation result before clearing CONFIRM_STATE.
The result reference points into GroundedCell storage, so clearing the
state first invalidates it before result.is_ok() reads it.

This keeps the poll contract: publish the completed result, clear the task
state, and report completion. It also matches rust_workflow_unlock_poll().


NickeZ (approved):
utACK

PR #2030, opened 2026-08-01
noise: use multi-source entropy for static key

Silexperience210:
Hi! I've been reading through the firmware to learn how the entropy paths fit together — the multi-source design in `random_32_bytes()` is a nice piece of work. While doing that I ran into something I couldn't explain to myself, so I thought I'd ask here.

In `memory_reset_hww()`, two persistent secrets are generated a few lines apart:

```c
// Set salt root
_interface_functions->random_32_bytes(chunk.fields.salt_root);

// Set a new noise static private key.
rust_noise_generate_static_private_key(...);
```

The salt root goes through `random_32_bytes()` — MCU TRNG XOR securechip TRNG XOR factory randomness, hashed with SHA-256. The Noise static key goes through `bitbox_noise::genkey()`, which reads `random.mcu_32_bytes()` directly, so it only gets the MCU TRNG.

I genuinely can't tell from the code whether that's deliberate. My best guess at a reason is the sync/async boundary: `random_32_bytes()` is async because the securechip is, while `rust_noise_generate_static_private_key` is a synchronous `extern "C"` entry point. If that's the rationale, please just close this — a one-line comment in `x25519.rs` would save the next reader the same detour, and I'd be happy to send that instead.

In case it wasn't deliberate, here's a patch that lines the two up:

- adds `genkey_from_entropy()`, which only applies the x25519 clamping to caller-supplied entropy
- the static key call site feeds it `random_32_bytes()` output via `block_on`, the same pattern already used in `securechip/imp.rs`
- `genkey()` keeps its current behaviour and is still used for the per-session ephemeral keys in `noise_xx::init()`, where a cheap synchronous source is exactly what you want. It's now a thin wrapper, and `test_genkey_matches_genkey_from_entropy` asserts both paths agree, so the ephemeral path is provably untouched.

This keeps `bitbox-noise` free of async. Two things it does cost, which you may not want: it adds a `bitbox-core-utils` dependency to `bitbox02` (no cycle — `bitbox-core-utils` only depends on `bitbox-hal`), and it makes `memory_reset_hww()` depend on the securechip being available. Very happy to drop this if either is unwelcome.

To be clear about scope: this isn't a security report. The key here is the device↔app pairing identity, not seed-related, and the current code is only weaker if the MCU TRNG itself is degraded. It's a consistency/hardening tidy-up, nothing more.

**Testing**: `cargo test --manifest-path src/rust/Cargo.toml --all-features` passes (full suite, 0 failures). Clippy with the flags from `src/CMakeLists.txt` reports no new lints, and rustfmt is clean on the changed files. I couldn't get the CMake/ARM cross-build working on my machine, so `thumbv7em-none-eabi` will need a CI run — apologies for leaving that to you.

**Two things I'd like your call on**:
1. I used `panic!` when the securechip can't provide randomness, mirroring the C `random_32_bytes()` which aborts. Would a proper `Abort()` be more consistent here?
2. Is making `memory_reset_hww()` securechip-dependent acceptable, or would you rather keep that path free of it?

Thanks for maintaining this in the open — it's been a pleasure to read.

---

*Updated after review from @adam2k — thanks!*

**Why the securechip dependency isn't new**: `salt_root`, generated two lines above this call
site, already goes `random_32_bytes` -> `random_32_bytes_sec` (`src/random.c:49`) ->
`rust_securechip_random` -> `block_on(random())` (`bitbox02/src/securechip/imp.rs:157`) ->
`Abort()` on failure. So `memory_reset_hww()` already cannot complete without a working
securechip, and already nests a `block_on`. This patch adds neither a new dependency nor a new
failure mode.

**No compatibility risk**: the static key is generated only in `memory_reset_hww()` and then
persisted, so existing devices keep their pairing identity.

**C unit tests are unaffected**: `rust_noise_generate_static_private_key` is `--wrap`ped at
`test/unit-test/CMakeLists.txt:51`, so `_test_memory_reset_hww_ble`'s `random_32_bytes` call
counts never reach the new path.

### DIFF
--- a/src/rust/bitbox02/src/random.rs
+++ b/src/rust/bitbox02/src/random.rs
@@ -33,7 +33,7 @@ pub extern "C" fn rust_noise_generate_static_private_key(
         Ok(entropy) => entropy,
         // Mirrors the C `random_32_bytes()`, which aborts if the securechip cannot provide
         // randomness rather than silently falling back to fewer sources.
-        Err(_) => panic!("securechip randomness unavailable"),
+        Err(e) => panic!("securechip randomness unavailable: {e:?}"),
     };
     let key = bitbox_noise::genkey_from_entropy(&entropy);
     private_key_out.as_mut().copy_from_slice(&key[..]);
@@ -57,8 +57,10 @@ mod tests {
         assert!([0; 32] != result);
     }
 
+    // `genkey()` is the ephemeral-key path; the static key is built with
+    // `genkey_from_entropy()`, whose clamping is covered in `bitbox-noise`.
     #[test]
-    fn test_generate_static_private_key() {
+    fn test_genkey() {
         let mut random = crate::hal::random::BitBox02Random;
         let key = bitbox_noise::genkey(&mut random);
         assert_eq!(key[0] & 0b111, 0);
--- a/src/rust/bitbox02/src/random.rs
+++ b/src/rust/bitbox02/src/random.rs
@@ -23,7 +23,19 @@ pub extern "C" fn rust_noise_generate_static_private_key(
     mut private_key_out: util::bytes::BytesMut,
 ) {
     let mut random = crate::hal::random::BitBox02Random;
-    let key = bitbox_noise::genkey(&mut random);
+    let mut securechip = crate::hal::securechip::BitBox02SecureChip;
+    // Use the same multi-source entropy as other persistent secrets (e.g. the salt root):
+    // sha256(MCU TRNG XOR securechip TRNG XOR factory randomness).
+    let entropy = match util::bb02_async::block_on(bitbox_core_utils::random::random_32_bytes(
+        &mut random,
+        &mut securechip,
+    )) {
+        Ok(entropy) => entropy,
+        // Mirrors the C `random_32_bytes()`, which aborts if the securechip cannot provide
+        // randomness rather than silently falling back to fewer sources.
+        Err(_) => panic!("securechip randomness unavailable"),
+    };
+    let key = bitbox_noise::genkey_from_entropy(&entropy);
     private_key_out.as_mut().copy_from_slice(&key[..]);
 }
 
--- a/src/rust/Cargo.lock
+++ b/src/rust/Cargo.lock
@@ -271,6 +271,7 @@ dependencies = [
  "bip39",
  "bitbox-aes",
  "bitbox-bytequeue",
+ "bitbox-core-utils",
  "bitbox-framed-serial-link",
  "bitbox-hal",
  "bitbox-noise",
--- a/src/rust/bitbox-noise/src/lib.rs
+++ b/src/rust/bitbox-noise/src/lib.rs
@@ -16,6 +16,6 @@ pub mod testing;
 mod x25519;
 
 pub use noise_xx::{Error, HandshakeHash, HandshakeResult, State};
-pub use x25519::{X25519, genkey};
+pub use x25519::{X25519, genkey, genkey_from_entropy};
 
 pub use noise_rust_crypto::sensitive::Sensitive;
--- a/src/rust/bitbox-noise/src/noise_xx.rs
+++ b/src/rust/bitbox-noise/src/noise_xx.rs
@@ -80,8 +80,8 @@ impl State {
 
     /// Can be called at any time to start waiting for a new communication channel.
     ///
-    /// `static_private_key` is the local static key. It can be generated using
-    /// `genkey()` with a HAL random source and then persisted. `random` is used to generate a
+    /// `static_private_key` is the local static key. It is generated using
+    /// `genkey_from_entropy()` and then persisted. `random` is used to generate a
     /// fresh ephemeral private key for each session.
     pub fn init(&mut self, static_private_key: Sensitive<PrivateKey>, random: &mut impl Random) {
         let ephemeral_private_key = genkey(random);
--- a/src/rust/bitbox-noise/src/x25519.rs
+++ b/src/rust/bitbox-noise/src/x25519.rs
@@ -15,11 +15,14 @@ pub struct X25519;
 pub type PrivateKey = [u8; 32];
 pub type PublicKey = [u8; 32];
 
-/// Generate a fresh x25519 private key by reading 32 random bytes from the HAL and applying
-/// the standard clamping.
-pub fn genkey(random: &mut impl Random) -> Sensitive<PrivateKey> {
+/// Turn 32 bytes of entropy into an x25519 private key by applying the standard clamping.
+///
+/// The caller is responsible for the quality of `entropy`. For long-lived keys, prefer
+/// `bitbox_core_utils::random::random_32_bytes()`, which mixes the MCU TRNG, the secure chip
+/// TRNG and the factory randomness.
+pub fn genkey_from_entropy(entropy: &[u8; 32]) -> Sensitive<PrivateKey> {
     let mut k: Sensitive<PrivateKey> = Sensitive::new();
-    random.mcu_32_bytes(&mut k);
+    k.copy_from_slice(entropy);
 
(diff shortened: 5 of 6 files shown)


src/rust/bitbox02/src/random.rs
```diff
@@ -23,7 +23,19 @@ pub extern "C" fn rust_noise_generate_static_private_key(
...
+    // Use the same multi-source entropy as other persistent secrets (e.g. the salt root):
+    // sha256(MCU TRNG XOR securechip TRNG XOR factory randomness).
+    let entropy = match util::bb02_async::block_on(bitbox_core_utils::random::random_32_bytes(
+        &mut random,
+        &mut securechip,
+    )) {
+        Ok(entropy) => entropy,
+        // Mirrors the C `random_32_bytes()`, which aborts if the securechip cannot provide
+        // randomness rather than silently falling back to fewer sources.
+        Err(_) => panic!("securechip randomness unavailable"),
```
adam2k: Err(_) drops the diagnostic details. I would recommend something like `Err(e) => panic!("securechip randomness unavailable: {e:?}")`, instead.

src/rust/bitbox02/src/random.rs
adam2k, line 1: `test_generate_static_private_key` no longer matches what it tests: it calls `bitbox_noise::genkey`, which after this PR is the ephemeral path, not the static one. Rename to `test_genkey`, or drop it, since its clamping assertions are now covered by `test_genkey_from_entropy_clamps` in `x25519.rs`.

adam2k (commented):
Not a maintainer, and starting to get familiar by reading through the codebase.  I left two notes left inline.

**Worth a maintainer's call before you invest more:** `random_32_bytes` is already exposed via bitbox02-sys (`bitbox02-sys/build.rs:136`). Calling it directly would be roughly five lines with no new dependency, no `block_on`, no panic arm, and it removes the `Cargo.toml` change entirely. Your version is the C-to-Rust direction the repo is heading, but it would help to get the maintainers direction here.

**Your Q2 (is securechip dependence acceptable): it already is.** `salt_root`, two lines above yours, goes `random_32_bytes` -> `random_32_bytes_sec` (`src/random.c:49`) -> `rust_securechip_random` -> `block_on(random())` (`bitbox02/src/securechip/imp.rs:157`) -> `Abort()` on failure. `memory_reset_hww()` can't complete without a working securechip today, and already nests a `block_on`.  Your patch adds no new dependency and no new failure mode. This belongs in the description, it's the strongest argument for the change.

On your other two: the sync/async explanation is consistent with `securechip/imp.rs`, which carries four `block_on` shims for the same reason; and `panic!` over `Abort()` looks right, since the handler at `bitbox02-rust-c/src/lib.rs:120` prints to screen and RTT before halting.

**Additional notes:**

1. `noise_xx.rs:84` still says the static key "can be generated using `genkey()`...and then persisted". No longer true, should point at `genkey_from_entropy()`. (Not inline, the file isn't in the diff.)
2. Please get CI to confirm flash impact. `bitbox-core-utils` now reaches every target linking `bitbox02`, including factory-setup, and neither of us could run the ARM build.
3. #1961 interacts, though there's no file overlap: it adds a `Timer` generic to `securechip::random()` and a `Hal::random_and_securechip()` accessor for exactly the borrow pair your call site constructs by hand. Worth checking ordering.
4. Add to the description: no compatibility risk. The static key is generated only in `memory_reset_hww()` and then persisted, so existing devices keep their pairing identity.

**What I ran:** 

`cargo test -p bitbox-noise --all-features` on toolchain 1.96.0, 4/4 pass; `cargo fmt --check` clean. Same `bindgen was not found` wall you hit on `cargo check -p bitbox02`, so that half needs CI, though I checked the `&Box<Zeroizing<[u8; 32]>>` to `&[u8; 32]` coercion separately and it compiles. Also worth noting in the description: the C unit tests can't break, since `rust_noise_generate_static_private_key` is `--wrap`ped at `test/unit-test/CMakeLists.txt:51` and `_test_memory_reset_hww_ble`'s `random_32_bytes` call counts never see the new path.

Silexperience210:
Thank you for this — it's a far more careful read than I managed myself, and the pointers to specific lines saved me a lot of guessing.

**On `bitbox02-sys`: you're right, and I should have looked.** `random_32_bytes` is indeed already allowlisted (`bitbox02-sys/build.rs:136`), and I checked that the C function overwrites rather than XORs into `buf` (it ends in `rust_sha256(random, sizeof(random), buf)`), so the direct call would be safe without zero-initialising. That version really is five lines with no `Cargo.toml` change. I built the Rust-side version without first checking whether the C one was already reachable, which was the wrong order to do this in.

I've left the current approach up rather than swapping it pre-emptively, since as you say it's a maintainer call. @benma — would you prefer the minimal `bitbox02_sys::random_32_bytes` version? Happy to send that instead; it's a smaller diff and I have no attachment to mine.

**Your answer to my Q2 is the thing I was missing.** I hadn't traced `salt_root` all the way through `random_32_bytes_sec` (`src/random.c:49`) to the `block_on` in `securechip/imp.rs:157`. You're right that `memory_reset_hww()` already can't complete without a working securechip and already nests a `block_on`, so this adds no new dependency and no new failure mode. I've added that to the description, along with the no-compatibility-risk point and the `--wrap` note — thanks for both.

**One small thing back on `{e:?}`**, since it's not obvious: CI sets `RUSTFLAGS: "-Dwarnings"` (`ci-common.yml:18`) and clippy runs with `-W clippy::use_debug`, so I was worried Debug formatting would turn into a hard error. I tested it in isolation — `use_debug` fires on `println!("{e:?}")` but not on `panic!("{e:?}")`, so your suggestion is safe. Applied.

Also applied: the stale `genkey()` reference in `State::init()`'s doc comment, and `test_generate_static_private_key` renamed to `test_genkey` with a note that it now covers the ephemeral path. Pushed as a separate commit so the review history stays readable.

**On #1961** — good catch, I hadn't seen it. `Hal::random_and_securechip()` is exactly the borrow pair I'm constructing by hand, so if that lands first this call site should use it. I'm happy for this to wait on #1961 if that's the cleaner ordering.

**On flash impact**: agreed, and I can't measure it either — I got as far as configuring the CMake cross-build but couldn't get bindgen to find the newlib headers for `thumbv7em-none-eabi`, so I'm relying on CI for both the ARM build and the size delta. If it's meaningful, that's another argument for the `bitbox02-sys` version.

Thanks again for taking the time on a first-time contribution.

adam2k:
Hey Silexperience210,

Correcting something I got wrong in my last review, sorry.  `memory_reset_hww()` doesn't call the C `random_32_bytes()`. It calls `_interface_functions->random_32_bytes` (`memory.c:439`), and `common_main.c:34` sets that to `random_32_bytes_mcu`, with the comment right above it explaining why:

```c
// common_main.c:32-34
// Use random_32_bytes_mcu over random_32_bytes as the latter mixes in
// randomness from the securechip, which is initialized only later.
.random_32_bytes = random_32_bytes_mcu,
```

`memory_setup()` runs at `common_main.c:47`, `rust_securechip_init()` at `common_main.c:55`.

So `random_32_bytes_sec` is never on this path: `salt_root` is MCU-only too, the existing Noise static key is consistent with it, and this patch does add a new dependency and a new failure mode.

On a fresh device the new path reaches `backend()` at `securechip/imp.rs:19`, which unwraps `None` and panics before your `Err(e) => panic!` arm is ever hit.  CI won't catch it, since `rust_noise_generate_static_private_key` is `--wrap`ped in the C unit tests and no Rust test exercises the FFI entry point.

The `bitbox02_sys::random_32_bytes` variant has the same problem for the same reason: it routes through `random_32_bytes_sec` -> `rust_securechip_random` -> the same uninitialized `backend()`.  Smaller diff, same failure.

Your original instinct was the right one: the one-line comment in `x25519.rs` pointing at `common_main.c:32-34` is the valuable change here.

benma:
Hi. Thanks, but we don't see a need to change this. Please open an issue in the future instead of opening a PR. We will probably document a policy to this effect at some point.

PR #2032, opened 2026-08-03
BUILD: document host tools for native builds

adam2k:
The macOS setup section covers the brew dependencies but not the three tools installed via cargo in the Dockerfile.  Without them, building the simulators natively fails with CBINDGEN-NOTFOUND, "bindgen was not found", and "prost-build-proto: command not found".

Pin the same versions as the container, since the section already warns that version drift from CI causes reproducibility problems.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -104,6 +104,20 @@ brew tap osx-cross/arm
 brew install arm-gcc-bin
 ```
 
+Install [rustup](https://rustup.rs). The toolchain version is pinned by
+`src/rust/rust-toolchain.toml`; rustup will download it automatically once you run `cargo` from
+ within `src/rust` (or after setting a rustup override for the repository).
+
+The build also invokes three tools that are not installed by the brew commands
+above. Install them with the same versions as the container (see the
+[Dockerfile](Dockerfile)):
+
+```sh
+cargo install cbindgen --version 0.29.2 --locked
+cargo install bindgen-cli --version 0.72.1 --locked
+cargo install --path tools/prost-build-proto --locked
+```
+
 ## Contributor instructions
 
 ### Check out the repository


NickeZ (approved):
lgtm

BUILD.md
```diff
@@ -104,6 +104,20 @@ brew tap osx-cross/arm
...
+ within `src/rust` (or after setting a rustup override for the repository).
+
+The build also invokes three tools that are not installed by the brew commands
+above. Install them with the same versions as the container (see the
+[Dockerfile](Dockerfile)):
+
+```sh
+cargo install cbindgen --version 0.29.2 --locked
+cargo install bindgen-cli --version 0.72.1 --locked
+cargo install --path tools/prost-build-proto --locked
```
benma, line 118: @NickeZ I slightly dislike that we repeat versions here - updating these then means we need to update the docs here too. Any better way?

NickeZ: This is just documentation. This section already comes with a healthy amount of warning that using brew is error prone.

I prefer that you can copy/paste the invocations rather than some abstract instruction like "go look for the version in the dockerfile".

We could make `make bootstrap` also install rustup and these tools and get rid of the versions in the Dockerfile. I think long term it would be good to have a `make bootstrap` that completely bootstraps the dev environment. Very good for short lived AI boxes.

But, for now, I still see this documentation change as an improvement over the status quo.

benma (changes requested):
@adam2k please squash the two commits into one. Thanks

adam2k:
> @adam2k please squash the two commits into one. Thanks

@benma done.  

@NickeZ I generated a new GH issue off of your suggestion.  I can pick this up next if you find it worthwhile. https://github.com/BitBoxSwiss/bitbox02-firmware/issues/2037

NickeZ:
@adam2k All commits in this repo need to be signed with your GPG key. Please sign your commit.

If you are serious about cryptography you take the time to set it up using an offline machine. See for example: https://github.com/drduh/yubikey-guide

adam2k:
> @adam2k All commits in this repo need to be signed with your GPG key. Please sign your commit.
> 
> If you are serious about cryptography you take the time to set it up using an offline machine. See for example: https://github.com/drduh/yubikey-guide

@NickeZ thanks, I did the Yubikey setup and signed the commit.

adam2k:
@benma I've squashed and signed the commits.

benma (approved):

PR #2033, opened 2026-08-03
btc: reject oversized policy keys

benma-agent:
Reject policy keys whose rendered confirmation exceeds the UI label limit.
This prevents registration from approving a key that cannot be displayed in full.

Add boundary coverage for a 640-byte key and rejection at 641 bytes.


cedwies (approved):
utACK

PR #2034, opened 2026-08-04
driver_init: Abort early on TRNG init failure

benma-agent:
Check both MCU TRNG initialization steps and halt before any random data is consumed if either fails.

For firmware, persist the bootloader auto-enter flag through an early-safe path.

The path does not depend on initialized screen or USB state.

Stage0 and stage1 halt without linking firmware memory code.

### DIFF
--- a/src/platform/driver_init.c
+++ b/src/platform/driver_init.c
@@ -4,6 +4,9 @@
 
 #include "driver_init.h"
 #include "bitbox02_pins.h"
+#if !defined(BOOTLOADER) && !defined(BOOTLOADER_STAGE0)
+    #include "memory/memory.h"
+#endif
 #include "memory/memory_shared.h"
 #include "util.h"
 #include <compiler.h>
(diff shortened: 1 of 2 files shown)


src/platform/driver_init.c
```diff
@@ -256,14 +259,38 @@ static void _delay_driver_init(void)
...
  */
 static void _rand_init(void)
 {
     hri_mclk_set_APBCMASK_TRNG_bit(MCLK);
-    rand_sync_init(&RAND_0, TRNG);
-    rand_sync_enable(&RAND_0);
+    if (rand_sync_init(&RAND_0, TRNG) != ERR_NONE) {
+        _rand_init_abort();
+    }
+    if (rand_sync_enable(&RAND_0) != ERR_NONE) {
```
NickeZ, line 291: I don't think rand_sync_enable can return anything other than ERR_NONE.

src/platform/driver_init.c
```diff
@@ -256,14 +259,38 @@ static void _delay_driver_init(void)
...
+
 /**
  * Initialize hardware random number generator
  */
 static void _rand_init(void)
 {
     hri_mclk_set_APBCMASK_TRNG_bit(MCLK);
-    rand_sync_init(&RAND_0, TRNG);
-    rand_sync_enable(&RAND_0);
+    if (rand_sync_init(&RAND_0, TRNG) != ERR_NONE) {
```
NickeZ, line 288: _rand_init is called by the bootloader init. So you effectively brick bitboxes. Perhaps we want to output some error message?

NickeZ (commented):
nACK, let's have a call about it.

PR #2035, opened 2026-08-04
backup: validate decoded seed length

benma-agent:
Reject backup data whose declared seed length exceeds the fixed
32-byte seed field. This prevents malformed SD-card backups from
triggering an out-of-bounds panic while listing backups.

Add a regression test with an otherwise valid protobuf and recomputed
checksum.

Tests:
- `cargo test --manifest-path src/rust/Cargo.toml --all-features -- --test-threads 1`
- `./scripts/dev_exec.sh make run-unit-tests`


NickeZ (approved):
utACK

PR #2038, opened 2026-08-11
ui: display full transaction amounts

benma-agent:
Remove lossy ellipsis truncation from Ethereum and swap amounts.

Use rendered font widths to retain compact confirmation screens when
amounts fit, and fall back to separate scrollable review screens when
they do not. Preserve final fee long-touch confirmation without
requiring users to scroll first.

Validate external swap amount strings before presenting them in full
and add a manual long-amount Ethereum transaction fixture.


cedwies (approved):
utACK

cedwies (approved):
re-ACK

PR #2039, opened 2026-08-11
payment_request: reject noncanonical spacing

benma-agent:
Require coin purchase amounts to use exactly one ASCII space
between the numeric value and unit, with no leading or trailing
whitespace.

This prevents signed destination strings containing newlines or
other whitespace from placing the visible amount outside the
display.


cedwies (approved):
utACK

PR #2040, opened 2026-08-11
bb03: bound and scroll confirmation bodies

benma-agent:
Keep confirmation navigation controls fixed while long bodies scroll
independently.

Use the shared 640-byte confirmation-body limit on both devices and
truncate BB03 bodies at a UTF-8 boundary with an ellipsis. This keeps
the truncation warning aligned with the value actually shown.

### DIFF
--- a/src/rust/bitbox03/src/ui/confirm.rs
+++ b/src/rust/bitbox03/src/ui/confirm.rs
@@ -1,13 +1,28 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use bitbox_hal::ui::{ConfirmParams, UserAbort};
+use alloc::string::String;
+use bitbox_hal::ui::{ConfirmParams, MAX_CONFIRM_BODY_SIZE, UserAbort};
 use bitbox_lvgl::{
-    self as lvgl, LabelExt, LvLabel, LvLabelLongMode, LvObj, LvOpacityLevel, ObjExt,
+    self as lvgl, LabelExt, LvLabel, LvLabelLongMode, LvObj, LvObjFlag, LvOpacityLevel, ObjExt,
 };
 use util::futures::completion::Responder;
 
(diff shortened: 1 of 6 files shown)


src/rust/bitbox03/src/ui/confirm.rs
```diff
@@ -1,14 +1,29 @@
...
 use bitbox_lvgl::{
-    self as lvgl, LabelExt, LvLabel, LvLabelLongMode, LvObj, LvOpacityLevel, ObjExt,
+    self as lvgl, LabelExt, LvLabel, LvLabelLongMode, LvObj, LvObjFlag, LvOpacityLevel, ObjExt,
 };
 use util::futures::completion::Responder;
 
 use super::nav_button::{NavIcon, build_close_button, build_nav_button};
 use super::slide_to_confirm::build_slide_to_confirm;
 
+fn truncate_body(body: &str) -> String {
```
cedwies, line 13: Nit: This always creates an extra heap copy of every confirmation body.

Codex suggests using copy on write:

```rs
use alloc::borrow::Cow;

fn truncate_body(body: &str) -> Cow<'_, str> {
    if body.len() <= MAX_CONFIRM_BODY_SIZE {
        return Cow::Borrowed(body);
    }

    let mut end = MAX_CONFIRM_BODY_SIZE;
    while !body.is_char_boundary(end) {
        end -= 1;
    }

    let mut truncated = String::from(&body[..end]);
    truncated.push_str("...");
    Cow::Owned(truncated)
}
```

cedwies (approved):
utACK

src/rust/bitbox03/src/ui/confirm.rs
benma: That was actually the first version, but I told it to not prematurely optimize it and stick to clear code instead. Maybe it was the wrong call :thinking:

PR #2041, opened 2026-08-12
docs: clarify signing fallback behavior

benma-agent:
Document the historical zero-contribution S2C fallback used by legacy
Bitcoin and Ethereum signing requests when anti-klepto is omitted.

Clarify that EIP-712 instead falls back to plain RFC6979 and that neither
fallback provides anti-klepto protection.

This is and aid to LLMs to not spend time analyzing this in reviews.

The alternative would be to fallback to RFC6979, but it's not worth the effort, considering that virtually all clients use antiklepto - this change could not retroactively change legacy apps.


cedwies (approved):
utACK

PR #2042, opened 2026-08-12
btc: warn on high all-change fees

benma-agent:
Use the verified input total as the fee-warning denominator when a
transaction has no external output value. This covers all-change and
OP_RETURN-only transactions without introducing an absolute fee policy.

Label the denominator explicitly and add regression coverage for both
high-fee and below-threshold cases.

### DIFF
--- a/src/rust/bitbox02-rust/src/workflow/transaction.rs
+++ b/src/rust/bitbox02-rust/src/workflow/transaction.rs
@@ -11,11 +11,49 @@ fn format_percentage(p: f64) -> String {
     util::decimal::format_no_trim(int, 1)
 }
 
+/// The denominator used to calculate and describe a transaction's fee percentage.
+pub enum FeePercentageBasis {
+    /// The amount sent to non-change/send outputs.
+    SendAmount,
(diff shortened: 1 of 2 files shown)


src/rust/bitbox02-rust/src/workflow/transaction.rs
```diff
@@ -11,25 +11,62 @@ fn format_percentage(p: f64) -> String {
     util::decimal::format_no_trim(int, 1)
 }
 
+/// The denominator used to calculate and describe a transaction's fee percentage.
+pub enum FeePercentageBasis {
+    /// The amount sent to external outputs.
```
cedwies: Nit: A self-transfer recognized as belonging to the same wallet can still be non-change, right? In this case `Non-change/send outputs` would be more accurate?

cedwies (approved):
utACK

src/rust/bitbox02-rust/src/workflow/transaction.rs
benma: Good catch, fixed

PR #2043, opened 2026-08-12
eth: require 20-byte EIP-712 addresses

benma-agent:
EIP-712 encodes addresses as uint160 values. The typed-message parser
previously passed any decoded value up to 32 bytes to leftpad32, which
allowed signing member words that cannot represent an Ethereum address.

Require the host-provided address to decode to exactly 20 bytes and add
tests for the 19-, 20-, and 21-byte boundaries.


Tomasvrba (approved):
ack, lgtm

PR #2044, opened 2026-08-12
keystore: zeroize derived encryption keys

benma-agent:
The final retained-seed encryption key was first stored in a plain stack
array before being copied into a Zeroizing vector. Its SHA-512 expansion
into the AES and HMAC keys was likewise stored in a plain array.

Wrap both derived-key arrays in Zeroizing so their storage is wiped on
every return path. Retained ciphertext and the unstretched KDF input are
intentionally unchanged because they cannot decrypt the seed without the
secure-chip KDF.


cedwies (approved):
utACK

PR #2047, opened 2026-08-18
btc: allow m/45' message signing

benma-agent:
## Summary

- allow Bitcoin message signing with keys in the m/45' application namespace
- keep the existing UI for standard keypaths and identify m/45' requests as external service keys
- add a send_message.py demo using the external-service keypath
- document that Casa, alongside Unchained, uses or exports m/45' keys


cedwies (approved):
utACK

PR #2048, opened 2026-08-19
btc: support timestamp locktimes

benma-agent:
Accept timestamp-based nLockTime values and display the exact signed value in UTC.

Keep confirmation for effective locktimes because they control when a transaction becomes final.

Remove RBF status because replacement is mempool policy, not signed consensus state.

Bitcoin Core's full-RBF policy also makes sequence-derived opt-in status misleading.

Litecoin did not show replacement status, so the per-coin RBF flag is no longer needed.

Document client support and add a BIP388 after(timestamp) vector that finalizes and validates.


cedwies (approved):
utACK

PR #2049, opened 2026-08-20
bb03: pin/passphrase entry and UI improvements

jadzeidan:
Adds the following to bb03 functionality and UI:

- PIN keypad entry for unlock
- Keyboard with bb02 supported special characters for passphrase
- Ability to show passphrase entry in simulator using `--passphrase`
- misc UI improvements


NickeZ (changes requested):
1. "lock" should be the default and there could be an argument `--unlock` to launch it unlocked.
2. Please wrap all secrets, like pins and passphrases, in Zeroize.

NickeZ (approved):
tACK

PR #2051, opened 2026-08-20
Document Python 3.10 requirement

NickeZ:
Since the introduction of https://pypi.org/project/pyhpke/ we depend on python 3.10.

State the minimum Python version in the build dependencies and Python script setup documentation.


cedwies (approved):
utACK

PR #2052, opened 2026-08-20
bb03: fix confirmation body color

NickeZ:
BB03 screens use black backgrounds, but LVGL's default light theme assigns dark text to nested containers. Enable the default dark theme so inherited text remains visible across the BB03 UI, including the pairing code.


jadzeidan (approved):
tACK, LGTM

PR #2053, opened 2026-08-20
Show ambiguous ERC-20 contract identities

benma-agent:
Show the full contract address for unknown ERC-20 targets and known
tokens whose symbols collide in the firmware registry. Keep the
existing shorter confirmation for known tokens with unique symbols.

Generate the ambiguous-symbol list from the registry at build time.
Ignore decimal differences: changing the raw amount can preserve the
same displayed amount for different tokens.

Cover known collisions with equal and different decimals, unknown
targets, and unchanged unique-token confirmations for legacy and
EIP-1559 transactions, including the existing payment-request case.


Tomasvrba (approved):
ack, lgtm

PR #2055, opened 2026-08-21
Show EIP-712 primary type

benma-agent:
Add a Message type confirmation before traversing typed data so the device
transcript binds the host-selected primary type.

Cover same-shaped requests with distinct primary types, update the end-to-end
signing screen expectation, and record the change in the firmware changelog.


Tomasvrba (approved):
nice improvement, lgtm

PR #2056, opened 2026-08-22
Warn before truncating EIP-712 array headers

benma-agent:
Route EIP-712 array-header confirmations through the shared
truncation-aware prompt so overlong member paths first show the warning.

Add regression coverage for an overlong valid member identifier.


Tomasvrba (approved):
The change itself looks good to me, is this this just a proactive improvement or did we have a real-world case of member names being too long? I've never come across that, but it can happen of course.

PR #2057, opened 2026-08-22
Make payment-request memo names scrollable

benma-agent:
Show the memo attribution and recipient on one scrollable line so long
authenticated recipient names remain reviewable.

Bump the firmware version to v9.27.2 and gate the updated test vectors
at that version, preserving historical transcripts. Regenerate the
canonical JSON and document the compatibility rule. Record scrollability
in the test UI and cover long SWAPKIT names.

### DIFF
--- a/src/rust/bitbox-test-vectors/src/btc_transaction/cases/screens.rs
+++ b/src/rust/bitbox-test-vectors/src/btc_transaction/cases/screens.rs
@@ -798,7 +798,12 @@ pub fn payment_request_owned_output() -> Vec<VersionExpectation> {
         success(
             Some("9.24.0"),
             Some("9.26.3"),
-            payment_request_screens("0.30000000 TBTC", "0.10000000 TBTC", 50),
+            payment_request_screens(
+                "0.30000000 TBTC",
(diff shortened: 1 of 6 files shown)


src/rust/bitbox-test-vectors/src/btc_transaction/cases/screens.rs
```diff
@@ -810,8 +816,23 @@ pub fn payment_request() -> Vec<VersionExpectation> {
         invalid_input_before("9.24.0"),
         success(
             Some("9.24.0"),
+            Some("9.27.0"),
```
cedwies: This needs to be `9.27.2`

cedwies (changes requested):

src/rust/bitbox-test-vectors/src/btc_transaction/cases/screens.rs
cedwies: There are two occurences of the wrong version in `screens.rs` the file.

cedwies (approved):
tACK

PR #2058, opened 2026-08-23
py: require antiklepto for signing

benma-agent:
Require firmware 9.4.0 or newer unconditionally for BTC transaction signing.
Reject BTC and ETH message signing and ETH transaction signing when firmware
predates each workflow's antiklepto support.

Keep Ed25519 workflows and the explicit EIP-712 opt-out unchanged, and bump
py-bitbox02 to version 8.0.0 for the breaking change.


bznein (approved):
utACK

PR #2059, opened 2026-08-23
py: validate antiklepto signatures

benma-agent:
Validate compact ECDSA signature scalars and require low-S encodings
before accepting firmware signing responses.

Validate recovery IDs for recoverable signatures and preserve validation
for the explicit EIP-712 anti-klepto opt-out.

Add regression tests for malformed signatures, high-S encodings, and
recovery IDs, and run them in CI.

### DIFF
--- a/py/bitbox02/bitbox02/bitbox02/bitbox02.py
+++ b/py/bitbox02/bitbox02/bitbox02/bitbox02.py
@@ -18,7 +18,12 @@ from bitbox02.communication import (
     ERR_DUPLICATE_ENTRY,
 )
 
-from .secp256k1 import antiklepto_host_commit, antiklepto_verify
+from .secp256k1 import (
+    antiklepto_host_commit,
+    antiklepto_verify,
+    antiklepto_verify_recoverable,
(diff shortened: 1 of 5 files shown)


py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -660,7 +664,8 @@ def btc_sign_msg(
         """
         # pylint: disable=no-member
 
-        self._require_atleast(semver.VersionInfo(9, 2, 0))
+        # Anti-klepto support for BTC message signing was added in v9.5.0.
+        self._require_atleast(semver.VersionInfo(9, 5, 0))
```
NickeZ, line 677: does this break `send_message.py` for versions < 9.5.0. In that case I think we need to add a compatibility flag that `send_message.py` can use.

benma: Was addressed in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2058

cedwies (approved):
utACK

PR #2060, opened 2026-08-24
send_message: add absolute locktime demos

benma-agent:
Split the existing locktime/RBF demo into height-based and time-based
absolute locktime entries. Use a non-final sequence so both locktimes take effect
without framing the demo around RBF.


NickeZ (approved):
tACK

PR #2064, opened 2026-08-24
Add bitbox-mcu-stm32u5 and vendor stm32u5

NickeZ:
extracted out from bb03 binaries.

This only adds a single rust crate with a single line of code to pick the MCU for bb03. The rest (2 million lines) is vendoring of a rust crate.


cedwies (approved):
utACK

PR #2065, opened 2026-08-24
Nickez/bb02 utf8 safe

NickeZ:
Doing truncation utf8 safe.

I think this is a prerequisite for utf8 fonts.

depends on https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2083


benma (approved):
utACK

Astra found one unimportant edge case about payment request recipient names being able to cause a crash instead of InvalidInput after your new assertion. Fixed in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2085

PR #2069, opened 2026-08-25
btc: allow m/48' message signing

benma-agent:
Accept Bitcoin message-signing keypaths under m/48' as external service keys, matching the
existing m/45' flow. Cover the new namespace in validation, signing, and developer tooling.

Bump firmware to v9.27.1 and move accumulated changelog entries into that release.

Tests:
- cargo test --manifest-path src/rust/Cargo.toml --all-features -- --test-threads 1
- ./scripts/dev_exec.sh make -j16 unit-test
- ./scripts/dev_exec.sh make run-unit-tests
- ./scripts/dev_exec.sh make run-rust-clippy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,10 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ## Firmware
 
 ### [Unreleased]
+
+### v9.27.1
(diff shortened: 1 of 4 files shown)


cedwies (approved):
utACK

PR #2070, opened 2026-08-25
Expose bootloader version in device info

NickeZ:
Read and validate the installed stage1 header, then expose its marketing version through the HAL and device info protobuf response.

Keep the field absent for legacy bootloaders and return None from the Python client when unavailable.


py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -162,9 +162,14 @@ def device_info(self) -> Dict[str, Any]:
...
         request.device_info.CopyFrom(device_info_request)
         response = self._msg_query(request, expected_response="device_info")
-        result = {
+        result: Dict[str, Any] = {
             "name": response.device_info.name,
             "version": response.device_info.version,
+            "bootloader_version": (
+                response.device_info.bootloader_version
+                if response.device_info.HasField("bootloader_version")
+                else None
```
benma, line 171: securechip_model and password_stretching_algo below are not added to the object if not available, while bluetooth is added as None. I wonder which is the better way 🤷‍♂️

py/bitbox02/bitbox02/bitbox02/bitbox02.py
```diff
@@ -162,9 +162,14 @@ def device_info(self) -> Dict[str, Any]:
         device_info_request = bitbox02_system.DeviceInfoRequest()
         request.device_info.CopyFrom(device_info_request)
         response = self._msg_query(request, expected_response="device_info")
-        result = {
+        result: Dict[str, Any] = {
             "name": response.device_info.name,
             "version": response.device_info.version,
+            "bootloader_version": (
```
benma, line 168: The Python lib should get a changelog entry in the unreleased section. 

https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2058/changes#diff-b32c19f14fc56266b2f0666efa143bcdf9cde82b298242a601c8cc14215a35ef bumps the version, no need to repeat it here

src/memory/memory.c
```diff
@@ -12,8 +12,13 @@
 #include "memory_shared.h"
 #include "random.h"
 #include "util.h"
+#include <platform/platform_config.h>
 #include <rust/rust.h>
 
+#if FACTORYSETUP == 0
```
benma: why gate on FACTORYSETUP? Should also not be compiled for stage0/stage1, right?

src/memory/memory.c
```diff
@@ -774,6 +792,73 @@ void memory_bootloader_hash(uint8_t* hash_out)
...
+        header->image_len <= header_len || header->image_len > BB02_STAGE1_MAX_LEN ||
+        version_len == 0 || version_len > BB02_STAGE1_HEADER_STAGE1_MARKETING_VERSION_MAX_LEN ||
+        version_len > version_out_len) {
+        return false;
+    }
+    for (size_t i = 0; i < version_len; i++) {
+        const uint8_t ch = header->stage1_marketing_version[i];
+        if (ch < 0x21 || ch > 0x7e) {
+            return false;
+        }
```
benma: Really needed here?

src/memory/memory.c
```diff
@@ -774,6 +792,73 @@ void memory_bootloader_hash(uint8_t* hash_out)
...
+
+    const uint32_t header_len = header->header_len;
+    const uint8_t version_len = header->stage1_marketing_version_len;
+    if (header->magic != BB02_STAGE1_HEADER_MAGIC || header_len < BB02_STAGE1_HEADER_LEN ||
+        header_len > BB02_STAGE1_MAX_LEN || (header_len % BB02_STAGE1_HEADER_ALIGNMENT) != 0 ||
+        header->image_len <= header_len || header->image_len > BB02_STAGE1_MAX_LEN ||
+        version_len == 0 || version_len > BB02_STAGE1_HEADER_STAGE1_MARKETING_VERSION_MAX_LEN ||
+        version_len > version_out_len) {
+        return false;
+    }
```
benma: Looks like copypaste from stage0 code, could be reused? Also isn't it enough to verify only the magic here?

py/bitbox02/bitbox02/bitbox02/bitbox02.py
NickeZ: I prefer setting it to None otherwise consuming code will have to deal with figuring out if the attribute is set or not which involves more gnarly code.

NickeZ: Actually, I have a pretty strong opinion that changing the shape of the type is much worse. All fields should be set, "undefined" values to None.

src/memory/memory.c
NickeZ: the reason is was gated was because the bootloader_upgrade.h header couldn't be included otherwise. Fixed

py/bitbox02/bitbox02/bitbox02/bitbox02.py
NickeZ: Fixed

src/memory/memory.c
NickeZ: no, fixed

src/memory/memory.c
NickeZ: simplified, fixed

src/bootloader/bootloader_product.h
```diff
@@ -10,12 +10,14 @@
 #define BB02_STAGE1_PRODUCT_ID_BITBOX_BTCONLY (2u)
 #define BB02_STAGE1_PRODUCT_ID_BITBOX_PLUS_MULTI (3u)
 #define BB02_STAGE1_PRODUCT_ID_BITBOX_PLUS_BTCONLY (4u)
+#define BB02_STAGE1_PRODUCT_ID_BITBOX02_FACTORYSETUP (5u)
```
benma: this looks quite wrong? there is no stage1 for factorysetup

codex suggestion:


     Better: make bootloader_upgrade.h product-neutral by removing its bootloader_product.h include, then explicitly include bootloader_product.h only in implementation files that actually use
     BB02_STAGE1_PRODUCT_ID. Alternatively, split the stage1 header format into a product-neutral header. Then the factory ID addition disappears.

py/bitbox02/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- `get_info()`: add the installed bootloader version to the returned device info
```
benma: typo, `get_device_info()`

src/memory/memory.c
```diff
@@ -774,6 +787,36 @@ void memory_bootloader_hash(uint8_t* hash_out)
 #endif
 }
 
+bool memory_get_bootloader_version(
+    uint8_t* version_out,
+    size_t version_out_len,
```
benma: could also check it is not larger than `MEMORY_BOOTLOADER_VERSION_MAX_LEN`, to avoid reading too much into the out buffer if it is oversized. or drop the arg completely and just document the out buffer has to be at least `MEMORY_BOOTLOADER_VERSION_MAX_LEN` and then copy exactly that amount?

a nit, because the caller uses `MEMORY_BOOTLOADER_VERSION_MAX_LEN` for the buffer size.

benma (changes requested):

src/bootloader/bootloader_product.h
NickeZ: done

py/bitbox02/CHANGELOG.md
NickeZ: done

src/memory/memory.c
NickeZ: done

benma (approved):
utACK 🙏  Please move the changelog entry to 9.27.1 when rebasing,

PR #2071, opened 2026-08-26
ci: Test individual PR commits

NickeZ:
Build each commit directly instead of creating a synthetic merge with the PR base. Include the head commit and label matrix jobs with their position in the PR history.


cedwies (approved):
utACK

PR #2072, opened 2026-08-26
Validate ATECC response lengths

benma-agent:
Preserve the receive buffer capacity and reject malformed ATECC
response counts before CryptoAuthLib can use them for CRC validation.

Match the status codes used by the standard CryptoAuthLib I2C HAL and add
regression tests for undersized and oversized response counts.

Tests:
- ./scripts/dev_exec.sh make firmware
- ./scripts/dev_exec.sh make unit-test
- ./scripts/dev_exec.sh make run-unit-tests


NickeZ (approved):
utACK

PR #2074, opened 2026-08-27
backup: reject malformed fields

benma-agent:
Treat missing optional metadata as invalid instead of unwrapping it,
and only accept the seed lengths supported by the keystore. This keeps
malformed backup files on the normal error path.

Add regression coverage for missing metadata and invalid seed lengths,
and correct the legacy create/load test's 12-byte fixture to the intended
16-byte size.

Tests:
- `cargo test --manifest-path src/rust/Cargo.toml --all-features -- --test-threads 1`
- `make unit-test`
- `make run-unit-tests`
- `make run-rust-clippy`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ customers cannot upgrade their bootloader, its changes are recorded separately.
 ## Firmware
 
 ### [Unreleased]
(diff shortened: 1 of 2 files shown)


cedwies (approved):
tACK

PR #2075, opened 2026-08-27
py: improve firmware loader UX

benma-agent:
Detect development bootloaders and automatically classify signed and unsigned firmware inputs.

Confirm the detected combination by default, with -y/--yes for non-interactive use.

Keep --debug as a deprecated no-op for backwards compatibility.

Warn about combinations expected to fail without blocking deliberate bootloader error tests.

Treat signature-data errors as nonfatal on development devices and fatal on production devices.

Improve bootloader errors and timeout handling, and document and test the flashing matrix.

### DIFF
--- a/py/load_firmware.py
+++ b/py/load_firmware.py
@@ -1,22 +1,94 @@
 #!/usr/bin/env python3
 # SPDX-License-Identifier: Apache-2.0
 
-"""TODO: document"""
+"""Flash signed or unsigned firmware onto a BitBox device.
+
+The input type is detected from the signed-firmware header. The supported combinations are:
+
+* Production device, signed firmware: flash the firmware and signature data.
+* Production device, unsigned firmware: warn that it will not boot, then flash the firmware.
+* Development device, signed firmware: flash the firmware, attempt to flash the signature data,
+  warn if the signature data is rejected, and reboot into the firmware anyway.
+* Development device, unsigned firmware: flash the firmware without signature data.
+
+Every flash requires confirmation unless ``-y``/``--yes`` is used. Edition mismatches and
+combinations that are expected to fail are called out before the confirmation.
+"""
 
 import argparse
-import sys
-
 import pprint
-from typing import Callable, Any, Tuple
-from time import sleep
+import sys
+import time
+from pathlib import Path
+from typing import Any, Callable, Dict, NamedTuple, Optional, Sequence, Tuple
 
 import hid
 
 from bitbox02.communication import devices, TransportLayer, u2fhid, bitbox_api_protocol
 from bitbox02.communication.devices import TooManyFoundException, NoneFoundException
 
-from bitbox02.bitbox02 import Bootloader, BitBox02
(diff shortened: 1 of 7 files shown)


py/load_firmware.py
```diff
@@ -1,24 +1,100 @@
...
 
-"""TODO: document"""
+"""Flash signed or unsigned firmware onto a BitBox device.
 
-import argparse
-import sys
+The supported combinations are:
+
+* Production device, signed firmware: flash the firmware and signature data.
+* Production device, unsigned firmware: refuse to flash.
```
NickeZ: why refuse to flash? I think being able to flash unsigned firmware is a valid test engineers can make.

py/load_firmware.py
```diff
@@ -1,24 +1,100 @@
...
+* Development device, signed firmware: flash the firmware, attempt to flash the signature data,
+  warn if the signature data is rejected, and reboot into the firmware anyway.
+* Development device, unsigned firmware with ``--debug``: flash the firmware without signature
```
NickeZ: Can we change the name of `--debug` to `--unsigned`?

NickeZ: btw, do we even need the `--unsigned` flag or could the tool check if the image contains signatures?

py/load_firmware.py
benma: Since fw images don't have an image header yet, this operation is not even possible basically. Unsigned now means not that the header is there and the sigs are missing, but that there is no header at all. We could still allow it and skip flashing the sigs, but it would not test failing sigs during flashing.

You can always use `--debug`/`--unsigned` if you want to flash invalid firmware to test that it wont' boot afterwards.

benma: Nevermind, next commit will allow it that drops the `--debug` flag, as it will proceed with any combo if accepted.

NickeZ: what do you mean not possible? I always thought you could write what you wanted through the write interface. Isn't writing the sigs a separate step from flashing?

py/load_firmware.py
```diff
@@ -106,46 +228,125 @@ def _find_and_open_usb_bitbox02(use_cache: bool) -> Tuple[devices.DeviceInfo, Tr
...
     parser = argparse.ArgumentParser(
-        description="Tool for flashing a new firmware on BitBox devices."
+        description="Tool for flashing a new firmware on BitBox devices.",
+        epilog=FLASHING_HELP,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
     )
     parser.add_argument(
         "--no-cache", action="store_true", help="Don't use cached or store noise keys"
     )
-    parser.add_argument("--debug", action="store_true", help="Flash a debug (unsigned) firmware.")
```
NickeZ, line 117: I think keeping `--debug` as a noop / deprecation notice would be more polite

benma: Re-added

NickeZ (approved):
tACK, nice!

PR #2080, opened 2026-09-03
factorysetup: make RTT framing robust

benma-agent:
Read the header and payload exactly so partial RTT reads cannot truncate API messages.

Send each response as one frame. Retry the complete buffer until NoBlockSkip publishes it.

Partial byte counts from NoBlockSkip are not committed and cannot be accumulated.


NickeZ (approved):
utACK, fixed all my nits

PR #2081, opened 2026-09-03
factorysetup: speed up certificate setup

benma-agent:
Select the attestation root key by its identifier before verifying the
certificate signature. This decreases the time to set the certificate to
less than 0.1 seconds. Previously it took up to roughly 2.4 seconds for a
root key near the bottom of the list.


NickeZ (approved):
utACK

PR #2082, opened 2026-09-03
reduce factorysetup image size

benma-agent:
Use libsecp256k1's static context for certificate verification while preserving its startup self-test, reuse RustCrypto SHA-256 for HMAC-SHA256, and store the root attestation keys as const compressed SEC1 points.

The unused-peripheral removal and screen-font changes have moved to #2092. Both PRs target `master` directly; this PR contains only the remaining three crypto-related commits. The figures below are measured without #2092.

Root identifiers remain SHA-256 of the uncompressed SEC1 encoding, so the existing host identifiers and single-signature certificate verification from #2081 are preserved.

| Factory setup | Before this PR | After | Saved |
|---|---:|---:|---:|
| Image | 195,296 bytes | 152,472 bytes | 42,824 bytes (21.9%) |
| RAM | 203,304 bytes | 196,152 bytes | 7,152 bytes (3.5%) |

Factory-setup sizes were remeasured on the current base and after each commit:

| Change | Image after | Image saved by this commit |
|---|---:|---:|
| Static secp256k1 context | 159,568 bytes | 35,728 bytes |
| Reuse SHA-256 for HMAC | 155,088 bytes | 4,480 bytes |
| Const compressed root keys | 152,472 bytes | 2,616 bytes |

The root-key change alone reduces RAM usage from 203,304 to 196,152 bytes, saving 7,152 bytes. All 110 keys retain their original values, indices, and identifiers; the comment above the table includes a Python conversion snippet using `ecdsa`. This change is confined to factory setup.

The HMAC change leaves RAM usage unchanged. Normal firmware still needs `bitcoin_hashes` for other callers, so its size increases slightly. These figures were also remeasured on the current base:

| Firmware | Before HMAC change | After | Change |
|---|---:|---:|---:|
| Multi | 686,080 bytes | 686,356 bytes | +276 bytes |
| Bitcoin-only | 525,392 bytes | 525,656 bytes | +264 bytes |

The three code patches are unchanged by the split. Factory setup builds successfully after every commit, and both normal firmware variants were built before and after the HMAC change. All 21 C test suites (`make run-unit-tests`) and 627 Rust tests (`cargo test --manifest-path src/rust/Cargo.toml --all-features -- --test-threads 1`) pass on the final branch. The tests cover identifier equivalence for compressed and uncompressed public keys, invalid keys and output lengths, and HMAC empty/long-key and overlapping-buffer cases. Hardware validation has not been performed.


NickeZ (changes requested):
See previous comment

NickeZ (approved):
utACK

PR #2083, opened 2026-09-03
backup: sanitize legacy device names

NickeZ:
Legacy backups may contain device names with characters outside the printable-ASCII set supported by the BitBox02 UI and device-name storage. Passing such names through can omit characters or abort backup checks and restores as those boundaries are hardened.

Replace unsupported characters when consuming a backup so it remains usable and the restored device gets a valid name. Preserve the original metadata for checksum validation.

Add regression coverage for Unicode and control characters.

### DIFF
--- a/src/rust/bitbox02-rust/src/backup.rs
+++ b/src/rust/bitbox02-rust/src/backup.rs
@@ -70,6 +70,19 @@ pub fn id(seed: &[u8]) -> String {
     hex::encode(mac.finalize_reset().into_bytes())
 }
 
+/// Replaces non-printable-ASCII characters in legacy backup names with `?`.
+pub(crate) fn sanitize_name(name: &str) -> String {
+    name.chars()
+        .map(|character| {
+            if util::ascii::Charset::All.contains(character) {
+                character
+            } else {
+                '?'
+            }
+        })
+        .collect()
+}
+
 fn compute_checksum(
     metadata: &pb_backup::BackupMetaData,
     data: &pb_backup::BackupData,
--- a/src/rust/util/src/ascii.rs
+++ b/src/rust/util/src/ascii.rs
@@ -12,12 +12,21 @@ pub enum Charset {
     AllNewline,
 }
 
+impl Charset {
+    /// Returns true if `character` is in this charset.
+    pub fn contains(&self, character: char) -> bool {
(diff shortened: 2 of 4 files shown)


src/rust/bitbox02-rust/src/backup.rs
```diff
@@ -70,6 +70,19 @@ pub fn id(seed: &[u8]) -> String {
     hex::encode(mac.finalize_reset().into_bytes())
 }
 
+/// Replaces non-printable-ASCII characters in legacy backup names with `?`.
+pub(crate) fn sanitize_name(name: &str) -> String {
+    name.chars()
+        .map(|character| {
+            if (' '..='~').contains(&character) {
```
benma, line 77: Could add a little helper to ascii.rs, which already has `is_printable_ascii` that does the same check inside.

benma (approved):
utACK, but see comment.

src/rust/bitbox02-rust/src/backup.rs
NickeZ: PTAL I had to do some refactoring to make the current is_printable_ascii work with chars.

src/rust/util/src/ascii.rs
```diff
@@ -27,6 +36,15 @@ mod tests {
 
     static ALL_ASCII: &[u8] = "! \"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~".as_bytes();
 
+    #[test]
+    fn test_contains() {
+        assert!(Charset::All.contains(' '));
+        assert!(Charset::All.contains('~'));
+        assert!(!Charset::All.contains('\n'));
+        assert!(!Charset::All.contains('ï'));
+        assert!(Charset::AllNewline.contains('\n'));
```
benma, line 45: At least one negative test would be good

benma (approved):
utACK, see unit test comment

cedwies (approved):
tACK

src/rust/util/src/ascii.rs
NickeZ: What do you call those tests on line 43/44?

PR #2084, opened 2026-09-04
Fix simulator stdout formatting

benma-agent:
Pass testing-only messages through constant printf formats so percent sequences remain data.


cedwies (approved):
utACK

PR #2085, opened 2026-09-07
payment_request: reject unsupported names

benma-agent:
Validate recipient names as printable ASCII before showing payment-request
screens. SWAPKIT provider names previously passed identity lookup without this
check and could reach the new UI assertion.

Return InvalidInput for non-ASCII text and all control characters, including
newlines, while preserving printable ASCII names.

refs #2065


NickeZ (approved):
utACK

PR #2086, opened 2026-09-07
docs: require context prefixes in commit subjects

benma-agent:
Require context-prefixed commit subjects and reuse the conventions in recent
commits touching the affected code. Apply the existing 50-character limit to
the complete subject, including the prefix.

State the rule directly in the agent guidelines and remove their redundant
contributor-guide reference and obsolete [WIP] draft-title instruction.


cedwies (approved):
utACK

PR #2088, opened 2026-09-07
bitcoin: reject missing policy xpubs

benma-agent:
Policy registration-status queries hash host-provided policies without parsing them first.
A missing optional xpub currently panics in the shared hash helper.

Return an error for absent xpubs and report invalid lookup keys as InvalidInput. Keep policy
parsing, valid hash bytes and existing registration lookups unchanged.

Cover missing and malformed xpubs in either key position, successful registration lookups and
the shared hash helper with regression tests.


cedwies (approved):
tACK

PR #2090, opened 2026-09-07
cardano: limit xpub batch size

benma-agent:
Limit Cardano xpub requests to 20 keypaths, matching the Bitcoin batch
limit. Reject larger batches before allocating response vectors or
accessing the keystore to bound derivation work and response size.

Cover the 20/21 boundary and document the new API limit in the firmware
changelog. The generic response-capacity check is handled separately in
#2087.


cedwies (approved):
utACK

PR #2096, opened 2026-09-09
usb: own HID receive buffers

benma-agent:
hid_hww_read() and hid_u2f_read() returned while hid_read() retained the
caller's pointer for an asynchronous USB OUT transfer. Their safe Rust wrappers
therefore ended the mutable borrow before a delayed host report DMA-wrote into
the buffer, which could already be freed or reused.

Give HWW and U2F separate 32-bit-aligned static 64-byte receive buffers. The USB
driver owns these for the whole transfer, and read polling copies a report into
the caller only after the completion callback.


NickeZ (approved):
utACK, I assume performance took a small hit, but this is easier than making the rust code compliant.

PR #2097, opened 2026-09-10
sd: validate transfer speed fields

benma-agent:
Reserved CSD and SDIO transfer-speed fields could index one past the
transfer-unit table when the unit code was 7.

Use a shared checked decoder, reject reserved units and multiplier code 0,
and fail card initialization on malformed values.


NickeZ (commented):
you are changing external code. How should we keep track of that if we upgrade?

PR #2098, opened 2026-09-10
oled: hold display in reset during startup

NickeZ:
Production stage0 verifies stage1 before initializing the OLED. Leaving the reset pin high can allow previous display content to remain visible at reduced brightness while the boost converter is disabled.

Set the reset output latch low before enabling the pin as an output. Keep the OLED in reset until oled_init() releases it and uploads a cleared framebuffer.

PR #2099, opened 2026-09-10
blupgrade: accept other stage0 image versions

benma-agent:
The production bootloader-upgrade firmware checks the installed
bootloader's development flags before writing stage0. Descriptor parsing
currently requires the installed stage0 image version to match the
installer's build, even though the descriptor layout is unchanged.

After a version bump, older descriptors fall back to legacy text-marker
scanning. This loses the stage0 development flag when stage1 has no
legacy development markers, such as a production stage1 paired with a
development stage0. Normal development stage1 images still carry those
markers and remain detectable.

Validate the descriptor using its magic and product ID regardless of the
stage0 image version. Extend the existing checks to cover v1 and a newer
image version for stage0 and stage1 development flags and production
descriptors.

refs #2098

### DIFF
--- a/src/bootloader_upgrade/firmware_installer_check.c
+++ b/src/bootloader_upgrade/firmware_installer_check.c
@@ -57,8 +57,8 @@ static bool _read_stage0_descriptor(
         return false;
     }
     memcpy(descriptor_out, descriptor, sizeof(*descriptor_out));
+    // The stage0 image version can differ from ours without changing the descriptor layout.
     return descriptor_out->magic == BB02_STAGE0_DESCRIPTOR_MAGIC &&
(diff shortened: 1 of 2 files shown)


src/bootloader_upgrade/firmware_installer_check.c
```diff
@@ -57,8 +57,8 @@ static bool _read_stage0_descriptor(
         return false;
     }
     memcpy(descriptor_out, descriptor, sizeof(*descriptor_out));
+    // The stage0 image version can differ from ours without changing the descriptor layout.
```
NickeZ, line 60: This comment is slop and doesn't make sense after the code change.

It should perhaps read "version is not checked as installer currently uses fields available since v1"?

NickeZ (approved):
utACK, please change the comment though.

PR #2100, opened 2026-09-10
platform: bound heap growth

benma-agent:
ASF4's _sbrk implementation advances the program break without checking the
linker-defined heap range. A large allocation can therefore move the break
past _heap_end and let subsequent writes reach outside the reserved heap.

Replace the vendored syscall source with a first-party _sbrk implementation
that checks growth and shrink requests using overflow-safe arithmetic.
Return ENOMEM or EINVAL without changing the break when a request crosses
the linker-defined bounds. Keeping this code outside ASF4 prevents a future
vendor update from silently dropping the checks again.

nosys.specs continues to provide the unused syscall stubs. Add a unit test
for exact-fit allocation, rejected growth and shrink, unchanged state after
failure, and INT_MIN handling.

This prevents writes beyond the heap. Exhausting memory still follows the
existing allocation-failure path.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -100,7 +100,6 @@ if(CMAKE_CROSSCOMPILING)
   # Advanced Software Framework Drivers
   # Only drivers needed for bootloader (minimal set)
   add_library(asf4-drivers-min STATIC
-    asf4-drivers/hal/utils/src/utils_syscalls.c
     asf4-drivers/hal/utils/src/utils_list.c
     asf4-drivers/hal/src/hal_atomic.c
     asf4-drivers/hal/src/hal_gpio.c
--- a/src/platform/sbrk.c
+++ b/src/platform/sbrk.c
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#include <errno.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <sys/types.h>
+
+extern uint8_t _heap_start;
+extern uint8_t _heap_end;
+
+caddr_t _sbrk(int incr);
+
+// Unlike ASF4's implementation, this confines the program break to the
+// linker-defined heap and checks growth and shrink without overflowing.
+caddr_t _sbrk(int incr)
+{
+    static uint8_t* heap;
+    const uintptr_t heap_start = (uintptr_t)&_heap_start;
+    const uintptr_t heap_end = (uintptr_t)&_heap_end;
+
+    if (heap == NULL) {
+        heap = &_heap_start;
+    }
+
+    const uintptr_t current = (uintptr_t)heap;
+    if (incr >= 0) {
+        const uintptr_t increment = (uintptr_t)incr;
+        if (current > heap_end || increment > heap_end - current) {
+            errno = ENOMEM;
+            return (caddr_t)-1;
+        }
+        heap = (uint8_t*)(current + increment);
+    } else {
+        const uintptr_t decrement = (uintptr_t)(-(int64_t)incr);
+        if (current < heap_start || decrement > current - heap_start) {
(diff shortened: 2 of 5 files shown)


test/unit-test/test_sbrk.c
```diff
@@ -0,0 +1,54 @@
...
+    ".global _heap_end\n"
+    "_heap_end = test_heap + 16\n");
+
+extern caddr_t _sbrk(int incr);
+
+static void test_sbrk_bounds(void** state)
+{
+    (void)state;
+
+    assert_ptr_equal(_sbrk(0), test_heap);
```
NickeZ, line 28: shouldn't the tests use malloc/free as that is the typical entrypoint to the heap?

external/CMakeLists.txt
```diff
@@ -99,8 +99,8 @@ if(CMAKE_CROSSCOMPILING)
 
   # Advanced Software Framework Drivers
   # Only drivers needed for bootloader (minimal set)
+  # ASF4's utils_syscalls.c has an unbounded _sbrk; src/platform/sbrk.c provides the bounded one.
```
NickeZ: this comment is unnecessary.

src/platform/sbrk.c
```diff
@@ -0,0 +1,41 @@
...
+#include <stddef.h>
+#include <stdint.h>
+#include <sys/types.h>
+
+extern uint8_t _heap_start;
+extern uint8_t _heap_end;
+
+caddr_t _sbrk(int incr);
+
+caddr_t _sbrk(int incr)
```
NickeZ, line 15: Add a comment here about what is unique about our implementation.

test/unit-test/test_sbrk.c
benma: Using malloc/free here would not use that _sbrk implementation:

> These cmocka tests are native Linux binaries, so malloc/free use the host libc and do not call the firmware’s _sbrk. Using them here would therefore leave this implementation untested. A meaningful malloc/free test would require running newlib-nano on the embedded target or an emulator, so I kept the direct _sbrk boundary tests.

src/platform/sbrk.c
benma: Added

external/CMakeLists.txt
benma: Removed

NickeZ (approved):
utACK

benma:
I did a manual on-device test to check that malloc/sbrk works as expected using:

```C

    // Temporary on-device heap bounds test. Remove before shipping.
    caddr_t before = _sbrk(0);
    uintptr_t remaining = (uintptr_t)&_heap_end - (uintptr_t)before;

    errno = 0;
    caddr_t result = _sbrk((int)(remaining + 1));
    int error = errno;
    caddr_t after = _sbrk(0);
    if (result != (caddr_t)-1 || error != ENOMEM || after != before) {
        // The old implementation advances the break. Restore it before Abort can allocate.
        if (after != before) {
            _sbrk(-(int)((uintptr_t)after - (uintptr_t)before));
        }
        Abort("FAIL: sbrk bounds");
    }

    // Also exercise the actual C allocator used by Rust.
    // A volatile function pointer prevents GCC from optimizing away the allocation checks.
    void* (*volatile test_malloc)(size_t) = malloc;
    size_t heap_size = (uintptr_t)&_heap_end - (uintptr_t)&_heap_start;
    void* ptr = test_malloc(heap_size + 1);
    if (ptr != NULL) {
        Abort("FAIL: oversized malloc");
    }

    ptr = test_malloc(32);
    if (ptr == NULL) {
        Abort("FAIL: allocation after OOM");
    }
    free(ptr);
    screen_print_debug("Heap test passed", 2000);

```

PR #2104, opened 2026-09-15
btc: reject invalid silent payment versions

benma-agent:
A valid Bech32m checksum does not guarantee that the data starts with a
valid witness version. Return the existing decoding error when the
version is absent or invalid instead of unwrapping it.

Keep support restricted to silent payment version zero. Cover missing
and invalid versions on both supported networks in the decoder tests.


cedwies (approved):
utACK

PR #2105, opened 2026-09-15
btc: reject unsupported policy hash fragments

benma-agent:
Reject Miniscript hash fragments during wallet-policy parsing, before
policy confirmation and registration. Apply the same validation to WSH
scripts and every Taproot leaf, preserving the encoding-depth limit.

Return InvalidInput from all four hash translation callbacks instead of
using the panicking helper. Hash fragments remain unsupported;
previously registered policies containing them are rejected on use.


cedwies (approved):
utACK

PR #2106, opened 2026-09-16
ui: remove unused cancel_is_backbutton

benma-agent:
No production caller enables cancel_is_backbutton. Remove the flag from the UI HAL,
BitBox02 Rust/C parameters, renderer state, and parameter-conversion tests.

Keep the existing default behavior: BitBox02 shows a cancel cross for empty input
and a back arrow while editing, and BitBox03 uses the cancel icon.


cedwies (approved):
utACK

PR #2073, opened 2026-08-27
bb03 ui updates

jadzeidan:
- add cancel/success icons
- update keyboard styling for recovery words entry and improve ux
- show all recovery words on same screen

PR #1986, opened 2026-06-03
bb03: placeholder UI for restore from mnemonic

jadzeidan:
Implements AI generated UI for restore
from mnemonic workflow.

PR #2107, opened 2026-09-16
workflow: remove mnemonic word retry

benma-agent:
Remove the redundant wordlist membership retry and its artificial test
introduced for the placeholder UI. Call enter_wordlist_word directly
with the same parameters and preset, preserving the dedicated word-entry
HAL and its back/cancel handling.

Both device UIs only confirm exact members of the supplied wordlist:
BitBox02 guards confirmation with can_confirm, and BitBox03's word-entry
UI from #2073 enables it only for complete matches. This also covers
restricted final-word choices, so the removed error/retry path is
unreachable through normal input. Final-word confirmation is unchanged.

Refs #1986.


benma-agent:
Superseded by the fresh draft https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2108, rebased onto master after #2073 merged.

PR #2109, opened 2026-09-16
hal: narrow recovery word entry parameters

benma-agent:
Accept the title, required wordlist, and preset directly instead of the
general EnterStringParams with its optional wordlist and unrelated flags.
Construct the parameters inside the UI implementations and always use
CanCancel::Yes, as recovery-word entry always allows aborting.

Preserve the BitBox02 enter_string fallback and existing back/cancel
behavior.


cedwies (approved):
utACK

PR #2110, opened 2026-09-16
bb03: wipe recovery text and wrap textarea access

benma-agent:
Follow up #2073 by wiping recovery-word UI text when it is replaced or deleted and keeping textarea reads behind safe LVGL wrappers.

- Store displayed recovery words, live word entry, and keyboard previews in fixed Rust-owned zeroizing buffers that LVGL borrows without copying. Clear previews when hidden.
- Zeroize the aggregate recovery-word text and use `ZeroizingText` for temporary strings and snapshots at the Rust/LVGL boundary. Reserve buffers before copying sensitive text to avoid leaving copies behind during reallocation.
- Replace direct textarea text reads in BitBox03 feature code with a safe accessor that retains an owned zeroizing snapshot while the callback runs.


src/rust/bitbox-lvgl/src/text.rs
```diff
@@ -0,0 +1,65 @@
...
+use crate::LvTextError;
+
+/// An owned text snapshot that wipes its allocation when dropped.
+///
+/// Temporary strings and snapshots crossing the LVGL boundary use this type so their
+/// Rust-owned storage is wiped automatically. Native widget text copies need separate
+/// protection; use [`crate::LvZeroizingLabel`] for labels with zeroizing backing storage.
+pub struct LvText(Zeroizing<Vec<u8>>);
+
+impl LvText {
```
NickeZ, line 17: I don't think using the `Lv` prefix is appropriate, we have only used it for lvgl types until now.

If you never intend to use the "resize" functionality of Vec. could you just store a `Box<[u8;N]>` instead?

benma: `Box<[u8]>` works :+1: renamed to `ZeroizingText` and changed to `Box<[u8]>`

PR #2111, opened 2026-09-17
hww: reset sessions on host reconnect

benma-agent:
When a host disconnects after receiving an intermediate response, a workflow can remain in
next_request().await while the device stays powered. The HWW lock has already been released, so
the USB timeout does not cancel it. On reconnect, the Python client's attestation request is
consumed as the old workflow's continuation and receives an error encrypted with the old Noise
session instead of an attestation response.

Add a top-level REQ_RESET that cancels the pending task, discards queued HWW responses, resets
Noise and restores the UI before acknowledging the new session. Drop task-owned UI components
before resetting the screen stack, and refuse reset while a U2F workflow owns the shared UI.
Allow REQ_INFO through the HWW busy gate without refreshing the operation timeout, so the host
can discover the firmware version before deciding whether to reset.

Bump the firmware to v9.28.0. The Python client sends REQ_RESET before attestation, unlock and
Noise setup through the private BitBoxCommonAPI._reset_session helper, which skips firmware
older than v9.28.0. Add regression tests for abandoned continuations, pending responses, U2F
arbitration, INFO timeout behavior and client compatibility, and document the API under
Unreleased. The powered-device reconnect flow still needs hardware validation.


py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -598,6 +612,9 @@ def __init__(
         else:
             self._bitbox_protocol = BitBoxProtocolV1(transport)
 
+        if self.version >= semver.VersionInfo(9, 28, 0):
```
NickeZ: Is this really necessary? won't an old device just respond with a HWW_RSP_NACK?

NickeZ: I don't like checking the version before we have asked the device. going forward I don't think we can assume that we know the version before we have asked the device for it.

benma: it's just cleaner and explicit, and very old devices didn't even have `HWW_` framing

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
```diff
@@ -300,6 +302,18 @@ def __init__(self, transport: TransportLayer):
...
 
+    def reset_session(self) -> None:
+        """Reset the previous session before attestation, unlock and the Noise handshake."""
+        cid = self._transport.generate_cid()
+        while True:
+            response = self._transport.query(HwwRequestCode.REQ_RESET, HWW_CMD, cid)
+            if response == HwwResponseCode.RSP_BUSY:
+                time.sleep(1)
+                continue
+            if response != HwwResponseCode.RSP_ACK:
```
NickeZ: check version here? NACK is the expected response for older fw.

benma: But it's not called for older firmware. Maybe just add it to the docstring?

NickeZ: Could check version earlier as well and just skip sending the reset.

NickeZ: the function isn't private? the client shouldn't have to check before calling this function

NickeZ: either check and skip sending the message or accept all valid responses. we can't throw an exception on a valid response from an old fw.

NickeZ (approved):
utACK

py/bitbox02/bitbox02/communication/bitbox_api_protocol.py
benma: Made it private and moved the version check inside.

PR #2113, opened 2026-09-17
build: remove C coverage tooling

benma:
The C coverage report targets no longer have instrumentation feeding them, so make coverage can succeed with an empty report.

Remove the Makefile entry and coverage flags, the CMake report targets and tool discovery, the dedicated CI job, and the corresponding build instructions and agent guidance.

Drop gcovr and lcov/genhtml from the development Docker image along with the gcov alternative setup. Bump the container version to 56 so CI rebuilds the image with these dependencies removed.


NickeZ (approved):
Lgtm

PR #2115, opened 2026-09-20
hww: support host passphrase entry

benma-agent:
Allow host entry of the optional BIP39 passphrase while keeping device entry as the default. The user approves switching to host entry on the device, then confirms the submitted passphrase before unlocking. Rejection, cancellation or invalid input restarts device entry.

Run the new unlock flow inside the paired Noise channel, preserving legacy unlock compatibility. Keep the existing request/response transport with polling for device entry, and reuse session reset on reconnect to abandon interrupted workflows safely.

Restrict host input to passphrases that can also be entered on the device. Withdraw the host-entry option before consent or confirmation starts, and prevent repeated unlock calls from changing an unlocked wallet.

Extend the Python API with callbacks for host input and optional entry controls, with polling owned by the library. Demonstrate entry, cancellation and withdrawal in `send_message.py`.

Keep the Python 3.7 requirement and supporting UI, request-buffer cleanup, U2F and test-helper changes in separate prerequisite commits.

Related PRs:

- [api-go #188](https://github.com/BitBoxSwiss/bitbox02-api-go/pull/188)
- [Wallet app #4445](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4445)


cedwies (commented):
Concept ACK.

There is one issue I would like to discuss (also mentioned in [#2000](https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2000):

Capital `i` and lowercase `l` are pixel identical. This issue predates this PR but IMO becomes a larger problem for host passphrase entry, since a malicious host can swap out letters while the user has no chance to detect it. We should fix this issue before merging this PR.

benma:
> Concept ACK.
> 
> There is one issue I would like to discuss (also mentioned in [#2000](https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2000):
> 
> Capital `i` and lowercase `l` are pixel identical. This issue predates this PR but IMO becomes a larger problem for host passphrase entry, since a malicious host can swap out letters while the user has no chance to detect it. We should fix this issue before merging this PR.

Tried here, but after discussing, we don't continue with this: https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2117

The search space to recover the right passphrase is very small, which makes an effective ransom attack here very unlikely.

cedwies (approved):
tACK

PR #2116, opened 2026-09-20
py: remove redundant HPKE casts

benma-agent:
The CI image installs pyhpke from the Python requirements, so mypy now knows its byte return
types and rejects the temporary casts as redundant. Remove the casts, unused import and
resolved dependency TODO without changing the wrapped DEK encoding.

New image made in https://github.com/BitBoxSwiss/bitbox02-firmware/pull/2113 includes the above dep, which introduced the lint failure.


cedwies (approved):
utACK

